PR #8, opened 2026-06-24
moe: PDL on the decode gate_up->down pair (gated SPARKINFER_PDL) [POC]

ai-hpc:
## Summary

**POC** — Programmatic Dependent Launch (PDL) on the decode MoE `gate_up → down` pair, gated behind
`SPARKINFER_PDL=1` (default **off**). Adds `si_pdl_lc()` / `si_pdl_sync()` device helpers and launches
the split-K `down` kernel via `cudaLaunchKernelEx` + `ProgrammaticStreamSerialization`, so `down`'s grid
spin-up overlaps `gate_up`'s tail — targeting the bs=1 decode launch-latency bound.

Opened mainly to **exercise the eval pipeline end-to-end**. Two caveats: (1) **untested** — not yet built
on a Blackwell box; (2) PDL is **gated off by default**, so the auto-eval (which runs the default build)
measures the unchanged path → expect `eval:none` unless run with `SPARKINFER_PDL=1`.

## Proof of speedup

```text
pending eval
```

| | decode tok/s |
|---|--:|
| before | 163.88 |
| after  | (pending) |

### DIFF
--- a/kernels/csrc/cuda/moe/expert_ffn_q4k.cu
+++ b/kernels/csrc/cuda/moe/expert_ffn_q4k.cu
@@ -26,6 +26,21 @@ namespace kernels {
 
 static constexpr int WPB = 8;   // warps per block
 
+// Programmatic Dependent Launch (PDL): overlap a kernel's grid spin-up with its
+// predecessor's tail to hide bs=1 decode launch latency (the ncu-confirmed bottleneck).
+// No-op unless the kernel is launched programmatic (cudaLaunchKernelEx +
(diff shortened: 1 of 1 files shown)


ai-hpc:
## ✅ sparkinfer auto-eval — `ffd7b42`

| metric | value |
|---|---|
| **label** | `eval:L` |
| decode | 187.61 tok/s |
| vs frontier | 163.88 tok/s → +14.5% (+23.7) |

Verified speedup over the live frontier.

PR #12, opened 2026-06-24
fix(runtime): make GGUF load safe when shared-expert tensors are absent

### LINKED ISSUE #7: runtime: GGUF decode dereferences null shared-expert weights when n_shared > 0 (the default)
## Summary

`Qwen35Model::load_gguf()` never populates the shared-expert weights (`w.shared_gate`, `w.shared_up`, `w.shared_down`), but `forward_token()` unconditionally launches the shared-expert FFN whenever `cfg.n_shared > 0` — and `n_shared` defaults to **1**. Any caller that loads a GGUF with the default config therefore dereferences three null device pointers inside `launch_moe_expert_ffn`, crashing decode.

## Details

`Qwen35Config::n_shared` defaults to `1`:

```cpp
// runtime/include/sparkinfer/models/qwen35.h:23
int   n_shared    = 1;
```

The per-token forward runs the shared expert when `n_shared > 0`:

```cpp
// runtime/src/models/qwen35.cpp:204-208
if (c.n_shared > 0) {
    kernels::launch_moe_expert_ffn(s.hn, w.shared_gate, w.shared_up, w.shared_down,
                                   s.d_shared_ids, s.d_shared_w, s.shared,
                                   1, 1, 1, H, c.moe_ffn, st);
    launch_residual_add(s.routed, s.shared, s.routed, H, st);
}
```

But `load_gguf()` (runtime/src/models/qwen35.cpp:307-383) only assigns the routed-expert tensors (`gate_q`/`up_q`/`down_q`); `w.shared_gate/up/down` are left at their `nullptr` initializers (qwen35.h:45-47). Only the directory-weight loader (lines 298-299) ever sets them. So on the GGUF path the three pointers are null and get dereferenced on the device.

## Why it isn't caught today

Every shipped GGUF example explicitly sets `cfg.n_shared = 0` (`qwen3_gguf_generate.cpp:48`, `qwen3_gguf_score.cpp:51`, `qwen3_gguf_bench.cpp:45`), which masks the bug. The target model (Qwen3-30B-A3B) has no shared expert, so `n_shared = 0` is in fact the correct value for it — but nothing in `load_gguf` enforces that, so it is a latent trap for any new caller that uses the default config.

## Proposed fix

Make the GGUF path self-consistent: in `load_gguf()`, detect whether the file actually contains shared-expert tensors (`blk.0.ffn_gate_shexp.weight` etc.). If it does not, force `s.cfg.n_shared = 0` (with a one-line warning) so `forward_token` never launches the shared FFN with null weights. This keeps existing callers working and makes default-config GGUF loads safe.

## Affected

- `runtime/src/models/qwen35.cpp` (`load_gguf`, `forward_token`)
- area: `runtime`

jason020818:
## Summary

Make the GGUF load path self-consistent for shared experts: clamp `n_shared` to 0 when the file has no `ffn_*_shexp` tensors (the Qwen3-30B-A3B case), and load those tensors when they are present so `forward_token` never launches `launch_moe_expert_ffn` with null weights.

Fixes #7.

## Proof of speedup

N/A — correctness/safety fix on the GGUF load path; no kernel or decode-path change.

| | decode tok/s |
|---|---:|
| before | N/A |
| after  | N/A |

## Test plan

- [ ] `bench/scripts/accuracy.sh` unchanged (Qwen3-30B-A3B already uses `n_shared=0`)

### DIFF
--- a/runtime/src/models/qwen35.cpp
+++ b/runtime/src/models/qwen35.cpp
@@ -311,6 +311,18 @@ bool Qwen35Model::load_gguf(const std::string& path) {
     GGUF g;
     if (!g.open(path)) return false;
 
+    // Shared-expert tensors are optional in GGUF (Qwen3-30B-A3B has none). The
+    // default config sets n_shared=1, so clamp it to what the file actually
+    // contains before forward_token can launch a null-weight FFN.
+    const bool gguf_has_shared =
+        g.tensor("blk.0.ffn_gate_shexp.weight") != nullptr;
+    if (c.n_shared > 0 && !gguf_has_shared) {
+        fprintf(stderr,
+                "[gguf] no shared-expert tensors; forcing n_shared=0 "
+                "(safe for models without a shared FFN)\n");
+        s.cfg.n_shared = 0;
+    }
+
     // upload raw quantized blocks, keep on device (for experts)
     auto dev_quant = [&](const std::string& name, int& qtype) -> const void* {
         const GGUFTensor* t = g.tensor(name);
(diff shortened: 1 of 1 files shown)


ai-hpc:
## ⚪ sparkinfer auto-eval — `aed5af2`

| metric | value |
|---|---|
| **label** | `eval:none` |
| decode | 184.61 tok/s |
| vs frontier | 187.61 tok/s → +0.0% (-3.0) |

PR #17, opened 2026-06-24
kernels/quant: reject unsupported GGUF types instead of silently decoding as F32

joaovictor712:
## Summary

Make `launch_gguf_dequant` fail loudly on a GGUF type it has no decoder for, instead of silently reinterpreting it as F32.

Fixes #16

**Why.** `kernels/csrc/cuda/quant/dequant_gguf.cu` decoded Q4_K/Q6_K/Q8_0/F16 and let every other ggml type fall through to the F32 decoder. That is reachable: `runtime/src/gguf.cpp` `block_info()` already sizes **Q5_K** (ggml type 13, 176 B / 256), so a Q5_K_M GGUF loads and its tensors reach `launch_gguf_dequant`, which has no Q5_K branch → the `else /* F32 */` path reinterprets 176-byte Q5_K blocks as raw float32 → silent garbage, no error (and a confusing downstream correctness failure rather than a fast, clear load error).

**Change.** Give F32 its own explicit branch and reject any type without a decoder:

```cpp
else if (ggml_type == GGML_F32) { deq_f32_kernel<<<...>>>(reinterpret_cast<const float*>(src),d,n_values); }
else { fprintf(stderr, "[gguf] launch_gguf_dequant: unsupported ggml type %d — no decoder ...\n", ggml_type); }
```

## Proof of speedup

- [ ] Tested on **RTX 5090** (`sm_120`)

N/A — this is a **correctness/robustness-only** change in the GGUF dequant dispatch, not a perf change. Valid Q4_K/Q6_K/Q8_0/F16/F32 weights are dequantized by the exact same kernels as before (the supported-type paths are untouched), so decode throughput and the accuracy gate are unaffected. Only genuinely-unsupported types change behavior (silent garbage → diagnosed refusal).

**Benchmark log** (before → after):

```text
N/A — dispatch hardening, no kernel math / perf change. Supported-type dequant output is bit-identical to baseline.
```

| | decode tok/s |
|---|--:|
| before | N/A |
| after  | N/A |

## Scope

Single focused change in `kernels/csrc/cuda/quant/dequant_gguf.cu`; one fix per PR per CONTRIBUTING.md, scoped to `kernels/`.

PR #20, opened 2026-06-24
fix(runtime): reject unsupported GGML types during GGUF load

### LINKED ISSUE #16: kernels/quant: launch_gguf_dequant silently decodes unsupported GGUF types (e.g. Q5_K) as F32 — garbage output, no error
## Summary

`launch_gguf_dequant` decodes Q4_K / Q6_K / Q8_0 / F16 and lets **every other** ggml type fall through to the F32 decoder. Any tensor of an unsupported-but-real type is then reinterpreted byte-for-byte as `float32`, producing silent garbage with no error.

This is reachable: the runtime's GGUF loader already *sizes* Q5_K. `runtime/src/gguf.cpp` `block_info()` returns a valid `(bytes=176, elems=256)` for ggml type **13 (Q5_K)**, so a Q5_K_M GGUF loads and its tensors get handed to `launch_gguf_dequant`, which has no Q5_K branch → the `else /* F32 */` path → 176-byte Q5_K blocks read as raw floats. The model then produces nonsense and (under the eval harness) fails the correctness gate for a non-obvious reason instead of failing fast at load.

## Location

`kernels/csrc/cuda/quant/dequant_gguf.cu` (`launch_gguf_dequant`)

```cpp
if (ggml_type == GGML_Q4_K) { ... }
else if (ggml_type == GGML_Q6_K) { ... }
else if (ggml_type == GGML_Q8_0) { ... }
else if (ggml_type == GGML_F16)  { ... }
else /* F32 */                   { deq_f32_kernel<<<...>>>(reinterpret_cast<const float*>(src),d,n_values); }
```

The enum in this file (`GGML_F32=0, F16=1, Q8_0=8, Q4_K=12, Q6_K=14`) has no `Q5_K=13`, so Q5_K (and any future/unknown type) hits the F32 fallthrough.

## Expected behavior

F32 should be its own explicit branch, and any type without a decoder should fail loudly rather than be silently misinterpreted as F32.

## Proposed fix

```cpp
else if (ggml_type == GGML_F32) { deq_f32_kernel<<<...>>>(reinterpret_cast<const float*>(src),d,n_values); }
else { fprintf(stderr, "[gguf] launch_gguf_dequant: unsupported ggml type %d — no decoder; refusing to reinterpret as F32\n", ggml_type); }
```

## Impact

Correctness / robustness: turns silent garbage (and a confusing downstream correctness failure) into a clear, diagnosed refusal. Valid Q4_K / Q6_K / Q8_0 / F16 / F32 weights are dequantized identically — no behavior change on supported types.

jason020818:
## Summary

Reject GGUF tensors whose ggml type `launch_gguf_dequant` cannot decode (e.g. Q5_K) during `load_gguf`, so load fails fast with a clear error instead of silently decoding as F32 and corrupting weights.

Complements the kernel-side guard in #17 — this catches unsupported types before any device dequant runs.

Fixes #16.

## Proof of speedup

N/A — correctness/safety on the GGUF load path.

| | decode tok/s |
|---|---:|
| before | N/A |
| after  | N/A |

## Test plan

- [ ] Load a known-good Qwen3-30B-A3B Q4_K_M GGUF — unchanged
- [ ] Attempting a Q5_K tensor on the dense/dequant path returns `load_gguf() == false`

### DIFF
--- a/runtime/src/models/qwen35.cpp
+++ b/runtime/src/models/qwen35.cpp
@@ -32,6 +32,21 @@ inline void cu(cudaError_t e, const char* what) {
     if (e != cudaSuccess) fprintf(stderr, "[qwen35] %s: %s\n", what, cudaGetErrorString(e));
 }
 using bf16 = unsigned short;
+
+// launch_gguf_dequant only implements F32/F16/Q8_0/Q4_K/Q6_K. Reject anything
+// else at load time so Q5_K (etc.) cannot silently fall through as F32.
+bool ggml_dequant_supported(int ggml_type) {
+    switch (ggml_type) {
+        case 0:  // F32
+        case 1:  // F16
+        case 8:  // Q8_0
+        case 12: // Q4_K
+        case 14: // Q6_K
+            return true;
+        default:
+            return false;
+    }
+}
 }
 
 struct Qwen35Model::Impl {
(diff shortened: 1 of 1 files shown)


ai-hpc:
## ⚪ sparkinfer auto-eval — `022fcd3`

| metric | value |
|---|---|
| **label** | `eval:none` |
| decode | 186.35 tok/s |
| vs frontier | 187.61 tok/s → +0.0% (-1.3) |

PR #21, opened 2026-06-24
fix(runtime): guard DecodeRunner batch size against scratch overflow

kiannidev:
## Summary
- Refuse `begin_step` / `decode_layer` when `num_seqs` is ≤0 or exceeds `max_batch`
- Prevents OOB writes into CUDA scratch buffers sized at construction time

## Why
`DecodeRunner` allocates per-token buffers for `max_batch` (default 256). A scheduler bug or misconfigured batch could pass a larger `num_seqs` and corrupt device memory silently.

## Test plan
- [x] Builds on paths that don't touch `bench/scripts/` (passes sensitive-paths-guard)
- [ ] `ctest --test-dir build` on Blackwell
- [ ] Existing decode GPU tests unchanged

### DIFF
--- a/runtime/src/decode_layer.cpp
+++ b/runtime/src/decode_layer.cpp
@@ -62,6 +62,11 @@ DecodeRunner::~DecodeRunner() {
 
 void DecodeRunner::begin_step(const std::vector<int>& seq_lens_before) {
     const int n = (int)seq_lens_before.size();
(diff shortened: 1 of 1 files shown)


ai-hpc:
## ⚪ sparkinfer auto-eval — `0680b3d`

| metric | value |
|---|---|
| **label** | `eval:none` |
| decode | 186.38 tok/s |

PR #26, opened 2026-06-24
runtime/gguf: clamp invalid general.alignment to default (fix divide-by-zero)

### LINKED ISSUE #24: runtime/gguf: general.alignment of 0 (or negative) causes divide-by-zero computing tensor data_start
## Summary

The GGUF loader reads `general.alignment` from file metadata but only guards against the key being *absent*, not against a present-but-invalid value:

```cpp
long alignment = ints_.count("general.alignment") ? ints_["general.alignment"] : 32;
```

It is then used as a divisor when resolving the tensor-data section:

```cpp
size_t data_start = (c.off + alignment - 1) / alignment * alignment;   // alignment == 0 -> divide by zero
```

A GGUF file that declares `general.alignment = 0` triggers an integer divide-by-zero (SIGFPE) at load. A negative alignment is equally invalid and produces a garbage (mis-aligned / underflowed) `data_start`, so every tensor's data pointer is wrong.

## Location

`runtime/src/gguf.cpp` — `GGUF::parse()` (the `alignment` read and the `data_start` computation).

## Why it matters

`general.alignment` is file-controlled. Valid GGUF alignment must be a positive power of two (the spec default is 32). The loader already fails loudly on other malformed metadata (unsupported array element types, out-of-range spans, `n_dims > 4`); a zero/negative alignment is the same class of untrusted-input hazard but is currently unguarded, turning a malformed file into a hard crash (SIGFPE) instead of a clean diagnosed failure.

## Proposed fix

Clamp an invalid alignment back to the default after reading it:

```cpp
long alignment = ints_.count("general.alignment") ? ints_["general.alignment"] : 32;
if (alignment <= 0) alignment = 32;
```

## Impact

Robustness only. Well-formed files (which carry a positive power-of-two alignment, or omit the key) are completely unaffected — the model-load path and accuracy gate are unchanged. Only a malformed `alignment <= 0` changes behavior: crash → safe default.

philluiz2323:
## Summary

Guard the GGUF loader against an invalid file-controlled `general.alignment`. The value is used as a divisor when resolving the tensor-data section, so a declared `general.alignment = 0` divides by zero (SIGFPE) and a negative value mis-aligns `data_start`. The existing guard only handled a *missing* key, not a present-but-invalid value.

Fixes #24

**Change.** After reading `general.alignment`, clamp any non-positive value back to the spec default (32):

```cpp
long alignment = ints_.count("general.alignment") ? ints_["general.alignment"] : 32;
if (alignment <= 0) alignment = 32;
```

This matches how the loader already fails safe on other malformed metadata (unsupported array element types, out-of-range spans, `n_dims > 4`).

## Proof of speedup

N/A — robustness-only change to the host-side GGUF loader, not a perf change. Well-formed files carry a positive power-of-two alignment (or omit the key, defaulting to 32), so they parse byte-identically; the model-load path and accuracy gate are unchanged. Only a malformed `alignment <= 0` changes behavior (SIGFPE → safe default). No decode-throughput impact.

```text
```

| | decode tok/s |
|---|--:|
| before | N/A |
| after  | N/A |

### DIFF
--- a/runtime/src/gguf.cpp
+++ b/runtime/src/gguf.cpp
@@ -97,6 +97,11 @@ bool GGUF::open(const std::string& path) {
     if (!c.ok) { fprintf(stderr, "[gguf] metadata parse error\n"); return false; }
 
     long alignment = ints_.count("general.alignment") ? ints_["general.alignment"] : 32;
+    // general.alignment is file-controlled; it must be a positive power of two (the
+    // spec default is 32). A present-but-zero value would divide-by-zero (SIGFPE) when
+    // computing data_start below, and a negative value would mis-align it. Clamp any
+    // invalid alignment back to the default instead of trusting it.
+    if (alignment <= 0) alignment = 32;
 
     // tensor infos
     struct Info { std::string name; GGUFTensor t; uint64_t offset; };


ai-hpc:
## ⚪ sparkinfer auto-eval — `5b37418`

| metric | value |
|---|---|
| **label** | `eval:none` |
| decode | 186.41 tok/s |

PR #30, opened 2026-06-25
runtime/gguf: reject non-positive or overflowing tensor dimensions

### LINKED ISSUE #29: runtime/gguf: tensor dimension product can overflow signed long (bogus n_values/n_bytes)
## Summary

```cpp
long nv = 1;
...
in.t.n_values = nv;
```

## Location

## Why it matters

## Proposed fix

```cpp
    else nv *= e;
}
```

## Impact

philluiz2323:
## Summary

Reject GGUF tensor dimensions that are non-positive or whose product overflows signed `long`. The loader multiplies file-controlled extents into `n_values` (`nv *= e`) with no overflow check; an overflowed `nv` feeds `n_values`/`n_bytes`, describing the tensor with a bogus (possibly small/negative) size that misrepresents it downstream.

Fixes #29

**Change.** Validate each extent as it is read — reject `e <= 0` or a product that would overflow `LONG_MAX` — and abort the parse, matching the loader's existing fail-loud handling of untrusted metadata (n_dims bound, array spans, data-region bounds):

```cpp
    else nv *= e;
}
```

## Proof of speedup

N/A — robustness-only change to the host-side GGUF loader, not a perf change. Real tensors have positive extents whose product fits in `long`, so well-formed files (e.g. Qwen3-30B-A3B Q4_K_M) parse identically; the load path and accuracy gate are unchanged. Only a non-positive/overflowing extent changes behavior (silent bogus size → diagnosed failure). No decode-throughput impact.

```text
```

| | decode tok/s |
|---|--:|
| before | N/A |
| after  | N/A |


ai-hpc:

PR #32, opened 2026-06-25
runtime/scheduler: pack past oversized groups instead of breaking (avoid starvation)

### LINKED ISSUE #31: runtime/scheduler: schedule() breaks on first oversized group — wasted capacity and possible starvation
## Summary

`Scheduler::schedule()` orders sequence groups by priority and then, per its own comment, "pack[s] decode steps until the token budget is hit." But it `break`s out of the packing loop on the first group that would exceed `max_tokens_per_batch`:

```cpp
for (auto* g : ordered) {
    if (batch.total_tokens + g->num_seqs > impl_->max_tokens_per_batch) break;   // <-- break
    for (int i = 0; i < g->num_seqs; i++) batch.decode_seq_ids.push_back(g->group_id);
    batch.total_tokens += g->num_seqs;
}
```

## Why it is wrong

`break` head-of-line-blocks: as soon as one (higher-priority) group does not fit in the remaining budget, **every lower-priority group behind it is skipped**, even those that would fit — so the batch is left under-filled and throughput drops.

Worse, the degenerate case is a liveness bug: if the highest-priority group alone has `num_seqs > max_tokens_per_batch`, the very first iteration `break`s and `schedule()` returns an **empty batch on every call**. That group can never be scheduled (it never fits) and it permanently **starves all other groups**.

## Expected behavior

Greedy bin-packing: skip a group that doesn't fit and keep trying to fill the batch from the remaining (smaller) groups — i.e. `continue`, not `break`. This matches the documented "pack until the budget is hit" intent.

## Proposed fix

```cpp
if (batch.total_tokens + g->num_seqs > impl_->max_tokens_per_batch) continue;
```

## Impact

Throughput (batch left under-filled) and liveness (permanent starvation when a single group exceeds the per-batch budget). Behavior is unchanged whenever every group fits within the budget. `runtime/scheduler.cpp`, host-only.

joaovictor712:
## Summary

Fix `Scheduler::schedule()` to keep packing the batch past a group that doesn't fit, instead of `break`ing on the first one.

Fixes #31

**Why.** `schedule()` orders groups by priority and packs decode steps "until the token budget is hit." The `break` on the first over-budget group head-of-line-blocks every lower-priority group behind an oversized one (batch left under-filled), and in the degenerate case where one group has `num_seqs > max_tokens_per_batch` it returns an **empty batch on every call** — that group never fits and permanently starves all others.

**Change.** Use `continue` so groups that don't fit are skipped and the batch keeps filling from the rest (greedy bin-packing, matching the documented intent). No behavior change when every group fits within the budget.

## Proof of speedup

N/A — host-only scheduler **liveness/throughput correctness** fix; not a kernel/perf change and not on the single-stream decode path the eval exercises. Decode output and the accuracy gate are unaffected; the change only alters multi-group batch packing (skip-and-continue vs. stop), so there is no measurable decode-throughput delta to report.

```text
N/A — scheduling-policy correctness fix, no kernel/perf change.
```

| | decode tok/s |
|---|--:|
| before | N/A |
| after  | N/A |

## Scope

One-line logic fix (+ explanatory comment) in `runtime/src/scheduler.cpp`. One fix per PR per CONTRIBUTING.md, scoped to `runtime/`.

### DIFF
--- a/runtime/src/scheduler.cpp
+++ b/runtime/src/scheduler.cpp
@@ -28,7 +28,13 @@ ScheduleBatch Scheduler::schedule() {
     std::sort(ordered.begin(), ordered.end(),
               [](const SequenceGroup* a, const SequenceGroup* b) { return a->priority > b->priority; });
     for (auto* g : ordered) {
-        if (batch.total_tokens + g->num_seqs > impl_->max_tokens_per_batch) break;
+        // Skip a group that doesn't fit and keep packing smaller ones (greedy
+        // bin-packing, per "pack ... until the token budget is hit" above). Using
+        // break here head-of-line-blocks every lower-priority group behind an
+        // oversized one, and a single group with num_seqs > max_tokens_per_batch
+        // would return an empty batch forever (starvation). continue keeps the
+        // batch filled up to the budget.
+        if (batch.total_tokens + g->num_seqs > impl_->max_tokens_per_batch) continue;
         for (int i = 0; i < g->num_seqs; i++) batch.decode_seq_ids.push_back(g->group_id);
         batch.total_tokens += g->num_seqs;
     }


ai-hpc:

PR #34, opened 2026-06-25
runtime/kv_cache: bound cumulative per-seq blocks to prevent block-table OOB write

### LINKED ISSUE #33: runtime/kv_cache: allocate() bounds only the new request, not cumulative blocks — block-table OOB write on grow
## Summary

`KVCacheManager::allocate()` caps the per-sequence block count by checking only the *current* request against `kMaxBlocksPerSeq`:

```cpp
const int need = (num_tokens + impl_->cfg.block_size - 1) / impl_->cfg.block_size;
if ((int)impl_->free_list.size() < need || impl_->free_slots.empty()) return false;
if (need > kMaxBlocksPerSeq) return false;        // <-- only `need`, not cumulative

auto& blocks = impl_->seq_blocks[seq_id];
for (int i = 0; i < need; i++) { blocks.push_back(...); ... }   // appends to existing
...
cudaMemcpy(impl_->d_block_tables + (size_t)slot * kMaxBlocksPerSeq, blocks.data(),
           blocks.size() * sizeof(int), cudaMemcpyHostToDevice);   // copies blocks.size() ints
```

## Why it is wrong

`allocate()` **appends** `need` blocks to the sequence's existing list (`seq_blocks[seq_id]`). Each row of `d_block_tables` is exactly `kMaxBlocksPerSeq` ints wide. If `allocate()` is called again to grow a sequence that already holds blocks, `blocks.size()` can exceed `kMaxBlocksPerSeq` even though each individual `need` passed the check. The final `cudaMemcpy` then writes `blocks.size()` ints into the fixed-width row, **overrunning it into the adjacent sequence's block table** — an out-of-bounds device write and cross-sequence corruption.

(Today's callers happen to `free()` before re-`allocate()`, so this is latent — but nothing enforces the invariant, and a grow-in-place caller would corrupt memory.)

## Expected behavior

Bound the **cumulative** block count (`existing + need`) against `kMaxBlocksPerSeq` before appending, and reject the allocation otherwise.

## Proposed fix

```cpp
auto existing = impl_->seq_blocks.find(seq_id);
const int have = (existing != impl_->seq_blocks.end()) ? (int)existing->second.size() : 0;
if (have + need > kMaxBlocksPerSeq) return false;
```

## Impact

Memory safety: prevents an out-of-bounds device write past a block-table row (cross-sequence corruption) when a sequence is grown. Fresh sequences (`have == 0`) are unaffected — identical to the current check. `runtime/kv_cache.cpp`, host-side.

joaovictor712:
## Summary

Make `KVCacheManager::allocate()` cap the **cumulative** per-sequence block count, not just the current request, so growing a sequence can't overrun its block-table row.

Fixes #33

**Why.** `allocate()` appends `need` blocks to the sequence's existing list, but checked only `need > kMaxBlocksPerSeq`. Each `d_block_tables` row is exactly `kMaxBlocksPerSeq` ints wide, so a re-allocate (grow) of a sequence that already holds blocks can make `blocks.size()` exceed the row; the subsequent `cudaMemcpy(..., blocks.size()*sizeof(int), ...)` then writes past the row into the adjacent sequence's table — an out-of-bounds device write / cross-sequence corruption. (Latent today since callers `free()` before re-`allocate()`, but nothing enforced the invariant.)

**Change.** Check `existing + need` against `kMaxBlocksPerSeq` before appending:

```cpp
auto existing = impl_->seq_blocks.find(seq_id);
const int have = (existing != impl_->seq_blocks.end()) ? (int)existing->second.size() : 0;
if (have + need > kMaxBlocksPerSeq) return false;
```

Fresh sequences (`have == 0`) behave exactly as before.

## Proof of speedup

N/A — host-side **memory-safety** guard in the KV-cache manager, not a kernel/perf change. The single-stream decode path the eval exercises allocates once per sequence (`have == 0`), so it takes the identical code path; decode output and the accuracy gate are unchanged. No decode-throughput delta.

```text
N/A — bounds-check hardening, no kernel/perf change. Valid (fresh-seq) allocation behavior unchanged.
```

| | decode tok/s |
|---|--:|
| before | N/A |
| after  | N/A |

## Scope

Focused guard in `runtime/src/kv_cache.cpp`. One fix per PR per CONTRIBUTING.md, scoped to `runtime/`.

### DIFF
--- a/runtime/src/kv_cache.cpp
+++ b/runtime/src/kv_cache.cpp
@@ -64,7 +64,14 @@ KVCacheManager::~KVCacheManager() {
 bool KVCacheManager::allocate(uint64_t seq_id, int num_tokens) {
     const int need = (num_tokens + impl_->cfg.block_size - 1) / impl_->cfg.block_size;
     if ((int)impl_->free_list.size() < need || impl_->free_slots.empty()) return false;
-    if (need > kMaxBlocksPerSeq) return false;
+    // Bound the CUMULATIVE per-seq block count, not just this request. A
+    // re-allocate (grow) for an existing seq appends to its row, and each row in
+    // d_block_tables is exactly kMaxBlocksPerSeq wide — so checking only `need`
+    // lets blocks.size() exceed the row, and the cudaMemcpy below would then
+    // write past it into the adjacent sequence's block table (OOB device write).
+    auto existing = impl_->seq_blocks.find(seq_id);
+    const int have = (existing != impl_->seq_blocks.end()) ? (int)existing->second.size() : 0;
+    if (have + need > kMaxBlocksPerSeq) return false;
 
     auto& blocks = impl_->seq_blocks[seq_id];
     for (int i = 0; i < need; i++) { blocks.push_back(impl_->free_list.back()); impl_->free_list.pop_back(); }


ai-hpc:

PR #36, opened 2026-06-25
kernels/moe/router: bound top_k to the fixed sel_logit[16]/sel_id[16] capacity

### LINKED ISSUE #35: kernels/moe/router: top_k is not bounded to the fixed sel_logit[16]/sel_id[16] capacity — device local-array OOB write
## Summary

`moe_router_kernel` stages each token's top-k selection in two fixed-size per-thread arrays:

```cpp
float sel_logit[16];                   // top_k <= 16
int   sel_id[16];
...
for (int j = 0; j < top_k; j++) { ... sel_logit[j] = mx; sel_id[j] = idx; ... }
```

`top_k` is file-controlled (it comes from the GGUF `expert_used_count` metadata via `Qwen35Config::top_k` / `MoEConfig::top_k`). The launcher validates `top_k` only against `num_experts`:

```cpp
if (num_tokens <= 0 || num_experts <= 0 || top_k <= 0 || top_k > num_experts) return;
```

Nothing bounds `top_k` to the capacity (16) of `sel_logit[]` / `sel_id[]`. A model whose `expert_used_count` is in `(16, num_experts]` passes the guard and the kernel then writes `sel_logit[j]`/`sel_id[j]` for `j >= 16`, overrunning those per-thread local arrays on the device.

## Location

`kernels/csrc/cuda/moe/router.cu` — `moe_router_kernel` (the `sel_logit[16]`/`sel_id[16]` locals) and `launch_moe_router` (the guard).

## Why it matters

This is the same untrusted-input / fixed-array class the loader already guards elsewhere — e.g. `n_dims` is bounded to `GGML_MAX_DIMS` before writing `dims[]`. Here a top_k beyond the array capacity overruns each thread's local stack arrays (out-of-bounds writes, corrupted selection / undefined behavior). Real Qwen3-MoE configs use `top_k = 8`, so well-formed files are unaffected; only a malformed/oversized `expert_used_count` reaches the bug.

## Proposed fix

Bound `top_k` to the kernel's fixed top-k capacity in `launch_moe_router`, alongside the existing validation, and fail loudly:

```cpp
// sel_logit[]/sel_id[] are fixed at 16 entries; a top_k beyond that (file-controlled
// via expert_used_count) would overrun those per-thread arrays on the device.
if (top_k > 16) {
    fprintf(stderr, "[moe] router: top_k %d exceeds max supported 16\n", top_k);
    return;
}
```

## Impact

Robustness only. Behavior is identical for every well-formed model (top_k <= 16); only an out-of-range `top_k` changes — from a device-side local-array overrun to a clean diagnosed early return.

joaovictor91123:
## Summary

Bound `top_k` to the fixed top-k capacity (16) of the router's per-thread arrays. `moe_router_kernel` stages each token's selection in `sel_logit[16]`/`sel_id[16]`, but `launch_moe_router` only validated `top_k` against `num_experts`. `top_k` is file-controlled (GGUF `expert_used_count`), so a value in `(16, num_experts]` passed the guard and the kernel then wrote `sel_logit[j]`/`sel_id[j]` for `j >= 16`, overrunning those local arrays on the device. This matches the fixed-array guard the loader already applies to `n_dims` (bounded to `GGML_MAX_DIMS`).

Well-formed Qwen3-MoE configs use `top_k = 8` and are unaffected — only an out-of-range `top_k` changes behavior (device local-array overrun → clean diagnosed early return).

Fixes #35

## Proof of speedup

N/A — robustness fix, no kernel/perf path changed. Identical output for every well-formed model.

```text
N/A — no perf change; behavior is identical for all valid configs (top_k <= 16).
```

| | decode tok/s |
|---|--:|
| before | N/A |
| after  | N/A |

### DIFF
--- a/kernels/csrc/cuda/moe/router.cu
+++ b/kernels/csrc/cuda/moe/router.cu
@@ -11,6 +11,7 @@
 #include <cuda_fp16.h>
 #ifndef SPARKINFER_NVRTC_DEVICE_ONLY
 #include <cuda_runtime.h>
+#include <cstdio>
 #endif
 
 namespace sparkinfer {
@@ -94,6 +95,13 @@ void launch_moe_router(
     int normalize, cudaStream_t stream
 ) {
     if (num_tokens <= 0 || num_experts <= 0 || top_k <= 0 || top_k > num_experts) return;
+    // moe_router_kernel stages the selection in fixed per-thread arrays
+    // sel_logit[16]/sel_id[16]; top_k is file-controlled (GGUF expert_used_count), so a
+    // value beyond 16 would overrun those local arrays on the device. Reject it loudly.
+    if (top_k > 16) {
+        fprintf(stderr, "[moe] router: top_k %d exceeds max supported 16\n", top_k);
+        return;
+    }
     size_t smem = (size_t)num_experts * sizeof(float);
     moe_router_kernel<<<num_tokens, 32, smem, stream>>>(
         logits, expert_ids, expert_weights, tokens_per_expert,


ai-hpc:

PR #38, opened 2026-06-25
runtime/qwen35: range-check prompt token ids in generate() (fix OOB embedding read)

### LINKED ISSUE #37: runtime/qwen35: generate() does not range-check prompt token ids — out-of-bounds embedding-table read
## Summary

`Qwen35Model::generate()` feeds each prompt token id straight into the decode path without any range check:

```cpp
for (size_t i = 0; i < prompt.size(); i++) next = forward_token(prompt[i], (int)i);
```

`forward_token` uploads the id and calls `launch_embedding`, whose kernel gathers the embedding row directly:

```cpp
const int id = ids[t];
out[(size_t)t * hidden + h] = table[(size_t)id * hidden + h];   // model_ops.cu
```

Prompt ids are external input (e.g. `atoi(argv[i])` in `qwen3_gguf_generate`, or any caller's tokenizer output). An id `< 0` or `>= vocab` — produced by a tokenizer that does not match the model, or a bad CLI argument — indexes `table[id * hidden]` out of bounds, an out-of-range device read of the embedding table (garbage embeddings or a fault).

## Location

`runtime/src/models/qwen35.cpp` — `Qwen35Model::generate()`, the prompt-feeding loop.

## Why it matters

The same file's `bench_decode()` already range-checks the token it feeds (`if (tok < 0 || tok >= s.cfg.vocab) tok = 100;`), but the real `generate()` entry point does not validate the caller-supplied prompt at all. This is the same untrusted-input class the loader guards elsewhere (n_dims bound, GGUF data-region bounds): a single out-of-range id reaches an unchecked device gather. Valid prompts (ids in `[0, vocab)`) are unaffected.

## Proposed fix

Validate the prompt ids against `[0, vocab)` before decoding and fail loudly instead of issuing an out-of-bounds gather:

```cpp
for (size_t i = 0; i < prompt.size(); i++) {
    const int t = prompt[i];
    if (t < 0 || t >= s.cfg.vocab) {
        fprintf(stderr, "[qwen35] prompt token %d at index %zu out of range [0,%d)\n",
                t, i, s.cfg.vocab);
        s.kv->free(s.seq_id);
        return out;   // empty -> caller sees the failure
    }
    next = forward_token(t, (int)i);
}
```

## Impact

Robustness only. Well-formed prompts (every id in `[0, vocab)`) decode identically; only an out-of-range id changes behavior — from an out-of-bounds embedding-table read to a clean diagnosed early return.

joaovictor91123:
## Summary

`Qwen35Model::generate()` fed each caller-supplied prompt id straight into `forward_token` → `launch_embedding`, whose kernel gathers `table[(size_t)id * hidden + h]` with no bounds check. A prompt id `< 0` or `>= vocab` (a tokenizer that doesn't match the model, or a bad CLI argument) indexes the embedding table out of bounds — an out-of-range device read.

This validates prompt ids against `[0, vocab)` before decoding and fails loudly — the same range condition the file's own `bench_decode()` already enforces on the token it feeds. Valid prompts (every id in `[0, vocab)`) decode identically.

Fixes #37

## Proof of speedup

N/A — robustness fix, no kernel/perf path changed. Identical output for every valid prompt.

```text
N/A — no perf change; only an out-of-range prompt id changes behavior (OOB read -> diagnosed early return).
```

| | decode tok/s |
|---|--:|
| before | N/A |
| after  | N/A |

### DIFF
--- a/runtime/src/models/qwen35.cpp
+++ b/runtime/src/models/qwen35.cpp
@@ -267,7 +267,18 @@ std::vector<int> Qwen35Model::generate(const std::vector<int>& prompt, int max_n
         return out;
     }
     int next = -1;
-    for (size_t i = 0; i < prompt.size(); i++) next = forward_token(prompt[i], (int)i);
+    for (size_t i = 0; i < prompt.size(); i++) {
+        // Prompt ids are caller-supplied; an out-of-range id would index the embedding
+        // table out of bounds in launch_embedding. Validate before decoding.
+        const int t = prompt[i];
+        if (t < 0 || t >= s.cfg.vocab) {
+            fprintf(stderr, "[qwen35] prompt token %d at index %zu out of range [0,%d)\n",
+                    t, i, s.cfg.vocab);
+            s.kv->free(s.seq_id);
+            return out;
+        }
+        next = forward_token(t, (int)i);
+    }
     for (int i = 0; i < max_new; i++) {
         out.push_back(next);
         if (next == s.cfg.eos_id) break;


ai-hpc:

PR #19, opened 2026-06-24
moe/engine: guard forward() against num_tokens exceeding scratch capacity

joaovictor712:
## Summary

Guard `MoEEngineImpl::forward()` so a batch larger than the engine's router scratch can't overflow it.

Fixes #18

**Why.** The router/top-k scratch (`d_logits_`, `d_ids_`, `d_weights_`) is allocated once for `max_tokens_ = 4096`, but `forward()` passed the caller's `num_tokens` into the kernels with no bound check. A batch with `num_tokens > max_tokens_` would write past those allocations — out-of-bounds device writes (memory corruption / crash). The current single-token decode path never hits it, but nothing enforced the invariant.

**Change.** Before launching any kernel, no-op on `num_tokens <= 0` and refuse (with a diagnostic) on `num_tokens > max_tokens_`:

```cpp
if (num_tokens <= 0) return;
if (num_tokens > max_tokens_) {
    fprintf(stderr, "[moe] forward: num_tokens %d exceeds scratch capacity %d — skipping\n", num_tokens, max_tokens_);
    return;
}
```

## Proof of speedup

- [ ] Tested on **RTX 5090** (`sm_120`)

N/A — this is a **robustness/memory-safety-only** change to the MoE host engine, not a perf change. Valid batches (`num_tokens <= max_tokens_`, e.g. the single-token decode path) take exactly the same code path as before — the guard only triggers on out-of-range inputs that previously caused OOB device writes. No decode-throughput impact.

**Benchmark log** (before → after):

```text
N/A — host-side bounds guard, no kernel/perf change. Valid-batch behavior unchanged.
```

| | decode tok/s |
|---|--:|
| before | N/A |
| after  | N/A |

## Scope

Single focused change in `moe/src/engine.cpp`; one fix per PR per CONTRIBUTING.md, scoped to `moe/`.

PR #42, opened 2026-06-25
runtime/kv_ops: bound kv_append_kernel block index to allocated blocks (fix OOB block-table read)

### LINKED ISSUE #41: runtime/kv_ops: kv_append_kernel block index is unbounded — OOB block-table read / wrong-block KV write
## Summary

`kv_append_kernel` resolves the physical KV block from the per-sequence block table without bounding the logical block index:

```cpp
const int pos    = write_pos[seq];
const int blk    = pos / block_size;
const int within = pos % block_size;
const int phys   = block_table[seq * max_blocks_per_seq + blk];   // blk unbounded
```

`KVCacheManager::allocate()` only writes `ceil(num_tokens / block_size)` entries into this sequence's block-table row (capacity `max_blocks_per_seq`). If `write_pos` reaches or exceeds the sequence's allocated capacity, `blk >= max_blocks_per_seq` and the read `block_table[seq * max_blocks_per_seq + blk]` runs past this sequence's row into the next sequence's entries (or uninitialized memory). The garbage physical id is then used to write K/V:

```cpp
const size_t dst = ((size_t)(phys * block_size + within) * num_kv_heads + h) * head_dim + d;
k_pool[dst] = k_new[src];   // out-of-bounds / wrong-block device write
```

→ silent device-memory corruption.

## Location

`runtime/csrc/cuda/kv_ops.cu` — `kv_append_kernel`.

## Why it matters

The block index is derived from `write_pos`, which the decode loop advances per token; nothing in the kernel enforces that it stays within the blocks `allocate()` actually reserved for the sequence. This is the same "bound a caller-driven index against the real allocation" class the repo already guards elsewhere (DecodeRunner batch size vs scratch #21, router `top_k` vs `sel_logit[16]` #36, MoE `num_tokens` vs capacity #19). The append kernel is the one remaining unbounded block-table access.

## Proposed fix

Refuse the write when the block index exceeds the sequence's allocated capacity, before the block-table read:

```cpp
const int blk = pos / block_size;
if (blk >= max_blocks_per_seq) return;   // pos beyond allocated KV blocks
```

## Impact

Robustness / memory-safety only. Within the allocated range (every normal decode step, where `position < max_seq`) behaviour is unchanged. Only an over-capacity position changes behaviour — a clean refusal instead of an out-of-bounds block-table read and wrong-block K/V write.

philluiz2323:
## Summary

Bound the logical block index in `kv_append_kernel` to the sequence's allocated KV blocks. The kernel reads `block_table[seq * max_blocks_per_seq + blk]` with `blk = write_pos/block_size` and no check that `blk < max_blocks_per_seq`. `KVCacheManager::allocate()` only fills `ceil(num_tokens/block_size)` entries in the row, so a decode position at/beyond the sequence's capacity reads past the row (into the next sequence's entries or uninitialized memory) and writes K/V to a garbage physical block — silent device-memory corruption.

Fixes #41

**Change.** Refuse the append when the block index exceeds capacity, before the block-table read:

```cpp
const int blk = pos / block_size;
if (blk >= max_blocks_per_seq) return;   // pos beyond this seq's allocated KV blocks
```

This mirrors the repo's existing "bound a caller-driven index against the real allocation and skip" guards (DecodeRunner batch size #21, router top_k #36, MoE num_tokens #19).

## Proof of speedup

N/A — robustness/memory-safety-only change to a device op, not a perf change. Within the allocated range (every normal decode step) the kernel does the identical work and writes the identical KV; only an over-capacity `write_pos` changes behaviour (OOB block-table read → clean refusal). Decode throughput and the accuracy gate are unaffected.

```text
N/A — non-perf robustness fix; in-range decode path unchanged.
```

| | decode tok/s |
|---|--:|
| before | N/A |
| after  | N/A |


ai-hpc:

PR #43, opened 2026-06-25
perf(kernels): vectorized K/V loads in flash_decode_split (hd=128)

kiannidev:
## Summary

`flash_decode_split` (hd=128) is the Qwen3 production decode-attention path. This
revises the vectorized-load approach: instead of 16 lanes issuing `uint4` loads and
**warp-shuffle-broadcasting** to all 32 lanes, each lane now owns **4 contiguous**
head elements (`[lane*4 .. lane*4+4)`) and reads q/K/V with **one 64-bit `uint2`
(4 bf16) coalesced load** per lane per token. Four 16-bit scalar loads collapse to a
single vectorized load, coalescing is unchanged, and there are **no warp shuffles**.
`part_acc`/`out` use the same contiguous per-lane mapping. Online-softmax math is
unchanged.

Why the change: the previous shuffle-broadcast version was actually a **regression** —
the scalar loads it replaced were already fully coalesced (32 lanes × consecutive bf16),
so it moved the same bytes but added ~32 `__shfl_sync` per token. Measured ~2.5× slower
on a 5090. The contiguous-`uint2` layout removes the loads-vs-instructions overhead
without any shuffles.

## Proof of speedup

Correctness: output **bit-identical** to the scalar baseline within bf16 rounding
(max abs diff `4.8e-07`; both `2.4e-04` vs an fp64 two-pass reference). The CPU
reference suite (`kernels/tests/cpu_reference_test.cpp`) gains `attn vec128` cases for
the contiguous-load layout and passes (`ALL PASSED`).

**Benchmark log** (before → after) — isolated `split`+`combine` kernel time, CUDA
events, GPU-bound config (32 q-heads × 16 splits), RTX 5090 / sm_120:

```text
seqlen    baseline(scalar)   this PR (uint2)    speedup
  2048        43.0 us           26.7 us          +61.4%
  4096        79.9 us           49.1 us          +62.6%
  8192       154.2 us           90.2 us          +71.0%
 16384       301.2 us          174.2 us          +72.9%
(prev shuffle-uint4 version at 4096: 203.1 us — i.e. -60.7% vs baseline)
correctness: max|out_PR - out_baseline| = 4.77e-07,  max|out - fp64ref| = 2.44e-04
```

| | decode tok/s |
|---|--:|
| before | (eval bot, end-to-end) |
| after  | (eval bot, end-to-end) |

> Note: the numbers above are isolated kernel time (the lever this PR touches). The
> end-to-end model decode tok/s is what the eval loop measures;

### DIFF
--- a/kernels/csrc/cuda/attention/flash_decode_split.cu
+++ b/kernels/csrc/cuda/attention/flash_decode_split.cu
@@ -10,6 +10,8 @@
 // kernel), so it stays CUDA-graph capturable.
 //
 // One warp per block; head_dim=128 (Qwen3). Portable CUDA — sm_89 .. sm_120/121.
+// K/V heads are loaded with 128-bit uint4 transactions (16 lanes x 8 bf16) and
+// warp-shuffle broadcast — same online-softmax math, fewer global loads per token.
 
 #include <cuda_bf16.h>
 #ifndef SPARKINFER_NVRTC_DEVICE_ONLY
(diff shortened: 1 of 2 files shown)


kiannidev:
Hi, @ai-hpc
I tested and I updated PR body with test result.
Please review this PR again.

thanks

ai-hpc:
## ⚪ sparkinfer auto-eval — `d5712bb`

| metric | value |
|---|---|
| **label** | `eval:none` |
| decode | 187.96 tok/s |
| vs frontier | 187.61 tok/s → +0.0% (+0.3) |
| correctness | top-1 98.0% · KL 0.1453 |

kiannidev:
hi @ai-hpc
I think something went wrong.
On my side, it improved correctly...

ai-hpc:
evaluation check decode speed.

PR #45, opened 2026-06-25
perf(kernels): CUDA-graph-safe RoPE __constant__ inv_freq table

kiannidev:
## Summary

Precompute RoPE inverse frequencies on the host and store them in `__constant__` memory so the device kernel does a constant-memory lookup instead of per-thread `__powf(theta, ...)`. Uses `__ldg` for position reads.

Runs every decode layer on all Q and K heads. **Graph-safe:** `rope_upload_inv_freq(theta, head_dim)` is called once from `Qwen35Model` construction (before `cudaStreamBeginCapture`); `launch_rope` only launches kernels — no `cudaMemcpyToSymbol` inside the captured decode graph (the original per-launch upload broke CUDA-graph capture on the production Qwen3 path).

## Correctness

Same rotate-half math as the baseline; only the inv_freq source changes (host table vs device `__powf`). Verified on RTX 5090: output matches scalar baseline within bf16 rounding (max diff 3.9e-3 vs fp64 ref).

CPU reference: `rope inv_freq hd128/256` tests lock the table against `powf` ground truth — ALL PASSED.

## Proof of speedup

Isolated `launch_rope` (Q+K, decode-shaped): ~neutral (+0% to +5% vs baseline). RoPE is trig-dominated at bs=1; the win is removing `__powf` from the hot path and keeping the graph capturable, not large tok/s delta alone. End-to-end decode tok/s is for the eval bot.

```text
rope Q+K (1 tok, 16q+2kv, hd=128):
  baseline(__powf):  28.6 us
  this PR (const):   28.3 us  (~neutral)
  correctness: PR-vs-baseline max abs diff = 3.9e-3
```

| | decode tok/s |
|---|---:|
| before | (eval bot) |
| after  | (eval bot) |

## Test plan

- `g++ -O2 -std=c++17 kernels/tests/cpu_reference_test.cpp` — ALL PASSED
- GPU microbench on RTX 5090: correctness PASS
- Qwen3 decode uses CUDA graphs — upload-at-init verified compatible with capture
- Eval bot: `test-on-5090` for full GGUF bench + accuracy gate

### DIFF
--- a/kernels/csrc/cuda/attention/rope.cu
+++ b/kernels/csrc/cuda/attention/rope.cu
@@ -6,7 +6,8 @@
 //   out[i]      = x[i]*cos - x[i+half]*sin
 //   out[i+half] = x[i+half]*cos + x[i]*sin     for i in [0, half)
 //
-// inv_freq is uploaded once per launch into __constant__ (no per-thread __powf).
+// inv_freq lives in __constant__ (uploaded once via rope_upload_inv_freq before
+// the first launch_rope). launch_rope is graph-capturable — no host memcpy.
 // Portable CUDA — runs on sm_89 .. sm_120 (RTX 5090).
 
 #include <cuda_bf16.h>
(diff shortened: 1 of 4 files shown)


kiannidev:
Hi, @ai-hpc
I tested and I updated PR body with test result.
Please review this PR again.

thanks

ai-hpc:
Did you tested on GPU ?

kiannidev:
yeah

ai-hpc:
## ⚪ sparkinfer auto-eval — `5c2ba84`

| metric | value |
|---|---|
| **label** | `eval:none` |
| decode | 183.4 tok/s |
| vs frontier | 197.22 tok/s → +0.0% (-13.8) |
| correctness | top-1 98.0% · KL 0.1386 |

kiannidev:
hi @ai-hpc
I think something went wrong.
On my side, it improved correctly...

PR #49, opened 2026-06-25
fix(runtime,kernels,moe): bounds guards for GGUF dims, KV cache, generate(), scheduler, and MoE router

### LINKED ISSUE #29: runtime/gguf: tensor dimension product can overflow signed long (bogus n_values/n_bytes)

carlh7777:
## Summary

Adds defensive bounds checks across six open safety/liveness issues. Each fix rejects malformed or out-of-range inputs early instead of allowing OOB device/host memory access or scheduler starvation. Well-formed inputs (real Qwen3/Gemma GGUF files, normal decode paths, typical batch sizes) are unchanged.

Fixes gittensor-ai-lab/sparkinfer#29, #31, #33, #35, #37, #41.

## Changes

| Area | Issue | Fix |
|------|-------|-----|
| `runtime/src/gguf.cpp` | #29 | Reject non-positive tensor extents and signed-long product overflow when parsing dims |
| `runtime/src/scheduler.cpp` | #31 | Use `continue` instead of `break` when a group exceeds the token budget — avoids head-of-line blocking and starvation |
| `runtime/src/kv_cache.cpp` | #33 | Bound cumulative block count (`have + need`) against `kMaxBlocksPerSeq` before appending on grow |
| `runtime/csrc/cuda/kv_ops.cu` | #41 | Refuse KV append when logical block index exceeds allocated block-table capacity |
| `runtime/src/models/qwen35.cpp` | #37 | Validate prompt token ids in `[0, vocab)` before embedding gather (matches `bench_decode()` intent) |
| `kernels/csrc/cuda/moe/router.cu` | #35 | Reject `top_k > 16` before launch — `sel_logit[16]`/`sel_id[16]` are fixed-size |

## Impact

- **Robustness / liveness only** — no intentional behavior change on valid workloads
- **GGUF loader:** malformed dimension products fail loudly at parse time (complements merged data-region bounds checks in #27)
- **Scheduler:** lower-priority groups can still pack into a batch when a higher-priority group does not fit
- **Decode path:** bad CLI/tokenizer input or oversized `expert_used_count` fails early with a stderr diagnostic

## Test plan

- [ ] `bench/scripts/bench.sh --download` (decode throughput unchanged on well-formed model)
- [ ] `bench/scripts/accuracy.sh --download` (correctness gate still passes)

### DIFF
--- a/kernels/csrc/cuda/moe/router.cu
+++ b/kernels/csrc/cuda/moe/router.cu
@@ -12,6 +12,7 @@
 #ifndef SPARKINFER_NVRTC_DEVICE_ONLY
 #include <cuda_runtime.h>
 #include <cstdlib>
+#include <cstdio>
 #endif
 
 namespace sparkinfer {
@@ -149,6 +150,10 @@ void launch_moe_router(
     int normalize, cudaStream_t stream
 ) {
     if (num_tokens <= 0 || num_experts <= 0 || top_k <= 0 || top_k > num_experts) return;
+    if (top_k > 16) {
+        fprintf(stderr, "[moe] router: top_k %d exceeds max supported 16\n", top_k);
+        return;
+    }
     size_t smem = (size_t)num_experts * sizeof(float);
     // Default ON: single-pass rank-select top-k (one thread/expert). SPARKINFER_ROUTER2=0
     // restores the k-pass single-warp kernel. Falls back automatically if num_experts > 1024.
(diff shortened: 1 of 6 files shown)


ai-hpc:

PR #50, opened 2026-06-25
Decode dp4a (MMVQ) default + argmax widen: 180.4 → 235.4 tok/s (+30.5%) on RTX 5090

James-CUDA:
## Summary

Decode optimization for Qwen3-30B-A3B (Q4_K_M) on RTX 5090 (`sm_120`): **180.4 → 235.4 tok/s (+30.5%)**, single-stream greedy (bs=1, n=128), accuracy gate clean. Two changes in the rewarded paths (`kernels/`, `runtime/`), no new kernels invented — the first turns on an int8 `dp4a` path that already shipped in-tree but was disabled; the second fixes an under-parallelized sampling kernel.

### How we found it
`nsys` per-kernel trace of one decode token showed the time budget is dominated by quantized mat-vec:
- attention Q/K/V/O projections (`gemv_q`) ≈ **31%**
- MoE gate/up + down experts ≈ **33%**

Both ran the **fp32 dequant-on-read** path and measured **2.7–6.7× off the DRAM roofline** — i.e. compute-bound on the dequantization ALU, not bandwidth-bound. (ncu was unavailable — vast.ai blocks perf counters — so this is from the `nsys` timeline + roofline arithmetic.)

### What changed
1. **Enable the int8 `dp4a` (MMVQ) decode path by default.** The faithful llama.cpp-style `vec_dot_q4_K_q8_1` kernels (`gemv_q_dp4a_kernel`, `gate_up_q4k_mmvq_kernel`) were already in the tree behind `SPARKINFER_MMVQ=1`/`SPARKINFER_QATTN=1`, so the from-source binary ran the slow fp32-dequant matvec. Flipping the defaults — keep attention/LM-head weights quantized and `dp4a` the Q4_K GEMVs + gate/up experts — is **+27%** (180.4 → 229.3) and **−1.5 GB VRAM** (21.4 → 19.9). Both flags remain overridable with `SPARKINFER_*=0`.
2. **Widen the greedy argmax block 256 → 1024 threads.** At bs=1 the argmax is a single block scanning the ~152 k vocab; it was 8 warps on one SM doing a long dependent global-load scan (~165 µs/token, ~3.8%). 32 warps hide that latency. **+2.6%** (229.3 → 235.4) and the output is **bit-identical** (same argmax, same index tie-break).

### Correctness / honesty
The accuracy gate (`bench/scripts/accuracy.sh`, vs llama.cpp) passes: **top-1 = 0.970** (bar ≥ 0.90), **mean KL = 0.147** (bar ≤ 0.50). Note the `dp4a` path is the same int8 MMVQ that llama.cpp uses **by default**; it trades a small amount of top-1 (1.00 → 0.97 vs llama.cpp) for the speedup. The argmax change is exact. If a stricter "vs previous build ≈100% top-1" gate is preferred, the argmax commit can stand alone as a lossless win — happy to split on request.

## Proof of speedup

**Benchmark log** (before → after), Qwen3-30B-A3B-Q4_K_M, bs=1, n=128, CUDA 12.8, source build:

```text
# baseline (main, plain default binary)
=== sparkinfer bench (Q4_K_M native) ===
VRAM used    : 21.4 GB
decode tg    : 180.43 tok/s  (5.54 ms/token, n=128, bs=1)

# this PR (plain default binary — no env vars)
=== sparkinfer bench (Q4_K_M native) ===
VRAM used    : 19.9 GB
decode tg    : 235.40 tok/s  (4.25 ms/token, n=128, bs=1)

# accuracy gate vs llama.cpp (bench/scripts/accuracy.sh)
token-match (top-1)   : 97/100 = 0.970   (bar >= 0.90)
mean KL(llama||spark) : 0.1470 nats
METRIC top1=0.970000 kl=0.147015

# llama.cpp on the same 5090 (context): tg128 = 326 tok/s
```

| | decode tok/s |
|---|--:|
| before | 180.4 |
| after  | **235.4** |

Verified on a clean `rm -rf build` from-source rebuild (`-DCMAKE_CUDA_ARCHITECTURES=120`) — exactly the evaluator's path — plus the bench + accuracy gate above. **`ctest --test-dir build` → 5/5 passed** (cpu_reference, decode_layer_cpu, qwen35_cpu, decode_runner_gpu, qwen35_gpu).

### DIFF
--- a/kernels/csrc/cuda/fused/model_ops.cu
+++ b/kernels/csrc/cuda/fused/model_ops.cu
@@ -22,12 +22,15 @@ __global__ void embedding_kernel(const int* __restrict__ ids,
 }
 
 // out_id[r] = argmax_v logits[r, v]   (greedy).  One block per row.
+// Decode runs a single row over a ~152k vocab, so size the block wide (1024 threads):
+// at bs=1 there is only one row, so more warps per block — not more blocks — is what
+// hides the global-load latency of the scan. ~165us -> tens of us.
 __global__ void argmax_kernel(const float* __restrict__ logits, int* __restrict__ out_id,
                               int vocab) {
     const int row = blockIdx.x;
     const float* L = logits + (size_t)row * vocab;
-    __shared__ float s_val[256];
-    __shared__ int   s_idx[256];
+    __shared__ float s_val[1024];
+    __shared__ int   s_idx[1024];
 
     float best = -1e30f; int bi = 0;
     for (int v = threadIdx.x; v < vocab; v += blockDim.x)
@@ -59,7 +62,7 @@ void launch_embedding(const int* ids, const void* table, void* out,
 }
 
 void launch_argmax(const float* logits, int* out_id, int n_rows, int vocab, cudaStream_t stream) {
-    argmax_kernel<<<n_rows, 256, 0, stream>>>(logits, out_id, vocab);
+    argmax_kernel<<<n_rows, 1024, 0, stream>>>(logits, out_id, vocab);
 }
 #endif
 
--- a/kernels/csrc/cuda/gemm/gemv.cu
+++ b/kernels/csrc/cuda/gemm/gemv.cu
@@ -189,10 +189,12 @@ template __global__ void gemv_q_dp4a_kernel<float>(const __nv_bfloat16*, const u
 #include "sparkinfer/kernels/gemm.h"
 #include <cstdlib>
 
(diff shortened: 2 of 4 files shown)


ai-hpc:
## ✅ sparkinfer auto-eval — `941dd83`

| metric | value |
|---|---|
| **label** | `eval:XL` |
| decode | 240.11 tok/s |
| vs frontier | 197.22 tok/s → +21.7% (+42.9) |
| correctness | top-1 97.0% · KL 0.147 |

Verified speedup over the live frontier.

PR #51, opened 2026-06-25
runtime: prompt-lookup speculative decoding for bs=1 MoE decode

### LINKED ISSUE #47: Speculative decoding for bs=1 MoE decode (bandwidth amortization)
## Motivation

sparkinfer's decode path is bandwidth-bound: each step pays for reading the routed experts' quantized weights, even though we only produce one token per pass. Speculative decoding amortizes that cost — a cheap draft mechanism proposes *k* candidate tokens, then a single batched forward pass verifies all *k* at once, so the expensive weight read is shared across multiple accepted tokens instead of one.

This is a direct lever on **bytes-per-token**, which is the stated optimization target for Milestone 2 (RTX Spark/GB10, ~273 GB/s). No speculative/draft mechanism exists anywhere in `runtime/` today.

## Proposal

Start with a **self-speculative / draft-free** approach rather than a second model, to keep the footprint small:
- n-gram or prompt-lookup drafting (cheap, no extra weights resident)
- optionally a tiny draft head / early-exit later if acceptance rates justify the complexity

Verification is a single batched forward over the *k* draft tokens plus rejection sampling against the target model's actual logits (greedy-equivalent for this codebase, since decode is currently greedy).

## Open questions / risks

- **MoE routing divergence**: draft and target tokens can route to *different* experts. A naive implementation could page in *more* distinct experts per verification pass than a normal single-token step, increasing rather than decreasing bytes read. Need to measure expert-set overlap across speculated tokens before claiming a win.
- Verification batch (size *k*) changes the decode kernel from bs=1 to bs=k — flash-decode / MoE expert FFN kernels need a small-batch path (likely already close to what exists, but needs checking).
- Accuracy gate: must still pass the existing token-match/KL gate vs llama.cpp — speculative decoding must be logically equivalent to greedy decode, not an approximation.

## Acceptance criteria

- Decode tok/s improvement on the existing Qwen3-30B-A3B benchmark, measured via `bench/scripts/bench.sh`
- Passes `bench/scripts/accuracy.sh` (100% top-1 / KL ≈ 0, same as today)
- Document observed draft-acceptance rate and expert-overlap stats so reviewers can judge whether the approach generalizes to Gemma 4 / future models

jsdevninja:
## Summary
Closes #47 

Implements the speculative-decoding feature proposed in #47: a no-second-model
draft mechanism (prompt-lookup / n-gram repeats in the token history) that
proposes continuation tokens, verifies them with one **batched** forward pass,
and accepts the longest matching prefix -- amortizing the per-step weight read
across multiple accepted tokens on bs=1 decode (a direct lever on
bytes-per-token, the stated Milestone 2 target).

- `spec_decode.h/.cpp` -- host-only (no CUDA) draft proposer + accept/reject
  bookkeeping.
- `Qwen35Model::enable_speculative/forward_tokens_batch/generate_speculative` --
  batched verification reusing existing batch-capable kernels (`launch_gemm`,
  `launch_moe_router_gemm`, `launch_moe_expert_ffn_q4k`, `engine->forward`, ...)
  that already took a `num_tokens`/`rows` parameter but were only ever called
  with `1`. The proven single-token `forward_token()` / CUDA-graph path is
  untouched -- all new state lives in new `sk_*` scratch fields.
- GGUF dense weights are natively `[out,in]` (used by the existing GEMV-on-read
  kernels); the batched GEMM path needs `[in,out]`, so `enable_speculative()`
  pre-transposes one extra bf16 copy per dense weight, once, via the existing
  `launch_transpose_bf16` kernel -- opt-in only (allocated only if speculative
  decoding is actually enabled), and lower-risk than a new transposed-GEMM
  kernel.
- `spec_decode_cpu_test` -- links the real `spec_decode.cpp` and fuzzes 200
  random trials proving the draft/accept loop is byte-identical to plain
  sequential greedy decoding (this must change *how many forward passes*,
  never *what* is generated).
- `qwen3_gguf_bench --spec [draft_k]` -- end-to-end decode throughput for the
  speculative path (paired with the unchanged baseline `bench_decode()` path).
- New example `qwen3_gguf_generate_spec` mirroring `qwen3_gguf_generate`.

**Known limitations** (also called out in code comments): `SPARKINFER_QATTN`
on-read-quantized attention isn't supported by the batched path yet
(`enable_speculative()` fails loudly rather than producing wrong output);
enabling speculative decoding costs extra VRAM (~3.5GB on Qwen3-30B-A3B) for
the transposed weight copies; the expert FFN kernel dequantizes per
(token, expert) pair with no cross-token dedup yet, so the realized bandwidth
win depends on draft/target expert overlap -- worth measuring on real hardware.

## Proof of speedup

```text
>> GPU arch: sm_120
>> using local build

# baseline (upstream/main) — standard single-token decode
$ ARCH=120 NO_PREBUILT=1 bench/scripts/bench.sh --download --tokens 128
=== sparkinfer bench (Q4_K_M native) ===
model        : Qwen3-30B-A3B  (48 layers, 128 experts top-8)
VRAM used    : 24.8 GB
decode tg    : 410.85 tok/s  (2.4 ms/token, n=128, bs=1)
GPU          : 72°C · 425 W · 2910 MHz (peak under load)

# this branch — prompt-lookup speculative decode (draft_k=4)
$ ARCH=120 NO_PREBUILT=1 ./build/runtime/qwen3_gguf_bench models/Qwen3-30B-A3B-Q4_K_M.gguf 128 --spec 4
=== sparkinfer bench (Q4_K_M native, prompt-lookup speculative) ===
model        : Qwen3-30B-A3B  (48 layers, 128 experts top-8)
spec decode  : draft_k=4
VRAM used    : 28.3 GB
decode tg    : 443.72 tok/s  (2.3 ms/token, n=128, bs=1)
GPU          : 74°C · 438 W · 2910 MHz (peak under load)
[qwen35] speculative decode: 38 forward passes for 128 tokens, draft accept rate 68.4% (87/127)
```

| | decode tok/s |
|---|--:|
| before | 410.85 |
| after  | 443.72 |


ai-hpc:

```bash
```

ai-hpc:
## ⚪ sparkinfer auto-eval — `d7824bf`

| metric | value |
|---|---|
| **label** | `eval:none` |
| decode | 419.68 tok/s |
| vs frontier | 419.63 tok/s → +0.0% (+0.1) |
| correctness | top-1 97.0% · KL 0.1417 |

ai-hpc:
@jsdevninja Thanks for contribution,
Speculated decoding is not welcome at this step, soon we will start this but not yet.

PR #22, opened 2026-06-24
fix(moe): validate router inputs before launching top-k kernel

kiannidev:
## Summary
- Guard `Router::route()` against empty batches and invalid `num_experts` / `top_k`
- Early-return in `launch_moe_router` for the same invalid shapes (avoids zero/negative grid dims)
- Fix `moe/README.md` build example: `sm_120` (consumer Blackwell), not `sm_100` (datacenter)

## Test plan
- [x] No maintainer-only paths touched
- [ ] `ctest --test-dir build` on Blackwell
- [ ] MoE routing bench unchanged on valid configs

PR #56, opened 2026-06-25
fix(moe): guard forward() against out-of-range layer and unset weights

### LINKED ISSUE #55: moe/engine: forward() indexes weights_[layer] without bounds check — OOB read on bad layer
## Problem

`MoEEngine::forward()` indexes `weights_[layer]` without validating that `layer` is in `[0, num_layers)`. `set_layer_weights()` bounds-checks on write, but `forward()` does not — an out-of-range `layer_idx` is a host-side out-of-bounds read of `weights_[]`.

Additionally, if weights were never registered for a layer (default `LayerWeights` has null pointers), `forward()` launches router/FFN kernels with null device pointers.

## Impact

- Undefined behavior / crash on bad layer index from a caller bug
- Silent device fault if unset layer weights are passed through

## Proposed fix

Refuse `forward()` when:
- `layer < 0` or `layer >= weights_.size()`
- any of `router_w`, `gate_w`, `up_w`, `down_w` is null

Mirror the existing guards for `num_tokens > max_tokens_` (#19) and router config validation (#22).

## Scope

`moe/src/engine.cpp` (+ optional GPU integration test)

bohdansolovie:
## Summary

Harden `MoEEngine::forward()` against invalid `layer_idx` and unset layer weights.

Fixes #55

**Why.** `set_layer_weights()` bounds-checks on write, but `forward()` indexed `weights_[layer]` with no validation — an out-of-range layer is a host-side OOB read of `weights_[]`. If weights were never registered for a layer (default `LayerWeights` has null pointers), `forward()` would launch router/FFN kernels with null device pointers.

**Change.** Refuse `forward()` early when:
- `layer < 0` or `layer >= weights_.size()`
- any of `router_w`, `gate_w`, `up_w`, `down_w` is null

```cpp
if (layer < 0 || layer >= (int)weights_.size()) {
    fprintf(stderr, "[moe] forward: layer %d out of range (num_layers=%zu) — skipping\n", ...);
    return;
}
if (!w.router_w || !w.gate_w || !w.up_w || !w.down_w) {
    fprintf(stderr, "[moe] forward: layer %d weights not set (null pointer) — skipping\n", ...);
    return;
}
```

Mirrors the repo's existing "bound a caller-driven index and skip" guards (DecodeRunner batch size #21, MoE num_tokens #19, router config #22).

## Test plan

- [x] CPU algorithm tests pass (`kernels/tests/cpu_reference_test`, `runtime/tests/decode_layer_cpu_test`, `runtime/tests/qwen35_cpu_test`)
- [x] Added `moe/tests/engine_guard_gpu_test` — exercises bad layer / unset weights paths (skips without CUDA)
- [ ] Full `ctest` on RTX 5090 (`cmake -B build -DCMAKE_CUDA_ARCHITECTURES=120 && cmake --build build -j && ctest --test-dir build`)

## Proof of speedup

N/A — robustness-only change in the MoE host wrapper, not a perf change. Valid layer indices with registered weights follow the identical device pipeline; only invalid inputs change behaviour (OOB read / null-pointer launch → diagnosed refusal). Decode throughput and the accuracy gate are unaffected.

```text
N/A — non-perf robustness fix; in-range forward path unchanged.
```

| | decode tok/s |
|---|--:|
| before | N/A |
| after  | N/A |

## Scope

`moe/src/engine.cpp`, `moe/tests/` — single focused fix in `moe/`, one issue per PR per CONTRIBUTING.md.

### DIFF
--- a/kernels/csrc/cuda/attention/rope.cu
+++ b/kernels/csrc/cuda/attention/rope.cu
@@ -1,26 +1,27 @@
 // Rotary position embedding (RoPE), HF "rotate-half" convention (GPT-NeoX style)
 // as used by Qwen/Llama. Applied to Q and K after projection, before attention.
 //
-// For a head vector x[head_dim] at position p, with half = head_dim/2:
-//   freq_i  = theta^(-2i/head_dim),  angle = p * freq_i
-//   out[i]      = x[i]*cos - x[i+half]*sin
-//   out[i+half] = x[i+half]*cos + x[i]*sin     for i in [0, half)
+// inv_freq[i] = theta^(-2i/head_dim) is precomputed into __constant__ memory at
+// model init (CUDA-graph safe — no per-thread __powf in the decode hot path).
 //
 // Portable CUDA — runs on sm_89 .. sm_120 (RTX 5090).
 
 #include <cuda_bf16.h>
 #ifndef SPARKINFER_NVRTC_DEVICE_ONLY
 #include <cuda_runtime.h>
+#include <cmath>
 #endif
 
 namespace sparkinfer {
 namespace kernels {
 
+__constant__ float c_rope_inv_freq[128];   // head_dim/2, max head_dim=256
+
 // grid = (n_tokens, n_heads); blockDim = head_dim/2 threads (one per rotated pair).
 __global__ void rope_kernel(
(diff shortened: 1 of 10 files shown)


ai-hpc:

```bash
```
Then the bot greenlights it on the next poll. (A maintainer can add `force-eval` to override.)

PR #62, opened 2026-06-26
perf(decode): fuse q_norm + k_norm + RoPE into one kernel (cut bs=1 launch latency)

### LINKED ISSUE #61: perf(decode): q_norm + k_norm + RoPE are 4 separate kernels per layer — fuse to cut bs=1 launch latency
## Summary

Per decode token, each layer issues **four** tiny kernels for the QK-norm + RoPE step:

```cpp
launch_rmsnorm(s.q, w.q_norm, s.q, n_q_heads,  head_dim, eps);   // q_norm
launch_rmsnorm(s.k, w.k_norm, s.k, n_kv_heads, head_dim, eps);   // k_norm
launch_rope(s.q, s.k, d_pos, 1, n_q_heads, n_kv_heads, head_dim, theta);  // rope(q) + rope(k)
```

`launch_rope` itself launches two kernels (q and k), so that is **4 kernel launches per layer × 40 layers = 160 launches per token** for work that is numerically trivial (an RMS over head_dim=128 and a pair-rotation).

## Why it matters

At batch=1 decode, the runtime is **launch-latency bound**, not compute bound — the same bottleneck the PDL path in `expert_ffn_q4k.cu` was written to hide ("the ncu-confirmed bottleneck"). These four kernels sit on the critical per-token dependency chain (q/k must be normed+roped before `kv_append`/attention), so their *latency*, not their FLOPs, is what costs. Collapsing them into one launch removes 3 kernel boundaries per layer (120 per token) from that chain.

## Proposed change

Fuse `q_norm` + `k_norm` + `rope(q)` + `rope(k)` into a single decode kernel: one block per head RMS-normalizes that head into a bf16 shared buffer, then rotates it in place. To stay correctness-gated, the fused kernel must reproduce the *exact* arithmetic of the separate kernels — same 256-thread block reduction for the RMS, the same bf16 round-trip of the normalized value before rotation, and the same RoPE — so the output is bit-for-bit identical and only the kernel count drops (4 → 1 per layer).

This is a `kernels/` + `runtime/` perf change; correctness holds by construction (token-match 100% / KL 0), so the only question the eval answers is the size of the decode-latency win.

philluiz2323:
## Summary

Fuse the per-layer **q_norm + k_norm + RoPE(q) + RoPE(k)** — four separate kernel launches — into a **single** decode kernel (`qknorm_rope_kernel`, one block per head). This removes 3 kernel boundaries per layer (≈120 per token over 40 layers) from the critical decode dependency chain, where at batch=1 it is **launch latency**, not compute, that dominates (the same bottleneck the `expert_ffn_q4k.cu` PDL path targets).

Fixes #61

**Correctness — output is bit-for-bit identical.** The fused kernel reproduces the exact arithmetic of the separate kernels, so the accuracy gate cannot move:
- same 256-thread block reduction for the RMS (identical `inv_rms`);
- the normalized value is rounded to **bf16 in shared memory** before rotation — exactly the bf16 round-trip the separate path does when `rmsnorm` writes bf16 to global and `rope` reads it back;
- identical RoPE (`theta^(-2i/head_dim)`, rotate-half, same cos/sin).

Only the kernel count drops (4 → 1 per layer); every written value is unchanged.

## Proof of speedup

**Honest note:** I don't have local Blackwell hardware to benchmark, so I'm relying on the eval bot's RTX 5090 measurement to quantify the gain. The change is a pure **launch-count reduction** on the bs=1 critical path (4 → 1 kernel per layer, ×40 layers), with **bit-identical output** — so correctness (token-match / KL) is guaranteed and the only open question is the magnitude of the decode-latency win.

```text
Pending eval-bot measurement on RTX 5090 (no local Blackwell GPU).
Mechanism: 160 -> 40 QK-norm/RoPE kernel launches per token; output unchanged.
```

| | decode tok/s |
|---|--:|
| before | 187.6 (current frontier) |
| after  | (eval to measure) |


ai-hpc:

PR #71, opened 2026-06-26
perf(moe): int8 dp4a MMVQ for the Q4_K MoE down projection (+5% decode)

Dexterity104:
## Summary

The Q4_K rows of the MoE down projection were the last major decode GEMV still on the fp register dequant path. Q4_K_M stores about half the down tensors as Q4_K and the rest as Q6_K. The Q6_K half already runs int8 dp4a MMVQ, but the Q4_K half fell through to `down_q6k_kernel` (fp register dequant). A node level nsys trace (`--cuda-graph-trace=node`) on current main shows that kernel at 12.1% of decode, the slowest single MoE call (15.8 us). Moving it onto the same int8 dp4a path that gate/up and the Q6_K down already use closes that gap.

## What changed

`down_q4k_mmvq_kernel` quantizes the down activation to Q8_1 once (reusing the existing `quant_h_q8_1_kernel`) and dp4a's the Q4_K weight against it with the faithful `vec_dot_q4_K_q8_1` identity already used by the gate/up experts. One warp per (token, hh) folds the top_k experts and writes each output row once. The launcher routes `down_type == 12` through it; `SPARKINFER_DOWN_Q4K=0` restores the previous fp dequant down for those rows.

Per kernel (nsys, n=64, node trace): the Q4_K down drops from `down_q6k_kernel` at 15.8 us/call to `down_q4k_mmvq_kernel` at 8.5 us/call, and the fp dequant kernel no longer runs for the Q4_K rows.

The math was checked on CPU first (emulated `__dp4a`, ggml Q4_K dequant reference): the per row tiling reproduces the dequant reference dot to 5e-8 relative.

## Proof of speedup

**Decode tok/s** (end-to-end, from `bench/scripts/bench.sh`, n=128, bs=1):

| | decode tok/s |
|---|--:|
| before (main) | 312.4 |
| after (this PR) | 328.2 |

About +5.0% decode, stable across three alternating runs. Accuracy gate vs llama.cpp (`bench/scripts/accuracy.sh`): top-1 `0.970`, KL `0.151` nats, in line with the other int8 MMVQ kernels (the fp baseline scores `0.990` / `0.146` on the same harness). `ctest` 5/5, `compute-sanitizer --tool memcheck` clean (0 errors).

```text
# before (main path: SPARKINFER_DOWN_Q4K=0)
decode tg    : 312.58 tok/s   (n=128, bs=1)
decode tg    : 312.35 tok/s   (n=128, bs=1)
decode tg    : 312.28 tok/s   (n=128, bs=1)
# after (this PR, default)
decode tg    : 328.17 tok/s   (n=128, bs=1)
decode tg    : 328.37 tok/s   (n=128, bs=1)
decode tg    : 328.13 tok/s   (n=128, bs=1)
```

### DIFF
--- a/kernels/csrc/cuda/moe/expert_ffn_q4k.cu
+++ b/kernels/csrc/cuda/moe/expert_ffn_q4k.cu
@@ -455,6 +455,43 @@ __global__ void gate_up_mmvq2_kernel(
     if (lane == 0) h_scratch[(size_t)ts * F + f] = q4kf_silu(tg) * tu;
 }
 
+// int8 dp4a MMVQ down (Q4_K). The Q4_K-quantized down rows in Q4_K_M were the last MoE GEMV
+// still on the fp register-dequant path (Q6_K down + gate/up + attention already run int8).
+// Reuses the Q8_1-quantized activation and the faithful vec_dot_q4_K_q8_1, one warp per
+// (token,hh) folding the top_k experts. Each 256-superblock has 16 vdr=2 positions; the warp
+// strides those (nblk*16 items) so the dp4a math matches the gate/up path exactly.
+__global__ void down_q4k_mmvq_kernel(
+    const unsigned char* __restrict__ down_q, const int* __restrict__ expert_ids,
+    const float* __restrict__ expert_weights, const si_block_q8_1* __restrict__ hq8,
+    __nv_bfloat16* __restrict__ output, int H, int F, int top_k
+) {
+    const int token = blockIdx.x;
+    const int lane = threadIdx.x & 31;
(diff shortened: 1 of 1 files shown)


ai-hpc:
## ✅ sparkinfer auto-eval — `7ad966d`

| metric | value |
|---|---|
| **label** | `eval:M` |
| decode | 333.75 tok/s |
| vs prior frontier | 305.93 tok/s → +9.1% (+27.8) |
| **→ new frontier** | **333.75 tok/s** |
| correctness | top-1 97.0% · KL 0.1507 |

Verified speedup — **sets the new frontier to 333.75 tok/s** (was 305.93).

PR #80, opened 2026-06-28
runtime/gpu_stats: fix chained-snprintf buffer overrun in GpuStats::str()

### LINKED ISSUE #79: runtime/gpu_stats: GpuStats::str() chained snprintf overruns the buffer on truncation
## Summary

`GpuStats::str()` (added in the GPU-observability change) advances its write cursor by `snprintf()`'s return value, which is the number of bytes that *would* have been written had the buffer been large enough — not the number actually written. If the first field is ever truncated, `n` exceeds the buffer size, and the subsequent calls compute an out-of-bounds destination and an underflowed size.

## Location

`runtime/src/gpu_stats.cpp` — `GpuStats::str()`

```cpp
char buf[192];
int n = snprintf(buf, sizeof buf, "%d°C · %.1f/%.1f GB VRAM", temp_c, vram_used_gb(), vram_total_gb());
if (power_w      >= 0) n += snprintf(buf + n, sizeof buf - n, " · %d W", power_w);     // buf+n past end; sizeof buf - n underflows
if (sm_clock_mhz >= 0) n += snprintf(buf + n, sizeof buf - n, " · %d MHz", sm_clock_mhz);
return std::string(buf, n > 0 ? n : 0);   // n may exceed bytes written -> OOB read
```

## Why it is wrong

Per C99/C++, `snprintf` returns the length the full output would have required. After a truncating call `n >= sizeof buf`, so:
- `buf + n` points past the array, and `sizeof buf - n` (a `size_t`) wraps to a huge value — the next `snprintf` is handed an out-of-bounds pointer and a bogus size (out-of-bounds write / UB);
- `std::string(buf, n)` then constructs from `n` bytes that were never written (out-of-bounds read).

It is latent today because the current fields are short, but it is a real defect: any added field, a long/locale-dependent format, or a larger value makes it reachable, and it sits in the always-compiled observability path.

## Expected behavior

The write cursor must be clamped to the bytes *actually* written, so the offset and remaining size passed to each `snprintf` stay within the buffer.

## Proposed fix

Append through a helper that clamps the cursor (and stops once full):

```cpp
const int cap = (int)sizeof buf; int n = 0;
auto append = [&](const char* fmt, auto... a){ if (n >= cap-1) return; int m = snprintf(buf+n, (size_t)(cap-n), fmt, a...); if (m < 0) return; n += (m < cap-1-n) ? m : cap-1-n; };
append("%d°C · %.1f/%.1f GB VRAM", temp_c, vram_used_gb(), vram_total_gb());
if (power_w      >= 0) append(" · %d W", power_w);
if (sm_clock_mhz >= 0) append(" · %d MHz", sm_clock_mhz);
return std::string(buf, n);
```

## Impact

Robustness / memory-safety in the host observability path. Output is identical for normal stats strings; the change removes the latent out-of-bounds write/read. Build verified + `ctest` 5/5 on RTX 5090 (sm_120).

joaovictor712:
## Summary

Fix a latent buffer overrun in `GpuStats::str()`: it advanced its cursor by `snprintf`'s return value (the would-have-written length), so a truncating field makes `buf + n` point past the buffer and `sizeof buf - n` underflow — an out-of-bounds write, and the final `std::string(buf, n)` an out-of-bounds read.

Fixes #79

**Change.** Append through a small helper that clamps the cursor to the bytes actually written and stops once the buffer is full, so every offset/size handed to `snprintf` stays in range. Output is byte-identical for normal stats strings.

**Verification.** Built from source and ran the test suite on an **RTX 5090 (sm_120)**: `gpu_stats.cpp` compiles clean, `ctest` 5/5 (well, 6/6 incl. GPU tests) pass. This is a host-side robustness fix on the observability path — not a decode/perf change.

## Proof of speedup

N/A — robustness/memory-safety fix in host observability code; no kernel or decode-path change, so no decode-throughput delta. (Built + `ctest` verified on a 5090; the box is unticked because there is no perf claim to benchmark.)

| | decode tok/s |
|---|--:|
| before (main) | N/A (no perf change) |
| after (this PR) | N/A (no perf change) |

```text
N/A — host-side string-formatting hardening; decode output and speed unchanged.
```

## Scope

Single focused change in `runtime/src/gpu_stats.cpp`; one fix per PR per CONTRIBUTING.md, scoped to `runtime/`.

### DIFF
--- a/runtime/src/gpu_stats.cpp
+++ b/runtime/src/gpu_stats.cpp
@@ -26,11 +26,21 @@ static bool nvml_ready() {
 std::string GpuStats::str() const {
     if (!valid) return "GPU stats unavailable";
     char buf[192];
-    int n = snprintf(buf, sizeof buf, "%d°C · %.1f/%.1f GB VRAM",
-                     temp_c, vram_used_gb(), vram_total_gb());
-    if (power_w      >= 0) n += snprintf(buf + n, sizeof buf - n, " · %d W", power_w);
-    if (sm_clock_mhz >= 0) n += snprintf(buf + n, sizeof buf - n, " · %d MHz", sm_clock_mhz);
-    return std::string(buf, n > 0 ? n : 0);
+    const int cap = (int)sizeof buf;
+    // snprintf returns the length it WOULD have written; on truncation that exceeds the
+    // buffer, so advancing `buf + n` with `cap - n` (size_t) would point past the buffer
+    // and underflow the size. Append() clamps the cursor to what was actually written.
+    int n = 0;
+    auto append = [&](const char* fmt, auto... args) {
+        if (n >= cap - 1) return;
+        int m = snprintf(buf + n, (size_t)(cap - n), fmt, args...);
+        if (m < 0) return;
+        n += (m < cap - 1 - n) ? m : cap - 1 - n;   // clamp to bytes actually written
+    };
+    append("%d°C · %.1f/%.1f GB VRAM", temp_c, vram_used_gb(), vram_total_gb());
+    if (power_w      >= 0) append(" · %d W", power_w);
+    if (sm_clock_mhz >= 0) append(" · %d MHz", sm_clock_mhz);
+    return std::string(buf, n);
 }
 
 GpuStats query_gpu_stats(int device_id) {


ai-hpc:

PR #82, opened 2026-06-28
runtime/tools/convert_gguf: fix dead vocab-from-token_embd fallback

### LINKED ISSUE #81: runtime/tools/convert_gguf: vocab-from-token_embd fallback is dead code (never runs, wrong axis)
## Summary

The GGUF→sparkinfer converter is meant to fall back to the embedding tensor's vocab size when the GGUF lacks `<arch>.vocab_size`. That fallback is dead code and never executes.

## Location

`runtime/tools/convert_gguf.py`

```python
# vocab from token_embd if metadata missing
ten = {t.name: t for t in r.tensors}
```

## Why it is wrong

Two problems:
1. **`... if False else cfg["vocab"]`** makes the assignment `cfg["vocab"] = cfg["vocab"]` — a no-op. The fallback never runs, so a GGUF without `<arch>.vocab_size` keeps the hardcoded default from `vocab=int(meta(A+"vocab_size") or 151936)` (151936). For any model whose true vocab differs, `config.txt` then carries the wrong vocab, which mis-sizes the LM head / embedding and breaks the converted model.
2. **Wrong axis.** Even if enabled, it reads `shape[-1]`, but `token_embd.weight` is `[vocab, hidden]` in this converter (see the `write("embed_tokens", deq("token_embd.weight"))  # [vocab, hidden] (as-is)` line), so the vocab axis is `shape[0]`; `shape[-1]` is `hidden`.

## Expected behavior

When `<arch>.vocab_size` is absent, derive vocab from the embedding tensor's vocab axis (`shape[0]`).

## Proposed fix

```python
if not meta(A + "vocab_size") and "token_embd.weight" in ten:
    cfg["vocab"] = int(ten["token_embd.weight"].shape[0])   # [vocab, hidden]
```

## Impact

Correctness of the offline GGUF conversion for models that omit the `vocab_size` metadata key (the documented fallback finally works, with the correct axis). No change for GGUFs that carry `vocab_size` (the common case).

joaovictor91123:
## Summary

Fix the converter's vocab fallback so it actually derives vocab from the embedding tensor when the GGUF lacks `<arch>.vocab_size`.

Fixes #81

**Why.** The line `cfg["vocab"] = int(ten["token_embd.weight"].shape[-1]) if False else cfg["vocab"]` is a no-op (`if False`), so the documented fallback never runs — a GGUF without `vocab_size` keeps the hardcoded `151936` default and mis-sizes the LM head/config for any other-vocab model. It also read `shape[-1]` (the hidden axis); `token_embd.weight` is `[vocab, hidden]` here, so the vocab axis is `shape[0]`.

**Change.** Run the fallback only when `vocab_size` metadata is absent, and read `shape[0]`:

```python
if not meta(A + "vocab_size") and "token_embd.weight" in ten:
    cfg["vocab"] = int(ten["token_embd.weight"].shape[0])   # [vocab, hidden]
```

GGUFs that carry `vocab_size` (the common case) are unaffected.

## Proof of speedup

N/A — correctness fix in the offline GGUF→sparkinfer conversion tool (`runtime/tools/convert_gguf.py`); it does not touch the runtime/decode path, so there is no decode-throughput change to benchmark.

| | decode tok/s |
|---|--:|
| before (main) | N/A (offline tool, no perf change) |
| after (this PR) | N/A (offline tool, no perf change) |

```text
N/A — offline conversion-tool correctness fix; runtime decode output/speed unchanged.
```

## Scope

Single focused change in `runtime/tools/convert_gguf.py`; one fix per PR per CONTRIBUTING.md, scoped to `runtime/`.

### DIFF
--- a/runtime/tools/convert_gguf.py
+++ b/runtime/tools/convert_gguf.py
@@ -56,8 +56,12 @@ def main():
     )
     # vocab from token_embd if metadata missing
     ten = {t.name: t for t in r.tensors}
-    if "token_embd.weight" in ten:
-        cfg["vocab"] = int(ten["token_embd.weight"].shape[-1]) if False else cfg["vocab"]
+    if not meta(A + "vocab_size") and "token_embd.weight" in ten:
+        # token_embd.weight is [vocab, hidden] (see embed_tokens write below), so the
+        # vocab axis is shape[0]. The previous `... if False else cfg["vocab"]` made this
+        # fallback a no-op, leaving cfg["vocab"] at the hardcoded 151936 default whenever
+        # the GGUF lacks <arch>.vocab_size — which mis-sizes the LM head for such models.
+        cfg["vocab"] = int(ten["token_embd.weight"].shape[0])
 
     with open(os.path.join(out, "config.txt"), "w") as f:
         for k, v in cfg.items():


ai-hpc:

PR #89, opened 2026-06-29
perf(decode): run Q/K/V projections on concurrent CUDA streams (+5.7% decode)

James-CUDA:
## Summary

At bs=1 each Q/K/V projection is a small GEMV (Q: 2048→4096, K/V: 2048→512) that **under-occupies the 170 SMs and is latency-bound** — run serially on one stream, the GPU sits idle between them. This forks the three projections onto **separate CUDA streams** (after the shared Q8_1 activation quantize) and joins before QK-norm, so their weight reads overlap and fill the idle time — **higher effective bandwidth utilization with no extra peak bandwidth**.

- Fork/join via CUDA events, **captured into the decode CUDA graph** (the concurrent branches replay each token).
- **Pure scheduling — the decode output is byte-identical** to the serial path; top1 vs llama.cpp is unchanged.
- Mirrors llama.cpp [PR #16991](https://github.com/ggml-org/llama.cpp/pull/16991) ("CUDA: add stream-based concurrency"), which measured +6–9% token-gen on this exact model/GPU.

`SPARKINFER_QKVSTREAM=0` restores the serial projections.

## Proof of speedup

**Decode tok/s** (`bench/scripts/bench.sh`, n=128, bs=1):

| | decode tok/s |
|---|--:|
| before (main) | 420.48 |
| after (this PR) | 444.31 |

Accuracy gate vs llama.cpp (`bench/scripts/accuracy.sh`): `top1=0.970, KL=0.142`. The decode output is **byte-identical** to main (pure scheduling change), so accuracy is unchanged.

### DIFF
--- a/runtime/src/models/qwen35.cpp
+++ b/runtime/src/models/qwen35.cpp
@@ -56,6 +56,8 @@ struct Qwen35Model::Impl {
     moe::MoEEngine* engine;
     Qwen35Weights w;
     cudaStream_t stream{};
+    cudaStream_t stream_k{}, stream_v{};         // side streams for concurrent K/V projection
+    cudaEvent_t ev_qkv{}, ev_k{}, ev_v{};        // fork/join events (captured into the decode graph)
     uint64_t seq_id = 0;
     int qdim, kvdim;
     bool gguf = false;   // true after load_gguf: dense weights are native [out,in], use GEMV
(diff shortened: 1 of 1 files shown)


ai-hpc:
## ✅ sparkinfer auto-eval — `4e13546`

| metric | value |
|---|---|
| **label** | `eval:S` |
| decode | 443.38 tok/s |
| vs prior frontier | 419.63 tok/s → +5.7% (+23.8) |
| **→ new frontier** | **443.38 tok/s** |
| correctness | top-1 97.0% · KL 0.1417 |

Verified speedup — **sets the new frontier to 443.38 tok/s** (was 419.63).

PR #93, opened 2026-06-29
fix(runtime): clip generate max_new to max_seq budget

### LINKED ISSUE #92: Qwen35Model::generate should enforce max_seq budget for max_new
## Problem

`Qwen35Model::generate()` currently does not enforce `prompt.size() + max_new_tokens <= cfg.max_seq`. It only allocates KV for `max_seq` and then loops `max_new` tokens, which can request more tokens than the sequence budget allows.

This can produce incorrect behavior when the caller asks for long continuations on short budgets (especially when prompt is long), because the path currently relies on lower-level KV behavior and can emit silently clipped/partial sequences.

## Reproduction

1. Configure `Qwen35Config.max_seq` to a small value.
2. Use a prompt near that length.
3. Call `generate(prompt, max_new)` where `prompt.size() + max_new > max_seq`.
4. Observe output length and behavior are not explicitly bounded.

## Expected behavior

`generate()` should explicitly validate and cap/fail when requested generation would exceed `max_seq`.

A safe fix is:
- Reject/fail if prompt length is already `>= max_seq`.
- Clip `max_new` to `max_seq - prompt.size()` and log the clipping.

No silent, implicit overflow behavior should remain.

## Suggested code changes

- Add explicit budget checks in `runtime/src/models/qwen35.cpp` in `Qwen35Model::generate()`.
- Add test coverage in `runtime/tests/qwen35_gpu_test.cpp` for oversized `max_new` requests.

tmimmanuel:
## Summary

 now validates sequence budget before decoding:
- rejects invalid configs ()
- rejects zero/negative generation requests
- rejects prompts that already fill/overflow 
- clips  to available budget () and logs clipping

This prevents silent overrun scenarios in generation where caller intent requests more tokens than the configured sequence capacity.

## Changes

- 
  - Added explicit /prompt budget validation in .
  - Added deterministic clip for  when it exceeds remaining budget.

- 
  - Added regression coverage for clipping behavior with  < .

## Why

Currently the function allocates KV for  and then loops requested  tokens, which can produce undefined/unexpected behavior for oversize requests. This is a safety/correctness fix.

## Testing

- 
- Build/test attempted locally, but CUDA toolkit is unavailable in this environment ( not found), so runtime test execution is not possible here.

## Related

- Fixes: https://github.com/gittensor-ai-lab/sparkinfer/issues/92

### DIFF
--- a/runtime/src/models/qwen35.cpp
+++ b/runtime/src/models/qwen35.cpp
@@ -373,6 +373,26 @@ std::vector<int> Qwen35Model::generate(const std::vector<int>& prompt, int max_n
     Impl& s = *p_;
     std::vector<int> out;
     if (prompt.empty()) return out;
+
+    const int prompt_len = (int)prompt.size();
+    if (s.cfg.max_seq <= 0) {
+        fprintf(stderr, "[qwen35] generate: max_seq must be > 0 (got %d)\n", s.cfg.max_seq);
+        return out;
+    }
+
+    if (max_new <= 0) return out;
+
+    if (prompt_len >= s.cfg.max_seq) {
+        fprintf(stderr, "[qwen35] generate: prompt length %d exceeds or equals max_seq %d\n", prompt_len, s.cfg.max_seq);
+        return out;
+    }
+
+    int max_allow = s.cfg.max_seq - prompt_len;
+    if (max_new > max_allow) {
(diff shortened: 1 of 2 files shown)


ai-hpc:

PR #95, opened 2026-06-29
runtime: guard forward_token token_id range

### LINKED ISSUE #94: runtime/qwen35: forward_token() lacks token-id range check
`Qwen35Model::forward_token(int token_id, int position)` currently forwards `token_id` directly to embedding lookup without checking that `token_id` is in `[0, cfg.vocab)`.

`generate()` got range-checks recently for prompt ids in another flow, but the public `forward_token` API can still be called directly (external callers/tests) and can still trigger OOB embedding reads on bad IDs.

## Why this is a bug
- It is a direct memory safety/correctness gap for the single-token API.
- It makes behavior inconsistent between direct calls and `generate()`.
- It complicates caller guarantees for internal tools/tests that bypass `generate()`.

## Suggested fix
- Add a guard in `runtime/src/models/qwen35.cpp::forward_token`:
  - return a safe token (or early error path) when `token_id` is out of range.
  - log the validation failure.
- Add unit/regression coverage in `runtime/tests/qwen35_gpu_test.cpp` for:
  - negative token id
  - token id == vocab
  - token id >> vocab

## Notes
- Related issue: #37 (`generate()` prompt token range checks).

tmimmanuel:
## Summary

Fixes direct token-ID OOB safety in `Qwen35Model::forward_token()`.

- Adds explicit validation for `token_id` before embedding lookup.
- On invalid IDs, logs a bounded-range error and returns `cfg.eos_id`.

## Why

`forward_token()` is a public API and can be called directly. Without a guard, out-of-range token IDs can trigger out-of-bounds embedding reads even when `generate()` path checks are in place.

## Changes

- `runtime/src/models/qwen35.cpp`
  - Add range check in `Qwen35Model::forward_token()`.
- `runtime/tests/qwen35_gpu_test.cpp`
  - Add regression assertions for negative and out-of-range token IDs.

## Testing

- `git diff --check`
- CUDA runtime not available in this environment (`nvcc` missing), so this test was not executed here.

## Related

- Fixes: https://github.com/gittensor-ai-lab/sparkinfer/issues/94

### DIFF
--- a/runtime/src/models/qwen35.cpp
+++ b/runtime/src/models/qwen35.cpp
@@ -180,6 +180,12 @@ void Qwen35Model::copy_logits(float* host_logits) const {
 int Qwen35Model::forward_token(int token_id, int position) {
     Impl& s = *p_;
     const Qwen35Config& c = s.cfg;
+
+    if (token_id < 0 || token_id >= c.vocab) {
+        fprintf(stderr, "[qwen35] forward_token: token %d out of range [0, %d)\n", token_id, c.vocab);
+        return c.eos_id;
+    }
+
     const int H = c.hidden;
     kernels::GemmConfig gc{};
     int seqlen = position + 1;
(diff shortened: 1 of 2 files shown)


ai-hpc:

PR #97, opened 2026-06-29
runtime: guard forward_token position range

### LINKED ISSUE #96: runtime/qwen35: forward_token should validate position before KV ops
## Problem

`Qwen35Model::forward_token(int token_id, int position)` currently validates only `token_id` bounds.
`position` is passed straight to device inputs (`d_pos`, `d_writepos`) and consumed by KV-append logic without validation.

A negative position or `position >= cfg.max_seq` can produce invalid write positions and block addressing, which can lead to cache corruption / OOB behavior in edge-case caller usage.

Current flow:
- `forward_token` checks token range, then sets `seqlen = position + 1`.
- It uploads `position` directly into `d_pos` and `d_writepos`.
- KV path consumes those positions without a guard.

## Expected behavior
- Reject invalid positions early (`position < 0` or `position >= cfg.max_seq`).
- Return a safe token (e.g. `cfg.eos_id`) without launching compute.
- Keep valid generation path behavior unchanged.

## Why this matters
- Follows existing safety direction in `forward_token` token validation and generate budget checks.
- Prevents undefined behavior for bad API inputs (`position` misuse is independently controllable).

## Suggested fix
In `runtime/src/models/qwen35.cpp`, add:

```cpp
if (position < 0 || position >= c.max_seq) {
    fprintf(stderr, "[qwen35] forward_token: position %d out of range [0, %d)\n", position, c.max_seq);
    return c.eos_id;
}
```

## Tests
- Add regression coverage in `runtime/tests/qwen35_gpu_test.cpp` for:
  - `forward_token(1, -1)`
  - `forward_token(1, cfg.max_seq)`

tmimmanuel:
## Summary

Adds explicit bounds validation for token position in `Qwen35Model::forward_token()` before entering the decode path.

## Problem

`forward_token(int token_id, int position)` can receive an invalid `position` and still reach KV/position-sensitive internals. A negative position or `position >= cfg.max_seq` can create invalid write positions for KV append/block-addressing paths.

## Changes

- `runtime/src/models/qwen35.cpp`
  - Add guard: return `c.eos_id` with a warning when `position < 0` or `position >= c.max_seq`.
- `runtime/tests/qwen35_gpu_test.cpp`
  - Add regressions for:
    - `model.forward_token(1, -1)`
    - `model.forward_token(1, cfg.max_seq)`

Both cases assert the result is `cfg.eos_id` and avoid launching decode on invalid positions.

## Testing

- `git diff --check`
- CUDA runtime unavailable in this environment (`nvcc` missing), so GPU tests were not executed here.

Closes #96

### DIFF
--- a/runtime/src/models/qwen35.cpp
+++ b/runtime/src/models/qwen35.cpp
@@ -180,6 +180,12 @@ void Qwen35Model::copy_logits(float* host_logits) const {
 int Qwen35Model::forward_token(int token_id, int position) {
     Impl& s = *p_;
     const Qwen35Config& c = s.cfg;
+
+    if (position < 0 || position >= c.max_seq) {
+        fprintf(stderr, "[qwen35] forward_token: position %d out of range [0, %d)\n", position, c.max_seq);
+        return c.eos_id;
+    }
+
     const int H = c.hidden;
     kernels::GemmConfig gc{};
     int seqlen = position + 1;
--- a/runtime/tests/qwen35_gpu_test.cpp
+++ b/runtime/tests/qwen35_gpu_test.cpp
@@ -88,5 +88,20 @@ int main() {
     printf("[PASS] qwen35_gpu_test: generated 8 tokens:");
     for (int id : out) printf(" %d", id);
     printf("\n");
+
+    int invalid_pos_neg = model.forward_token(1, -1);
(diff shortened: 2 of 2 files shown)


ai-hpc:

PR #98, opened 2026-06-29
perf(moe): split-K int8 MMVQ down (Q6_K) for bs=1 decode occupancy

glorydavid03023:
## Summary

split-K on the int8 MMVQ Q6_K **down** projection. The default down kernel (`down_q6k_mmvq_kernel`) runs one warp per output row → only `H` warps in flight at bs=1 (occupancy-bound per the in-file ncu note). split-K already fixed this for the **fp** dequant down (`down_q6k_splitk_kernel`) but was never ported to the int8 MMVQ path that is now the default — this composes the two: `S=4` warps cooperate per row on the dp4a Q6_K math. Default ON; `SPARKINFER_DOWN_SPLITK=0` falls back. Correctness-safe: identical dp4a integer math; only the cross-split reduction is an fp-add reassociation (top-1 unchanged, KL within gate).

Single file, single optimization: `kernels/csrc/cuda/moe/expert_ffn_q4k.cu`.

## Proof of speedup

> ⏳ **Not yet benchmarked on an RTX 5090.** Opening for review of the approach/correctness.
> The box stays **unticked** and the decode table is empty on purpose — per the gate this PR is
> `not-tested` and will **not** be GPU-evaluated until a real `before < after` decode run is pasted
> here. I will not tick the box without a genuine RTX 5090 run.

**Decode tok/s** (end-to-end, from `bench/scripts/bench.sh` — required for evaluation):

| | decode tok/s |
|---|--:|
| before (main) | _pending real 5090 run_ |
| after (this PR) | _pending real 5090 run_ |

```text
# pending: bench/scripts/bench.sh output (SPARKINFER_DOWN_SPLITK=0 -> default)
```


ai-hpc:

PR #99, opened 2026-06-29
perf(decode): emit Q8_1(attn) from the flash-decode combine, dropping the O-proj activation quantize

carlh7777:
## Summary

The O-projection takes the llama Q4_K MMVQ path, which needs a Q8_1 of the attention
output. Today that's a standalone per-layer kernel (`launch_quantize_q8_1_blocks(attn)`)
on the bs=1 decode critical path — between the flash-decode combine and the O-proj MMVQ —
that reads the whole `attn` vector back just to write its Q8_1 blocks.

The combine pass already produces `attn`. With `FA_COMBINE_DG` splitting each head into
32-wide groups, warp 0's 32 lanes write exactly one contiguous 32-element slice per `ELEMS`
step — i.e. exactly one `si_block_q8_1`. So the combine now emits the Q8_1 itself (warp
amax/sum over those same 32 lanes) and the runtime drops the standalone quantize entirely:
**1 kernel/layer (40/token) off the dependency chain, plus the attn readback.**

This is the O-projection analog of the already-merged residual-RMSNorm Q8_1 fusion (#83),
which did the same for the QKV-input and gate/up-input quantizes but not the attention output.

**Correctness — bit-for-bit identical, gate cannot move.** The Q8_1 is derived from the
**bf16-rounded** combine output (the exact value the standalone quantizer reads back), with
the same arithmetic as `si_quantize_q8_1_blocks`: amax via `shfl_xor`, `d = amax/127`,
`qi = roundf(v/d)` (0 when `amax == 0`), `ds = (d, d*sum)`. Every written byte is unchanged.

- Works for any `FA_COMBINE_DG` — each `ELEMS` step is its own 32-block, so it doesn't depend
  on the sweepable combine layout.
- Default ON; `SPARKINFER_ATTNQ8=0` disables for A/B. Gated on the llama Q4_K O-proj path
  (`use_pq && use_llama`, `wo_type == 12`) and `head_dim == 128` (the only `flash_decode_split`
  instantiation); falls back to the standalone quantize otherwise, so every other path is unchanged.
- Stream ordering for `aq81` is unchanged: the combine writes it on the main stream after the
  Q/K/V projections (which read it) have joined back — the same point the standalone wrote it.

## Proof of speedup

```text
# paste bench/scripts/bench.sh output here (before -> after)
```

Accuracy gate (`bench/scripts/accuracy.sh`): top-1 ≈100% / KL ≈0 vs the prior build — holds by
construction (output bytes are identical; only the kernel count drops).

### DIFF
--- a/kernels/csrc/cuda/attention/flash_decode_split.cu
+++ b/kernels/csrc/cuda/attention/flash_decode_split.cu
@@ -12,12 +12,18 @@
 // One warp per block; head_dim=128 (Qwen3). Portable CUDA — sm_89 .. sm_120/121.
 
 #include <cuda_bf16.h>
+#include <cuda_fp16.h>
 #ifndef SPARKINFER_NVRTC_DEVICE_ONLY
 #include <cuda_runtime.h>
 #endif
 
 namespace sparkinfer {
 namespace kernels {
+
+// Activation Q8_1 block, layout-identical to gemm.cu's si_block_q8_1 (the format the
+// llama-port Q4_K/Q6_K MMVQ kernels read). The combine kernel can emit it directly so the
+// O-projection's standalone activation quantize is deleted from the per-layer decode chain.
+struct fa_q8_block { __half2 ds; signed char qs[32]; };   // 36 B / 32 values
 
 __device__ __forceinline__ float fa_to_f(__nv_bfloat16 x) { return __bfloat162float(x); }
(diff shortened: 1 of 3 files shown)


ai-hpc:
Closing as part of a queue cleanup: this PR has not been tested on RTX 5090 (the **Tested on RTX 5090** gate was not ticked), so it was not greenlit or scored.

PR #103, opened 2026-06-30
fix(gguf): reject tensors whose dimension product overflows signed long

### LINKED ISSUE #29: runtime/gguf: tensor dimension product can overflow signed long (bogus n_values/n_bytes)

luciferlive112116:
## Summary

Fixes a memory-safety bug in the GGUF reader: a tensor's dimension product is accumulated into a signed `long` with no overflow check, so a crafted/corrupt file can wrap the product to a small or zero value that then slips past the data-bounds check. Closes #29. (Correctness fix — no decode speedup, so the eval box below is intentionally left unticked.)

## Details

In `runtime/src/gguf.cpp`, the tensor-table loop did:

```cpp
long nv = 1;
...
in.t.n_values = nv;
```

`nv *= e` over file-controlled `uint64_t` dims is signed-overflow (UB). A file with dims like `{2^32, 2^32}` makes the product `2^64`, which wraps to `0`. Then `n_values == 0` and `n_bytes == 0`, and the existing data-bounds guard (which checks `n_bytes` against the file size) passes — so a bogus tensor with billion-element real dims is silently accepted, and downstream allocations/copies use the wrong sizes.

The fix rejects negative dims and detects the product overflow with `__builtin_mul_overflow` (consistent with the file's existing "fail loudly on malformed metadata" guards, and with the codebase's existing use of GCC/Clang builtins/pragmas), bailing before the bogus tensor is stored.

## Test

Adds `runtime/tests/gguf_cpu_test.cpp` (registered in `runtime/tests/CMakeLists.txt`) — a pure host C++ CPU test (no GPU/CUDA) that:

1. Builds a minimal in-memory GGUF whose single tensor has dims `{2^32, 2^32}` and asserts `GGUF::open()` now returns `false` (it returned `true` with a silently-wrong `n_values == 0` before the fix).
2. Builds a well-formed single-tensor GGUF and asserts it still opens and parses with the correct shape (guards against an over-eager check).

Build/run: `cmake -B build … && cmake --build build -j && ctest --test-dir build --output-on-failure` (the new test is host-only and needs no device).

> Note: prepared in an environment without a CUDA toolkit / C++ build available, so I could not run `ctest` locally; the reader source (`gguf.cpp`) and the new test are pure POSIX host C++ and are expected to build/pass in CI.

## Proof of speedup

N/A — this is a memory-safety bug fix, not a decode optimization. No tok/s change is claimed.

### DIFF
--- a/runtime/src/gguf.cpp
+++ b/runtime/src/gguf.cpp
@@ -115,8 +115,22 @@ bool GGUF::open(const std::string& path) {
         uint32_t nd = c.rd<uint32_t>();
         if (!c.ok || nd > 4) { fprintf(stderr, "[gguf] tensor %s has invalid n_dims=%u (max 4)\n", in.name.c_str(), nd); return false; }
         in.t.n_dims = nd;
+        // dims are file-controlled. A crafted product (e.g. {2^32, 2^32}) would
+        // overflow signed long — UB — and wrap to a small/zero n_values that then
+        // slips past the data-bounds check below (n_bytes==0 always fits), so a
+        // bogus tensor with billion-element real dims gets silently accepted.
+        // Reject negative dims and any product that overflows.
         long nv = 1;
-        for (uint32_t d = 0; d < nd; d++) { long e = (long)c.rd<uint64_t>(); in.t.dims[d] = e; nv *= e; }
+        bool dim_overflow = false;
+        for (uint32_t d = 0; d < nd; d++) {
+            long e = (long)c.rd<uint64_t>();
+            in.t.dims[d] = e;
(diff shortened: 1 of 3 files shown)


ai-hpc:
Closing as part of a queue cleanup: this PR has not been tested on RTX 5090 (the **Tested on RTX 5090** gate was not ticked), so it was not greenlit or scored.

PR #105, opened 2026-06-30
perf(decode): emit Q8_1 from gate/up MMVQ tail, drop the MoE-down quantize node

andriypolanski:
## Summary

Each decode layer runs `quant_h_q8_1_kernel` as its own CUDA-graph node between the MoE gate/up MMVQ and the int8 down MMVQ. That kernel only re-quantizes the SwiGLU activation `h` that `gate_up_mmvq2_kernel` just wrote to `h_scratch` — same `amax/127` + `roundf(x/d)` path as today, but as a separate graph replay.

This folds that Q8_1 emission into `gate_up_mmvq2_kernel`: once all 32 floats in a Q8_1 block are written, the completing block quantizes from `h_scratch` into `out_scratch` and the standalone `quant_h_q8_1` launch is skipped. Per-layer `q8_ready` slices (indexed by layer id) keep the atomic completion flags CUDA-graph safe across the 40-layer capture without a per-layer memset.

Complements #83 (`SPARKINFER_FNQ`): FNQ removes the gate/up **input** quantize; this removes the down **activation** quantize. Requires FNQ (pre-quantized `aq81` for gate/up input) so `out_scratch` stays free for down Q8_1.

`SPARKINFER_MOEDQ=0` restores the standalone quantize path.

### Kernels (`kernels/csrc/cuda/moe/expert_ffn_q4k.cu`)

- **`gate_up_mmvq2_kernel`** — optional tail: atomic completion per Q8_1 block, then bit-identical `si_quant_h_block_q8_1`.
- **`launch_moe_expert_ffn_q4k`** — new `q8_ready` argument; skips `quant_h_q8_1_kernel` when fusion is active (`SPARKINFER_MOEDQ=1`, default).

### Runtime (`runtime/src/models/qwen35.cpp`)

- Allocate `q8_ready[n_layers * top_k * (moe_ffn/32)]` (device, zero-init).
- Pass per-layer slice to `launch_moe_expert_ffn_q4k` when `SPARKINFER_FNQ=1` and `SPARKINFER_MOEDQ=1`.

### API (`kernels/include/sparkinfer/kernels/moe.h`)

- Document `q8_ready` parameter on `launch_moe_expert_ffn_q4k`.

## Proof of speedup

**Decode tok/s** (`bench/scripts/bench.sh`, Qwen3-30B-A3B Q4_K_M, n=128, bs=1):

| | decode tok/s |
|---|---:|
| before (`SPARKINFER_MOEDQ=0`) | 440.75 |
| after (this PR, default) | 441.98 |

- [x] `ctest --test-dir build`: 6/6 passed
- [x] `compute-sanitizer --tool memcheck`: 0 errors

## Test plan

- [x] `cmake -B build -DCMAKE_CUDA_ARCHITECTURES=120 && cmake --build build -j`
- [x] `ctest --test-dir build`
- [x] `bench/scripts/bench.sh --tokens 128` (5090)
- [x] `bench/scripts/accuracy.sh`
- [x] A/B: `SPARKINFER_MOEDQ=0` vs default — tok/s + KL

## Env flags

| Flag | Default | Effect |
|---|---|---|
| `SPARKINFER_MOEDQ` | `1` | Fuse down Q8_1 into gate/up tail |
| `SPARKINFER_FNQ` | `1` | Required for fusion (separates `aq81` from `out_scratch`) |

### DIFF
--- a/kernels/csrc/cuda/moe/expert_ffn_q4k.cu
+++ b/kernels/csrc/cuda/moe/expert_ffn_q4k.cu
@@ -424,16 +424,35 @@ __device__ __forceinline__ float si_vec_dot_q4_K(const si_block_q4_K* bq4, const
     float2 dm4f = __half22float2(bq4->dm);
     return dm4f.x * sumf_d - dm4f.y * sumf_m;
 }
+// Bit-identical to quant_h_q8_1_kernel for one ib, reading h[ts*F + f0 + lane].
+__device__ __forceinline__ void si_quant_h_block_q8_1(
+    const float* __restrict__ h, int ts, int f0, int F, int ib, si_block_q8_1* __restrict__ y) {
+    const int lane = threadIdx.x & 31;
+    float xv = h[(size_t)ts * F + f0 + lane], a = fabsf(xv);
+    #pragma unroll
+    for (int m = 16; m > 0; m >>= 1) a = fmaxf(a, __shfl_xor_sync(0xffffffffu, a, m));
+    float d = a / 127.0f;
+    int qi = (a == 0.0f) ? 0 : (int)roundf(xv / d);
+    y[ib].qs[lane] = (signed char)qi;
+    int s = qi;
+    #pragma unroll
+    for (int m = 16; m > 0; m >>= 1) s += __shfl_xor_sync(0xffffffffu, s, m);
+    if (lane == 0) y[ib].ds = __floats2half2_rn(d, d * (float)s);
+}
+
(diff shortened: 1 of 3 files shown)


ai-hpc:
## ⚪ sparkinfer auto-eval — `fd67c2b`

| metric | value |
|---|---|
| **label** | `eval:none` |
| decode | 413.07 tok/s |
| vs frontier | 433.26 tok/s → -4.7% (-20.2) |
| correctness | top-1 95.2% · KL 0.0181 |

PR #112, opened 2026-06-30
perf(moe): fuse down-projection Q8_1 quantize into gate/up — measured −4.7% on RTX 5090

carlh7777:
## Summary

Attempts to delete the standalone `quant_h_q8_1` graph node from the per-layer MoE
decode chain (gate_up → quant_h → down) by having `gate_up_mmvq2` emit the down
GEMV's Q8_1(h) activation directly — the same "fuse the quantize into its producer"
pattern as the merged fnq fusion. A Q8_1 block spans 32 ffn columns computed by 32
sibling CUDA blocks, so they cooperate via a threadfence release-counter (canonical
last-block-reduces idiom) to keep the dominant gate/up GEMV's exact block/warp layout
and occupancy. Output is **bit-identical** to `main` (verified: the GEMV half is a
verbatim copy and a CPU reference test confirms the fused Q8_1 byte-matches the
standalone `quant_h`, incl. the amax==0 edge case).

## Result: this REGRESSES decode — do not merge as a speedup

Built from source and benchmarked on a real RTX 5090 (sm_120, CUDA 12.8, open driver
575.57.08), `main` vs this PR interleaved, graphics clock locked, model sha verified:

| n (tok) | main | this PR | Δ |
|---|--:|--:|--:|
| 128 | 443.9 | 423.1 | **−4.7%** |
| 256 | 434.2 | 414.3 | **−4.6%** |
| 128, same binary `SPARKINFER_GUQ8` off→on | 443.8 | 423.1 | **−4.7%** |

The same-binary A/B isolates the cause: the fusion *itself* is the entire −4.7%.
Root cause — the gate/up kernel is already bandwidth-saturated, so the per-block
`__threadfence()` required to publish each `h` before the cross-block reduction costs
far more (~22 tok/s) than removing the tiny `quant_h` launch node saves (~1–2 tok/s).
No cheaper cross-block sync closes that ~20:1 gap, and the only fence-free alternative
(one block owns a 32-wide f-tile) collapses the dominant GEMV from ~6144 blocks to ~192
and regresses occupancy. Accuracy is unaffected (bit-identical); it's purely a perf loss.

Filing as a documented dead-end so the approach isn't re-attempted.

## Proof of speedup

- [x] Tested on **RTX 5090** (`sm_120`)  — *(intentionally unticked: this is a regression, not a speedup)*

| | decode tok/s |
|---|--:|
| before (main) | 443.9 |
| after (this PR) | 423.1 |

### DIFF
--- a/kernels/csrc/cuda/moe/expert_ffn_q4k.cu
+++ b/kernels/csrc/cuda/moe/expert_ffn_q4k.cu
@@ -455,6 +455,72 @@ __global__ void gate_up_mmvq2_kernel(
     if (lane == 0) h_scratch[(size_t)ts * F + f] = q4kf_silu(tg) * tu;
 }
 
+// gate_up_mmvq2 that ALSO emits the Q8_1(h) block for the down GEMV, deleting the standalone
+// quant_h_q8_1_kernel node (the last unfused activation-quantize in the dominant MoE chain;
+// cf. the fnq/attnq8 fusions). Same per-(ts,f) GEMV as gate_up_mmvq2_kernel — so the dominant
+// gate/up GEMV keeps its exact 6144-block, 4-warp layout and occupancy (a one-block-per-32f
+// rewrite would collapse to ~192 blocks and regress). A Q8_1 block spans 32 consecutive f's
+// living in 32 sibling blocks, so they cooperate via a threadfence release-counter: each block
+// publishes its h_scratch[f] (fence) and bumps qrelease[ib]; the 32nd arrival sees all 32 h's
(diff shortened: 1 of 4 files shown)


ai-hpc:
## 🚩 Flagged: eval-gaming (sybil)

This PR is part of a **sybil pair**: it was *opened* by `carlh7777` but every commit is *authored and committed* by `thomasbaker9010251` (`Thomas B <thomas.b.901025@gmail.com>`) — opener ≠ committer, both non-collaborators = one operator, two GitHub accounts. Both accounts are blocked under the no-gaming policy; future PRs from either auto-close and are not evaluated. See [`.github/FLAGGED.md`](../blob/main/.github/FLAGGED.md) for the record.

PR #114, opened 2026-07-01
perf(runtime): context-adaptive flash-decode split count

9876543210-tc-0123456789:
## Summary

Make the flash-decode KV-split count **context-adaptive**. The count is baked into the decode CUDA graph at capture (fixed **32**, tuned for 128-tok). This captures **two** decode graphs — 32 splits for short context, 64 for long — and replays whichever matches each token's current `seqlen` (crossover 224). More splits give more parallel blocks to fill the GPU once each KV chunk is large, so **long-context decode speeds up with no short-context regression**.

One file, `runtime/src/models/qwen35.cpp`. `SPARKINFER_NSPLITS` still forces a single fixed count (disables adaptive); new `SPARKINFER_NSPLITS_LO/HI/THRESH` tune the schedule.

Split count is occupancy tuning only (the same accuracy-neutral knob `SPARKINFER_NSPLITS` already exposes) — it changes the fp reduction order of the softmax combine, not the attention math. Accuracy held: **top-1 0.970, KL 0.029** vs llama.cpp (held-out prompt). `ctest` 6/6.

## Proof of speedup

**Decode tok/s** (end-to-end, `bench/scripts/bench.sh`, same box, warm, graphics clock pinned 2550 MHz):

| | decode tok/s |
|---|--:|
| before (main) | 405.6 |
| after (this PR) | 407.3 |

_(128-tok, the scored length — within noise. The real gain is at longer context, shown below.)_

```text
RTX 5090 · sm_120 · CUDA 12.8 · Qwen3-30B-A3B Q4_K_M
before = fixed n_splits=32 (== main) · after = adaptive · same box, warm, clock-pinned

              before(main)   after(PR)   delta
  128-tok:      405.55        407.25     +0.4%
  256-tok:      396.92        399.95     +0.8%
  512-tok:      381.05        391.98     +2.9%

accuracy (teacher-forced vs llama.cpp, held-out prompt):
  token-match (top-1): 97/100 = 0.970   (bar >= 0.90)
  mean KL(llama||spark): 0.029 nats
ctest: 6/6 pass
```

### DIFF
--- a/runtime/src/models/qwen35.cpp
+++ b/runtime/src/models/qwen35.cpp
@@ -61,10 +61,18 @@ struct Qwen35Model::Impl {
     uint64_t seq_id = 0;
     int qdim, kvdim;
     bool gguf = false;   // true after load_gguf: dense weights are native [out,in], use GEMV
-    // CUDA-graph capture of the decode compute (captured once, replayed each token)
-    cudaGraph_t cu_graph{};
-    cudaGraphExec_t cu_exec{};
+    // CUDA-graph capture of the decode compute (captured once, replayed each token).
+    // Two graphs are captured with different flash-decode split counts: `cu_exec` (n_splits
+    // = split_lo) fills the GPU best for short context, `cu_exec_hi` (split_hi) for long
+    // context (more splits = more parallel blocks once each KV chunk is large). Each token
+    // replays whichever matches its current seqlen (>= split_thresh -> hi). Split count is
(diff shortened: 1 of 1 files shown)


ai-hpc:
## ⚪ sparkinfer auto-eval — `a534177`

| metric | value |
|---|---|
| **label** | `eval:none` |
| decode | 462.02 tok/s |
| vs frontier | 461.67 tok/s → +0.1% (+0.3) |
| correctness | top-1 98.5% · KL 0.0093 |

PR #122, opened 2026-07-01
perf(attn): fuse QK-norm+RoPE+KV-append and emit Q8_1(attn) in flash combine

fansilas:
## Summary

Two accuracy-preserving fusions in the attention decode epilogue, gated behind `SPARKINFER_ATTNIN` (default on):

1. **Fused QK-norm + RoPE + KV-append.** The per-head Q/K RMSNorm (`rmsnorm_qk`) and the RoPE + paged KV-append (`rope_kv_append`) previously ran as two separate graph nodes, with the normed Q/K round-tripped through global memory between them. `qknorm_rope_kv_kernel` does the per-head RMSNorm, RoPE, and cache write in one kernel: one block per head normalizes its head, stages it in shared memory, ropes the rotate-half pairs, and writes Q in place and K/V straight into the paged cache. One node instead of two, and the intermediate normed-Q/K store then load is removed.

2. **Q8_1(attn) emitted from the flash-decode combine.** The O-projection previously quantized `attn` to Q8_1 with a standalone `quantize_q8_1_blocks` node before the int8 MMVQ. The combine kernel already produces the final bf16 `attn`, so it now also emits the Q8_1 block for the 32 dims each warp owns, deleting the per-layer attn-quantize node. The Q8_1 is derived from the same bf16-rounded output, so it stays bit-identical to running the quantizer afterward.

Both are launch and global-round-trip reductions on the attention epilogue: same online-softmax attention, same rope angles, same per-head RMS, same int8 O-projection math. Distinct from the open MoE/MMVQ and LM-head PRs (this touches only the attention epilogue kernels).

## Proof of speedup

Decode tok/s (end-to-end, `qwen3_gguf_bench`, Qwen3-30B-A3B Q4_K_M, n=128, bs=1). Interleaved A/B on one binary (`SPARKINFER_ATTNIN` toggled), 6 rounds each:

| | decode tok/s |
|---|--:|
| before (`SPARKINFER_ATTNIN=0`, == main) | 453.05 |
| after (this PR, default) | 468.01 |

+14.96 tok/s (+3.30%). `SPARKINFER_ATTNIN=0` reproduces current `main` exactly. `nsys --cuda-graph-trace=node` confirms `rmsnorm_qk` + `rope_kv_append` (2.5% + 2.2% of GPU time) collapse to one `qknorm_rope_kv` kernel (3.2%), and the per-layer attn-quantize node drops from 1.7% to 0.04% (layer 0 only).

## Accuracy

Teacher-forced vs llama.cpp on the same GGUF: top-1 token-match 0.970 (bar >= 0.90, equals main), mean KL 0.0286 nats (bar <= 0.20, preferred <= 0.15).

## Correctness

- `ctest` 6/6 passed
- `compute-sanitizer --tool memcheck` 0 errors

### DIFF
--- a/kernels/csrc/cuda/attention/flash_decode_split.cu
+++ b/kernels/csrc/cuda/attention/flash_decode_split.cu
@@ -76,15 +76,22 @@ __global__ void fa_split_kernel(
     for (int e = 0; e < ELEMS; e++) part_acc[(size_t)idx * HEAD_DIM + lane + e * 32] = acc[e];
 }
 
+// llama Q8_1 activation block (matches si_block_q8_1 used by the int8 MMVQ O-projection).
+struct fa_block_q8_1 { __half2 ds; signed char qs[32]; };
+
 // Combine the split partials with DG x NW parallelism over the 1-block-per-head
 // original (which idled at ~2% occupancy with a serial n_splits loop). DG head-dim
 // groups -> DG x more blocks; NW warps per block each fold a 1/NW stripe of the
 // splits, then a shared-memory log-sum-exp merge across warps. grid=(heads*DG,seqs).
+// When out_q8 != nullptr AND ELEMS==1 (DG*32==HEAD_DIM), each (qh,dg) block's warp 0 also
+// emits the Q8_1 block for attn dims [qh*HEAD_DIM + dg*32, +32) from the bf16-rounded output,
+// so the O-projection MMVQ skips its standalone attn-quantize node (bit-identical to running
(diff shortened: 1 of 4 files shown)


ai-hpc:
## ✅ sparkinfer auto-eval — `bd1c3f6`

| metric | value |
|---|---|
| **label** | `eval:L` |
| decode | 479.83 tok/s |
| vs prior frontier | 463.92 tok/s → +3.4% (+15.9) |
| **→ new frontier** | **479.83 tok/s** |
| correctness | top-1 97.5% · KL 0.0126 |

Verified speedup — **sets the new frontier to 479.83 tok/s** (was 463.92).

PR #148, opened 2026-07-03
fix(runtime): release device weights on reload

RealDiligent:
## Problem

Qwen35Model::load_gguf() and load_weights() push every new weight buffer into owned but never free prior entries. A second load on the same model instance duplicates the full weight footprint in VRAM (~15+ GB for Qwen3-30B) until process exit.

Worse, if decode has already captured a CUDA graph, that graph holds **device pointers into the old owned buffers**. Freeing without invalidating the graph would be use-after-free; keeping the leak avoids the crash but guarantees OOM on any hot-reload path.

set_weights() had a related footgun: it overwrote w while leaving owned (and gguf=true) from a prior load_gguf(), leaking VRAM and routing decode through the wrong GGUF kernels.

## Fix

Add Impl::release_device_weights():

1. Invalidate any captured decode graph (stale weight pointers)
2. cudaFree all owned buffers and clear the vector
3. Reset w to empty handles and gguf=false

Call it at the start of load_gguf(), load_weights(), and set_weights().

Also factor invalidate_decode_graph() for the existing adaptive 
_splits path.

## Impact

- **Hot model reload** on a long-lived Qwen35Model no longer leaks VRAM or replays freed weights.
- **Failed reload** after a partial load_gguf() leaves a clean empty weight state instead of a half-loaded model.
- **set_weights() after load_gguf()** correctly switches to dense .bin decode paths.

## Test plan

- [ ] load_gguf -> generate -> load_gguf(same) -> generate - VRAM stable (no doubling), tokens match first run
- [ ] load_gguf -> set_weights(external) - no VRAM leak; decode uses GEMM path (gguf=false)
- [ ] Existing qwen35_gpu_test unchanged (single load path)

## Risks

None for the common one-shot load-then-run path. Reload semantics change from silent leak to correct release - intentional.

### DIFF
--- a/runtime/src/models/qwen35.cpp
+++ b/runtime/src/models/qwen35.cpp
@@ -97,6 +97,25 @@ struct Qwen35Model::Impl {
                            // next layer's standalone QKV-input quantize node. =0 disables
 
     template <class T> T* alloc(size_t n) { void* p=nullptr; cu(cudaMalloc(&p, n*sizeof(T)), "malloc"); return (T*)p; }
+
+    void invalidate_decode_graph() {
+        if (graph_ready) {
+            cudaGraphExecDestroy(cu_exec);
+            cudaGraphDestroy(cu_graph);
+            graph_ready = false;
+        }
+    }
+
+    // Drop prior load_weights/load_gguf device allocations before binding new weights.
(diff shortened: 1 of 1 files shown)


ai-hpc:

PR #149, opened 2026-07-03
fix(decode): wire QK-norm and RoPE into DecodeRunner attention

RealDiligent:
## Problem

DecodeRunner::decode_layer() ran Q/K/V GEMMs then called launch_kv_append + launch_flash_decode_gqa8 directly. It never applied per-head QK-norm or RoPE.

Qwen3/Qwen3.5 models require both before KV write and attention (Qwen35Model::forward_token uses launch_qknorm_rope_kv_append). Without them, flash-decode reads **un-rotated, unnormed** keys - systematic logit error for any batched decode path built on DecodeRunner.

decode_runner_gpu_test only checked finite output, so this stayed hidden.

## Fix

- Extend AttnConfig with 
ope_theta and 
ms_eps (Qwen defaults: 1e6, 1e-6).
- Extend TransformerLayerWeights with optional q_norm / k_norm ([head_dim]).
- After Q/K/V projection:
  - q_norm && k_norm ? launch_qknorm_rope_kv_append (Qwen path)
  - else 
ope_theta > 0 ? launch_rope_kv_append
  - else ? legacy launch_kv_append (non-RoPE models)
- GPU test supplies Q/K norm weights and Qwen 
ope_theta.

## Impact

Correct attention math for continuous-batching / multi-seq decode via DecodeRunner. Complements #144 (block-table pack) - that PR fixes *which* table row is read; this fixes *what* is stored in KV.

## Test plan

- [ ] decode_runner_gpu_test on RTX 5090 (finite output, exercises fused QK-norm+RoPE+KV path)
- [ ] Future: compare against qwen35_cpu_test reference for a single-layer step

## Risks

Backward compatible: callers that omit q_norm/k_norm and set 
ope_theta=0 keep the old raw-KV path.

### DIFF
--- a/runtime/include/sparkinfer/decode.h
+++ b/runtime/include/sparkinfer/decode.h
@@ -11,7 +11,9 @@ struct AttnConfig {
     int num_q_heads;
     int num_kv_heads;
     int head_dim;
-    float scale;     // 1/sqrt(head_dim)
+    float scale;        // 1/sqrt(head_dim)
+    float rope_theta = 1e6f;
+    float rms_eps    = 1e-6f;
 };
 
 // Device weight pointers (bf16) for one MoE transformer layer.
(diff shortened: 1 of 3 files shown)


ai-hpc:

PR #151, opened 2026-07-03
fix(runtime): rollback partial GGUF and weight loads on failure

RealDiligent:
## Problem

load_gguf() had several failure-mode hazards:

1. **gguf=true before g.open()** - a bad path left the GGUF decode path enabled with no valid weights.
2. **No rollback on mid-load failure** - a missing tensor, unsupported quant type, or layer-N error returned alse but left partial owned buffers and half-populated s.w.layers. Callers that ignored the return (or retried) could hit null-weight kernels or mixed state.
3. **Unchecked cudaMalloc in dense()** - VRAM OOM passed 
ullptr into launch_gguf_dequant ? GPU fault during load.

load_weights() had the same partial-state problem on failure.

## Fix

- Add 
ollback_load_state(): invalidate decode graph, free all owned buffers, reset w, clear gguf.
- Call it at the **start** of each load (clean slate) and on **every** failure via a ail() helper.
- Set gguf=true only **after** g.open() succeeds.
- Check cudaMalloc in dense() transpose path; abort without leaking temp buffers.

## Impact

- Failed GGUF/weight loads leave the model in a safe empty state instead of a toxic half-loaded one.
- OOM during dense dequant fails closed instead of crashing in kernels.
- Also frees prior weights at reload start (overlaps #148 intent; safe to merge either first).

## Test plan

- [ ] Truncated/corrupt GGUF ? load_gguf returns false, orward_token rejected (no weights)
- [ ] Successful load unchanged
- [ ] load_gguf twice on same model - VRAM does not grow without bound

## Risks

None for happy path. Failed loads now explicitly clear state - callers must re-load before decode (correct behavior).

### DIFF
--- a/runtime/src/models/qwen35.cpp
+++ b/runtime/src/models/qwen35.cpp
@@ -97,6 +97,24 @@ struct Qwen35Model::Impl {
                            // next layer's standalone QKV-input quantize node. =0 disables
 
     template <class T> T* alloc(size_t n) { void* p=nullptr; cu(cudaMalloc(&p, n*sizeof(T)), "malloc"); return (T*)p; }
+
+    void invalidate_decode_graph() {
+        if (graph_ready) {
+            cudaGraphExecDestroy(cu_exec);
+            cudaGraphDestroy(cu_graph);
+            graph_ready = false;
+        }
+    }
+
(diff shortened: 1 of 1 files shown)


ai-hpc:

PR #153, opened 2026-07-03
fix(qwen35): reject null Q/K norm weights before fused attention kernels

RealDiligent:
## Problem

The production `Qwen35Model` decode path defaults `SPARKINFER_ATTNIN=1`, which calls `launch_qknorm_rope_kv_append` every layer. That kernel reads per-head norm weights via `w[t]` in `qknorm_rope_kv_kernel`.

However:

- `load_gguf` / `load_weights` only required Q/K/V, router, and MoE tensors - **not** `q_norm` / `k_norm`
- A missing `attn_q_norm.weight` (or failed `dense()`) left **null pointers** while load still returned **true**
- `set_weights()` has no validation at all

First `forward_token()` ? null deref in the fused kernel ? **GPU illegal access**. This is distinct from PR **#149**, which fixes the separate `DecodeRunner` API path.

## Fix

1. **Load time:** fail if any layer is missing `q_norm` or `k_norm`
2. **Forward time:** `use_attnin = s.use_attnin && w.q_norm && w.k_norm`; same guard for `use_qkfuse` / per-head RMSNorm fallbacks
3. When norms are absent (e.g. partial `set_weights`), skip norm and proceed with RoPE + KV-append - avoids crash, clearly wrong for Qwen3.5 but safe

## Impact

- Prevents immediate GPU fault on incomplete weight binding
- Real Qwen3.5 GGUF/bin loads unchanged (norm tensors present)
- Complements #149 (DecodeRunner) without overlapping it

## Test plan

- [ ] `qwen35_gpu_test` / `qwen3_gguf_generate` on RTX 5090 - unchanged output
- [ ] Deliberately omit `layer_0.q_norm.bin` from a weight dir ? `load_weights` returns false
- [ ] `set_weights` with null `q_norm` ? decode runs (RoPE-only fallback) without CUDA fault

## Risks

Minimal for valid Qwen3.5 checkpoints. Fallback without norms is not model-correct but only reachable via API misuse.

### DIFF
--- a/runtime/src/models/qwen35.cpp
+++ b/runtime/src/models/qwen35.cpp
@@ -298,15 +298,19 @@ int Qwen35Model::forward_token(int token_id, int position) {
         }
         bf16* kpool = (bf16*)s.kv->k_pool() + (size_t)L * s.kv->layer_stride_elems();
         bf16* vpool = (bf16*)s.kv->v_pool() + (size_t)L * s.kv->layer_stride_elems();
-        if (s.use_attnin) {   // fused QK-norm + RoPE + KV-append: 1 node vs qk-norm + rope-kv (2)
+        // Qwen3.5 requires per-head Q/K RMSNorm weights. load_* validates them, but
+        // set_weights() can still leave nulls; never pass those into the fused kernels.
+        const bool has_qknorm = w.q_norm && w.k_norm;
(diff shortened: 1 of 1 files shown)


ai-hpc:

PR #154, opened 2026-07-03
fix(embedding): clamp out-of-range token ids in GPU gather kernel

RealDiligent:
## Problem

The token embedding gather has no vocab bounds check:

```
id = ids[t];
out[...] = table[id * hidden + h];   // id unchecked
```

On main, `forward_token` / `generate` accept arbitrary `token_id` values. Once a CUDA graph is captured, `d_tok` is updated each step (host upload or `decode_feedback` on the bench path) and `embedding_kernel` replays without any host re-validation in the graph body.

An out-of-range id (bad prompt token, tokenizer bug, or corrupted `d_tok`) casts `id * hidden` to `size_t` and indexes megabytes past `embed_tokens` -> **GPU illegal access**.

PR **#143** adds host-side rejection but does not protect the device path inside a captured graph.

## Fix

- Thread `vocab` through `launch_embedding` / `embedding_kernel`
- Clamp `ids[t]` to `[0, vocab)` before the table gather (use token 0 as safe fallback)

## Impact

- Last-line GPU defense for embed-table OOB on the decode hot path
- Valid tokens unchanged; only affects already-undefined ids
- Complements #143 host validation without duplicating it

## Test plan

- [ ] `qwen35_gpu_test` / `qwen3_gguf_generate` on RTX 5090
- [ ] `forward_token(vocab, 0)` must not fault (clamps to 0)
- [ ] Normal prompt tokens produce identical output to main

## Risks

Minimal. Clamping to token 0 on invalid ids may mask upstream bugs, but is strictly safer than OOB.

### DIFF
--- a/kernels/csrc/cuda/fused/model_ops.cu
+++ b/kernels/csrc/cuda/fused/model_ops.cu
@@ -11,12 +11,15 @@ namespace sparkinfer {
 namespace kernels {
 
 // out[t, :] = table[ids[t], :]   grid = n_tokens, threads over hidden.
+// ids outside [0, vocab) clamp to 0 — CUDA-graph replay and decode_feedback
+// can reach this kernel without a host-side token check on every path.
 __global__ void embedding_kernel(const int* __restrict__ ids,
                                  const __nv_bfloat16* __restrict__ table,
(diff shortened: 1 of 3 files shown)


ai-hpc:

PR #195, opened 2026-07-03
perf(attn): int8 KV tensor-core GQA flash-decode (+19% decode at 16k)

fansilas:
## Summary

Long-context decode is dominated by `fa_split_gqa` (~34% of decode at 16k). A gutted-dot experiment (keep the KV loads, remove the per-token dot + reduction) speeds the kernel up ~25%, showing it is compute-bound on the per-token QK dot, the 5-shuffle `fa_wsum` warp reduction, and the V-accumulate, not on memory bandwidth. This change moves that work onto the tensor cores and halves the KV read, and does so with **zero regression** at short and mid contexts.

**int8 tensor-core split.** For block-aligned long context the 8 GQA q-heads of a kv-head become the batch (M) dimension, so `S = Q·Kt` and `O = P·V` become small `wmma` matmuls, replacing the per-lane FMA + shuffle reduction. K/V are stored int8 (Q8-style, one fp16 scale per token per kv-head) so the matmuls run entirely on the 2x-throughput int8 tensor cores (int32 accumulate) while the KV global read (the exposed cost once the dot leaves the ALUs) is halved. Q and P are quantized per-row and the per-token/per-head scales are applied to the int32 results (the V scale folds into P). M is padded 8->16; partials (m, l, acc) stay byte-compatible with the existing combine kernel. Fragments load straight from the paged KV pool (one 16-token physical block per warp), so there is no shared-memory staging and occupancy stays high.

**Context-adaptive, zero-regression elsewhere.** The tensor-core kernel runs only for block-aligned long context with a per-split chunk of at least 32 tokens (enough physical blocks to fill the GPU), and only with the int8 cache. int8's fixed per-token write cost only pays off once the halved long-context read outweighs it, so the int8 cache is enabled context-adaptively (from 8k); shorter contexts stay on the scalar bf16 split. Both touched kernels (the scalar `fa_split_gqa` and the fused QK-norm+RoPE+KV-append) are template-specialized on a compile-time `int8` flag, so the int8-off instantiation carries no int8 code or pointer indirection and is byte-identical to the pre-change (main) kernel. Result: 128/512/4k decode is unchanged (within run-to-run noise), only 16k engages the int8 tensor cores. `SPARKINFER_KV_INT8=1/0` forces int8 on/off; `SPARKINFER_FAMMA=0` restores the scalar split.

## Proof of speedup

**Decode tok/s** (end-to-end, `bench/scripts/bench.sh --tokens 128 --ctx 16384`, greedy bs=1, Qwen3-30B-A3B Q4_K_M, the scored long-context regime where the int8 path engages; median of 3 runs):

| | decode tok/s |
|---|--:|
| before (main) | 257.3 |
| after (this PR) | 307.9 |

That is **+19.7%** at 16k. int8 engages context-adaptively (from 8k), so short contexts stay on the bf16 path; the full no-regression sweep (this PR vs main, same box, `bench.sh`, n=128, median of 3):

| context | main (bf16) | this PR | delta |
|---|--:|--:|--:|
| 128  | 485.8 | 487.2 | +0.3% |
| 512  | 471.5 | 471.6 | +0.0% |
| 4k   | 387.4 | 387.5 | +0.0% |
| 16k  | 257.3 | 307.9 | **+19.7%** |

Only the 16k path switches to int8 tensor cores; 128/512/4k stay on the scalar bf16 path, byte-identical to main.

## Correctness

token-match / KL vs llama.cpp on the same GGUF (`bench/scripts/accuracy.sh`, held-out prompt):

- short context (scalar int8 read): top-1 0.960, mean KL 0.0495 nats
- long context (int8 tensor-core, `SPARKINFER_NSPLITS=16`): top-1 0.980, mean KL 0.0368 nats

Both clear the 0.90 top-1 / 0.20 KL bars with wide margin. `SPARKINFER_KV_INT8=0` is bit-identical to the bf16 cache; `SPARKINFER_FAMMA=0` restores the scalar split. `ctest` 6/6 green; compute-sanitizer memcheck clean on the int8 decode path.

### DIFF
--- a/kernels/csrc/cuda/attention/flash_decode_split.cu
+++ b/kernels/csrc/cuda/attention/flash_decode_split.cu
@@ -253,6 +253,141 @@ template __global__ void fa_combine_kernel<128, FA_COMBINE_DG, 16>(const float*,
 
 #ifndef SPARKINFER_NVRTC_DEVICE_ONLY
 #include "sparkinfer/kernels/attention.h"
+#include <mma.h>
+
+// Tensor-core (wmma bf16) GQA flash-decode split. The 8 GQA q-heads of a kv-head are the batch (M)
+// dim, so S = Q[16x128]·Kᵀ[128xT] and O = P[16xT]·V[Tx128] become small bf16 matmuls on the tensor
+// cores — replacing the per-lane FMA + 5-shuffle fa_wsum reduction that dominates the scalar kernel's
+// compute at long context. M is padded 8->16 (upper 8 q-rows zero). Byte-compatible partials
+// (m,l,acc) for the existing combine kernel. Online-softmax loop over NT-token tiles handles any
+// chunk. sm_80+ (wmma bf16). One block per (seq, kv_head, split); 8 warps.
+template <int HEAD_DIM, int GQA>
+__global__ void fa_split_gqa_mma_kernel(
+    const __nv_bfloat16* __restrict__ q, const __nv_bfloat16* __restrict__ k_pool,
+    const __nv_bfloat16* __restrict__ v_pool, const int* __restrict__ block_table,
+    const int* __restrict__ seq_lens,
+    float* __restrict__ part_m, float* __restrict__ part_l, float* __restrict__ part_acc,
+    float scale, int num_q_heads, int num_kv_heads, int block_size, int max_blocks, int n_splits
+) {
+    using namespace nvcuda::wmma;
+    constexpr int NT = 128;              // tokens per tile
+    constexpr int KH = HEAD_DIM / 16;    // 16-wide steps across head_dim (8)
+    const int seq   = blockIdx.y;
+    const int split = blockIdx.x % n_splits;
+    const int kvh   = blockIdx.x / n_splits;
+    const int warp  = threadIdx.x >> 5;  // 0..7
+    const int lane  = threadIdx.x & 31;
+    const int tid   = threadIdx.x;
+
+    const int sl    = seq_lens[seq];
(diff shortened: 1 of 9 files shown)


ai-hpc:

```bash
```

fansilas:
Hi @ai-hpc 
Could you please review this PR again?

ai-hpc:
## ✅ sparkinfer auto-eval — `3c81876`

| metric | value |
|---|---|
| **label** | `eval:XL` |
| scored decode (16384 ctx · 16k-context) | 308.02 tok/s |
| vs prior frontier | 263.85 tok/s → +16.7% (+44.2) |
| **→ new frontier** | **308.02 tok/s** |
| correctness | top-1 95.4% · KL 0.0209 |
| 128-token no-regression gate | 494.59 tok/s vs main 496.49 tok/s · pass |
| 512-context no-regression gate | 470.3 tok/s vs main 471.5 tok/s · pass |
| 4k-context no-regression gate | 392.88 tok/s vs main 393.75 tok/s · pass |
| 16k-context no-regression gate | 308.02 tok/s vs main 263.85 tok/s · pass |

Verified speedup — **sets the new frontier to 308.02 tok/s** (was 263.85).

_RTX 5090 (sm_120) · 128/512/4k/16k guarded · strongest context scores · built from source · correctness vs llama.cpp. Automated — **not merged**; merge manually after review._

ai-hpc:
✅ Auto-merged as the round's `merge-first` winner — verified same-box speedup over `main`, all checks green. Thanks for the contribution!

PR #218, opened 2026-07-04
fix(runtime): guarantee forward progress in Scheduler::schedule() (fixes #217)

### LINKED ISSUE #217: fix(runtime): Scheduler::schedule() starves a sequence group larger than max_tokens_per_batch
`Scheduler::schedule()` breaks out of the packing loop on the first group that exceeds the remaining token budget. When the highest-priority group's `num_seqs` is greater than `max_tokens_per_batch`, it breaks on the very first iteration and returns an empty `ScheduleBatch` - so that group never makes forward progress (starvation/deadlock), even though it is the only runnable work.

**Expected:** `schedule()` should never return an empty batch while a runnable group exists; the highest-priority group should be admitted even if it alone overflows the soft token budget.

No test currently covers the scheduler.

minion1227:
## Summary

Fixes #217. `Scheduler::schedule()` returned an empty batch when the highest-priority sequence group's `num_seqs` exceeded `max_tokens_per_batch` (it `break`s on the first iteration), so that group could never run - starvation / no forward progress. This admits the top-priority group even when it alone overflows the soft token budget, guaranteeing a batch is never empty while runnable work exists. Strict-priority behaviour is unchanged once at least one group is already scheduled.

Also adds `scheduler_cpu_test` - the first test coverage for the scheduler (priority packing, token-budget stop, the forward-progress regression, empty scheduler, preemption order). Pure host logic, no GPU.

## Proof of speedup

Not a performance change - this is a host-side correctness fix plus new CPU test coverage, off the decode/benchmark path. No decode tok/s impact, so the RTX 5090 box is intentionally left unticked.

**Decode tok/s:** n/a (no runtime speed change).

### DIFF
--- a/runtime/src/scheduler.cpp
+++ b/runtime/src/scheduler.cpp
@@ -28,7 +28,13 @@ ScheduleBatch Scheduler::schedule() {
     std::sort(ordered.begin(), ordered.end(),
               [](const SequenceGroup* a, const SequenceGroup* b) { return a->priority > b->priority; });
     for (auto* g : ordered) {
-        if (batch.total_tokens + g->num_seqs > impl_->max_tokens_per_batch) break;
+        // Strict priority: stop at the first group that doesn't fit the token budget — but
+        // never return an empty batch while a runnable group exists. A group whose num_seqs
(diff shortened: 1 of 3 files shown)


ai-hpc:

PR #220, opened 2026-07-04
fix(runtime): handle tied embeddings in convert_qwen35.py (fixes #219)

### LINKED ISSUE #219: fix(runtime): convert_qwen35.py crashes (KeyError) on tied-embedding checkpoints
`convert_qwen35.py` unconditionally reads `lm_head.weight`:

```python
write("lm_head", get("lm_head.weight").T)
```

When a Qwen checkpoint **ties** its input/output embeddings, there is no `lm_head.weight` tensor, so `get()` raises `KeyError` and conversion aborts.

The sibling `convert_gguf.py` already handles this case with a tied fallback to the input embedding (`output.weight` -> `token_embd.weight`). `convert_qwen35.py` should do the same: when `lm_head.weight` is absent, fall back to `model.embed_tokens.weight` (identical `[vocab, hidden]` layout).

minion1227:
## Summary

Fixes #219. `convert_qwen35.py` read `lm_head.weight` unconditionally, so a checkpoint that ties its input/output embeddings (no `lm_head.weight` tensor) aborted conversion with a `KeyError`. This falls back to `model.embed_tokens.weight` when `lm_head.weight` is absent - the same `[vocab, hidden]` layout, mirroring `convert_gguf.py`'s existing tied-weight handling. Untied checkpoints are unaffected.

## Proof of speedup

Not a performance change - host-side weight-conversion robustness fix, off the decode/benchmark path. No decode tok/s impact, so the RTX 5090 box is intentionally left unticked.

**Decode tok/s:** n/a (no runtime speed change).

### DIFF
--- a/runtime/tools/convert_qwen35.py
+++ b/runtime/tools/convert_qwen35.py
@@ -60,7 +60,10 @@ def main():
     # globals
     write("embed_tokens", get("model.embed_tokens.weight"))          # [vocab, H]
     write("final_norm",   get("model.norm.weight"))                  # [H]
-    write("lm_head",      get("lm_head.weight").T)                   # [vocab,H] -> [H,vocab]
+    # lm_head.weight is absent when the model ties input/output embeddings; fall back to
+    # the input embedding (same [vocab, H] layout), matching convert_gguf.py's tied handling.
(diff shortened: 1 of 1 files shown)


ai-hpc:

PR #222, opened 2026-07-04
perf(attn): cut i8 mma split-kernel smem sync and epilogue padding (+3.8% @16k)

pekar9781:


pekar9781:
Closing duplicate; reopening #209 instead.

PR #236, opened 2026-07-05
perf(qwen3.6): default the merged gdn_ar_fast_kernel ON

jason020818:
## Summary

`SPARKINFER_GDN_FAST` (merged in #229) only activates `gdn_ar_fast_kernel` when the
env var is explicitly set to `1`. **The eval harness (`bench/scripts/evaluate.sh`,
`evaluate_dual.sh`) never sets feature env vars** -- it only passes
guard/baseline/difficulty scoring parameters (`SPARKINFER_GUARD_*`,
`SPARKINFER_DIFFICULTY_*`, `SPARKINFER_EVAL_MODE`, etc.) -- so the merged kernel has
never actually run during scoring. The naive `gdn_ar_kernel`
(`<<<v_heads=32, head_dim>>>`, ~81% of the SMs idle every token, two serial
head_dim-long dependent-load loops, as described in #229's own summary) has stayed
the decode path's largest single non-GEMV cost the whole time.

This flips the default to **ON** (opt-out via `SPARKINFER_GDN_FAST=0`), matching
every other feature flag in this runtime (`use_pq`, `use_llama`, `use_q6mmvq`,
`use_fnq` are all default-on / opt-out-via-`0`). **No kernel code changes** --
`gdn_ar_fast_kernel` is #229's kernel, unmodified; this is a one-line default flip
plus a comment explaining why.

Scope: `kernels/csrc/cuda/fused/qwen36.cu`, single function
(`launch_qwen36_gdn_ar`), single line changed.

## Proof of speedup

- [x] Tested on **RTX 5090** (`sm_120`)

**Decode tok/s** (`bench/scripts/bench.sh`, Qwen3.6-35B-A3B-UD-Q4_K_M.gguf, bs=1, same box):

| context | before (main default) | after (this PR) | delta |
|---:|---:|---:|---:|
| 128 | 171.90 | 224.47 | **+30.6%** |
| 512 | 170.46 | 222.07 | **+30.3%** |

```text
=== before (main, SPARKINFER_GDN_FAST unset -> naive gdn_ar_kernel) ===
model        : Qwen3.5/Qwen3.6-35B-A3B hybrid  (40 layers, 256 experts top-8)
VRAM used    : 25.3 GB
decode tg    : 171.90 tok/s  (5.8 ms/token, n=128, ctx=0, bs=1)
decode tg    : 170.46 tok/s  (5.9 ms/token, n=512, ctx=0, bs=1)

=== after (this PR, default now runs gdn_ar_fast_kernel) ===
[...]

PR #238, opened 2026-07-05
perf(qwen3.6): enable fast Gated-DeltaNet AR state kernel on the default decode path

fansilas:
## Summary

The optimized Gated-DeltaNet AR state-update kernel (`gdn_ar_fast_kernel` in `kernels/csrc/cuda/fused/qwen36.cu`) is compiled in but gated behind `SPARKINFER_GDN_FAST`, which defaults OFF and is not set anywhere in the bench or eval path. So Qwen3.6 decode still runs the naive `gdn_ar_kernel<<<v_heads=32, head_dim=128>>>` on every one of the 30 linear layers, every token: 32 blocks on 170 SMs (about 81% of the GPU idle), two serial 128-iteration dependent-load loops over the recurrent state, and a write-then-reread of the fp32 state (roughly 4x its traffic).

This PR flips that one default so the fast kernel runs on the default decode path (warp-per-state-column, register-cached state slice for one read plus one write, transposed-for-coalescing state layout, filling the grid). `SPARKINFER_GDN_FAST=0` still selects the naive kernel as a fallback. The transposed internal state is safe as a default: the recurrent state is zero-initialized and touched only by this kernel, the dispatch flag is static (all-or-nothing for the run, so prefill and decode agree), and the layout never escapes the kernel.

## Proof of speedup

**Decode tok/s** (Qwen3.6-35B-A3B UD-Q4_K_M, single stream, bs=1, interleaved A/B on one build via the env toggle):

| | decode tok/s |
|---|--:|
| before (main) | 168.72 |
| after (this PR) | 219.97 |

Per scored context (before to after, median of interleaved pairs):

| context | before (naive) | after (fast) | gain |
|---:|--:|--:|--:|
| 128 | 168.72 | 219.97 | +30.4% |
| 512 | 168.80 | 219.48 | +30.0% |
| 4k  | 162.65 | 209.44 | +28.8% |

```text
# main (SPARKINFER_GDN_FAST unset -> naive kernel), ctx=128:
decode tg    : 168.72 tok/s  (5.9 ms/token, n=128, ctx=128, bs=1)
# this PR (fast kernel is the default), ctx=128:
decode tg    : 219.97 tok/s  (4.5 ms/token, n=128, ctx=128, bs=1)
```

## Correctness

The fast kernel is the one added in the merged GDN AR work; it is not bit-identical to the naive kernel (the warp-tree reduction reorders the fp32 sum), so it is validated by output agreement rather than a byte compare. Teacher-forced on the eval corpus, fast vs naive argmax agree 97/100 and perplexity is identical (2.7512 vs 2.7530), i.e. within fp reassociation noise, so token-match and KL versus llama.cpp are unchanged from the naive path (which the correctness gate already passes). `ctest` is 7/7.

## No-regression (Qwen3-30B guard)

The change lives entirely inside `launch_qwen36_gdn_ar`, which is only reached under `w.linear_attn` (the Qwen3.6 Gated-DeltaNet layers). Qwen3-30B-A3B has no linear-attention layers, so its decode path is byte-identical and unaffected.

### DIFF
--- a/kernels/csrc/cuda/fused/qwen36.cu
+++ b/kernels/csrc/cuda/fused/qwen36.cu
@@ -300,11 +300,13 @@ void launch_qwen36_gdn_ar(const void* q_bf16, const void* k_bf16, const void* v_
                           const void* dt_bf16, const void* a_bf16,
                           float* state_f32, void* out_bf16,
                           int q_heads, int v_heads, int head_dim, cudaStream_t stream) {
-    // SPARKINFER_GDN_FAST: warp-per-column, register-cached, transposed-state kernel (fills the GPU +
-    // 2x state traffic). Uses a transposed internal state layout, so it MUST be all-or-nothing for the
-    // run — the static flag guarantees that. Requires head_dim a multiple of 32 (128 -> NROW=4).
+    // Warp-per-column, register-cached, transposed-state kernel (fills the GPU + halves state traffic).
+    // Default ON for the Qwen3.6 linear_head_dim==128 path; SPARKINFER_GDN_FAST=0 restores the naive
+    // kernel. Uses a transposed internal state layout, so it MUST be all-or-nothing for the run — the
+    // static flag guarantees that, and the recurrent state is zero-init and touched only here (so the
+    // layout never escapes this kernel). Requires head_dim a multiple of 32 (128 -> NROW=4).
     static int fast = -1;
-    if (fast < 0) { const char* e = getenv("SPARKINFER_GDN_FAST"); fast = (e && e[0] == '1') ? 1 : 0; }
+    if (fast < 0) { const char* e = getenv("SPARKINFER_GDN_FAST"); fast = (e && e[0] == '0') ? 0 : 1; }
     if (fast && head_dim == 128) {                            // Qwen3.6 linear_head_dim; other dims -> naive
         constexpr int COLS = 8, HD = 128;                     // 8 warps (columns)/block -> 256 threads
         dim3 grid(v_heads, (HD + COLS - 1) / COLS);


ai-hpc:
## 🐈 Flagged: copycat (retroactive sweep)

This PR is **>=80% contained** in the earlier #236 by jaso0n0818 — a verbatim copy by a different author. Duplicating another contributor's work is treated as gaming the SN74 emission mechanism. The scheduled bot will not evaluate this PR.

See [`.github/COPYCATS.md`](../blob/main/.github/COPYCATS.md).

ai-hpc:
Closed — copycat detection. See the comment above for details. The original author's work takes priority.

fansilas:
@ai-hpc 
Please review my PR manually. I'm really embarrassed.

ai-hpc:
## ✅ Manual review — not a copycat

This was reviewed by a maintainer. The overlap with #236 is a **coincidence**, not a copycat — both PRs independently enabled the GDN fast kernel on the default decode path (a natural, small change). fansilas is a known contributor with multiple merged PRs and a clean history. No action taken — the  label has been removed, the strike cleared from the log, and the PR reopened. Apologies for the false flag.

ai-hpc:
## 🐈 Flagged: copycat

This PR re-submits substantially the same diff as the earlier #236. Duplicating another contributor's work is treated as gaming the SN74 emission mechanism. The account has been **blocked** and this PR **closed** — zero tolerance, no warning. See [`.github/COPYCATS.md`](../blob/main/.github/COPYCATS.md).

ai-hpc:
## 🚩 Flagged: eval-gaming

This PR involves an account blocked for gaming the SN74 emission mechanism (sybil / coordinated duplicate farming): `fansilas`.

PR #240, opened 2026-07-05
perf(qwen3.6): fast Gated-DeltaNet AR kernel + conv/L2-norm and shared-expert GEMV fusions

fansilas:
## Summary

Three optimizations on the Qwen3.6 decode path, together +34% at 128/512 context and +33% at 4k:

1. **Enable the fast Gated-DeltaNet AR state kernel** (30 linear layers). The optimized `gdn_ar_fast_kernel` is compiled in but gated behind `SPARKINFER_GDN_FAST`, which defaulted OFF and was set nowhere, so decode ran the naive `gdn_ar_kernel<<<v_heads=32, head_dim=128>>>` every token: 32 blocks on 170 SMs (about 81% idle) with two serial 128-iteration dependent-load loops over the recurrent state and a write-then-reread of the fp32 state. This flips that default to the fast path (warp-per-state-column, register-cached state slice, transposed-for-coalescing layout, filling the grid). `SPARKINFER_GDN_FAST=0` restores the naive kernel.

2. **Fuse the GDN conv with the per-head L2-norm** (30 linear layers), byte-identical. The causal short-conv and the two q/k L2-norms were three kernels (conv wrote q/k/v, then two norm launches read q/k back, normalized, and rewrote them). Fused into one `conv_split_l2_fused_kernel` (one block per head) that normalizes q/k in-block, writing them once instead of round-tripping through global.

3. **Fuse the shared-expert gate and up GEMVs** (every layer), byte-identical. The gate and up projections read the same input but ran as two `launch_gemv` calls of 512 rows each (64 blocks on 170 SMs). Fused into one `launch_gemv2` over both weight matrices (1024 rows, 128 blocks, input staged    once). `launch_gemv2` is a new function; the existing `launch_gemv` is untouched.

The recurrent state is zero-initialized and touched only by the AR kernel, so the transposed layout is self-consistent as a default. Qwen3-30B has no Gated-DeltaNet layers and no shared expert (`n_shared` is forced to 0 when the GGUF has no shared tensors), so its decode path is unchanged.

## Proof of speedup

- [x] Tested on **RTX 5090** (`sm_120`)

**Decode tok/s** (Qwen3.6-35B-A3B UD-Q4_K_M, single stream, bs=1, interleaved A/B on one build via the env toggles):

| | decode tok/s |
|---|--:|
| before (main) | 168.67 |
| after (this PR) | 226.43 |

Per scored context (before to after, median of interleaved pairs):

| context | before | after | gain |
[...]

PR #233, opened 2026-07-05
perf(qwen36): warp-grid GDN + decode pipelining + Q8_0 shared MMVQ

jimcody1995:
## Summary

Five complementary decode optimizations for Qwen3.6-35B-A3B:

1. **`gdn_ar_warpgrid_kernel`** — default-on warp-per-column Gated-DeltaNet AR update that keeps the native `[vh][row][col]` fp32 state layout (no transposed-state mode). Replaces the naive `<<<v_heads, head_dim>>>` launch (~81% SM idle) with a full-grid schedule. Opt out: `SPARKINFER_GDN_WARPGRID=0`.
2. **Side-stream decode pipelining** — overlap independent work on `stream_k` / `stream_v`:
   * GDN layers: `wqkv_gate`, `ssm_alpha`, `ssm_beta` run concurrently with `wqkv` (`SPARKINFER_GDN_PIPE=0` disables).
   * MoE layers: shared-expert FFN runs concurrently with routed top-8 expert FFN (`SPARKINFER_SHEXP_PIPE=0` disables).
3. **Fused conv q/k L2 norm** — one launch after `conv_split` instead of two separate `l2_norm_heads` kernels (30 GDN layers).
4. **Qwen3.6 ffn=512 MoE MMVQ templates** — specialized `gate_up_mmvq2_qwen_kernel<2048,512,8>` (+ pack2) and split-K down for the routed expert path.
5. **Q8_0 shared-expert int8 MMVQ** — keeps shared weights quantized, reuses FNQ `Q8_1(hn)` for gate/up dp4a, fuses SwiGLU, dp4a down (`SPARKINFER_QMOE=0` falls back to bf16 GEMV).

Qwen3-30B guard is unaffected (no GDN layers, no shared expert).

## Proof of speedup

- [x] Tested on **RTX 5090** (`sm_120`)

**Decode tok/s** (end-to-end, from `bench/scripts/bench.sh` — required for evaluation):

| | decode tok/s |
|---|--:|
| before (main) | 162.17 |
| after (this PR) | 246.82 |

Model: `Qwen3.6-35B-A3B-UD-Q4_K_M.gguf` (unsloth), `n=128`, `bs=1`, `SPARKINFER_BENCH_DEVICE_LOOP=0`, default env (no feature flags).

### Full sweep (RTX 5090, same settings)

| context | before (main) | after (this PR) | delta |
|--------:|--------------:|----------------:|------:|
[...]

PR #229, opened 2026-07-04
feat(qwen36): add optimized Gated-DeltaNet AR state update kernel wit…

James-CUDA:
## Summary

Add an optimized Gated-DeltaNet autoregressive state-update kernel (`gdn_ar_fast_kernel`, gated behind `SPARKINFER_GDN_FAST`) for the Qwen3.6 hybrid-linear path. Same recurrence math as the existing `gdn_ar_kernel`, but restructured to stop being launch- and latency-bound during decode:

- **One warp per state column.** The naive kernel launched `<<<v_heads=32, head_dim>>>` — only 32 blocks on ~170 SMs (~81% idle), 30x/token, serialized on two dependent 128-iter loops. The fast kernel splits the `head_dim` rows across a warp's 32 lanes, so the grid becomes `v_heads x (head_dim/COLS)` blocks x `COLS` warps and fills the GPU.
- **Register-cached state slice.** Each column's state rows live in registers (`sloc[]`) -> **one** global read + **one** global write of the 2 MB/layer state, vs. the naive kernel's write-then-reread (4x traffic).
- **Transposed state layout `[vh][col][row]`.** Under warp-per-column this makes each lane read a contiguous 32-row run (coalesced) instead of a 32-way scatter.
- `HEAD_DIM` is a template param so `NROW` is compile-time and the r-loops fully unroll, keeping `sloc[]` in registers.

Scope: single file, `kernels/csrc/cuda/fused/qwen36.cu` — inside the rewarded `kernels/` path. Dispatch is a static, all-or-nothing flag (required, since the transposed layout must be self-consistent for the whole run); non-128 head dims and unset flag fall back to the naive kernel unchanged.

**Correctness / accuracy gate.** Not byte-identical to the naive kernel — the warp-tree fp32 reduction reorders the sum — so it is validated by **top-1 agreement + low mean KL** over a long sequence (per CONTRIBUTING's accuracy gate), not by `cmp`. The recurrent state is zero-init and touched only by this kernel, so the internal layout stays self-consistent.

## Proof of speedup

> Tick the box **only if you actually ran it on an RTX 5090** and pasted the real
> `bench/scripts/bench.sh` output below. A ticked box with an empty/placeholder table gets
> `needs-benchmark` and is **not** evaluated.

- [x] Tested on **RTX 5090** (`sm_120`)

**Decode tok/s** (end-to-end, from `bench/scripts/bench.sh` — required for evaluation):

| | decode tok/s |
|---|--:|
| before (main) | 166.78 |
[...]

PR #244, opened 2026-07-05
perf(qwen3.6): bf16 recurrent state for the Gated-DeltaNet AR kernel (+31% decode)

cleanjunc:
## Summary

The Gated-DeltaNet (linear-attention) recurrent state update is the single hottest kernel in Qwen3.6-35B-A3B decode. Skip-toggle differential profiling on an RTX 5090 puts it at ~25% of per-token decode time: the naive kernel launches only `<<<v_heads=32, head_dim>>>` = 32 blocks on ~170 SMs and is latency-bound on two serial 128-iteration dependent-load loops over the ~2 MB/layer **fp32** recurrent state, 30x per token.

This PR stores that recurrent state in **bf16** instead of fp32, halving its HBM read+write traffic (the kernel's dominant cost). The new kernel fills the grid one warp per state column; each lane owns a contiguous 4-row run moved as a single vectorized bf16x4 (`uint2`) load/store, so a warp reads and writes its whole 128-row column in one coalesced 256 B transaction, with the gated decay folded into the read so the column is touched exactly once.

bf16 precision is safe for the recurrent state because the gated decay (`g < 1`) bounds accumulation over the sequence. Value-closeness to the fp32 path is verified below.

## Decode throughput (RTX 5090, sm_120, Qwen3.6-35B-A3B-UD-Q4_K_M)

Tightly interleaved A/B on one build via `SPARKINFER_GDN_BF16` (default on; `=0` restores the exact fp32 path, which reproduces current main):

| context | before fp32 state | after bf16 state | gain |
| --- | --- | --- | --- |
| 128  | 168.83 | 221.39 | +31.1% |
| 512  | 168.73 | 221.16 | +31.1% |
| 4096 | 162.83 | 211.10 | +29.6% |

| stage | Qwen3.6 decode tok/s |
| --- | --- |
| before (main) | 168.73 |
| after (this PR) | 221.16 |

## Accuracy vs llama.cpp (the eval gate)

Teacher-forced against the pinned llama.cpp reference over seed-fuzzed prompts (`accuracy_compare.py`), across three seeds:

| seed | tokens | top-1 (bar >= 0.90) | KL llama-vs-spark (bar <= 0.20) |
| --- | --- | --- | --- |
| 42   | 271 | 0.930 | 0.017 |
| 7    | 250 | 0.956 | 0.036 |
| 2026 | 189 | 0.958 | 0.018 |

For reference, current main (fp32 state) scores top-1 0.956 / KL 0.018 on seed 42, so bf16 costs a few near-tie flips but stays well inside the gate. `ctest` 7/7 pass.

## Scope and safety

- **Qwen3.6-only.** The bf16 kernel and the bf16 state buffer are reached solely from the hybrid GDN path (`w.linear_attn`, `head_dim == 128`). Qwen3-30B is not hybrid, never enters this code, and is byte-for-byte unchanged, so the no-regression guard is untouched.
- **Env-gated** (`SPARKINFER_GDN_BF16`, default on) with a clean fallback to the fp32 kernel.
- The bf16 state layout is internal to this kernel and all-or-nothing for the run (static flag), so it stays self-consistent across the captured decode graph. The state buffer is allocated fp32-sized and reinterpreted, so no allocation path changes.

### DIFF
--- a/kernels/csrc/cuda/fused/qwen36.cu
+++ b/kernels/csrc/cuda/fused/qwen36.cu
@@ -214,6 +214,75 @@ __global__ void gdn_ar_fast_kernel(const __nv_bfloat16* __restrict__ q,
     if (lane == 0) out[(size_t)vh * HEAD_DIM + j] = __float2bfloat16(y);
 }
 
+// bf16 recurrent-state Gated-DeltaNet AR (SPARKINFER_GDN_BF16, default ON). Same math as the naive
+// kernel, but the per-layer recurrent state S is stored in BF16 rather than FP32 — halving the ~2 MB/
+// layer state HBM traffic (read+write, 30x/token) that dominates this kernel. Grid-filled one warp per
+// state column. Each lane owns a CONTIGUOUS run of ROWS_PER_LANE rows and moves them as ONE vectorized
+// bf16x4 (uint2) load/store, so a warp reads/writes its whole HEAD_DIM column in a single coalesced
+// 256 B transaction; the decay is folded into the read so the column is touched exactly once each.
+// bf16 state is value-close to the fp32 path over a full sequence (the recurrent decay g<1 bounds
+// accumulation): measured top-1 agreement 0.978 / KL 0.010 vs the naive fp32 kernel. Requires HEAD_DIM
+// a multiple of 128 (ROWS_PER_LANE = HEAD_DIM/32 must be a multiple of 4 for the uint2 packing); the
+// bf16 state layout is internal to this kernel, so it is all-or-nothing for the run (static flag).
+template <int WPB, int HEAD_DIM>
+__global__ void gdn_ar_bf16state_kernel(const __nv_bfloat16* __restrict__ q,
+                                        const __nv_bfloat16* __restrict__ k,
+                                        const __nv_bfloat16* __restrict__ v,
+                                        const __nv_bfloat16* __restrict__ alpha,
+                                        const __nv_bfloat16* __restrict__ beta,
+                                        const __nv_bfloat16* __restrict__ dt,
+                                        const __nv_bfloat16* __restrict__ a,
+                                        __nv_bfloat16* __restrict__ state,   // BF16 [head][col][row]
+                                        __nv_bfloat16* __restrict__ out,
+                                        int q_heads, int v_heads) {
+    constexpr int RPL = HEAD_DIM / 32;            // rows per lane (compile-time -> registers)
+    constexpr int NVEC = RPL / 4;                 // uint2 (4 bf16) chunks per lane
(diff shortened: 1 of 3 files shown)


ai-hpc:
## 🐈 Copycat warning (retroactive sweep)

A specific kernel in this PR (GDN AR state update) is **41% contained at the function level** in #233's `gdn_ar_warpgrid_kernel` by a different author (jimcody1995). AI semantic analysis (DeepSeek) confirmed: identical GDN recurrent attention computation. This is a **warning** (strike 1/2). A second copycat-like submission will result in an automatic block.

If this is a legitimate independent implementation, comment on this PR and a maintainer will review.

ai-hpc:
Closed — copycat detection. See the comment above for details. The original author's work takes priority.

cleanjunc:
@ai-hpc I can't understand this result. Could you please review again?

cleanjunc:
Requesting a maintainer review — I believe this is a false positive, for three concrete reasons.

**1. "Identical GDN recurrent attention computation" is the model's defined math, not evidence of copying.** Every Gated-DeltaNet AR kernel in this repo computes the identical recurrence, because it has to or the output is wrong:
- `gdn_ar_kernel` — the naive base kernel already on `main`
- `gdn_ar_fast_kernel` — #229 (James-CUDA), merged
- `gdn_ar_warpgrid_kernel` — #233 (jimcody1995)
- `gdn_ar_rowwarp_kernel` — #240

A semantic-similarity check will report "identical computation" for any two of these, including the naive base kernel, so that signal can't tell independent work from copying.

**2. The 41% function-level overlap is the unavoidable shared surface, not the optimization.** It is the 7-pointer parameter signature (`q, k, v, alpha, beta, dt, a, state, out`), the per-head `g` / `beta` / `scale` setup (again, the model's math), pointer arithmetic, `#pragma unroll`, and the warp-reduce helper — all of which appear identically in the naive base kernel. Measured at the line-diff level the overlap with #233 is ~29%, and none of the overlapping lines are the state-update logic.

**3. The contribution is a mechanism #233 does not have: a bf16 recurrent state.** #229/#233/#240 all declare `float* state` (fp32). This PR declares `__nv_bfloat16* state` and stores/reads the recurrent state in **bf16**, halving its ~2 MB/layer HBM traffic — which is the source of the +31% (the fp32 kernels top out near +30%). The access pattern also differs from #233's warpgrid: each lane owns a contiguous 4-row run moved as one `uint2`-vectorized bf16x4 load/store, with the decay folded into the read.

Happy to have the two kernels diffed side by side — the bf16 `state` declaration + allocation and the vectorized access pattern are straightforward to verify as distinct from #233. Thanks for taking a look.

PR #242, opened 2026-07-05
perf(qwen3.6): int8 Q8_0 dp4a MMVQ for attention/GDN projections (+24% decode)

nickmopen:
## What

Adds the missing **Q8_0 × Q8_1 dp4a MMVQ** path so Qwen3.6's attention and Gated-DeltaNet projections run int8 on-read instead of being expanded to bf16 at load.

## Why

Profiling **Qwen3.6-35B-A3B** (the dual-eval primary, Unsloth `UD-Q4_K_M`) on an RTX 5090 at ctx=512 showed `gemv_kernel<bf16>` at **59.7% of decode**. Cause: the UD quant stores the attention/GDN projections (`attn_q/k/v/output`, `attn_qkv`, `ssm_out`) as **Q8_0**, and sparkinfer had MMVQ only for Q4_K/Q6_K — so every Q8_0 weight was dequantized to bf16 at load (`deq_q8_0_kernel`, +2.7 GB VRAM) and ran the slow generic GEMV. The routed experts (Q4_K) already run int8.

## How

Q8_0 is already int8, so this is the simplest MMVQ (no 4-bit unpack) and byte-faithful to llama.cpp `vec_dot_q8_0_q8_1` (Q8_0 is symmetric → no Q8_1 offset term: `dot = d_w·d_a·Σ w_i·a_i`). Q8_0 blocks are 34 B / 2-byte aligned, so weight ints are read 2-byte-aligned like the Q6_K path.

- **kernels:** `si_mmvq_q80_kernel` + `launch_mmvq_q80/_f32`.
- **loader:** `attn_w` keeps Q8_0 raw (joins Q4_K/Q6_K); `SPARKINFER_QATTN=0` still reverts to dense bf16.
- **dispatch:** `proj_xn` / `proj_from` / O-projection route ggml type 8 → `mmvq_q80`; `prepare_xn_quant` emits the shared Q8_1 activation when any projection is Q8_0.

## Results (RTX 5090, Qwen3.6-35B-A3B UD-Q4_K_M, ctx=512, bs=1)

| | decode tok/s | VRAM |
|---|---|---|
| main (Q8_0 → bf16 GEMV) | 221.3 | 25.3 GB |
| **this PR (Q8_0 MMVQ)** | **273.96** | 23.6 GB |

**+24% decode**, and lower VRAM (weights stay int8).

[...]

PR #280, opened 2026-07-07
perf(gdn): fuse conv + L2-norm + DeltaNet AR into a single Qwen3.6 GDN megakernel

Dexterity104:
## Summary

Fuse the Qwen3.6-35B-A3B Gated-DeltaNet per-layer main-stream chain **conv_split -> l2_norm_qk -> gdn_ar** (3 launch wrappers / 4 kernels, on 30 linear layers per token) into **one persistent CUDA-graph node**, `gdn_mega_kernel` (new file `kernels/csrc/cuda/fused/gdn_mega.cu`). Those four kernels are tiny non-matmul ops on grids far too small to fill the 5090, so under graph replay their per-node scheduling latency and the `lin_q/lin_k/lin_v/lin_gdn` global round-trips between them are fully exposed on the main stream (an env-gated skip of just this segment is worth ~5.4% wall at ctx=512). The megakernel keeps every per-token transient (normalized q/k, conv'd v) in shared memory, runs conv+L2 at the AR's GPU-filling warp-per-column occupancy instead of the conv's own 32-block launch, and advances `conv_state` with an in-kernel position ring (the single per-channel owner writes `ring[pos % K]`, a slot no block reads that token, so the redundant per-block conv reads never race the write and no separate launch is needed). `gated_norm` stays a separate kernel: its post-AR per-v-head reduction spans this grid's column tiles, so folding it in would need a grid sync that costs what it saves. New kernel only; the per-kernel implementations in `qwen36.cu` are untouched.

**Bit-identical** to the existing `conv_split_kernel + l2_norm_qk_kernel + gdn_ar_fast_kernel` composed: same `__expf/__logf/rsqrtf` intrinsics, same warp-butterfly reduction order, same two bf16 round-trips. Teacher-forced self-consistency (`bench/scripts/self_consistency.py`) of the megakernel vs the fallback trio on the same tokens, real text:

```
positions            : 100
argmax agreement     : 100/100 = 1.0000   (want 1.0000)
max |logprob diff|   : 0.00000
mean self-KL(a||b)   : 0.000000 nats
```
(PPL 2.76718 on both paths; a 256-token fuzz gives 255/255 = 1.0000 as well.)

**Qwen3.6-only, zero regression on Qwen3-30B:** dispatched behind `SPARKINFER_GDN_MEGA` (default on) and the hybrid shape (`c.hybrid && linear_head_dim==128 && linear_conv_kernel==4`). Qwen3-30B is non-hybrid, so it never allocates GDN state (`if (cfg.hybrid)`) nor enters the linear branch (`is_linear_layer` requires `hybrid`); the new code is unreachable and its decode path is byte-for-byte unchanged. `SPARKINFER_GDN_MEGA=0` restores the three separate kernels (used as the "before" A/B leg below). The only other change is sizing `lin_conv_state` to `conv_kernel` slots/layer (was `conv_kernel-1`); the fallback conv_split uses the first `conv_kernel-1` slots unchanged.

## Proof of speedup

**Decode tok/s** (one build, A/B via `SPARKINFER_GDN_MEGA`, interleaved, medians):

| | decode tok/s |
|---|--:|
| before (main) | 385.81 |
| after (this PR) | 394.99 |

Full sweep (best scored context = 512):

| ctx | before (main) | after (this PR) | speedup |
|--:|--:|--:|--:|
| 128 | 391.21 | 399.01 | +1.99% |
| 512 | 385.81 | 394.99 | +2.38% |
| 4096 | 370.19 | 377.39 | +1.95% |

```text
# RTX 5090 (sm_120), Qwen3.6-35B-A3B-UD-Q4_K_M, n=128, bs=1
# before = SPARKINFER_GDN_MEGA=0 (three separate kernels), after = megakernel (default)
--- ctx=128 ---  before=391.21 tok/s   after=399.01 tok/s   (+1.99%)
--- ctx=512 ---  before=385.81 tok/s   after=394.99 tok/s   (+2.38%)
--- ctx=4096 --- before=370.19 tok/s   after=377.39 tok/s   (+1.95%)
```

### DIFF
--- a/kernels/csrc/cuda/fused/gdn_mega.cu
+++ b/kernels/csrc/cuda/fused/gdn_mega.cu
@@ -1,0 +1,223 @@
+// Qwen3.6-35B-A3B Gated-DeltaNet MEGAKERNEL.
+//
+// Collapses the per-layer main-stream serial chain
+//     conv_split (depthwise causal conv + SiLU + q/k/v split)
+//   -> l2_norm_qk (per-head L2 normalize of q and k)
+//   -> gdn_ar_fast (DeltaNet recurrent state update)
+//   -> (conv_state advance)
+// into ONE persistent CUDA-graph node. In the decode graph these were 3 host launch
+// wrappers firing 4 tiny, GPU-underfilling kernels on the MAIN stream (~30 GDN layers/
+// token) whose q/k/v/gdn intermediates round-tripped through global memory between each
+// launch. This kernel keeps every per-token transient (normalized q/k, conv'd v) in
+// shared memory / registers, does the conv at the GPU-filling warp-per-column occupancy
+// of the AR (instead of the conv's own 32-block launch), and never re-materializes the
+// intermediates in gmem. gated_norm stays a separate kernel: its post-AR per-v-head
+// reduction spans this grid's column tiles (cross-block), so it cannot fold in without a
+// grid sync that would cost what it saves.
+//
+// BIT-IDENTICAL to conv_split_kernel + l2_norm_qk_kernel + gdn_ar_fast_kernel<8,128>
+// composed, for the Qwen3.6 shape (head_dim=128, conv_kernel=4). The device math below is
+// copied op-for-op from those kernels (same __expf/__logf/rsqrtf intrinsics, same warp
+// butterfly reduction order, same two bf16 round-trips) so top1/KL vs the current main is
+// exactly 1.0 / 0.0.
+//
+// conv_state is treated as a size-K position ring (ring[p % K] = raw input at position p)
+// instead of the fallback's tap-major shift. That makes the single per-channel owner write
+// exactly ring[pos % K] -- a slot no block reads this token (the taps read pos-1,pos-2,
+// pos-3) -- so the redundant per-block conv reads never race the owner's write, and the
+// state advance needs neither a separate launch nor a side stream. The caller sizes and
+// zero-inits conv_state to K slots/layer; the fallback path uses the first K-1 slots
+// (tap-major) unchanged.
+//
+// Qwen3.6-only: dispatched behind SPARKINFER_GDN_MEGA and the hybrid shape guard; the
+// Qwen3-30B (non-hybrid) decode path never allocates GDN state nor reaches this kernel.
+
+#include <cuda_bf16.h>
+#ifndef SPARKINFER_NVRTC_DEVICE_ONLY
+#include <cuda_runtime.h>
+#endif
+
+namespace sparkinfer {
+namespace kernels {
+
+// --- numeric primitives, identical to qwen36.cu's q36_* (load-bearing for bit-identity) ---
+__device__ __forceinline__ float gm_to_f(__nv_bfloat16 x) { return __bfloat162float(x); }
+__device__ __forceinline__ float gm_silu(float x) { return x / (1.f + __expf(-x)); }
+__device__ __forceinline__ float gm_sigmoid(float x) { return 1.f / (1.f + __expf(-x)); }
+__device__ __forceinline__ float gm_softplus(float x) { return x > 20.f ? x : __logf(1.f + __expf(x)); }
+__device__ __forceinline__ float gm_wsum(float v) {
+    #pragma unroll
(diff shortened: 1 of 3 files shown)


Dexterity104:
Hi @ai-hpc I wonder why this is evaluated as "copycat".

Dexterity104:
I think this is a false-positive from both PRs editing the same function. The 71% match is Qwen35Model::forward_token in qwen35.cpp — the one big ~400-line Qwen3.6 decode driver that every Qwen3.6 change has to touch. #242 and I each only add a handful of lines to it; the rest is identical upstream code, so any two PRs that edit it will look ~90% "contained" in each other. That's also why the PR-level containment came out low.

The actual work doesn't overlap. #242 adds int8 Q8_0 MMVQ to the projection GEMVs (gemm/gemv.cu), and this PR adds a brand-new file, kernels/csrc/cuda/fused/gdn_mega.cu, that fuses conv_split + l2_norm_qk + gdn_ar into a single megakernel — #242 doesn't touch any fused/GDN kernel at all. Different file, different technique.

Happy to pull the GDN dispatch out into its own small helper if that clears the match on forward_token.

ai-hpc:
## ❌ sparkinfer auto-eval — `66fccf8`

| metric | value |
|---|---|
| **label** | `eval:REJECT` |
| scored decode (4096 ctx · 4k-context · Qwen3.6) | 354.31 tok/s |
| correctness (Qwen3.6 vs llama.cpp) | top-1 94.4% · KL 0.0161 |
| vs same-box main | 361.34 tok/s → -1.9% (-7.0) |
| Qwen3.6 128-token no-regression gate | 373.94 tok/s vs main 383.96 tok/s · fail |
| Qwen3.6 512-context no-regression gate | 367.97 tok/s vs main 376.53 tok/s · fail |
| Qwen3.6 4k-context no-regression gate | 354.31 tok/s vs main 361.34 tok/s · pass |
| **Qwen3-30B guard — accuracy** | top-1 95.8% · KL 0.0198 · pass |
| Qwen3-30B guard — 128-token | 458.57 tok/s · **fail** |
| Qwen3-30B guard — 512-context | 436.39 tok/s · **fail** |
| Qwen3-30B guard — 4k-context | 359.45 tok/s · **fail** |
| regressions | regression-128, regression-512, regression-qwen3-128, regression-qwen3-512, regression-qwen3-4k |

**Rejected** — Qwen3-30B decode regressed at context(s): 128, 512, 4k.

_RTX 5090 (sm_120) · 128/512/4k/16k/32k guarded · scored vs same-box main · strongest context scores · built from source · correctness vs llama.cpp. Automated — **not merged**; merge manually after review._

ai-hpc:
## ❌ sparkinfer auto-eval — `66fccf8`

| metric | value |
|---|---|
| **label** | `eval:REJECT` |
| scored decode (4096 ctx · 4k-context · Qwen3.6) | 354.27 tok/s |
| correctness (Qwen3.6 vs llama.cpp) | top-1 94.1% · KL 0.0265 |
| vs same-box main | 370.12 tok/s → -4.3% (-15.9) |
| Qwen3.6 128-token no-regression gate | 374.00 tok/s vs main 394.32 tok/s · fail |
| Qwen3.6 512-context no-regression gate | 368.04 tok/s vs main 385.42 tok/s · fail |
| Qwen3.6 4k-context no-regression gate | 354.27 tok/s vs main 370.12 tok/s · fail |
| **Qwen3-30B guard — accuracy** | top-1 97.1% · KL 0.0245 · pass |
| Qwen3-30B guard — 128-token | 458.51 tok/s · **fail** |
| Qwen3-30B guard — 512-context | 436.28 tok/s · **fail** |
| Qwen3-30B guard — 4k-context | 359.47 tok/s · **fail** |
| regressions | regression-128, regression-512, regression-4k, regression-qwen3-128, regression-qwen3-512, regression-qwen3-4k |

**Rejected** — Qwen3-30B decode regressed at context(s): 128, 512, 4k.

_RTX 5090 (sm_120) · 128/512/4k/16k/32k guarded · scored vs same-box main · strongest context scores · built from source · correctness vs llama.cpp. Automated — **not merged**; merge manually after review._

ai-hpc:
Closed — eval:REJECT (regression at all contexts, Qwen3-30B guard failed).

PR #290, opened 2026-07-09
fix(runtime): thread seq_id through decode_layer's KV block-table lookup

### LINKED ISSUE #289: # runtime/decode: decode_layer hardcodes seq_id 0 for block-table lookup — breaks under KV slot reuse
## Summary

`DecodeRunner::decode_layer` looks up the KV-cache block table for a whole batch via
`block_table(0)` — i.e. "whatever slot seq_id 0 currently occupies" — then indexes into it by
batch row. This is only correct on a cold start where sequences are allocated in order and none
have been freed. Under normal continuous-batching slot reuse it either crashes (null-deref) or
silently cross-contaminates unrelated sequences' KV history.

## Where

- `runtime/src/decode_layer.cpp:99`
- `runtime/src/kv_cache.cpp:64-96` (root cause: slot assignment)
- `runtime/include/sparkinfer/decode.h:43-44` (API gap: no seq_id/slot param)

## Code

```cpp
// decode_layer.cpp:96-102
bf16* kpool = (bf16*)s.kv->k_pool() + (size_t)layer * s.kv->layer_stride_elems();
bf16* vpool = (bf16*)s.kv->v_pool() + (size_t)layer * s.kv->layer_stride_elems();
int* btable = s.kv->block_table(0);   // batch occupies slots 0..num_seqs-1
launch_kv_append(kpool, vpool, s.k, s.v, btable, s.d_write_pos,
                 num_seqs, s.attn.num_kv_heads, s.attn.head_dim,
                 s.kv->block_size(), s.kv->max_blocks_per_seq(), stream);
```

```cpp
// kv_cache.cpp:92-96
int* KVCacheManager::block_table(uint64_t seq_id) const {
    auto it = impl_->seq_slot.find(seq_id);
    if (it == impl_->seq_slot.end()) return nullptr;
    return impl_->d_block_tables + (size_t)it->second * kMaxBlocksPerSeq;
}
```

`block_table(uint64_t seq_id)` resolves `seq_id` → slot via a map, not slot index → table. Passing
the literal `0` fetches the table for whichever sequence currently owns slot `seq_id==0`, then the
kernels walk forward from there by `row * max_blocks_per_seq`, implicitly assuming batch row *i*
lives in slot *i*.

`KVCacheManager::allocate()` / `free()` (`kv_cache.cpp:64-90`) recycle slots through a **LIFO**
`free_slots` stack keyed by `seq_id`, independent of batch position:

```cpp
// allocate(): first-time slot assignment
else { slot = impl_->free_slots.back(); impl_->free_slots.pop_back(); impl_->seq_slot[seq_id] = slot; }
// free(): slot returned to the pool
if (s != impl_->seq_slot.end()) { impl_->free_slots.push_back(s->second); impl_->seq_slot.erase(s); }
```

## Failure modes

1. **Crash**: once seq_id `0` finishes and is freed, `block_table(0)` returns `nullptr`. That
   null pointer is passed straight into `launch_kv_append` / `launch_flash_decode_gqa8` as the
   block-table argument.
2. **Silent corruption**: if seq_id `0` is still live but other batch rows' slots aren't
   contiguous after it (guaranteed once any earlier slot is freed and reused out of order), each
   row reads/writes a *different* sequence's KV cache — wrong attention output, and the erroneous
   K/V append permanently corrupts the mismatched sequences' history. No bounds check catches
   this because every access stays within the allocated `[kMaxSeqs, kMaxBlocksPerSeq]` table.

## Why this isn't caught by CI today

`DecodeRunner` currently has only one real caller, `runtime/tests/decode_runner_gpu_test.cpp`,
which allocates sequences once, in order, and never frees/reuses a slot mid-test — so the bug is
latent until continuous-batching scheduling is wired up to call `decode_layer` across sequence
churn.

## Suggested fix direction

Thread per-row sequence identity through the call instead of assuming row == slot:

- Add a `const std::vector<uint64_t>& seq_ids` (or equivalent slot list) parameter to
  `DecodeRunner::decode_layer`.
- Build the per-row block-table pointer/offset from that list rather than a single
  `block_table(0)` call.
- Extend `decode_runner_gpu_test.cpp` to cover the slot-reuse case: allocate A/B/C, free B,
  allocate D (recycles B's slot), decode a batch ordered `[A, C, D]`, and assert no
  cross-contamination between sequences.

## Scope note

This is a correctness fix, not a speedup — per CONTRIBUTING.md it scores `eval:none` (0 SN74
reward) but is still eligible for normal review/merge if scoped only to `runtime/`.

## Status

The fix described above has already been implemented (uncommitted, in this working tree) in:
`runtime/include/sparkinfer/decode.h`, `runtime/src/decode_layer.cpp`,
`runtime/tests/decode_runner_gpu_test.cpp`. This document is now most useful as the PR
description when opening/reopening that fix, or as a comment adding detail to issue #202.

dale053:
Closes #289

## Summary

`DecodeRunner::decode_layer` looked up the batch's KV block table via a hardcoded
`block_table(0)`, implicitly assuming batch row *i* always lives in physical KV slot *i*. That
only holds on a cold start; once any sequence is freed and a slot is recycled (normal
continuous-batching churn — `KVCacheManager` reuses slots via a LIFO free-list keyed by
`seq_id`, independent of batch position), it either null-derefs (if seq_id 0 isn't currently
live) or silently cross-contaminates unrelated sequences' KV history (if slots aren't
contiguous after seq_id 0).

Fix: `DecodeRunner::begin_step` now takes the batch's `seq_ids` and gathers each row's real
block table (`kv->block_table(seq_ids[i])`) into a contiguous per-batch scratch buffer once
per step (paging is layer-independent, so this is reused across all layers). `decode_layer`
reads from that gathered buffer instead of assuming row == slot.

## Changes

- `runtime/include/sparkinfer/decode.h` — `begin_step(seq_ids, seq_lens_before)`.
- `runtime/src/decode_layer.cpp` — per-batch block-table scratch buffer + gather in
  `begin_step`; `decode_layer` uses the gathered buffer instead of `block_table(0)`.
- `runtime/tests/decode_runner_gpu_test.cpp` — updated call site; added a slot-reuse
  regression (free a sequence, allocate a new one into the recycled slot, decode a batch
  ordered out of slot order, assert no crash / finite output).

## Proof of speedup

N/A — correctness/API fix, no decode-path performance change. Per CONTRIBUTING.md this scores
`eval:none` (0 SN74 reward) but is eligible for normal review/merge.

```text
N/A
```

| | decode tok/s |
|---|--:|
| before | N/A |
| after  | N/A |

## Verification checklist (fill in before opening)

- [ ] `cmake -B build -DCMAKE_CUDA_ARCHITECTURES=120 && cmake --build build -j`
- [ ] `ctest --test-dir build` — 5/5, including the new slot-reuse case in
      `decode_runner_gpu_test`
- [ ] `bench/scripts/accuracy.sh --download` — confirm no regression (this PR shouldn't
      change accuracy at all; it only fixes a batching-identity bug not yet exercised by any
      production caller)
- [ ] `compute-sanitizer` clean, if available

*Not yet run in this environment — no CUDA toolkit or GPU available here. Run the above on
your RTX 5090 box before submitting.*

### DIFF
--- a/runtime/include/sparkinfer/decode.h
+++ b/runtime/include/sparkinfer/decode.h
@@ -35,9 +35,12 @@ public:
                  int max_batch = 256);
     ~DecodeRunner();
 
-    // Begin a decode step: set the new-token position for each sequence
-    // (host seq_lens = length BEFORE this token). Uploads positions to device.
-    void begin_step(const std::vector<int>& seq_lens_before);
+    // Begin a decode step: bind each batch row to its sequence and set its new-token
+    // position (host seq_lens = length BEFORE this token). Uploads positions to device and
+    // gathers each seq_id's KV block table into a contiguous per-batch scratch buffer —
+    // batch row order is not guaranteed to match physical KV-slot order once sequences have
+    // been freed/reallocated, so decode_layer can no longer assume row i lives in slot i.
+    void begin_step(const std::vector<uint64_t>& seq_ids, const std::vector<int>& seq_lens_before);
 
     // Run one layer in-place on x: [num_seqs, hidden] (bf16, device).
     void decode_layer(int layer, void* x, int num_seqs,
--- a/runtime/src/decode_layer.cpp
+++ b/runtime/src/decode_layer.cpp
@@ -29,6 +29,8 @@ struct DecodeRunner::Impl {
     // scratch (bf16 unless noted)
     bf16 *xn, *q, *k, *v, *attnout, *ao, *h, *hn, *moeout;
     int *d_seq_lens, *d_write_pos;
+    int mbps = 0;                    // kv->max_blocks_per_seq(), cached at construction
+    int* d_batch_block_table = nullptr;  // [max_batch, mbps] gathered per begin_step, shared across layers
 
     template <class T> T* alloc(size_t n) { void* p = nullptr; cu(cudaMalloc(&p, n * sizeof(T)), "malloc"); return (T*)p; }
 };
@@ -39,6 +41,7 @@ DecodeRunner::DecodeRunner(int hidden, AttnConfig attn, KVCacheManager* kv,
     p_->hidden = hidden; p_->attn = attn; p_->kv = kv; p_->moe = moe; p_->max_batch = max_batch;
     p_->qdim  = attn.num_q_heads  * attn.head_dim;
     p_->kvdim = attn.num_kv_heads * attn.head_dim;
+    p_->mbps  = kv->max_blocks_per_seq();
     const int M = max_batch;
     p_->xn      = p_->alloc<bf16>((size_t)M * hidden);
     p_->q       = p_->alloc<bf16>((size_t)M * p_->qdim);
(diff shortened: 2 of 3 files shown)


skyrocket2026:
Closing due to inactivity (>2 days since last update). Reopen or push new commits when ready to continue.

PR #284, opened 2026-07-08
perf(qwen3.6): int8 KV + tensor-core flash-decode for the hd256 full-attn layers (long-context)

nickmopen:
## Summary

Qwen3.6's 10 full-attention layers (head_dim=256) ran **bf16 KV + a scalar dot** — the int8 tensor-core MMA GQA flash-decode was only ever instantiated for head_dim=128. This ports it to hd256 (the kernel is already `HEAD_DIM`-generic) and adds an int8 partial-RoPE KV append, **halving the KV read** for those 10 layers. Context-adaptive: turns on at ≥ 8k (where KV read dominates); short contexts stay bf16, byte-identical to main.

Prior "int8-KV-on-hybrid fails accuracy" was a **mis-dispatch** (hd256 fed through the hd128 kernel); a correct hd256 int8 path had never been tried.

## Proof of speedup

- [x] Tested on **RTX 5090** (`sm_120`)

**Decode tok/s** (end-to-end, `bench/scripts/bench.sh --download --ctx 16384`, Qwen3.6-35B-A3B UD-Q4_K_M, bs=1 — this is a **long-context** optimization; contexts < 8k are unchanged from main):

| | decode tok/s |
|---|--:|
| before (main) | 354.62 |
| after (this PR) | 370.74 |

At 32k the gain is larger (KV read is a bigger share): main 316.93 → this PR **349.87** (+10.4%).

```text
== ctx=16384 ==
before (main, bf16 KV) : decode tg  354.62 tok/s  (2.8 ms/token, n=128, ctx=16384, bs=1)
after  (this PR, int8 KV): decode tg  370.74 tok/s  (2.7 ms/token, n=128, ctx=16384, bs=1)
== ctx=32768 ==
before (main, bf16 KV) : decode tg  316.93 tok/s  (3.2 ms/token, n=128, ctx=32768, bs=1)
after  (this PR, int8 KV): decode tg  349.87 tok/s  (2.9 ms/token, n=128, ctx=32768, bs=1)
```

## Accuracy

[...]

PR #300, opened 2026-07-09
fix(kernels): correct hd256 int8-MMA flash-decode output beyond the MMA engagement depth

### LINKED ISSUE #299: hd256 int8-MMA flash-decode returns incorrect attention output once the MMA gate engages (Qwen3.6, >16k ctx)
## Summary

The tensor-core int8 flash-decode split kernel (`fa_split_gqa_mma_i8_kernel`), enabled for Qwen3.6's ten hd256 full-attention layers in #284, produces numerically wrong output whenever it actually engages. On current main (`ee2ff71`), a teacher-forced probe diverges from the exact tile path at ~100% of positions past the engagement depth — the model is effectively decoding with corrupted attention for every token beyond ~16k context.

The eval pipeline cannot see this: the accuracy gate teacher-forces at short context, where the MMA gate (`chunk >= 32` physical blocks per split) never engages with the current n_splits policy, and the 16k/32k bench legs measure decode speed only. So the current 16k/32k frontier numbers are produced by a kernel whose output at exactly those contexts does not match the model.

## Evidence (RTX 5090 sm_120, CUDA 13.0, main @ ee2ff71, Qwen3.6-35B-A3B UD-Q4_K_M)

Same build, same box, same 17,500 forced tokens (`ids[i] = 100 + (i*37) % 7900`); the only difference between arms is `SPARKINFER_FAMMA`:

| region | positions | argmax mismatches (FAMMA=1 vs FAMMA=0) | mean abs. delta logprob |
| --- | --- | --- | --- |
| tile path both arms (<= 16384) | 16384 | **1** | 0.0009 |
| MMA engaged (16385..17499) | 1114 | **1113 (99.9%)** | 5.32 |

The clean region is where the MMA gate has not engaged (both arms run the identical tile path). From the first engaged position (16385) the MMA arm diverges immediately — |delta logprob| at 16385 alone is 3.87, and that position is a single-kernel-call comparison on a bit-identical KV prefix, so this is kernel math, not error accumulation.

Speed side-effect: the broken kernel also skips half its PV math and half its V reads, so it is artificially fast — same box, `SPARKINFER_FAMMA=1` (default) vs `=0`: 363.84 vs 348.67 tok/s @16k (+4.35%) and 341.27 vs 304.10 @32k (+12.22%). The current dashboard 16k/32k frontier carries that inflation.

### Repro

```bash
IDS=$(python3 -c "print(' '.join(str(100 + (i*37)%7900) for i in range(17500)))")
SPARKINFER_FAMMA=0 build/runtime/qwen3_gguf_score models36/Qwen3.6-35B-A3B-UD-Q4_K_M.gguf 1 $IDS > ref.txt
SPARKINFER_FAMMA=1 build/runtime/qwen3_gguf_score models36/Qwen3.6-35B-A3B-UD-Q4_K_M.gguf 1 $IDS > mma.txt
# compare the am= (argmax) column per position; divergence starts at position 16385
```

## Why hd128 (Qwen3-30B) is unaffected

The kernel is HEAD_DIM-generic but was only ever numerically exercised at hd128, where several internal strides coincide by accident: the score-tile row stride equals HEAD_DIM equals the 128-token group width, a single 128-wide PV pass happens to cover the whole head, and 4 elems/lane Q quantization happens to cover the whole row. At hd256 each of those coincidences breaks. Qwen3-30B and the Qwen3.6 GDN layers never touch this path.

Fix PR incoming (linked).

reyanthony062001-ops:
## Summary

The hd256 instantiation of `fa_split_gqa_mma_i8_kernel` — the int8 tensor-core flash-decode split enabled for Qwen3.6's ten full-attention layers in #284 — computes wrong output wherever its gate engages (>16k ctx with the current n_splits policy). #299 has the black-box evidence: on current main, 1113/1114 argmax mismatches vs the exact tile path past position 16384, |Δlogprob| 3.9 at the first engaged position on a bit-identical KV prefix.

Four hd256-specific defects, all in the kernel body (hd128 was correct by coincidence of strides):

1. The QK score `store_matrix_sync` used `HEAD_DIM` as ldm. The score tile is [16 q-rows × 128 group tokens], so the row stride is the token width 128; at hd256 rows were stored 256 apart and read 128 apart — the softmax consumed interleaved garbage.
2. The P′ fragment `load_matrix_sync` in the PV phase had the same ldm bug.
3. The PV phase ran a single 128-wide dim pass: O dims 128..255 were never computed and stayed at their zero init.
4. Q quantization hardcoded 4 elems/lane (=128 dims): at hd256, `s_qi` dims 128..255 fed the QK mma uninitialized shared memory, and the per-row amax scale ignored half the vector.

All four fixes are generic in HEAD_DIM; hd128 behavior is unchanged by construction (EPT = HEAD_DIM/32 = 4, ldm 128 == HEAD_DIM, single dh pass), so Qwen3-30B and the Qwen3.6 GDN layers are untouched.

Correctness validation (same probe as #299, RTX 5090, base ee2ff71):

| comparison | ≤16384 | 16385..17499 (MMA engaged) |
| --- | --- | --- |
| main FAMMA=1 vs exact tile (FAMMA=0) | 1 mismatch | 1113/1114 (99.9%), mean abs Δlp 5.32 |
| **this PR** FAMMA=1 vs exact tile | 0 | 168/1114 (15.1%), mean abs Δlp 0.60 |

The residual 15% is flat across the window (14–17% in every 300-position bucket, not accumulating) — the int8 Q/P quantization noise the scheme carries by design, vs a bf16-exact reference at 17k-token attention depth.

ctest: 7/7. accuracy.sh (Qwen3-30B vs pinned llama.cpp): top-1 96/100 = 0.960 (bar >= 0.90), mean KL 0.0495 nats (bar <= 0.20).

## Speed: what this costs vs current main, and what it keeps vs the correct baseline

Straight talk: the broken kernel is artificially fast — it skips half the PV math and half the V reads. Correct math costs some of that back, but keeps most of #284's long-context win over the tile path:

**Decode tok/s** (`qwen3_gguf_bench <gguf> 128 <ctx>`, median of 3, same box, base ee2ff71):

| ctx | main (broken MMA) | main FAMMA=0 (exact tile) | this PR | vs main | vs tile |
| --- | --- | --- | --- | --- | --- |
| 128 | 407.61 | — | 407.72 | +0.03% | — |
| 512 | 400.95 | — | 400.99 | +0.01% | — |
| 4k | 383.70 | — | 383.71 | +0.00% | — |
| 16k | 363.84 | 348.67 | 359.54 | −1.18% (98.82%) | **+3.12%** |
| 32k | 341.27 | 304.10 | 334.91 | −1.86% (98.14%) | **+10.13%** |

All contexts hold ≥98% of main on both models. Qwen3-30B guard (hd128, behavior unchanged by construction): 16k 99.80%, 32k 99.92%, short contexts flat across the full 5-context sweep.

Put differently: the broken kernel inflates the current 16k/32k frontier by +4.35% and +12.22% over the exact tile path. This PR keeps +3.12% / +10.13% of #284's long-context win while producing output that actually matches the model.

## Proof of speedup

```text
# qwen3_gguf_bench <gguf> 128 <ctx>, interleaved A/B, per-rep decode tok/s (RTX 5090 locked 2550 MHz)
# Qwen3.6-35B-A3B UD-Q4_K_M
ctx      main(1,2,3)              FAMMA=0 tile(1,2,3)      this PR(1,2,3)
128      407.64 407.61 407.61     -                        407.81 407.72 407.72
512      400.90 400.95 400.96     -                        401.07 400.99 400.81
4096     383.71 383.70 383.64     -                        383.69 383.75 383.71
16384    363.79 363.84 363.76*    348.67 348.67 348.68     359.53 359.54 359.38*
32768    341.35 341.27 341.26*    304.10 304.09 304.33     334.91 334.94 334.71*
# Qwen3-30B-A3B Q4_K_M (guard)
16384    309.96 309.69            -                        308.70 309.69
32768    242.28 242.09            -                        242.08 241.91
# *16k/32k main + PR values from the 3-arm run (benchD2); short-ctx rows from the 45-run sweep (benchD).
```

Fixes #299.

### DIFF
--- a/kernels/csrc/cuda/attention/flash_decode_split.cu
+++ b/kernels/csrc/cuda/attention/flash_decode_split.cu
@@ -353,12 +353,16 @@ __global__ void __launch_bounds__(GQA * 32, 5) fa_split_gqa_mma_i8_kernel(
     float* s_l  = s_m + 16;                                           // [16]
 
     // Quantize Q per q-head row (warp w owns rows 2w, 2w+1; rows >= GQA are zero pad).
+    // EPT spans the whole head vector: 4 elems/lane at hd128, 8 at hd256. Hardcoding 4 left
+    // s_qi dims 128..255 uninitialized at hd256 (and computed amax over half the row), so the
+    // QK mma k-tiles 8..15 multiplied against stale shared memory.
+    constexpr int EPT = HEAD_DIM / 32;
     #pragma unroll
     for (int rr = 0; rr < 2; rr++) {
         const int r = warp * 2 + rr;
-        float qv[4], amax = 0.f;
+        float qv[EPT], amax = 0.f;
         #pragma unroll
-        for (int e = 0; e < 4; e++) {
+        for (int e = 0; e < EPT; e++) {
             qv[e] = (r < GQA) ? __bfloat162float(q[(size_t)(seq * num_q_heads + kvh * GQA + r) * HEAD_DIM + lane + e * 32]) : 0.f;
             amax = fmaxf(amax, fabsf(qv[e]));
         }
@@ -367,7 +371,7 @@ __global__ void __launch_bounds__(GQA * 32, 5) fa_split_gqa_mma_i8_kernel(
         const float d = amax / 127.0f;
         if (lane == 0) s_qs[r] = d;
         #pragma unroll
-        for (int e = 0; e < 4; e++)
+        for (int e = 0; e < EPT; e++)
             s_qi[r * HEAD_DIM + lane + e * 32] = (signed char)((amax == 0.f) ? 0 : (int)roundf(qv[e] / d));
     }
     for (int i = tid; i < GQA * HEAD_DIM; i += blockDim.x) s_o[i] = 0.f;
@@ -403,7 +407,13 @@ __global__ void __launch_bounds__(GQA * 32, 5) fa_split_gqa_mma_i8_kernel(
                 load_matrix_sync(bf, kb + ks * 16, KVLD);
                 mma_sync(cf, af, bf, cf);
             }
-            store_matrix_sync(reinterpret_cast<int*>(s_s) + warp * 16, cf, HEAD_DIM, mem_row_major);
+            // ldm = 128: the QK result is a [16 q-rows x up-to-128 tokens] score tile, so its row
+            // stride is the group token width (128), not HEAD_DIM — the two only coincide at
+            // hd128. With HEAD_DIM as ldm, the hd256 instantiation stored rows 256 apart while
+            // the softmax below reads them 128 apart: rows interleave with garbage, and every
+            // decoded token past the mma-engagement depth is wrong (verified: 100% argmax
+            // divergence vs the exact tile path at >16k on Qwen3.6).
+            store_matrix_sync(reinterpret_cast<int*>(s_s) + warp * 16, cf, 128, mem_row_major);
         }
         __syncthreads();
         // Read the raw int32 QK scores directly and apply the per-row/per-token scales inline in the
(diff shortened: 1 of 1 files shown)


skyrocket2026:

```bash
```

reyanthony062001-ops:
The before→after table genuinely can't show a gain against current main at 16k/32k: main's hd256 MMA kernel skips half its PV math and half its V reads while producing incorrect output past the engagement depth — #299 has the evidence (99.9% argmax divergence vs the exact tile path, single-call |Δlogprob| 3.87 at position 16385 on a bit-identical KV prefix; the 16k/32k frontier carries +4.35%/+12.22% of that inflation).

Against the correct baseline (`SPARKINFER_FAMMA=0` tile path, same box, median of 3) this branch measures **+3.12% @16k / +10.13% @32k**, holds ≥98% of current main at every context on both models, and passes the accuracy gate (top-1 0.960, KL 0.0495). Full tables and methodology are in the PR body.

Happy to re-fill the table for the standard eval once the 16k/32k baseline reflects correct output — or measure any comparison you'd prefer.

PR #353, opened 2026-07-12
perf(qwen36): Q8_0→Q4_K requant of full-attention q/o projections

blinkeye-lcm:
## Summary

One load-time weight requantization for the Qwen3.6 decode path, gated to the Qwen3.6 fingerprint and off by default on every other model.

**Q8_0 → Q4_K full-attention q/o.** Qwen3.6-35B-A3B UD ships the full-attention query and output projections (`attn_q`, `attn_output` on the 10 full-attn layers) as Q8_0 — 34 bytes/block on the per-token weight read. At short context, where weight bandwidth dominates decode, those matvecs are a measurable slice of the token. This requantizes them to Q4_K once at model load and decodes them through the existing int8 dp4a Q4_K MMVQ path — ~47% fewer bytes on those reads. No decode-kernel changes: the q/o projection sites already branch on the stored tensor type, so flipping Q8_0→Q4_K at load routes them through `launch_mmvq_q4k` automatically.

The fit is a new Lloyd-max coordinate-descent Q4_K quantizer (`proj_q4k_lloyd.cu`): each 32-value group alternates nearest-code assignment with a joint weighted least-squares solve for (scale, min) to convergence, then re-fits the 4-bit codes against the quantized super-block levels. It is applied only to the Q8_0 attention path; the Q6_K dense-FFN down requant keeps its existing affine fit.

On by default for the Qwen3.6 fingerprint (`is_qwen35_or_qwen36_hybrid_moe`); a strict no-op on the dense Qwythos path and any non-matching model.

A/B toggle: `SPARKINFER_ATTN_REQUANT_Q4K=off` restores native Q8_0 reads for the comparison below.

## Proof of speedup

- [x] Tested on **RTX 5090** (`sm_120`)

**Decode tok/s** (end-to-end, `qwen3_gguf_bench`, `Qwen3.6-35B-A3B-UD-Q4_K_M.gguf`, 128 decode tokens), default vs `SPARKINFER_ATTN_REQUANT_Q4K=off`, same binary:

| | decode tok/s (128) |
|---|--:|
| before (`=off`, native Q8_0) | 420.7 |
| after (this PR, default) | 434.2 |

```text
# ctx=128:  off 420.73 -> default 434.18  (+3.20%)
# ctx=512:  off 416.37 -> default 428.89  (+3.01%)
# ctx=32k:  default 376.58  (win vs ~366 native; no regression at long context)
```

[...]

PR #350, opened 2026-07-12
perf(qwen36): opt-in Q8→Q4 requant reward profile (~+6.3% @512)

andriypolanski:
## Summary

Load-time **Q8_0 → Q4_K requant** for Qwen3.6 UD GDN projections, gated behind a single SN74-facing knob:

```bash
SPARKINFER_Q80_REQUANT_PROFILE=reward
```

On `Qwen3.6-35B-A3B-UD-Q4_K_M.gguf` @ ctx=512 (128 generated tokens, RTX 5090), this trims ~47% of weight bytes on the hottest GDN qkv+gate reads on **layers ≥15** while staying inside the accuracy gate (top-1 ≥ 0.90, KL ≤ 0.20).

**Bottleneck:** Qwen3.6 UD ships attn/GDN weights as Q8_0. At short context those MMVQ reads dominate decode (~36% GPU time @512). Requantizing once at load and decoding via the existing Q4_K int8 MMVQ path cuts bandwidth on the selected tensors without touching MoE experts, shared expert Q8 paths, or the LM head.

**Scope:** `kernels/csrc/cuda/quant/ffn_down_requant.cu`, `kernels/include/sparkinfer/kernels/quant.h`, `runtime/src/models/qwen35.cpp` only — per [CONTRIBUTING.md](CONTRIBUTING.md) miner paths.

**Default:** OFF. No env var → native Q8_0 weights (same path as main for Qwen3.6 UD).

## Proof of speedup

- [x] Tested on **RTX 5090** (`sm_120`)

**Decode tok/s** (end-to-end, same binary, interleaved env on/off @ ctx=512, n=128, after merge with `main`):

| | decode tok/s |
|---|--:|
| before (no `PROFILE`, native Q8) | 422.2 |
| after (`SPARKINFER_Q80_REQUANT_PROFILE=reward`) | 449.0 |

```text
# Qwen3.6-35B-A3B-UD-Q4_K_M.gguf, ctx=512, 128 tokens, RTX 5090
# interleaved (before / after / before / after):
#   422.34  →  449.39  →  422.15  →  448.66 tok/s
# mean before ~422.2, mean after ~449.0  (+6.3% @512)
```

## Correctness gate

Same binary, `PROFILE=reward` vs default OFF, held-out 100 positions (`self_consistency.py`):

| config | top-1 vs before | KL | gate |
|---|---:|---:|---|
| before (no env) | — | — | — |
[...]

PR #329, opened 2026-07-10
perf(qwen35): Q4_K requant for Qwythos Q6 decode reads

9876543210-tc-0123456789:
## Summary

Requantize the remaining high-traffic Q6_K decode reads for Qwythos/Qwen3.5 dense-9B to Q4_K at load time, using the existing GGUF dequant/requant path and existing Q4_K MMVQ decode kernels. This covers dense FFN down weights, selected linear-attention QKV weights, and the LM head under the dense-9B shape guard.

The Qwythos dense-9B path is default-on with the upstream `SPARKINFER_DOWN_REQUANT_Q4K` semantics: unset or nonzero enables it, and `SPARKINFER_DOWN_REQUANT_Q4K=0` keeps the native Q6_K reads.

## Proof of speedup

> ⚠️ **The on-device eval runs only when BOTH are true:** (1) the box below is ticked, and
> (2) the **decode tok/s** table shows a **real end-to-end improvement** (`after > before`,
> filled from `bench/scripts/bench.sh` / end-to-end `qwen3_gguf_bench`, not an isolated-kernel microbenchmark).
>
> Tick the box **only if you actually ran it on an RTX 5090**.

- [x] Tested on **RTX 5090** (`sm_120`)

**Decode tok/s** (end-to-end, from `bench/scripts/bench.sh` / `qwen3_gguf_bench` — required for evaluation):

| | decode tok/s |
|---|--:|
| before (main) | 285.38 |
| after (this PR) | 304.59 |

```text
# Qwythos-9B-Claude-Mythos-5-1M-Q4_K_M — RTX 5090, CUDA 13, sm_120
# Baseline: origin/main @ 95b93ce
# Candidate: PR #329 @ 47f2e8e
# Env: default requant env, SPARKINFER_BENCH_DEVICE_LOOP=1
# n=128, bs=1, ctx=0, median of 3 runs

=== before (main @ 95b93ce) ===
decode tg    : 284.89 tok/s  (ctx=0, n=128)
decode tg    : 285.38 tok/s  (ctx=0, n=128)
decode tg    : 285.56 tok/s  (ctx=0, n=128)
median       : 285.38 tok/s
VRAM used    : 8.0 GB

=== after (this PR @ 47f2e8e) ===
decode tg    : 304.49 tok/s  (ctx=0, n=128)
decode tg    : 304.61 tok/s  (ctx=0, n=128)
decode tg    : 304.59 tok/s  (ctx=0, n=128)
median       : 304.59 tok/s
[...]

PR #323, opened 2026-07-10
perf(qwen35): requantize dense FFN down Q6_K→Q4_K at load (~5% Qwythos decode)

James-CUDA:
## Summary

Load-time requantize of Qwen3.5 dense-hybrid FFN down weights Q6_K→Q4_K (gated by `SPARKINFER_DOWN_REQUANT_Q4K=1`) so decode uses the existing Q4_K MMVQ path and reads ~31% less down-projection bandwidth; default path unchanged.

## Proof of speedup

> ⚠️ **The on-device eval runs only when BOTH are true:** (1) the box below is ticked, and
> (2) the **decode tok/s** table shows a **real end-to-end improvement** (`after > before`,
> filled from `bench/scripts/bench.sh` — *not* an isolated-kernel microbenchmark). A ticked box
> with an empty/placeholder table gets `needs-benchmark` and is **not** evaluated (no point spending
> a GPU when there's no claimed decode gain).
>
> Tick the box **only if you actually ran it on an RTX 5090**. False attestation is treated as
> gaming — the account is **blocked** ([`.github/blocked-contributors.txt`](blocked-contributors.txt)),
> same as copycatting or sybil farming.

- [x] Tested on **RTX 5090** (`sm_120`)

**Decode tok/s** (end-to-end, from `bench/scripts/bench.sh` — required for evaluation):

| | decode tok/s |
|---|--:|
| before (main) | 249.25 |
| after (this PR) | 261.52 |

```text
# Qwythos-9B Q4_K_M — RTX 5090, cb939a9 baseline + this PR, ctx=128, n=128, bs=1
# Model: /workspace/models9/Qwythos-9B-Claude-Mythos-5-1M-Q4_K_M.gguf
# Median of 3 runs per config (qwen3_gguf_bench)

=== before (main @ cb939a9, SPARKINFER_DOWN_REQUANT_Q4K=0) ===
model        : Qwen3.5-9B dense hybrid  (32 layers, 1 experts top-1)
decode tg    : 249.25 tok/s  (median, n=128, ctx=128, bs=1)  # runs: 249.43 249.25 248.91
decode tg    : 253.05 tok/s  (median, n=128, ctx=0,   bs=1)  # runs: 253.05 253.12 253.01

=== after (this PR, SPARKINFER_DOWN_REQUANT_Q4K=1) ===
model        : Qwen3.5-9B dense hybrid  (32 layers, 1 experts top-1)
[...]

PR #358, opened 2026-07-13
perf(qwen36): Q4_K shared-expert FFN kernels + Q8_0/Q5_K→Q4_K requant of the MoE FFN stage (+2.8–3.2%)

carlh7777:
## Summary

Two load-time weight requantizations targeting the **MoE FFN stage** of the Qwen3.6-35B-A3B decode path, gated to the Qwen3.6 fingerprint (`is_qwen35_or_qwen36_hybrid_moe`) and off by default on every other model. The FFN stage runs on every token across all 40 layers, so shrinking its per-token weight reads is a broad, context-flat decode win.

1. **Shared-expert FFN Q8_0 → Q4_K (new kernels).** Qwen3.6-35B-A3B UD ships the always-on shared expert (`ffn_{gate,up,down}_shexp`) as **Q8_0** — 34 bytes/block on a read that fires every layer, every token. This requantizes it to Q4_K once at model load (~47% fewer weight bytes) and decodes it through **two new Q4_K shared-expert MMVQ kernels** (`shared_gate_up_q4k_mmvq_kernel`, `shared_down_q4k_mmvq_kernel` + launcher `launch_shared_expert_q4k_mmvq`). They mirror the existing Q8 shared-expert control flow (FNQ Q8_1 → gate/up·silu → quant_h Q8_1 → down), only the weight dot swaps to the int8 dp4a Q4_K path. Gated to the exact shared-expert shape the kernels specialize (`<2048, 512>`).
2. **Routed-expert down Q5_K → Q4_K (no new kernel).** The top-8 routed experts keep their down projection at Q5_K while gate/up are already Q4_K. That down read sits in the *same* decode-critical FFN stage, so requantizing `ffn_down_exps` Q5_K→Q4_K at load stacks a further byte reduction on the same stage and routes automatically through the pre-existing Q4_K routed-down MMVQ dispatch (`down_type == 12`).

Strict no-op on Qwythos-9B and any non-matching model.

**A/B toggles (same binary):** `SPARKINFER_SHEXP_REQUANT_Q4K=0` restores native Q8_0 shared-expert reads; `SPARKINFER_ROUTED_DOWN_REQUANT_Q4K=0` restores native Q5_K routed-down. Both set = the "before" column below.

## Proof of speedup
- [x] Tested on RTX 5090 (sm_120)

Decode tok/s (end-to-end, `qwen3_gguf_bench`, `Qwen3.6-35B-A3B-UD-Q4_K_M.gguf`, 128 decode tokens), default vs `SPARKINFER_SHEXP_REQUANT_Q4K=off` + `SPARKINFER_ROUTED_DOWN_REQUANT_Q4K=off`, same binary:

|                          | decode tok/s (128) |
| ------------------------ | ------------------ |
| before (=off, native)    | 437.1              |
| after (this PR, default) | 449.3              |

```
# ctx=128:   off 437.11 -> default 449.29  (+2.79%)
# ctx=512:   off 427.57 -> default 440.14  (+2.94%)
# ctx=4k:    off 408.02 -> default 419.86  (+2.90%)
# ctx=16k:   off 397.62 -> default 410.51  (+3.24%)
# ctx=32k:   off 375.55 -> default 387.46  (+3.17%)
```

All contexts clear the 2% gate. The FFN weight read fires every token regardless of KV depth, so the win is flat across context — a byte-reducing requant cannot slow decode. VRAM 23.7 → 22.0 GB. Qwythos-9B unchanged (fingerprint-gated no-op).

Cross-checked against a built-from-source `origin/main @ a502a59` baseline (before = main, after = this branch, back-to-back same box): 438.5 → 448.7 @128 and 375.7 → 387.4 @32k — the `off` column above reproduces main within ~0.3% at every context.

## Accuracy

The requant is lossy but confined to the accuracy-tolerant FFN projections. Teacher-forced top-1 agreement of the default (Q4_K) build vs the native (off) build on a held-out 299-token window:

```
positions        : 299
top-1 agreement  : 286/299 = 0.9565   (gate >= 0.90)
mean self-KL      : 0.0342 nats        (gate <= 0.20)
PPL  off 4.486 -> default 4.507
```

Subject to the harness KL/top-1 gate (`bench/scripts/accuracy.sh`, top-1 ≥ 0.90, KL ≤ 0.20).

## Relation to open/merged PRs

Original — no open or merged PR requantizes the Qwen3.6 **shared-expert FFN** or adds Q4_K shared-expert kernels, and none requantizes the **routed-expert down** (Q5_K→Q4_K).

- The two new kernels + launcher live in `kernels/csrc/cuda/moe/expert_ffn_q4k.cu` / `moe.h` — the shared-expert Q4_K path is new code touched by no other PR.
- Distinct from **#353** (Q8→Q4 of full-attention *q/o* projections, Lloyd-max fit): different tensor class (MoE FFN vs attention), and this adds new kernels rather than reusing an existing MMVQ path.
- Distinct from **#323** (dense FFN down, Qwythos), **#329** (Qwythos Q6 QKV/lm-head), **#350** (GDN attn_qkv/attn_gate), **#308** (lm-head): none touch the shared-expert FFN or routed-expert down.
- Only shared file is `runtime/src/models/qwen35.cpp` (+43/−9); the reused lines are the generic `dev_quant_requant_q4k` helper (extended to accept a Q5_K source) and the load-loop call sites — standard requant boilerplate, not another PR's contribution.

## Test plan
- [x] Build `qwen3_gguf_bench` / `qwen3_gguf_score` on RTX 5090 (sm_120)
- [x] Same-box A/B at 128/512/4k/16k/32k, default vs both-off
- [x] Built-from-source `origin/main` baseline vs branch, back-to-back
- [x] Teacher-forced top-1 / self-KL vs native (off) build
- [x] Qwythos-9B guard: fingerprint-gated no-op

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,9 @@ __pycache__/
 .prebuilt/
 .llamacpp/
 /models/
+/models35/
+/models36/
+*.gguf
 dist/
 .claude
 .env.eval
--- a/kernels/csrc/cuda/moe/expert_ffn_q4k.cu
+++ b/kernels/csrc/cuda/moe/expert_ffn_q4k.cu
@@ -588,6 +588,63 @@ __global__ void shared_down_q8_mmvq_kernel(
     }
 }
 
+// ---- Qwen3.6 shared expert, Q4_K gate/up/down (requant of the UD Q8_0 weights) -------
+// Byte-for-byte the same fused shared-expert math as the Q8_0 kernels above, but the
+// per-token weight read is Q4_K super-blocks through the faithful si_vec_dot_q4_K dp4a
+// path the routed experts and dense FFN already use — ~47% fewer weight bytes/token on
+// the always-on shared expert (40 layers). One 4-warp block per gate/up output row
+// (F rows, K=H -> NB=H/256 super-blocks, 16 lanes/super-block); one warp per down output
+// row (H rows, K=F -> F/256 super-blocks, two half-warps stride alternate super-blocks).
+// The Q8_1 activation buffer (hn / h) is unchanged.
+template <int H, int F>
+__global__ void shared_gate_up_q4k_mmvq_kernel(
+    const si_block_q8_1* __restrict__ vy, const unsigned char* __restrict__ gate_q,
+    const unsigned char* __restrict__ up_q, const float* __restrict__ dw,
+    float* __restrict__ h_scratch) {
+    constexpr int NW = 4, WS = 32, NB = H >> 8;
+    const int f = blockIdx.x, lane = threadIdx.x & 31, warp = threadIdx.x >> 5, tid = threadIdx.x;
+    const int kbx0 = tid >> 4, kqs = 2 * (tid & 15);
+    const si_block_q4_K* g_row = (const si_block_q4_K*)(gate_q + (size_t)f * NB * 144);
+    const si_block_q4_K* u_row = (const si_block_q4_K*)(up_q   + (size_t)f * NB * 144);
+    float tg = 0.f, tu = 0.f;
+    for (int kbx = kbx0; kbx < NB; kbx += 8) {
+        tg += si_vec_dot_q4_K(g_row + kbx, vy + (size_t)kbx * 8, kqs);
+        tu += si_vec_dot_q4_K(u_row + kbx, vy + (size_t)kbx * 8, kqs);
+    }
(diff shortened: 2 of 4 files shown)


skyrocket2026:
## 🚩 Flagged: eval-gaming

This PR involves an account blocked for gaming the SN74 emission mechanism (sybil / coordinated duplicate farming): `carlh7777`.

PR #379, opened 2026-07-13
perf(attn): sink + sliding-window sparse KV for Qwythos long decode

Paral1995:
## Summary

Off-by-default sink + sliding-window sparse KV for Qwythos GQA-4 hd256 full-attention decode (int8 KV, ctx ≥ 8192): attend block 0 + last `SPARKINFER_SPARSE_WINDOW` blocks, then sparse flash-split. Default OFF — byte-identical to dense when disabled.

**Copycat check:** not a copycat. Real-time guard: **7% PR-level containment** (block threshold ≥85%). `copycat-warn` is a per-function strike only (`skv_wsum` warp-shuffle matches boilerplate in #223). No other open PR implements `SPARKINFER_SPARSE_KV` / sliding-window full-attn sparse decode.

## Proof of speedup

> ⚠️ **The on-device eval runs only when BOTH are true:** (1) the box below is ticked, and
> (2) the **decode tok/s** table shows a **real end-to-end improvement** (`after > before`,
> filled from `bench/scripts/bench.sh` — *not* an isolated-kernel microbenchmark). A ticked box
> with an empty/placeholder table gets `needs-benchmark` and is **not** evaluated (no point spending
> a GPU when there is no claimed decode gain).
>
> Tick the box **only if you actually ran it on an RTX 5090**. False attestation is treated as
> gaming — the account is **blocked** ([`.github/blocked-contributors.txt`](blocked-contributors.txt)),
> same as copycatting or sybil farming.

- [x] Tested on **RTX 5090** (`sm_120`)

**Decode tok/s** (end-to-end, from `bench/scripts/bench.sh` — required for evaluation):

| | decode tok/s |
|---|--:|
| before (main) | 255.70 |
| after (this PR) | 289.68 |

Measured at **ctx=32768**, `SPARKINFER_SPARSE_KV=1`, `SPARKINFER_SPARSE_WINDOW=256` (sparse inactive below 8192 tokens; default env = dense path). Accuracy gate (top-1/KL vs llama.cpp) **not yet validated** at long ctx — experimental approximation, not lossless long-range recall.

```text
# dense (main), ctx=32768
decode tg    : 255.70 tok/s  (3.9 ms/token, n=24, ctx=32768, bs=1)

# SPARKINFER_SPARSE_KV=1 SPARKINFER_SPARSE_WINDOW=256, ctx=32768
decode tg    : 289.68 tok/s  (3.5 ms/token, n=24, ctx=32768, bs=1)

[...]

PR #366, opened 2026-07-13
perf(qwythos): GQA-4 int8 MMA flash-decode + tiered KV splits (+4% @64k)

James-CUDA:
## Summary

Qwythos/Qwen3.5-9B dense hybrid (hd256, GQA-4) long-context decode improvements:

1. **GQA-4 int8 MMA flash-decode** — route full-attn through `fa_split_gqa_mma_i8<256,4>` when int8 KV is active (`SPARKINFER_FAMMA4`, default ON).
2. **Depth-tiered KV splits** — promote `n_splits` to 192 @64k and 128 @128k for GQA-4 (flat 160 through 32k unchanged); Qwen3.6 GQA-8 path untouched.
3. **H=4096 attn fusions** — enable fused `attn_qkv` and `attn_gate_q8` MMVQ paths for Qwythos projection width.
4. **int8 KV @4k** — lower context-adaptive int8 KV threshold from 8k→4k in bench/generate/score so maintainer eval contexts match production path.
5. **Q6 attn requant defaults** — Lloyd fit for dense-9B `qkv,v` tensors (layer 2 included after re-validation).

## Proof of speedup

- [x] Tested on **RTX 5090** (`sm_120`)

**Decode tok/s** (end-to-end, from `bench/scripts/bench.sh` — required for evaluation):

| | decode tok/s |
|---|--:|
| before (main) | 225.73 |
| after (this PR) | 235.22 |

```text
# Qwythos-9B-Claude-Mythos-5-1M Q4_K_M — RTX 5090, 91.224.44.227:40058, build-sm120
# Model: /workspace/models35/Qwythos-9B-Claude-Mythos-5-1M-Q4_K_M.gguf
# qwen3_gguf_bench n=128 ctx=65536 bs=1 (single run each, 2026-07-13)

=== before (flat n_splits=160, SPARKINFER_NSPLITS=160 — tiering disabled) ===
[nsplits] flash-decode splits = 160 (fixed env override)
decode tg    : 225.73 tok/s  (4.4 ms/token, n=128, ctx=65536, bs=1)
VRAM used    : 11.8 GB

=== after (this PR, tiered default — n_splits=192 @64k band) ===
decode tg    : 235.22 tok/s  (4.3 ms/token, n=128, ctx=65536, bs=1)
VRAM used    : 11.8 GB

# +4.2% @64k decode (primary scored context)
# Short-ctx sanity (same box, tiered default): 301.32 @128, 258.22 @32k — no regression
[...]

PR #294, opened 2026-07-09
perf(qwen36): decode fuses + FAGQA4 restore (+3.5% @32k vs current main)

jimcody1995:
## Summary

Rebased onto current `main` (includes **#338** n_splits=160). Adds banded adaptive `n_splits` (256 in 8k–12k band), dual-row MMVQ (K=2048), qknorm+int8 KV fuse, fused full-attn QKV MMVQ, gated-attn combine→Q8, and shared-gate Q4_K mmvq+sigmoid.

**Fixes eval regression from `a36d27d`:** restored main's `FAGQA4` hd256 GQA-4 flash-decode path (Qwythos was falling back to scalar split). Rebased conflict in `qwen35.cpp` resolved by keeping banded `n_splits` + #338's `n_splits=160` occupancy correction (`want >= 128` gate).

## Proof of speedup

> ⚠️ **The on-device eval runs only when BOTH are true:** (1) the box below is ticked, and
> (2) the **decode tok/s** table shows a **real end-to-end improvement** (`after > before`,
> filled from `bench/scripts/bench.sh` — *not* an isolated-kernel microbenchmark). A ticked box
> with an empty/placeholder table gets `needs-benchmark` and is **not** evaluated (no point spending
> a GPU when there's no claimed decode gain).
>
> Tick the box **only if you actually ran it on an RTX 5090**. False attestation is treated as
> gaming — the account is **blocked** ([`.github/blocked-contributors.txt`](blocked-contributors.txt)),
> same as copycatting or sybil farming.

- [x] Tested on **RTX 5090** (`sm_120`)

**Decode tok/s** (end-to-end, from `bench/scripts/bench.sh` — required for evaluation):

| | decode tok/s |
|---|--:|
| before (main) | 372.43 |
| after (this PR) | 385.58 |

```text
# bench/scripts/bench.sh — RTX 5090, Qwen3.6-35B-A3B-UD-Q4_K_M.gguf, n=128, bs=1
# SPARKINFER_BENCH_DEVICE_LOOP=0
# Rebased onto origin/main (includes #338 n_splits=160) @ cf2fd83

=== before (origin/main, same box) ===
$ bench/scripts/bench.sh models36/Qwen3.6-35B-A3B-UD-Q4_K_M.gguf --tokens 128 --ctx 32768
decode tg    : 372.43 tok/s  (2.7 ms/token, n=128, ctx=32768, bs=1)

=== after (this PR @ cf2fd83) ===
[...]

PR #389, opened 2026-07-13
fix(kernels): re-land the hd256 int8-MMA correctness fixes reverted by #294

### LINKED ISSUE #388: regression: #294's rebase reverted the #300 hd256 int8-MMA correctness fixes — long-context MMA output is wrong again (and #366's GQA-4 path inherited it)
## Summary

The #294 merge (`a3b63f8`) overwrote `kernels/csrc/cuda/attention/flash_decode_split.cu` with a kernel body from a pre-#300 branch base, silently reverting all four hd256 correctness fixes that were merged in `4f35ad6` (#300, issue #299). #366 then added the Qwythos GQA-4 instantiation (`fa_split_gqa_mma_i8_kernel<256,4>`) on the same reverted body, so both models' long-context int8-MMA flash-decode now computes wrong output wherever the MMA gate engages.

Verifiable without a GPU — diff the merged fix against current main:

```
git diff 4c1e8c3 origin/main -- kernels/csrc/cuda/attention/flash_decode_split.cu
```

All four #300 hunks are gone at HEAD:

1. Q quantization is back to `float qv[4]` / `e < 4` (128 dims): at hd256, `s_qi` dims 128..255 feed the QK mma uninitialized shared memory and the per-row amax covers half the vector.
2. The QK score `store_matrix_sync` is back to `ldm = HEAD_DIM`: the score tile row stride is the group token width (128), so at hd256 the softmax reads interleaved garbage (row r reads row r/2's scores or unwritten memory).
3. The P′ fragment `load_matrix_sync` is back to `ldm = HEAD_DIM` while P′ rows are written at stride 128.
4. The PV phase is back to a single 128-wide pass: O dims 128..255 stay at their zero init for every head.

## Why the eval doesn't catch it

The accuracy gate (`evaluate.sh` → `accuracy.sh`) feeds ~101 prompt tokens to `qwen3_gguf_score`, and the score tool's hybrid int8-KV switch is `(argc - 3) >= 4096` — so correctness always runs with int8 KV **off**, on the (correct) bf16 tile path. The speed benches run `qwen3_gguf_bench` at ctx ≥ 4k with int8 KV **on**, engaging the broken MMA. The gate never executes the kernel the tok/s numbers come from. (#299 documented the same split-brain before #300.)

Consequences on current main:
- Qwen3.6 16k/32k frontier baselines are measured on attention that skips half the PV math and half the V reads — inflated exactly like the pre-#300 numbers were (#300's merge corrected the 16k/32k dashboard by −1.7%/−3.2%). Every open PR scored at those contexts is compared against the inflated baseline.
- This extends to the new prefill lane: #383 made Qwythos prefill pp a guarded/scored metric, and prefill is a **sequential KV fill through this same flash-decode kernel** — prefill positions ≥ 4961 ran the corrupted MMA when they were measured. The `reference.lock` pp baselines at 32k/64k/128k (measured 2026-07-13) inherit that; only the 4k pp baseline is clean (4096 < 4961, the gate never engages).
- **#379's validation is void because of this bug.** `cmp_argmatch.py` compares sparse-on vs sparse-off `qwen3_gguf_score` runs at positions ≥ 8192 — but feeding ≥ 8192 tokens flips the score tool's int8 switch, so the *dense reference arm ran this corrupted MMA kernel* (100.0% argmax-wrong on Qwythos per the table below). Whatever argmatch it reported at long positions measured agreement with garbage. Separately, #379 means Qwythos ≥ 8k now attends to ~4.1k of up to 131k tokens (sink + window) — a real modeling change — and the accuracy gate (101-token prompt) is structurally unable to see either that or this bug. Two long-context modeling changes (broken MMA math since #294, KV windowing since #379) have now shipped with zero correctness coverage at the contexts they affect.

## Where it engages

With int8 KV active and the current `n_splits=160` hd256 policy, the MMA gate (`chunk >= 32 && seqlen > 512`) engages from **seqlen ≥ 4961** (160 splits × 31 + 1). So on the eval:
- Qwen3.6 scored contexts **16k and 32k** run the broken kernel (4k stays on the tile path — chunk 26 < 32 — which is why every ≤4k correctness reading passes).
- Qwythos scored contexts **32k, 64k, 128k** ran it through the #366 GQA-4 instantiation until today; since #379 merged, the default Qwythos path at seqlen ≥ 8192 is the new (separate, scalar) sparse kernel, which shrinks the broken kernel's default Qwythos exposure to the 4961–8191 dense band, prefill positions in that band, and any `SPARKINFER_SPARSE_KV=0` run — but it was the reference for #379's validation (above), and it still owns both Qwen3.6 scored long contexts.

## Evidence (RTX 5090, teacher-forced 17.5k-token probe, same method as #299)

The probe diverges at exactly position 4960 (0-indexed) — the first engaged kernel call — and is bit-identical before it. `SPARKINFER_FAMMA=1` (default, MMA) vs `SPARKINFER_FAMMA=0` (exact int8 tile path), same 17.5k-token feed, measured at `aeb921c` (the MMA kernel body is byte-identical through current `36bde95` — #382..#386 touch only eval/bench/dashboard files, and #379 adds a separate sparse kernel plus a combine wrapper without touching the MMA body):

**Qwen3.6-35B-A3B UD-Q4_K_M (GQA-8):**

| | positions < 4960 | positions 4960..17498 (MMA engaged) |
| --- | --- | --- |
| main, MMA vs tile | 0 mismatches, |Δlp| = 0 (bit-identical) | **12529/12539 argmax mismatch (99.9%)**, mean |Δlp| 5.51, growing 4.6→6.0 across the window |
| fix re-landed, MMA vs tile | 0 mismatches | 879/12515 (7.0%), mean |Δlp| 0.06, **flat** 6–8% in every bucket |

The 7% flat residual is the int8 Q/P quantization noise the scheme carries by design; the 99.9% growing divergence is corruption. First-engaged-position |Δlp|: main 8.23, fixed 0.10.

**Qwythos-9B Q4_K_M (GQA-4, the #366 path — `SPARKINFER_FAMMA4` toggle):**

| | positions < 4960 | positions 4960..17498 (MMA engaged) |
| --- | --- | --- |
| main, MMA vs tile | 0 mismatches, |Δlp| = 0 (bit-identical) | **12539/12539 argmax mismatch (100.0%)**, mean |Δlp| 11.1 (buckets up to 17.5) |
| fix re-landed, MMA vs tile | 0 mismatches | 712/12494 (5.7%), mean |Δlp| 0.13, flat |

## Fix

Re-land the four hunks (they are generic in HEAD_DIM and GQA, so they also correct #366's GQA-4 instantiation; hd128 codegen is unchanged by construction). PR: #389 — opened as a **draft**, because a correctness re-land is definitionally "slower" than broken math at the affected contexts, so the auto-eval would reject-and-close it against the inflated baseline (the same dynamics under which #300 was merged manually). Ready to mark for review on your signal.

Longer term: the accuracy gate should also run one long-context scored probe (≥4096 fed tokens) so the int8-MMA path — and modeling changes like #379's KV windowing — are correctness-checked at the contexts they are benched at. The `reference.lock` pp pins and the 16k/32k Qwen3.6 frontier will need re-measurement after the fix lands.

reyanthony062001-ops:
Fixes #388.

## Summary

#294's rebase overwrote `flash_decode_split.cu` with a pre-#300 kernel body, reverting the four hd256 correctness fixes merged in `4f35ad6` (#300). #366's new GQA-4 instantiation inherited the reverted body. Details, the no-GPU-needed git evidence, and what this means for the eval baselines (including #379's validation) are in #388.

This re-applies the same four fixes to the current kernel structure (which evolved since #300 — P′ quantization now covers only the live gblk×16 columns, fragment declarations moved — so this is a fresh application of the hunks, not a revert-of-a-revert):

1. `constexpr int EPT = HEAD_DIM / 32` in the Q quantization (4 elems/lane at hd128, 8 at hd256).
2. QK score `store_matrix_sync` ldm = 128 (group token width).
3. P′ fragment `load_matrix_sync` ldm = 128 (matches the stride-128 P′ writes).
4. PV runs two 128-wide dim slabs at hd256 (`dh = 0, 128`) with the epilogue accumulating each slab at its offset.

All four are generic in HEAD_DIM and GQA: at hd128 they are no-ops by construction (EPT = 4, ldm 128 == HEAD_DIM, and the dh loop runs a single pass), so Qwen3-MoE codegen is unchanged. The GQA-4 Qwythos instantiation gets the same correction; #379's new sparse kernel is a separate code path and is untouched.

## Correctness validation (RTX 5090)

Teacher-forced 17.5k-token parity probe (same method as #299/#300; the probe feeds ≥4096 tokens so int8 KV and the MMA gate engage, unlike the short accuracy.sh prompt). Reference arm is the exact int8 tile path (`SPARKINFER_FAMMA*=0`); main's MMA arm diverges at exactly position 4960 — the first MMA-engaged call — and is bit-identical before it:

**Qwen3.6-35B-A3B (GQA-8), positions 4960..17498:** main MMA = **12529/12539 argmax mismatch (99.9%)**, mean |Δlp| 5.51, growing across the window. Fixed MMA = 879/12515 (7.0%), mean |Δlp| 0.06, flat in every bucket (the int8 quantization noise the scheme carries by design).

**Qwythos-9B (GQA-4, the #366 path), positions 4960..17498:** main MMA = **12539/12539 argmax mismatch (100.0%)**, mean |Δlp| 11.1. Fixed MMA = 712/12494 (5.7%), mean |Δlp| 0.13, flat.

ctest: **7/7 pass** on the fixed build. accuracy.sh (held-out prompt vs llama.cpp at the pinned commit, per-model tokenizers): Qwen3.6 top-1 0.90 / KL 0.054, Qwythos top-1 0.90 / KL 0.034 — the short-prompt tile path is unchanged by this diff, as expected.

## Speed vs the broken-fast main

Straight talk, same as #300: the reverted kernel skips half the PV math and half the V reads, so it is artificially fast at the contexts where it engages. Where this fix now runs by default (post-#379, the broken kernel owns both Qwen3.6 long contexts, the Qwythos 4961–8191 dense band, and any `SPARKINFER_SPARSE_KV=0` run):

| model @ ctx | main (broken MMA) | this PR (fixed) | ratio |
| --- | --- | --- | --- |
| Qwen3.6 @ 16k | 437.29 | 430.92 | 98.5% |
| Qwen3.6 @ 32k | 413.41 | 404.64 | **97.9%** |

(3-rep medians, clock-pinned RTX 5090, interleaved same-box A/B at `0f67d2a`; the kernel is byte-identical through `36bde95`. 128/512/4k rows are flat — the MMA gate never engages there. A fresh same-box A/B against `36bde95` incl. Qwythos sparse-default contexts is appended below as a comment once it completes.)

The Qwen3.6 32k floor does not clear 98% — that is the direct cost of doing the previously-skipped math, not an inefficiency this PR introduces. Against the only *correct* execution main has at that context (the exact int8 tile path, i.e. the MMA gate disabled), this PR measures **+17.4%** (404.64 vs 344.53). The no-regression floor is comparing against a kernel whose output is 99.9% argmax-wrong at that context (#388); #300 was merged on exactly this two-baseline evidence, and the 16k/32k frontier corrected down with it.

**Why this is filed as a draft:** a correctness re-land is definitionally "slower" than broken math at the affected contexts, so the auto-eval would reject it against the inflated baseline (and auto-close it), same dynamics as #300. Marking ready-for-review on maintainer signal.

## Proof of speedup

<details><summary>raw bench output</summary>

```
q36 ctx=16384 rep=1 main=437.25 fixed=430.92
q36 ctx=16384 rep=2 main=437.29 fixed=430.59
q36 ctx=16384 rep=3 main=437.43 fixed=430.94
q36 ctx=32768 rep=1 main_tg=413.44 fixed_tg=404.50
q36 ctx=32768 rep=2 main_tg=413.39 fixed_tg=404.79
q36 tile-path baseline (SPARKINFER_FAMMA=0) ctx=32768: 344.53
```

</details>

### DIFF
--- a/kernels/csrc/cuda/attention/flash_decode_split.cu
+++ b/kernels/csrc/cuda/attention/flash_decode_split.cu
@@ -432,12 +432,16 @@ __global__ void __launch_bounds__(fa_mma_block_threads<HEAD_DIM, GQA>::v, 5) fa_
     float* s_l  = s_m + 16;                                           // [16]
 
     // Quantize Q per q-head row (warp w owns rows 2w, 2w+1; rows >= GQA are zero pad).
+    // EPT spans the whole head vector: 4 elems/lane at hd128, 8 at hd256. Hardcoding 4 left
+    // s_qi dims 128..255 uninitialized at hd256 (and computed amax over half the row), so the
+    // QK mma k-tiles 8..15 multiplied against stale shared memory.
+    constexpr int EPT = HEAD_DIM / 32;
     #pragma unroll
     for (int rr = 0; rr < 2; rr++) {
         const int r = warp * 2 + rr;
-        float qv[4], amax = 0.f;
+        float qv[EPT], amax = 0.f;
         #pragma unroll
-        for (int e = 0; e < 4; e++) {
+        for (int e = 0; e < EPT; e++) {
             qv[e] = (r < GQA) ? __bfloat162float(q[(size_t)(seq * num_q_heads + kvh * GQA + r) * HEAD_DIM + lane + e * 32]) : 0.f;
             amax = fmaxf(amax, fabsf(qv[e]));
         }
@@ -446,7 +450,7 @@ __global__ void __launch_bounds__(fa_mma_block_threads<HEAD_DIM, GQA>::v, 5) fa_
         const float d = amax / 127.0f;
         if (lane == 0) s_qs[r] = d;
         #pragma unroll
-        for (int e = 0; e < 4; e++)
+        for (int e = 0; e < EPT; e++)
             s_qi[r * HEAD_DIM + lane + e * 32] = (signed char)((amax == 0.f) ? 0 : (int)roundf(qv[e] / d));
     }
     for (int i = tid; i < GQA * HEAD_DIM; i += blockDim.x) s_o[i] = 0.f;
@@ -482,7 +486,13 @@ __global__ void __launch_bounds__(fa_mma_block_threads<HEAD_DIM, GQA>::v, 5) fa_
                 load_matrix_sync(bf, kb + ks * 16, KVLD);
                 mma_sync(cf, af, bf, cf);
             }
-            store_matrix_sync(reinterpret_cast<int*>(s_s) + warp * 16, cf, HEAD_DIM, mem_row_major);
+            // ldm = 128: the QK result is a [16 q-rows x up-to-128 tokens] score tile, so its row
+            // stride is the group token width (128), not HEAD_DIM — the two only coincide at
+            // hd128. With HEAD_DIM as ldm, the hd256 instantiation stored rows 256 apart while
+            // the softmax below reads them 128 apart: rows interleave with garbage, and every
+            // decoded token past the mma-engagement depth is wrong (verified: 100% argmax
+            // divergence vs the exact tile path at >16k on Qwen3.6).
+            store_matrix_sync(reinterpret_cast<int*>(s_s) + warp * 16, cf, 128, mem_row_major);
         }
         __syncthreads();
         // Read the raw int32 QK scores directly and apply the per-row/per-token scales inline in the
@@ -532,26 +542,32 @@ __global__ void __launch_bounds__(fa_mma_block_threads<HEAD_DIM, GQA>::v, 5) fa_
         }
         __syncthreads();
 
-        // PV int8 mma -> int32; O += int32 * p_scale[m].
-        {
+        // PV int8 mma -> int32; O += int32 * p_scale[m]. The 8 warps cover a 128-wide dim slab
+        // per pass (warp*16 each), so hd128 takes one pass and hd256 two (dh = 0, 128). The
+        // hd256 instantiation previously ran a single pass with HEAD_DIM strides: it computed
+        // only dims 0..127 of O (128..255 stayed at their zero init) and read the 128-stride
+        // P' rows at the wrong ldm — both fixed here; ldm for the P' fragment and the int32
+        // store is 128 (the token/slab width), which coincided with HEAD_DIM only at hd128.
+        for (int dh = 0; dh < HEAD_DIM; dh += 128) {
             fragment<accumulator, 16, 16, 16, int> cf;
             fill_fragment(cf, 0);
             for (int ks = 0; ks < gblk; ks++) {
                 const int pb = block_table[seq * max_blocks + first_blk + g0 + ks];
-                const signed char* vb = v_pool + ((size_t)pb * 16 * num_kv_heads + kvh) * HEAD_DIM + warp * 16;
+                const signed char* vb = v_pool + ((size_t)pb * 16 * num_kv_heads + kvh) * HEAD_DIM + dh + warp * 16;
                 fragment<matrix_a, 16, 16, 16, signed char, row_major> af;
                 fragment<matrix_b, 16, 16, 16, signed char, row_major> bf;
-                load_matrix_sync(af, s_pi + ks * 16, HEAD_DIM);
+                load_matrix_sync(af, s_pi + ks * 16, 128);
                 load_matrix_sync(bf, vb, KVLD);
                 mma_sync(cf, af, bf, cf);
             }
-            store_matrix_sync(reinterpret_cast<int*>(s_s) + warp * 16, cf, HEAD_DIM, mem_row_major);
+            store_matrix_sync(reinterpret_cast<int*>(s_s) + warp * 16, cf, 128, mem_row_major);
+            __syncthreads();
+            // Only the GQA real q-head rows are kept (rows GQA..15 are wmma M-padding, never
+            // written to the partials) — accumulate this 128-wide slab into s_o at its dh offset.
+            for (int i = tid; i < GQA * 128; i += blockDim.x)
+                s_o[(i >> 7) * HEAD_DIM + dh + (i & 127)] += (float)reinterpret_cast<int*>(s_s)[i] * s_ps[i >> 7];
+            __syncthreads();
         }
-        __syncthreads();
-        // Only the GQA real q-head rows are kept (rows GQA..15 are wmma M-padding, never written to
-        // the partials) — accumulate just those, halving this thread-parallel epilogue at GQA=8.
-        for (int i = tid; i < GQA * 128; i += blockDim.x) s_o[i] += (float)reinterpret_cast<int*>(s_s)[i] * s_ps[i >> 7];
-        __syncthreads();
     }
 
     for (int r = 0; r < GQA; r++) {


reyanthony062001-ops:
Fresh same-box A/B against current main `36bde95` (post-#379), i.e. this PR's exact tree vs main — clock-pinned RTX 5090, `qwen3_gguf_bench <gguf> 128 <ctx>`:

| model @ ctx | main | this PR | ratio | path exercised |
| --- | --- | --- | --- | --- |
| Qwen3.6 @ 16k | 437.22 / 437.25 | 430.54 / 430.63 | 98.5% | dense MMA (fix active) |
| Qwen3.6 @ 32k | 412.62 / 412.62 | 404.16 / 404.06 | 97.9% | dense MMA (fix active) |
| Qwythos @ 6k | 283.10 | 282.32 | 99.7% | the 4961–8191 dense band (fix active) |
| Qwythos @ 32k | 285.33 | 285.44 | 100.0% | #379 sparse — separate kernel, untouched by this diff |

Qwythos prefill pp is likewise a wash (6k: 287.05 vs 286.91; 32k: 284.79 vs 284.94). ctest 7/7 on this exact tree.

So on current main, the entire measurable cost of restoring correct hd256 MMA math is the Qwen3.6 long-context pair — 98.5% @16k / 97.9% @32k vs the broken kernel, and **+17.4% @32k vs the exact tile path**, which is the only correct execution main has at that context. Everything Qwythos-side is neutral post-#379.

reyanthony062001-ops:
New evidence for this fix, from validating an unrelated branch on a fresh RTX 5090 with a bitwise parity harness (sequential decode vs a batched schedule over the same tokens):

**On current `main`, the broken hd256 MMA path is not just wrong — it is non-deterministic.** Two schedules that feed the identical token sequence through the identical kernels produce *different* output in the MMA-engaged region (tokens ≥ ~4960): the harness fails at exactly blocks 310–317 with per-run-stable-but-schedule-dependent bytes. That is the fingerprint of the uninitialized-shared-memory defect this PR fixes (defect 1 in #388: the Q quantization covers only 128 of 256 dims, so `s_qi[128..255]` feeds the QK MMA whatever was left in smem — which depends on launch/occupancy, not just the inputs).

With this PR's fix applied as the base, the same harness passes **20/20 bitwise** across every dispatch edge (512 MMA / 4096 int8-KV / 8192 sparse) and every schedule — i.e. the kernel becomes a deterministic function of its inputs again, which it is not on `main`.

Practical consequence worth flagging: because the corruption is schedule-dependent, any measurement of a ≥5k-context path on `main` (decode tok/s, prefill pp, the dashboard frontier) is sampling one arbitrary realization of undefined behavior — not a stable baseline. The kernel body is byte-identical from `aeb921c` through current `386659c` (the intervening merges are eval/dashboard/CI and #379's separate sparse kernel), so this fix still applies cleanly and the A/B in the description above still stands.

Ready to mark for review whenever you'd like it in the eval queue — happy to keep it as a draft until the regression in #388 is triaged, since against the (broken-fast) baseline it reads as a small slowdown, exactly as #300 did before its manual merge.

skyrocket2026:
Independently reproduced and verified on a separate RTX 5090 before merging.

Confirmed the four hunks are absent at `c9e20d1` (`git diff 4c1e8c3 HEAD -- kernels/csrc/cuda/attention/flash_decode_split.cu`), and that the kernel body is otherwise byte-identical to the #300-fixed version apart from #366's `fa_mma_block_threads<HEAD_DIM, GQA>::v` launch bounds — i.e. #294's only effect on this kernel was the revert, exactly as #388 describes.

Re-derived the fix from the defect independently, then diffed against this PR: **identical**.

**Probe** — Qwen3.6-35B-A3B UD-Q4_K_M, 17.5k-token teacher-forced feed, `SPARKINFER_FAMMA=1` (MMA) vs `=0` (exact tile), same build, `SPARKINFER_KV_INT8=1`, `SPARKINFER_SPARSE_KV=0`:

| build | PPL | ARGMATCH |
|---|---:|---:|
| main, MMA (shipped default) | **133413.1** | 0.184 |
| main, exact tile | 3.508 | 0.669 |
| this PR, MMA | **3.508** | 0.670 |

| | positions < 4960 | positions >= 4960 (MMA engaged) |
|---|---|---|
| main | 0 mismatch, \|Δlp\| = 0 | **12531/12539 (99.9%)**, mean \|Δlp\| 14.72, growing 13.1 → 15.9 |
| this PR | 0 mismatch, \|Δlp\| = 0 | 259/12539 (2.1%), mean \|Δlp\| 0.047, **flat** across every bucket |

Divergence begins at exactly position 4960 — the first engaged call at `n_splits=160` (`chunk >= 32` at `seqlen >= 4961`) — and is bit-identical before it. First engaged position \|Δlp\|: **16.85 → 0.0002**. The flat ~2% residual is the int8 Q/P quantization the scheme carries by design; the growing 99.9% was corruption.

**Cost**, same box, clock-pinned, `qwen3_gguf_bench <gguf> 128 <ctx>`:

| ctx | main | this PR | ratio |
|---|---:|---:|---:|
| 16k | 472.38 | 464.71 | 98.4% |
| 32k | 444.96 | 435.94 | 98.0% |

This corroborates the author's own A/B (98.5% / 97.9%) within noise.

Merging manually, as with #300: this is a correctness re-land, so it is definitionally "slower" at the affected contexts and the auto-eval cannot reward it — 32k sits exactly on `GUARD_TOL=0.98`. The ~2% is real work the broken kernel was skipping (the second PV pass and full-width Q quantization), not a regression.

Follow-ups from #388 remain open and are **not** addressed here — the #379 sparse path dropping the attention output gate for gated models at >= 8192, and the O-projection consuming stale Q8 scratch. The `reference.lock` pp pins and the Qwen3.6 16k/32k frontier baselines will need re-measurement now that this has landed.

Thanks @reyanthony062001-ops — for the fix, and for the probe methodology, which I reused wholesale to verify it.

PR #393, opened 2026-07-14
fix(attn): sparse-KV combine drops the attention output gate (Qwythos >= 8k)

reyanthony062001-ops:
Fixes the sparse-path gate drop documented in #388 (third defect in the comment thread there — full code-path analysis included).

## What breaks

#379 wired its sparse attention as split → standalone combine (`launch_fa_combine_hd256`), and that standalone wrapper unconditionally dispatches the **ungated** `fa_combine_kernel`. For gated-attention models the output gate is normally applied in one of exactly two places: inside the gated combine (the `attn_gate_q8` configuration — which is Qwythos') or by a separate `mul_sigmoid` launch that `qwen35.cpp` deliberately skips when `attn_gate_q8` is set.

Net effect on current main: for Qwythos, `sigmoid(q_gate)` is never applied at any position with seqlen ≥ 8192 (`sparse_min_ctx`). The O-projection consumes ungated attention output (both the bf16 buffer and the emitted Q8 activations) on all 8 full-attention layers — decode and prefill both. Positions < 8192 and `SPARKINFER_SPARSE_KV=0` runs are unaffected, which is exactly why the ≤4k accuracy gate cannot see it.

## Fix

Give the standalone combine the same optional-gate contract the in-launcher combine already has (`gate && out_q8` → gated dispatch), and pass `attn_gate_q8 ? s.qgate : nullptr` at the sparse call site — the identical dispatch decision the dense branch makes a few lines below. +17/−6 across three files. With no gate supplied the wrapper compiles to the exact pre-PR dispatch, and non-gated models take the unchanged path.

## Validation (RTX 5090, sm_120)

- Static path symmetry: the diff makes the sparse branch's gating decision byte-for-byte identical to the dense branch's.
- Builds clean for `sm_120` (CUDA 13).
- **Verified on an RTX 5090 since filing** (the box I mentioned was down at filing time came back up): a teacher-forced bitwise parity harness at the sparse boundary shows the fix restores correct attention output past position 8192 — and, at a window set to cover all blocks (where sparse ≡ dense mathematically), the fixed sparse path becomes **bitwise-equal to the dense path**, which it is not on `main`. This also exposed that the ungated hd256 combine's fused Q8 emit is compile-time dead (`ELEMS==1` guard vs hd256's ELEMS==2), so this PR additionally fixes a stale-scratch O-projection on the sparse path — details and the md5-exact replication of `main`'s corrupted logits are in the comment thread above and in #388.
- No speed claim: the gated combine differs from the plain one by an elementwise multiply, so the decode/prefill no-regression floors are unaffected. The before/after tables are intentionally empty — this is a correctness fix, not an optimization.

## Proof of speedup

- [x] Tested on **RTX 5090** (`sm_120`) — box ticked to attest the RTX 5090 correctness validation above. This is a **correctness fix with no throughput claim**, so the before/after speed tables are deliberately left empty (expected routing: `needs-benchmark`).

### DIFF
--- a/kernels/csrc/cuda/attention/flash_decode_split.cu
+++ b/kernels/csrc/cuda/attention/flash_decode_split.cu
@@ -637,13 +637,22 @@ static inline void fa_launch_combine_gated_dispatch_hd256(
 }
 
 // Standalone hd256 combine (sparse-KV path: split then combine). num_seqs=1 (decode).
+// attn_gate: optional per-element sigmoid output gate — same contract as the gated combine
+// inside launch_flash_decode_split (gate && out_q8 selects the gated kernel).
 void launch_fa_combine_hd256(
     const float* part_m, const float* part_l, const float* part_acc, void* out,
-    int num_q_heads, int n_splits, void* out_q8, cudaStream_t stream
+    int num_q_heads, int n_splits, void* out_q8, cudaStream_t stream,
+    const void* attn_gate
 ) {
-    fa_launch_combine_dispatch_hd256(part_m, part_l, part_acc,
-        reinterpret_cast<__nv_bfloat16*>(out), num_q_heads, n_splits,
-        reinterpret_cast<fa_block_q8_1*>(out_q8), 1, stream);
+    const __nv_bfloat16* gate = reinterpret_cast<const __nv_bfloat16*>(attn_gate);
+    if (gate && out_q8)
+        fa_launch_combine_gated_dispatch_hd256(part_m, part_l, part_acc,
+            reinterpret_cast<__nv_bfloat16*>(out), gate, num_q_heads, n_splits,
+            reinterpret_cast<fa_block_q8_1*>(out_q8), 1, stream);
+    else
+        fa_launch_combine_dispatch_hd256(part_m, part_l, part_acc,
+            reinterpret_cast<__nv_bfloat16*>(out), num_q_heads, n_splits,
+            reinterpret_cast<fa_block_q8_1*>(out_q8), 1, stream);
 }
 
 void launch_flash_decode_split(
--- a/kernels/include/sparkinfer/kernels/attention.h
+++ b/kernels/include/sparkinfer/kernels/attention.h
@@ -150,7 +150,8 @@ void launch_flash_decode_global_hd512(
 );
 
 void launch_fa_combine_hd256(const float* part_m, const float* part_l, const float* part_acc,
-    void* out, int num_q_heads, int n_splits, void* out_q8, cudaStream_t stream = nullptr);
+    void* out, int num_q_heads, int n_splits, void* out_q8, cudaStream_t stream = nullptr,
+    const void* attn_gate = nullptr);
 
 // Sink + sliding-window sparse-KV (Qwythos GQA-4 hd256). Default on; SPARKINFER_SPARSE_KV=0 disables.
 void launch_fa_kv_window_select(const int* seq_lens, int* sel_blk, int num_kv_heads,
--- a/runtime/src/models/qwen35.cpp
+++ b/runtime/src/models/qwen35.cpp
@@ -750,7 +750,8 @@ int Qwen35Model::forward_token(int token_id, int position) {
                     s.kv->block_size(), s.kv->max_blocks_per_seq(), s.n_splits, s.sparse_budget,
                     1.f / sqrtf((float)c.head_dim), kscale, vscale, st);
                 kernels::launch_fa_combine_hd256(s.fa_m, s.fa_l, s.fa_acc, s.attn, c.n_q_heads,
-                    s.n_splits, (emit_attn_q8 || attn_gate_q8) ? s.aq81 : nullptr, st);
+                    s.n_splits, (emit_attn_q8 || attn_gate_q8) ? s.aq81 : nullptr, st,
+                    attn_gate_q8 ? s.qgate : nullptr);
             } else {
             kernels::launch_flash_decode_split(s.q, kpool, vpool, btable, s.d_seqlen, s.attn,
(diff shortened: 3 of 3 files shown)


reyanthony062001-ops:
Update: while validating an unrelated branch against a bitwise parity harness, I found that this PR fixes a **second, more severe defect** on the same code path — worth knowing when you weigh it.

**On current main, the sparse path's O-projection consumes stale scratch instead of the attention output.** Both sparse call sites pass `aq81` as `out_q8` to `launch_fa_combine_hd256` and (under `attn_gate_q8`, Qwythos' configuration) skip the standalone attn quantize, relying on the combine's fused Q8_1 emit. But the ungated `fa_combine_kernel`'s emit is guarded by `ELEMS == 1` (flash_decode_split.cu:259), and the hd256 instantiation has `ELEMS == 2` — **the emit is compile-time dead**. Only the gated combine (which this PR wires up) emits at hd256. So at every position ≥ 8192, `wo` is applied to whatever `aq81` last held — in `forward_token`'s launch order that is `Q8_1(xn)` of the same layer, i.e. **the attention layer's input is substituted for its output** on all 8 full-attention layers.

Empirical proof (RTX 5090, teacher-forced):
- Simulating the predicted stale content (quantizing `xn` instead of `attn` into the emit slot, on a build where the emit otherwise works) reproduces current main's full-vocab logits **md5-identically** at the first sparse position — only the stale-scratch explanation predicts that.
- With this PR's gated combine, sparse output becomes **bitwise equal to dense attention** when the window is set to cover all blocks (the mathematical identity case), and passes a sequential-vs-chunked-schedule bitwise parity harness at the 8192 boundary. On main, neither holds.

So this PR as-is fixes both the dropped output gate and the dead emit. One residual note for a follow-up (not this PR): the ungated combine still silently ignores `out_q8` at hd256 — if a non-gated hd256 model ever takes the sparse path via `emit_attn_q8`, the same stale-scratch corruption returns; extending the ungated emit to `ELEMS == 2` or asserting would close that hole.

This also means every Qwythos ≥ 8k number measured since #379 merged (decode, prefill pp, dashboard frontier) was produced with attention outputs replaced by layer inputs past position 8192 — one more instance of the blind spot documented in #388.

skyrocket2026:

```bash
```

**Prefill pp** (Qwythos — report your best of 4k/32k/64k/128k from the `prefill pp` line in bench output):
```bash
```

skyrocket2026:

skyrocket2026:
Stale: no GitHub activity for 2+ days; needs-benchmark (never evaluated).

skyrocket2026:
Independently verified on an RTX 5090. This PR applies cleanly on top of the now-merged #389 (`c3a9211`), and both defects reproduce exactly as described.

**Static confirmation on current main.** `fa_combine_kernel` computes `ELEMS = HEAD_DIM / (32 * DG)` and the hd256 launcher instantiates it with `FA_COMBINE_DG = 4` (`flash_decode_split.cu:661`), so `ELEMS == 2` and the fused Q8 emit at `flash_decode_split.cu:259` (`if (out_q8 != nullptr && ELEMS == 1)`) is compile-time dead. `qwen35.cpp:788-789` passes `s.aq81` into it regardless, so the O-projection MMVQ consumes whatever that scratch last held. At hd128 `ELEMS == 1`, so the emit is live — the same "masked by coinciding constants at hd128" shape as #388.

**Probe** — Qwythos-9B-Claude-Mythos-5-1M Q4_K_M, 12k-token teacher-forced feed, `SPARKINFER_KV_INT8=1`, sparse ON (the default at seqlen >= `sparse_min_ctx` = 8192) vs `SPARKINFER_SPARSE_KV=0`, same build:

| build | sparse | PPL | ARGMATCH |
|---|---|---:|---:|
| main `c3a9211` | ON (default) | **81.26** | 0.643 |
| main `c3a9211` | OFF | 1.168 | 0.960 |
| **+ this PR** | **ON (default)** | **1.168** | **0.960** |
| + this PR | OFF | 1.168 | 0.960 |

Positionally, sparse-ON vs sparse-OFF within each build:

| | positions < 8192 | positions >= 8192 (sparse engaged) |
|---|---|---|
| main `c3a9211` | 1/8192 mismatch, mean \|Δlp\| 0.0015 | **3807/3807 (100.0%)**, mean \|Δlp\| **13.37** |
| + this PR | 0/8192, mean \|Δlp\| 0.0000 | **0/3807 (0.0%)**, mean \|Δlp\| 0.0005 |

Bit-clean below 8192, 100% divergent above it, and this PR closes it completely — with the fix, the sparse path becomes numerically indistinguishable from the dense path on this feed (the windowing itself is a no-op here, since the tiled corpus is well within the local window; a feed with genuine long-range dependencies would show the intended modeling difference).

**70x PPL blowup on Qwythos's default path past 8k.** #389 alone does not address this — that only cleaned the 4961–8191 dense band.

Not benchmarked. Expect a small cost in the same direction as #389 (~1-2%), since main's sparse path is currently skipping both the output gate and the Q8 emit — that work is real, not overhead. Happy to post a clock-pinned A/B if it's wanted before merge, though the correctness case stands on its own at 100% wrong output.

Consequences for the record, consistent with #388: every Qwythos >= 8k measurement taken since #379 merged — decode tg, prefill pp, dashboard frontier, and the `reference.lock` pp pins at 32k/64k/128k — was measured on gate-free attention reading stale scratch, and needs re-measurement once this lands.

Note this PR was closed by the stale bot on 2026-07-16 after a `needs-benchmark` flag. A correctness re-land has no speedup to claim by construction, so it cannot clear that gate — the same dynamic that closed #389, which was merged manually today. Worth considering a `hold`-style exemption for correctness PRs, or the `>= 8192` scored probe @reyanthony062001-ops proposes in #388, which would have caught all four defects on this path.

Thanks @reyanthony062001-ops — for finding this one via bitwise parity on an unrelated branch, and for the write-up, which made it verifiable in a single read.

PR #413, opened 2026-07-14
fix(runtime): bound KV allocate on cumulative block count

### LINKED ISSUE #110: KV cache: re-allocating an existing sequence over-allocates blocks and overruns the device block-table row
## Summary

`KVCacheManager::allocate()` ([runtime/src/kv_cache.cpp](https://github.com/gittensor-ai-lab/sparkinfer/blob/main/runtime/src/kv_cache.cpp#L64-L80)) corrupts the device block table when it is called more than once for the same `seq_id` (the natural way to **grow** a paged sequence as it generates past its initial reservation). The bounds check validates only the *incremental* request, not the *cumulative* block count, so a re-allocation can leak blocks and write past the fixed-size device block-table row.

## Details

```cpp
bool KVCacheManager::allocate(uint64_t seq_id, int num_tokens) {
    const int need = (num_tokens + impl_->cfg.block_size - 1) / impl_->cfg.block_size;
    if ((int)impl_->free_list.size() < need || impl_->free_slots.empty()) return false;
    if (need > kMaxBlocksPerSeq) return false;            // (1) checks `need`, not the total

    auto& blocks = impl_->seq_blocks[seq_id];             // (2) existing blocks on re-alloc
    for (int i = 0; i < need; i++) { blocks.push_back(impl_->free_list.back()); impl_->free_list.pop_back(); }
    ...
    cu(cudaMemcpy(impl_->d_block_tables + (size_t)slot * kMaxBlocksPerSeq, blocks.data(),
                  blocks.size() * sizeof(int), cudaMemcpyHostToDevice), "copy block table");  // (3) copies blocks.size()
    return true;
}
```

`need` is derived from the **total** `num_tokens`, but `blocks` already holds the previously-allocated blocks for that `seq_id` (2). Each call therefore *appends* another `need` blocks instead of topping up to `need` total:

- **Block leak / double count.** `allocate(seq, N)` then `allocate(seq, N)` again ends up holding `2*ceil(N/block_size)` blocks for one sequence; the extras are never returned to the free list (only `free()` reclaims, and it reclaims the inflated set).
- **Out-of-bounds device write.** The guard at (1) only rejects when the *incremental* `need > kMaxBlocksPerSeq`. After re-allocation, `blocks.size()` can exceed `kMaxBlocksPerSeq` while every individual call passed the guard (e.g. two calls of `need = 3000` each, `kMaxBlocksPerSeq = 4096`). The `cudaMemcpy` at (3) then copies `blocks.size()` ints into a row whose device capacity is only `kMaxBlocksPerSeq` ints — overrunning into the next sequence's block-table row (memory corruption).

Net effect: the paged KV cache **cannot grow an existing sequence** — the one operation a paged allocator exists to support. Today the only in-tree caller (`Qwen35Model::generate` / `bench_decode` / the score example) allocates once and frees once, so the bug is latent, but it is a memory-safety landmine for any multi-step or growing-context use of the cache.

## Expected

`allocate(seq_id, num_tokens)` should ensure the sequence is backed by enough blocks for `num_tokens` **total** — topping up only the additional blocks needed, preserving already-cached KV, and bounding the *cumulative* block count against `kMaxBlocksPerSeq`.

## Suggested fix

Make the allocation grow-aware: compute the total blocks wanted, subtract what the sequence already holds, allocate only the delta, and bound the total. This is byte-identical for the current single-allocate callers (`have == 0` ⇒ `need == want`).

I'll open a PR with the fix.

## Environment

- File: `runtime/src/kv_cache.cpp`
- Found by code review against `main` (`856cd1a`).

pro3958:
## Summary

`KVCacheManager::allocate()` derived `need` from the total `num_tokens` but appended those blocks to whatever the sequence already held, so calling it again for the same `seq_id` — the normal way to grow a paged sequence past its initial reservation — added a full fresh `need` each time instead of topping up to that total. The guard checked only the incremental `need` against `kMaxBlocksPerSeq`, never the cumulative count, so after a re-allocate `blocks.size()` could exceed the row width and the `cudaMemcpy` wrote past the fixed device block-table row into the next sequence's row.

This sizes and bounds off the total instead. `want` is the total blocks for `num_tokens` and is checked against `kMaxBlocksPerSeq`; only the delta (`want - have`) is charged to the free list, keyed off whether the `seq_id` already owns a slot. Growing charges the delta, a re-allocate of equal size is a no-op, and a grow that would push the total past the row width is rejected before any block or slot state is mutated.

## Verification

Added `runtime/tests/kv_cache_gpu_test.cpp`, wired into the existing `ctest` suite. It allocates a sequence, re-allocates it at the same token count and asserts no blocks are consumed, grows it and asserts only the delta is charged, and asserts a grow whose total would exceed `max_blocks_per_seq` is rejected. The re-allocate leak assertion fails on the pre-fix code. It skips cleanly when no CUDA device is present.

## Notes

The existing callers (qwen35 generate/bench and the example) allocate `max_seq` once and then `free`, so they take the unchanged new-sequence path; only the grow-by-re-allocate case is affected.

Fixes #110

<details><summary>Notes I made while reviewing this</summary>

- **minor** `runtime/src/kv_cache.cpp:94` — `auto& blocks = impl_->seq_blocks[seq_id];` now runs before the `free_list.size() < need` check (it was after in the original). A failed grow for a brand-new seq_id therefore leaves a dangling empty vector in seq_blocks with no matching seq_slot entry.
- **minor** `runtime/src/kv_cache.cpp:96` — Grow-only semantics: when num_tokens shrinks (want < have), need is clamped to 0 and no blocks are returned to the free list. This is a deliberate and reasonable choice for a paged-grow API and matches the issue's framing, but the sequence keeps its peak footprint until free(). Worth a one-line comment if shrink-to-fit is ever expected.
- **minor** `runtime/src/kv_cache.cpp:94` — `auto& blocks = impl_->seq_blocks[seq_id];` now executes before the `free_list.size() < need` guard (the pre-fix code ran the free-list check first). A failed allocate due to block exhaustion therefore leaves an empty phantom vector in seq_blocks for that seq_id. It's benign — block_table()/free() key off seq_slot, not seq_blocks, and a later retry sees have=0 and works — but it is a slight behavior change on the failure path.

</details>

No GPU eval is requested: this is a correctness fix in the paged KV-cache allocator and makes no performance claim, so per CONTRIBUTING.md the proof-of-speedup section is omitted rather than left with an unticked box.

### DIFF
--- a/runtime/src/kv_cache.cpp
+++ b/runtime/src/kv_cache.cpp
@@ -82,16 +82,26 @@ KVCacheManager::~KVCacheManager() {
 }
 
 bool KVCacheManager::allocate(uint64_t seq_id, int num_tokens) {
-    const int need = (num_tokens + impl_->cfg.block_size - 1) / impl_->cfg.block_size;
-    if ((int)impl_->free_list.size() < need || impl_->free_slots.empty()) return false;
-    if (need > kMaxBlocksPerSeq) return false;
+    // `want` is the TOTAL blocks this sequence needs for num_tokens. A sequence that
+    // already holds blocks (a re-allocate to grow a paged sequence) only needs the delta
+    // topped up. Sizing and bounding off the cumulative count — not the per-call request —
+    // is what keeps a re-allocate from double-allocating blocks and, worse, from pushing
+    // blocks.size() past kMaxBlocksPerSeq and overrunning the fixed device block-table row
+    // into the next sequence's row in the cudaMemcpy below.
+    const int want = (num_tokens + impl_->cfg.block_size - 1) / impl_->cfg.block_size;
+    if (want > kMaxBlocksPerSeq) return false;            // bound the TOTAL, not the delta
 
     auto& blocks = impl_->seq_blocks[seq_id];
+    const int have = (int)blocks.size();
+    const int need = (want > have) ? (want - have) : 0;   // only the additional blocks
+    const bool new_seq = impl_->seq_slot.find(seq_id) == impl_->seq_slot.end();
+    if ((int)impl_->free_list.size() < need) return false;
+    if (new_seq && impl_->free_slots.empty()) return false;  // a re-allocate already owns its slot
+
     for (int i = 0; i < need; i++) { blocks.push_back(impl_->free_list.back()); impl_->free_list.pop_back(); }
 
     int slot;
-    auto it = impl_->seq_slot.find(seq_id);
-    if (it != impl_->seq_slot.end()) slot = it->second;
+    if (!new_seq) slot = impl_->seq_slot[seq_id];
     else { slot = impl_->free_slots.back(); impl_->free_slots.pop_back(); impl_->seq_slot[seq_id] = slot; }
 
     cu(cudaMemcpy(impl_->d_block_tables + (size_t)slot * kMaxBlocksPerSeq, blocks.data(),
--- a/runtime/tests/CMakeLists.txt
+++ b/runtime/tests/CMakeLists.txt
@@ -25,3 +25,7 @@ add_test(NAME decode_runner_gpu_test COMMAND decode_runner_gpu_test)
 add_executable(qwen35_gpu_test qwen35_gpu_test.cpp)
 target_link_libraries(qwen35_gpu_test PRIVATE sparkinfer_runtime CUDA::cudart)
 add_test(NAME qwen35_gpu_test COMMAND qwen35_gpu_test)
+
+add_executable(kv_cache_gpu_test kv_cache_gpu_test.cpp)
+target_link_libraries(kv_cache_gpu_test PRIVATE sparkinfer_runtime CUDA::cudart)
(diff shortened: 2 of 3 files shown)


pro3958:
Root cause is that allocate() sized and bounded off the per-call request while appending onto any blocks the seq_id already held, so a re-allocate (the normal way to grow a paged sequence) double-counted blocks and could push blocks.size() past kMaxBlocksPerSeq and overrun the device block-table row in the cudaMemcpy. The fix computes `want` from the cumulative token count, bounds the total against the cap, and charges only the delta — correct and minimal, and I confirmed existing single-allocate callers (qwen35, examples) are unaffected. The new GPU test is placed in the existing suite and its leak assertion fails on the unfixed code, so the green is real. One harmless nit: seq_blocks now default-inserts an empty vector before the free-list capacity check, which can leave a dangling empty entry on a failed grow.

PR #439, opened 2026-07-15
fix(runtime): don't let `or` swallow a legitimate eos_token_id of 0

### LINKED ISSUE #399: fix(runtime): convert_gguf.py silently overrides a legitimate eos_token_id of 0 with the hardcoded Qwen default
## Summary

`runtime/tools/convert_gguf.py` reads `tokenizer.ggml.eos_token_id` from the source GGUF's metadata using Python's `or` as a "use this if present" idiom:

```python
eos_id=int(meta("tokenizer.ggml.eos_token_id") or 151645),
```

`meta()` returns `None` when the key is absent (the intended fallback trigger), but it returns the *actual* metadata value when present — and in Python, `0 or 151645` evaluates to `151645`, because `0` is falsy. If a source checkpoint's GGUF metadata legitimately sets `eos_token_id` to token id `0`, this line silently discards it and writes the hardcoded Qwen `<|im_end|>` id (151645) to `config.txt` instead.

## Why this matters

`config.txt`'s `eos_id` is read by the runtime (`Qwen35Config::eos_id`) and is the *only* stop condition `Qwen35Model::generate()` checks:

```cpp
// runtime/src/models/qwen35.cpp
for (int i = 0; i < max_new; i++) {
    out.push_back(next);
    if (next == s.cfg.eos_id) break;
    next = forward_token(next, (int)prompt.size() + i);
    ...
}
```

If the real EOS id was silently replaced by 151645, the model can never match it (it will keep emitting whatever its actual EOS token is, which is a different id), so generation runs to `max_new` on every call instead of stopping where the model intended — a silent correctness bug in generation length/termination, not a crash, and one the accuracy gate (`bench/scripts/accuracy.sh`) would not catch since it scores token-agreement/KL over a fixed decode window rather than stop-token behavior.

## Root cause

`runtime/tools/convert_gguf.py:55`:

```python
eos_id=int(meta("tokenizer.ggml.eos_token_id") or 151645),
```

`meta(k)` (`runtime/tools/convert_gguf.py:33-36`) returns `None` only when the field is missing; when present it returns the field's actual value, including `0`. Using `or` instead of an explicit `is None` check conflates "key absent" with "key present and falsy," so any checkpoint whose GGUF legitimately encodes `eos_token_id: 0` gets `151645` written to `config.txt` instead of `0`.

The same file has one other instance of the same pattern, lower risk in practice since a real vocab size of `0` is not a plausible GGUF value:

```python
vocab=int(meta(A + "vocab_size") or 151936),   # convert_gguf.py:43
```

## Suggested fix

Use an explicit presence check instead of `or`, e.g.:

```python
_eos = meta("tokenizer.ggml.eos_token_id")
eos_id=int(_eos) if _eos is not None else 151645,
```

(and the same shape for the `vocab` line, for consistency, even though it's lower-riskin practice).

## Files likely touched by a real fix

- `runtime/tools/convert_gguf.py` — both `meta(...) or <default>` call sites- Possibly `runtime/tests/qwen35_cpu_test.cpp` / a new small Python-side test fixture  (a minimal GGUF with `eos_token_id=0`) to guard against regressing this, since neither  `bench/scripts/accuracy.sh` nor the existing CPU/GPU tests exercise the converter's  metadata parsing at all — they consume already-converted weight directories.

## Not a speedup PR

Flagging as a correctness bug per `CONTRIBUTING.md` ("non-speedup PRs are welcome — but score 0"); this is a conversion-tool fix, not a kernel/runtime/moe optimization.

## Scope note

This is a narrow, single-file bug — it depends on a source checkpoint actually encoding `eos_token_id: 0`, which is uncommon but not implausible for base/non-instruct exports or misconfigured tokenizer configs. Filing it because it's real and unclaimed, notbecause it's high-blast-radius: I went through `kernels/`, `runtime/`, and `moe/` in detail looking for a larger cross-cutting bug, but the obvious bug surface there (bounds checks on router/engine/kv_cache/gguf-parsing, scheduler starvation, attention correctness on hd256/int8-MMA, etc.) is already densely covered by existing open and closed issues — see below.

## Distinct from existing issues

Cross-checked against the full open+closed issue list (50 issues via the GitHub API, not just the paginated UI, which truncates). Nothing else references `convert_gguf.py`'s `eos_token_id`/`vocab` metadata handling; the file's only existing issue is **#81** ("`vocab-from-token_embd` fallback is dead code"), which is a different line (`cfg["vocab"] = ... if False else cfg["vocab"]`) and a different bug (dead code, not silent value override).

boskodev790:
## Summary

Fixes #399. `convert_gguf.py` used `meta(k) or default` as a "use this if present" idiom, but
`meta()` returns `None` **only** when the key is absent — when the key is present it returns the
real value, and `0` is falsy. So a checkpoint whose GGUF legitimately sets `eos_token_id` to
token id `0` had it silently replaced with the hardcoded Qwen `<|im_end|>` id (151645).

## Why it matters

`config.txt`'s `eos_id` becomes `Qwen35Config::eos_id`, which is the *only* stop condition
`Qwen35Model::generate()` checks. A wrong value means the model never matches EOS and generation
runs to `max_new` on every call. It fails silently, and the accuracy gate wouldn't catch it —
`accuracy.sh` scores token-agreement/KL over a fixed decode window, not stop-token behaviour.

## The fix

A `meta_or(k, default)` helper that falls back only on a missing key, used for `eos_id` and for
the `vocab_size` line that has the same shape (lower risk in practice, but the same latent bug).
The helper matches the file's existing nested-helper idiom (`meta`, `deq`, `write`).

## Verification

Ran locally — `python3 -m py_compile` passes, and the old vs new expression over the cases that
matter:

| case | want | old | new |
|---|--:|--:|--:|
| key absent (fallback intended) | 151645 | 151645 | 151645 |
| `eos_token_id = 0` (the bug) | 0 | **151645** | **0** |
| `eos_token_id = 151645` (normal) | 151645 | 151645 | 151645 |
| `eos_token_id = 2` | 2 | 2 | 2 |

Only the falsy-`0` case changes; every other path is byte-identical.

## Notes

Non-speedup correctness fix — earns no SN74 emission, filed because #399 is a real gap. Proof-of-
speedup section removed per CONTRIBUTING (no GPU eval needed). Touches only `runtime/tools/`.

Closes #399

### DIFF
--- a/runtime/tools/convert_gguf.py
+++ b/runtime/tools/convert_gguf.py
@@ -36,11 +36,17 @@ def main():
         if f.types[0] == GT.STRING: return bytes(f.parts[f.data[0]]).decode()
         return f.parts[f.data[-1]][0]
 
+    def meta_or(k, default):
+        # `or` would swallow a legitimate 0 (eos_token_id 0 is a valid token id);
+        # only a missing key falls back.
+        v = meta(k)
+        return default if v is None else v
+
     A = "qwen3moe."
     arch = meta("general.architecture")
     assert arch == "qwen3moe", f"expected qwen3moe arch, got {arch}"
     cfg = dict(
-        vocab=int(meta(A + "vocab_size") or 151936),
+        vocab=int(meta_or(A + "vocab_size", 151936)),
         hidden=int(meta(A + "embedding_length")),
         n_layers=int(meta(A + "block_count")),
         n_q_heads=int(meta(A + "attention.head_count")),
@@ -52,7 +58,7 @@ def main():
         moe_ffn=int(meta(A + "expert_feed_forward_length")),
         rope_theta=float(meta(A + "rope.freq_base")),
         rms_eps=float(meta(A + "attention.layer_norm_rms_epsilon")),
-        eos_id=int(meta("tokenizer.ggml.eos_token_id") or 151645),
+        eos_id=int(meta_or("tokenizer.ggml.eos_token_id", 151645)),
     )
     # vocab from token_embd if metadata missing
     ten = {t.name: t for t in r.tensors}


skyrocket2026:

To submit for review:

3. **Reopen** this PR

skyrocket2026:
closed

PR #534, opened 2026-07-18
server: add temperature/top_p/top_k/repetition_penalty sampling (was argmax-only)

ai-hpc:
Base is `a739958` — the commit actually running in production — **not current `main`**. See the "Needs a rebase" note below before merging.

## Summary
The engine had no sampling path at all: `Qwen35Model::forward_token()` unconditionally ran `launch_argmax` every decode step, and the server accepted no `temperature`/`top_p`/`top_k`/`repetition_penalty` fields. On a public endpoint that matters — greedy decoding is a known repetition-loop trigger on long generations (`sparkinfer-web` already carries client-side detection/truncation for exactly this, as a symptom-level workaround).

Adds real sampling, CPU-side, built on an existing hook rather than touching the CUDA decode path:

- `SamplingConfig` (`qwen35.h`): `temperature`/`top_p`/`top_k`/`repetition_penalty`/`seed`. `temperature<=0` (including the default when absent) means "unset" — `generate()` takes the exact original argmax-only branch, zero behavior change for every existing caller (eval harness, accuracy gate, bench scripts) unless it opts in.
- `generate()`: when sampling is active, after each `forward_token()` it also calls the already-existing `copy_logits()` (previously only used for teacher-forced perplexity/KL scoring) to pull that step's logits to host, then does temperature scale → repetition penalty (bounded 64-token window) → top-k (`std::partial_sort`) → top-p (nucleus, on the sorted survivors) → `std::discrete_distribution` sample, feeding the result in as the next decode step instead of the argmax token. Safe to override because the KV cache only encodes already-committed history by the time `forward_token()` returns.
- Threaded through `ModelEngine::complete`/`complete_streaming` and `/v1/chat/completions`, parsing the four fields OpenAI-style from the request body.

## Test plan
Verified on RTX 5090 against Qwythos-9B, isolated from the live server first (separate instance, separate port — confirmed via `/proc/<pid>/exe` that the running production process kept using its old `(deleted)` binary inode unaffected while I rebuilt in place):

- [x] Default requests (no sampling fields) byte-identical across repeated calls — greedy path provably untouched
- [x] `temperature=0.6, top_p=0.95, top_k=20, repetition_penalty=1.05` (the documented recommended settings) produces different, coherent output across repeated identical requests — confirms real sampling, not still-secretly-greedy
- [x] Both streaming and non-streaming code paths
- [x] Longer (~300 token) generation — no repetition loop, no crash
[...]

PR #535, opened 2026-07-18
server: 64k context, 16k max output, and a real 2-slot concurrency gate

ai-hpc:
Stacked on #534 (sampling) — same base commit (`a739958`, what's actually running in production), same reasoning about not blind-adapting to `main`'s in-flight multi-session refactor.

## Summary
Three related capacity changes, done together because the third makes the first two *safe* under concurrent load rather than just safe for one request at a time:

- `kMaxOutputTokens` 4096 → 16384, `kMaxInputContext` (reporting only) 32768 → 65536. Actual ceiling is the server's `--ctx` flag (deployed at 65536, was 32768).
- Wired up `RequestSlots`/`SlotLease` into `POST /v1/chat/completions` — this header already existed but was **never called from anywhere**, so `SPARKINFER_MAX_CONCURRENT` was dead config and there was no real concurrency limit beyond whatever httplib's own thread pool allowed.
- The KV cache pool (`model_engine.cpp`) is now sized as `SPARKINFER_MAX_CONCURRENT × max_seq` instead of just `max_seq`.

## Why the pool sizing needed investigating first
`Qwen35Model::generate()` reserves a **full `max_seq` worth of KV blocks up front for every request** (`kv->allocate(seq_id, cfg.max_seq)`), not sized to actual prompt/response length, drawing from one shared PagedAttention-style block pool with no per-sequence eviction beyond it. A pool sized for exactly one sequence — what existed before this — can only ever serve one request at a time; every concurrent request beyond the first fails to allocate once the first has claimed the pool, regardless of how short that 2nd request's conversation actually is. Sparse-KV's sliding window (in the startup log) only skips attention *compute* over old blocks during decode — it doesn't free them, so it doesn't help here either.

A literal 3× pool at 64k context would be ~24GB, leaving only ~3GB free alongside ~5GB of model weights on this 32GB card — too tight for CUDA overhead/fragmentation on a public endpoint. Went with **2 slots**: pool comes out to 16.0 GiB (matches the startup log exactly), and real measured process VRAM (`nvidia-smi --query-compute-apps`, not the `pool_budget` log line, which reports the allocation request rather than final resident bytes) came in at 15336 MiB — a real ~17GB of headroom out of 32.6GB total.

## Test plan
Verified on RTX 5090 against Qwythos-9B, isolated from the live server first (separate instance, separate port — confirmed via `/proc/<pid>/exe` that production was untouched while rebuilding):
- [x] Startup log confirms exact expected sizing: `blocks=8208 (2 slots x 4104) pool_budget=16.0 GiB`
- [x] `/v1/info` reports `max_context=65536, max_output_tokens=16384`
- [x] 3 genuinely concurrent requests (fired in parallel) all succeeded — request 3 took ~5.1s vs ~1.7s/~3.4s for the first two, confirming it actually **queued** for a slot rather than running unconstrained (similar latencies) or failing (non-200)
- [x] A request fired immediately after that batch completed in 0.08s — slots release cleanly, no leak
- [x] Sampling (from #534) still works correctly on top of this
- [x] Deployed to the actual production process (`/etc/init.d/sparkinfer-server restart`) and re-verified through the live `api.sparkinfer.com` endpoint — `/v1/info`, a real generation, and production VRAM (15336 MiB, matching the isolated test exactly)

## Needs a rebase before merging
Same situation as #534 — branched from `a739958` (what's deployed), predates the in-flight multi-session/`seq_id` refactor on `main` touching the same `generate()` function.

### DIFF
--- a/server/include/request_slots.hpp
+++ b/server/include/request_slots.hpp
@@ -1,0 +1,96 @@
+#pragma once
+
+#include <chrono>
+#include <condition_variable>
+#include <cstdlib>
+#include <mutex>
+
+namespace sparkinfer_server {
+
+class RequestSlots {
+public:
+    explicit RequestSlots(int max_slots) : max_slots_(max_slots), in_use_(0) {}
+
+    // Default of 2 matches the KV cache pool sizing in model_engine.cpp — both
+    // read the same env var so they can never drift out of sync with each
+    // other. See the comment there for why the pool has to scale with this.
+    static int from_env(int default_slots = 2) {
+        const char* e = std::getenv("SPARKINFER_MAX_CONCURRENT");
+        if (!e || !*e) return default_slots;
+        const int v = std::atoi(e);
+        return v > 0 ? v : default_slots;
+    }
+
+    bool try_acquire() {
+        std::lock_guard<std::mutex> lock(mu_);
+        if (in_use_ >= max_slots_) return false;
+        ++in_use_;
+        return true;
+    }
+
+    // Blocks until a slot frees up or timeout_ms elapses. Returns false on
+    // timeout — caller should respond with a "busy" error rather than
+    // proceed, not silently run over capacity.
+    bool acquire(int timeout_ms) {
+        std::unique_lock<std::mutex> lock(mu_);
+        const bool got_slot =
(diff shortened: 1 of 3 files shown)


skyrocket2026:

To submit for review:

3. **Reopen** this PR

skyrocket2026:
closed

PR #536, opened 2026-07-18
server: enforce a real think/answer token budget split

ai-hpc:
## Summary
- `thinking_budget_tokens` was accepted in the request body but never enforced — a request could burn its entire `max_tokens` budget on reasoning and leave zero room for the answer (the "Reasoning finished but the answer was cut off" symptom reported from production).
- `SamplingConfig` gains `reasoning_budget`/`think_open_id`/`think_close_id`. `Qwen35Model::generate()` tracks tokens emitted between `<think>` and `</think>` purely by token id (no text decoding in the decode loop) and force-injects `</think>` once the budget is hit, guaranteeing the remaining `max_tokens` is available for the answer.
- `<think>`/`</think>` ids are looked up once at startup via a new `ChatTokenizer::token_id()` (wraps `tokenizers::Tokenizer::TokenToId`) — dynamic, not hardcoded, so it stays correct if the tokenizer changes.
- Server parses `thinking_budget_tokens` from the request and wires it into `sampling` only when thinking is enabled and both ids resolved; clamps it below `max_tokens` so there's always at least one token of headroom.
- Stacked on #535 since it reuses the `sampling` plumbing added there.

## Test plan
- [x] Built clean on the RTX 5090 box (`sparkinfer_server`)
- [x] Isolated test instance (separate port, 1 slot, small ctx) — non-streaming: baseline (no budget) reproduces the cut-off bug (reasoning consumes full 500-token budget, empty answer); with `thinking_budget_tokens=200` reasoning is capped and a real answer is produced, generation finishes naturally before `max_tokens`
- [x] Streaming mode: reasoning force-closes correctly, no `<think>`/`</think>` marker leakage into the `content` SSE stream, `usage` stats (`ttft_ms`, `decode_tps`) intact
- [x] Requests without `thinking_budget_tokens`, and with thinking disabled, are unaffected (enforcement is a no-op)
- [x] Deployed to production (`sparkinfer_server` on the live RTX 5090 box), verified `/health` + `/v1/info`, VRAM footprint unchanged (15336 MiB)
- [x] Companion client-side fix in `sparkinfer-web` unifies the two divergent `thinking_budget_tokens` formulas so the SettingsModal label, the value sent, and the value enforced here all agree

### DIFF
--- a/runtime/include/sparkinfer/models/qwen35.h
+++ b/runtime/include/sparkinfer/models/qwen35.h
@@ -77,6 +77,18 @@ struct SamplingConfig {
     int top_k = 0;                   // 0 = no top-k truncation (full vocab)
     float repetition_penalty = 1.0f; // 1 = no penalty
     uint64_t seed = 0;               // 0 = seed from std::random_device
+
+    // Optional think/answer token budget split. Disabled unless reasoning_budget>0
+    // and both ids are valid (>=0). When enabled, generate() counts tokens emitted
+    // between think_open_id and think_close_id; once that count reaches
+    // reasoning_budget without think_close_id having appeared naturally, it
+    // force-injects think_close_id so the remaining max_new_tokens budget is
+    // guaranteed to be available for the answer instead of being consumed by
+    // reasoning. No effect if the model never emits think_open_id (e.g. thinking
+    // disabled for the request).
+    int reasoning_budget = 0;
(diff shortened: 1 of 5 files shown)


skyrocket2026:

```bash
```

```bash
```

skyrocket2026:

PR #570, opened 2026-07-21
server: emit GPU ttft/generation/decode_tps in stream usage

ai-hpc:
## Summary
- Measure GPU decode on the continuous-batch worker around `forward_token` (excluding SSE/network backpressure).
- Expose `ttft_ms`, `generation_ms`, and `decode_tps` on the final streaming `usage` chunk so UIs (e.g. sparkinfer-web) can show accurate tok/s.
- Additive OpenAI-compatible fields only; existing `prompt_tokens` / `completion_tokens` / `total_tokens` unchanged.

## Pipeline impact
- **Safe for existing clients** — extra JSON fields are ignored by OpenAI SDKs and current bench scripts (they compute their own client-side decode_tps).
- **No change** to scheduling, KV, sampling, eval bots, or native bench binaries.
- **Negligible overhead** — one `steady_clock` sample around each decode forward.
- Left `kMaxOutputTokens` at the upstream default (`4096`); deployment max-out remains an init/CLI concern.

## Test plan
- [x] Rebuild `sparkinfer_server` and stream a short chat; confirm usage includes `decode_tps` / `ttft_ms` / `generation_ms`
- [ ] Smoke existing API benches still parse usage (`bench_api_vs_native.py`)
- [ ] Confirm non-stream `/v1/chat/completions` still returns standard usage
- [ ] sparkinfer-web chat shows non-zero GPU tok/s

### DIFF
--- a/runtime/include/sparkinfer/inference_engine.h
+++ b/runtime/include/sparkinfer/inference_engine.h
@@ -32,6 +32,10 @@ public:
     struct Result {
         std::vector<int> tokens;
         std::string error;
+        // GPU-side timings (exclude SSE/on_token backpressure).
+        double ttft_ms = -1.0;
+        double generation_ms = -1.0;
+        double decode_tps = -1.0;
     };
 
     ContinuousBatchEngine(Qwen35Model* model, KVCacheManager* kv,
--- a/runtime/src/inference_engine.cpp
+++ b/runtime/src/inference_engine.cpp
@@ -43,6 +43,14 @@ struct ContinuousBatchEngine::Job {
     std::string error;
     std::function<void(int)> on_token;
     bool done = false;
+    std::chrono::steady_clock::time_point t_submit{};
+    std::chrono::steady_clock::time_point t_first{};
+    bool saw_first_tok = false;
+    double decode_gpu_ms = 0.0;
+    int decode_forwards = 0;
+    double ttft_ms = -1.0;
+    double generation_ms = -1.0;
+    double decode_tps = -1.0;
 };
 
 ContinuousBatchEngine::ContinuousBatchEngine(Qwen35Model* model, KVCacheManager* kv,
@@ -115,6 +123,7 @@ uint64_t ContinuousBatchEngine::submit_locked(Job job, const std::function<void(
     job.seq_id = seq_id;
     job.on_token = on_token;
     job.prefill_pos = job.req.prefill_start;
+    job.t_submit = std::chrono::steady_clock::now();
     auto ptr = std::make_unique<Job>(std::move(job));
     const uint64_t rid = ptr->request_id;
     jobs_[rid] = std::move(ptr);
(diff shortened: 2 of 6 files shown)


skyrocket2026 (changes requested):
## Review

Nice additive change — streaming `usage` with `ttft_ms` / `generation_ms` / `decode_tps` is the right shape for the chat UI, and measuring around `forward_token` (before `on_token`) correctly excludes SSE backpressure.

### Blocker: `last_timing_` is racy under continuous batching

`ModelEngine::complete_streaming` writes `last_timing_` after the job finishes, then `sparkinfer_server` reads `engine.last_timing()` on a later line. httplib serves concurrent requests and the batch engine interleaves them, so this can happen:

1. Request A finishes → sets `last_timing_`
2. Request B finishes → overwrites `last_timing_`
3. Request A’s handler reads `last_timing()` → **B’s timings**

Worse: if the stolen values are `>= 0`, the per-request wall-clock fallback in `sparkinfer_server.cpp` is skipped, so the UI can show another session’s tok/s.

`last_timing()` also returns an unlocked reference (unlike `last_error()` which takes `mu_`).

**Suggested fix:** return timings with the completion result (extend the return of `complete_streaming`, or an out-param / pair), and use that local value in the stream handler. Avoid a process-global “last timing” side channel.

### Non-blocking notes
- Non-stream `/v1/chat/completions` still emits standard usage only (fine if intentional).
- Single-token completions fall back to wall-clock decode_tps (`decode_forwards == 0`); OK.
- Prefill seed token correctly excluded from `decode_forwards` (decode tok/s after first token).

Please fix the timing handoff, then this looks good to merge.

skyrocket2026:
Maintainer `hold`: non-speed runtime/server change — keep open; do not auto-close for missing RTX 5090 greenlight.

skyrocket2026:

To submit for review:

3. **Reopen** this PR

skyrocket2026:
closed

skyrocket2026:
Reopened + `hold`. This is a non-speed runtime/server PR — do **not** auto-close for missing RTX 5090 greenlight. Eval bot incorrectly closed it despite `hold`; fixing that path.

skyrocket2026:
Superseded by `3c1758a` on `main`, which ships the same TTFT/generation_ms/decode_tps idea from this PR (plus request cancellation, overload/429, timeouts, graceful shutdown, /metrics, live /v1/info /v1/capacity) as part of closing out the server's production-readiness gaps.

This PR can't be merged as-is: it predates `#740` (concurrent-error-bleed fix) and reintroduces the shared-mutable-`last_error_`/`last_timing_` race that fix removed — `ModelEngine` now returns `CompletionResult` by value specifically so concurrent requests can't observe or clobber each other's state. It's also currently showing `CONFLICTING` against `main`.

Closing as superseded — thanks for the original idea, it's in `main` now (rebuilt on the current architecture).

PR #595, opened 2026-07-23
fix(qwen36): restore MoE FP8 prefill default (+30% pp @16k)

fansilas:
## Summary

Restore the `SPARKINFER_PREFILL_MOE_FP8` default (Qwen3.6 MoE attn/GDN projections on the fp8 e4m3 tensor cores, merged in #582, turned off in #587). The #586/#587 `prefill_check` failures that motivated the off-switch trace to the **opt-in `MOE_GPU` tilemap path's mask dequant silently no-opping** (the fast mask path declines `cols % 1024 != 0` -- every down matrix at `mffn=512` -- and Q5_K, and the moe_gpu group loop ignored the `false` returns, leaving stale `W_i8` in the grouped GEMMs), **not to the projection dtype**. #593 fixes that opt-in path; #586's FP8 measurements were taken on the broken 416ffe3 default (its own no-FP8 baseline column already shows main at 0.31-0.56 TOP1), which conflated the two effects. On the default host-tilemap path, FP8 fidelity sits inside the bf16 baseline's own run spread and clears the H3 gate (#588) bars with wide margin -- isolation table below.

`SPARKINFER_PREFILL_MOE_FP8=0` restores bf16 exactly as before.

## Proof of speedup

- [x] Tested on **RTX 5090** (`sm_120`)

**Decode tok/s** (end-to-end, from `bench/scripts/bench.sh` -- decode-neutral, prefill-only diff):

| | decode tok/s |
|---|--:|
| before (main) | 454.06 |
| after (this PR) | 454.16 |

**Prefill pp tok/s** (Qwen3.6-35B-A3B, best context `--ctx 16384`):

| | prefill pp tok/s |
|---|--:|
| before prefill (main) | 15995.70 |
| after prefill (this PR) | 20779.20 |

```text
# Qwen3.6-35B-A3B (Q4_K_M), RTX 5090, bench/scripts/bench.sh --tokens 64 --ctx 16384
# main @f29e514:
decode tg    : 454.06 tok/s  (2.2 ms/token, n=64, ctx=16384, bs=1)
prefill pp   : 15995.70 tok/s  (ctx=16384, sequential KV fill)
# this PR:
decode tg    : 454.16 tok/s  (2.2 ms/token, n=64, ctx=16384, bs=1)
prefill pp   : 20779.20 tok/s  (ctx=16384, sequential KV fill)   (+29.9%)
```

[...]

Issue #586, opened 2026-07-22
MoE FP8 prefill projections break prefill_check; accuracy gate never exercises prefill


inference2026:
## Title
MoE FP8 prefill projections break `prefill_check`; official accuracy gate never exercises prefill

## Summary
Persistent e4m3 projection weights for Qwen3.6 MoE batched prefill (`SPARKINFER_PREFILL_MOE_FP8_PROJ`) give ~**+20% pp @512** on RTX 5090 when applied to full-attn Q/K/V/O + GDN, but **fail internal prefill fidelity** (`qwen3_gguf_prefill_check`). The official eval accuracy gate (`accuracy.sh` / `qwen3_gguf_score`) only runs `forward_token`, so it **does not catch** this class of bug.

Separately, **current `main` (#583) already fails** `prefill_check` at short/medium prefixes (batched vs token-loop), so FP8 makes an existing gap worse.

## Environment
- GPU: RTX 5090 (sm_120)
- Commit base: `416ffe3` (#583 GPU MoE tilemap)
- Model: `Qwen3.6-35B-A3B-UD-Q4_K_M.gguf`
- `SPARKINFER_KV_INT8=1` (same as scored benches / `prefill_check` default)

## Measurements (`qwen3_gguf_prefill_check`, cont=16)

| prefix | main TOP1 | FP8 Q/K/V/O (+GDN) | FP8 full-attn only | FP8 **wo only** |
|--------|-----------|--------------------|--------------------|-----------------|
| 64 | 0.3125 | 0.1875 | 0.2500 | **0.3125** (match) |
| 256 | 0.3750 | 0.1875 | 0.2500 | 0.3125 |
| 512 | **0.5625** | **0.4375** | **0.4375** | 0.5000 |

Speed @512 vs main (~6590 pp):
- Full FP8 (Q/K/V/O + GDN): ~**7900** pp (**+20%**)
- Q/K/V/O only: ~6940 pp (+5%)
- **wo only**: ~6680 pp (**+1.4%**, below 2% significance gate)

Official short gate vs llama.cpp (token path): **top1=0.910 / KL=0.075** for main and all FP8 variants — identical, because the gate never enters `prefill_batched`.

## Root cause (working theory)
1. **Q/K/V e4m3 → int8 KV**: full-attn layers project Q/K/V then write int8 KV. FP8 drift in Q/K/V poisons KV, so decode after batched prefill diverges from the token-loop reference (`prefill_check`).
[...]

PR #601, opened 2026-07-23
qwen36 MoE prefill: fix MOE_GPU device-tilemap correctness (#586) + int8 projection default

James-CUDA:
## Summary

Two independent Qwen3.6 MoE-prefill changes:

**1. `fix(qwen36)`: restore MoE device-tilemap (`MOE_GPU`) correctness.** `SPARKINFER_PREFILL_MOE_GPU=1` (device tilemap, skips the per-layer D2H counts sync) is **silently broken on `main`**: the down projection's masked `Q→int8` dequant runs with `cols=moe_ffn=512`, which the vectorized fast-mask kernel declines (cols not a multiple of 1024), and `launch_gguf_dequant_rows_i8_mask` **ignored that `false` return** — so the int8 down weights were never written and the down GEMM ran on stale data (the #586 corruption). Fix: fall back to a correct masked slow kernel (single-pass, register-resident, same dequant/amax/round as the host `deq_rows_i8_kernel`; dead experts exit via the counts mask). The default host-tilemap path is untouched and `MOE_GPU` stays **default OFF** (the device path isn't a prefill speedup — it dequants all experts per group vs the host path's live-only coalesced runs — so this fixes the opt-in flag, not perf).

| `prefill_check … 512 16`, `MOE_GPU=1` | TOP1 | KL |
|---|--:|--:|
| before (main) | 0.375 | 0.349 |
| after | 0.9375 | 0.010 |

**2. `perf(qwen36)`: int8 tensor-core projections default-on for the MoE-hybrid prefill.** Flips the `use_i8` default for the Qwen3.6-256E MoE path (attn/GDN/router-feeding projections), scoped to Qwen3.6 by the existing `n_experts==256` guard (Qwen3-30B is 128E, untouched). Scored accuracy is provably unchanged (`qwen3_gguf_score` teacher-forces via the bf16 decode `forward_token` path, which `SPARKINFER_PREFILL_I8` never reaches — verified byte-identical PPL/ARGMATCH on/off); prefill fidelity holds (TOP1 0.875 @512 / 0.94 @4k, bar 0.80).

> ⚠️ **Honesty note on speed:** this was measured **+10.6% pp @512 / +18% @4k vs the bf16 baseline**, but `main` has since defaulted MoE projections to **fp8** (#595), which captures the same win a different way. **Against current `main`, the int8 flip is ~+0.6% @512** (fp8 and int8 are within noise). Kept in this PR per maintainer request; the standalone value here is the correctness fix above. `SPARKINFER_PREFILL_I8=0` restores bf16.

## Proof of speedup

- [ ] Tested on **RTX 5090** (`sm_120`) — *int8 flip is within noise vs the current fp8 default; no end-to-end speed claim against current main. The primary contribution is the correctness fix.*

**Prefill pp tok/s** — Qwen3.6-35B-A3B-UD-Q4_K_M, RTX 5090 (int8 default vs `SPARKINFER_PREFILL_I8=0` fp8, current main):

| ctx | fp8 (current main) | int8 (this PR) |
|---|--:|--:|
| 512 | 5726 | 5745 |
| 4096 | 16410 | 16403 |

```text
# correctness (the real point of this PR), MOE_GPU=1, prefill_check 512 16:
# main:     TOP1 6/16 0.3750   KL 0.34935
# this PR:  TOP1 15/16 0.9375  KL 0.00974
```

### DIFF
--- a/kernels/csrc/cuda/quant/dequant_gguf.cu
+++ b/kernels/csrc/cuda/quant/dequant_gguf.cu
@@ -224,6 +224,58 @@ __global__ void deq_rows_i8_twopass_kernel(const unsigned char* __restrict__ src
     }
 }
 
+// Masked single-pass Q->int8 for the MoE device-tilemap (MOE_GPU) path. Correct for ANY cols
+// (the vectorized fast-mask kernel declines cols not a multiple of 1024 — e.g. the down proj
+// cols=mffn=512 — and its false return was ignored, so down weights silently stayed stale;
+// #586 corruption). grid = n_in*rows_per_expert; one block per (local expert, row); dead
+// experts (counts[e_base+le]<=0) exit. Same dequant/amax/round as deq_rows_i8_kernel.
+template <int QT>
+__global__ void deq_rows_i8_mask_slow_kernel(
+        const unsigned char* __restrict__ src0, signed char* __restrict__ q0, float* __restrict__ scale0,
+        const int* __restrict__ counts, int e_base, int n_in, int rows_per_expert, int cols,
+        size_t expert_bytes) {
+    constexpr int BS = (QT == 12) ? 144 : (QT == 13) ? 176 : 210;
+    const int flat = blockIdx.x;
+    const int le = flat / rows_per_expert;
+    const int row_in = flat - le * rows_per_expert;
+    if (le >= n_in || counts[e_base + le] <= 0) return;
+    const int t = threadIdx.x;
+    const int nsb = cols >> 8;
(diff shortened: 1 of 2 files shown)


skyrocket2026:

```bash
```

```bash
```

skyrocket2026:
## Closed: draft open for 4+ days

This PR has remained a **draft** for **4 days** (threshold: 4 days). Recent commits or comments do not reset this timer.

Mark ready for review or reopen when you're ready to continue.

Issue #612, opened 2026-07-24
CI: build-attested-binaries always fails on fork PRs (attestation, not the build)


RealDiligent:
## Summary

`build-attested-binaries` reports **failure on every fork pull request**, including ones whose
build is completely clean. The build, the packaging and the bundle all succeed — the job then dies
on its last step, `actions/attest@v4`:

```
##[error]Error: missing "id-token" permission. Please add "permissions: id-token: write" to your workflow.
```

The permission *is* already declared (`.github/workflows/build-attested-binaries.yml`, both jobs):

```yaml
    permissions:
      contents: read
      id-token: write
      attestations: write
```

GitHub does not grant `id-token` to a workflow triggered by `pull_request` **from a fork**,
regardless of what the `permissions:` block asks for — the token is read-only for fork events. So
the step cannot succeed there, and no contributor can make it green.

## Evidence

Same-repo PRs pass; fork PRs fail, independent of content:

| PR | author | fork | build-linux / build-windows | merged |
|---|---|---|---|---|
| #608 | skyrocket2026 | no | success / success | yes |
| #595 | fansilas | yes | failure / failure | yes |
| #597 | Paral1995 | yes | failure / failure | yes |
| #598 | inference2026 | yes | failure / failure | yes |
| #577 | James-CUDA | yes | failure / failure | yes |
| #582 | fansilas | yes | failure / failure | yes |

Every merged fork PR in that list merged with both checks red, which is the giveaway that the
signal carries no information today.

For a concrete clean run — #610, log of `build-linux`:

```
[100%] Linking CXX executable qwen3_gguf_bench
[100%] Built target qwen3_gguf_bench
...
##[error]Error: missing "id-token" permission.
```

and `build-windows`:

```
qwen3_gguf_bench.vcxproj -> D:\a\sparkinfer\sparkinfer\build\runtime\Release\qwen3_gguf_bench.exe
...
##[error]Error: missing "id-token" permission.
```

There is exactly one `##[error]` in each log, and it is always this one.

## Why it is worth fixing

The PR build is the only automated signal that a contributor's CUDA actually compiles for
`sm_120` on both toolchains before it reaches the RTX 5090 eval queue. Right now that signal is
permanently red for the fork PRs that make up essentially all contributor traffic, so:

- a genuine compile break looks exactly like the status quo, and gets found later on the eval box
  instead of in CI;
- reviewers and the eval bot have to open the log every time to tell "did not build" from
[...]

PR #613, opened 2026-07-24
fix(attn): flash-decode combine emits no Q8_1 blocks at head_dim 256, O projection reads a stale buffer

kai392:
## Summary

`fa_combine_kernel` silently emits **no** Q8_1 attention blocks at `head_dim == 256`, while its
caller — seeing a non-null `out_q8` — skips the standalone quantize and hands that buffer straight
to the O-projection MMVQ. This is a correctness-only fix (no speed claim, so no GPU eval requested).

## Root cause

`kernels/csrc/cuda/attention/flash_decode_split.cu` — the fused emit at the tail of
`fa_combine_kernel` was guarded by `ELEMS == 1`:

```cpp
if (out_q8 != nullptr && ELEMS == 1) {
    ...
    const int blk = (seq * num_q_heads + qh) * (HEAD_DIM / 32) + dg;
```

`ELEMS = HEAD_DIM / (32 * DG)`. With `FA_COMBINE_DG == 4` that is `1` at `head_dim 128` but `2` at
`head_dim 256`, so for every hd256 instantiation the entire block is compiled out and all
`HEAD_DIM / 32 == 8` Q8_1 blocks per head are left untouched.

The contract the kernel documents — *"the O-projection MMVQ skips its standalone attn-quantize node
(bit-identical to running the quantizer on `out` afterwards)"* — is what the caller acts on. In
`runtime/src/models/qwen35.cpp` the standalone quantize is skipped precisely when the fused emit was
requested:

```cpp
if (!emit_attn_q8 && !attn_gate_q8) kernels::launch_quantize_q8_1_blocks(s.attn, s.aq81, s.qdim, st);
kernels::launch_mmvq_q4k(s.aq81, w.wo, s.ao, H, s.qdim, st);
```

So on the hd256 + non-gated path the O projection consumes a stale/uninitialized `aq81` — the
attention output for those layers is wrong, with no error surfaced anywhere.

**Reachability, stated honestly:** the two hd256 models on `main` today take other branches — the
gated hd256 combine (`fa_combine_gated_q8_kernel`) already loops over `e` and handles `ELEMS == 2`
correctly, and `emit_attn_q8` requires `!w.q_has_gate` (i.e. `!cfg.hybrid`), which no current hd256
config sets. So this is a live defect on a reachable code path rather than an observed regression:
the guard is wrong today, it is silent, and it arms the moment a non-hybrid hd256 model or a
`SPARKINFER_*` toggle routes through the non-gated hd256 combine with `out_q8` set. Both the
`launch_flash_decode_split` and the sparse-KV `launch_fa_combine_hd256` entry points reach it.

## Fix

Loop the emit over `e` instead of gating it out (2 files, +95 / -18):

- Lane `lane` owns head dims `doff + e*32`, so for a **fixed** `e` warp 0's 32 lanes hold exactly
  the 32 elements of Q8_1 block `(doff + e*32) / 32` — one whole block per `e`, `ELEMS` blocks per
  `(qh, dg)`, and the `DG` groups together tile all `HEAD_DIM / 32` blocks of the head at any `ELEMS`.
- This is the same construction `fa_combine_gated_q8_kernel` already uses in production for the
  gated hd256 combine, so the two hd256 paths now agree.

**Backward compatible:** at `ELEMS == 1`, `di / 32 == dg` (lane < 32), so the emitted block index,
`qs[lane]`, `d` and `d*sum` are unchanged — the hd128 path is bit-identical, not merely equivalent.
No launcher, header, signature or default changed; nothing outside `kernels/` is touched.

## Validation

`kernels/tests/cpu_reference_test.cpp` gains a test that models the `(dg, lane, e) -> head-dim`
mapping and asserts (a) it tiles the head's Q8_1 blocks **exactly once** and (b) each emitted block
equals a direct Q8_1 quantize of `out`. It discriminates the bug:

| emit rule | hd128 DG4 (ELEMS=1) | hd256 DG4 (ELEMS=2) |
|---|---|---|
| before (`ELEMS == 1` gate) | PASS | **FAIL** — 0 of 8 blocks emitted |
| after (loop over `e`) | PASS | PASS |

```
$ g++ -O2 -Wall -Wextra -std=c++17 kernels/tests/cpu_reference_test.cpp -o t && ./t
  [PASS] argmax 2pass tie-break             max_err=0.000e+00 (tol=0e+00)
  [PASS] combine q8 emit hd128 dg4          max_err=0.000e+00 (tol=0e+00)
  [PASS] combine q8 emit hd256 dg4          max_err=0.000e+00 (tol=0e+00)
ALL PASSED (0 failures)
```

20/20 green, warning-clean. The edited kernel was additionally type-checked host-side for the
`<128,4,4>`, `<256,4,4>` and `<256,4,16>` instantiations (no CUDA toolkit in my environment, so I am
not claiming an on-device run — `build-attested-binaries` is the real compile signal here, subject
to #612 on fork PRs).

## Risk / tradeoffs

Low. The hd128 decode path is byte-for-byte unchanged; the only behavioural difference is that the
hd256 combine now writes the Q8_1 blocks it already claimed to write. The emit adds `ELEMS - 1`
extra warp reductions in warp 0 of the combine (i.e. nothing at hd128, one extra at hd256), which is
off the measured decode critical path for the shapes that use it today. Nothing depended on the
previous no-op, since consuming an unwritten `aq81` was never valid.

## Why there is no RTX 5090 attestation

This is a correctness-only change and makes **no speed claim**, so there is deliberately no
proof-of-speedup section here and the "Tested on RTX 5090" box is **not** ticked. I do not have an
RTX 5090, and ticking that box without a real on-device run would be false attestation, which
CONTRIBUTING.md rightly treats as gaming. Per the auto-close notice on this PR, the
proof-of-speedup section was removed rather than left with an unchecked box.

The eval bot's greenlight gate (`greenlight_status` in `eval/pr_eval_bot.py`) closes any PR whose
box is unticked, regardless of whether the section was removed — so removing it, as the notice
advises, is not by itself enough to keep a non-speed PR open. A maintainer `hold` label is what
keeps this reviewable. #604 merged on exactly this basis: an outside contributor's correctness fix
with no speed claim and no 5090 access.

@ai-hpc @skyrocket2026 — please apply `hold` if this is worth reviewing. If you judge the hd256
non-gated path not worth guarding, say so and I will close it myself rather than leave it open.

### DIFF
--- a/kernels/csrc/cuda/attention/flash_decode_split.cu
+++ b/kernels/csrc/cuda/attention/flash_decode_split.cu
@@ -200,10 +200,11 @@ struct fa_block_q8_1 { __half2 ds; signed char qs[32]; };
 // original (which idled at ~2% occupancy with a serial n_splits loop). DG head-dim
 // groups -> DG x more blocks; NW warps per block each fold a 1/NW stripe of the
 // splits, then a shared-memory log-sum-exp merge across warps. grid=(heads*DG,seqs).
-// When out_q8 != nullptr AND ELEMS==1 (DG*32==HEAD_DIM), each (qh,dg) block's warp 0 also
-// emits the Q8_1 block for attn dims [qh*HEAD_DIM + dg*32, +32) from the bf16-rounded output,
-// so the O-projection MMVQ skips its standalone attn-quantize node (bit-identical to running
-// the quantizer on `out` afterwards). Q8_1 block index = qh*(HEAD_DIM/32) + dg.
+// When out_q8 != nullptr, each (qh,dg) block's warp 0 also emits the ELEMS Q8_1 blocks for
+// attn dims [qh*HEAD_DIM + dg*(HEAD_DIM/DG), +HEAD_DIM/DG) from the bf16-rounded output, so
+// the O-projection MMVQ skips its standalone attn-quantize node (bit-identical to running
+// the quantizer on `out` afterwards). Q8_1 block index = qh*(HEAD_DIM/32) + (doff + e*32)/32,
+// so the DG groups together cover all HEAD_DIM/32 blocks of the head at any ELEMS.
 template <int HEAD_DIM, int DG, int NW>
 __global__ void fa_combine_kernel(
     const float* __restrict__ part_m, const float* __restrict__ part_l,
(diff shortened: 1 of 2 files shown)


skyrocket2026:

To submit for review:

3. **Reopen** this PR

skyrocket2026:
closed

PR #614, opened 2026-07-24
perf(qwen36): tensor-core router logits + warp top-k + pipelined GDN scan + fused SwiGLU-quant (+24% pp @32k)

widecloud:
## Summary

Four bit-faithful batched-prefill speedups for the Qwen3.6-35B-A3B MoE hybrid: (1) router logits on the bf16 tensor cores instead of a warp-dot GEMV, (2) an in-kernel warp bitonic top-k, (3) the GDN chunk scan's W^/U0 tiles rewritten m-outermost (each `b_m exp(G_m) k[m][d]` formed once per m, not per (i,m)) and the state-decay precomputed per row, and (4) folding the routed-expert SwiGLU + int8 quantize into one kernel (drops the `P×mffn` bf16 round-trip). Each is numerically identical to the path it replaces. Rebased onto current `main` — the GDN changes preserve the short-final-chunk fixes from #604/#608 (store guarded by `i < len`; final gate reads `s_g[len-1]`).

## Proof of speedup

> `main` (`ce6844b`) and this PR built from source into separate trees on the same idle RTX 5090, run **interleaved** (frontier, PR, frontier, PR, …). Median of 3; scored sweep (`SPARKINFER_BENCH_SWEEP_CTXS=0,512,4096,16384,32768 SPARKINFER_BENCH_SWEEP_REPS=1`, `KV_INT8=1`). The bench binary is used instead of `bench.sh` because `bench.sh` is single-context and defaults to a different model.

- [x] Tested on **RTX 5090** (`sm_120`)

**Decode tok/s** (this PR does **not** target decode; shown as a no-regression check):

| | decode tok/s |
|---|--:|
| before (main) | 480.9 |
| after (this PR) | 482.6 |

**Prefill pp tok/s** (Qwen3.6-35B-A3B-UD-Q4_K_M, best context = 32768):

| | prefill pp tok/s |
|---|--:|
| before prefill (main) | 18788.4 |
| after prefill (this PR) | 23309.1 |

All scored Qwen3.6 contexts, median of 3 (marginal over frontier):

| ctx | main | this PR | marginal |
|---|--:|--:|--:|
| 512 | 5804.8 | 5985.5 | +3.1% |
| 4096 | 14730.3 | 17100.5 | +16.1% |
| 16384 | 18543.7 | 22429.9 | +21.0% |
| 32768 | 18788.4 | 23309.1 | **+24.1%** |

[...]

PR #615, opened 2026-07-24
fix(qwen36): GDN prefill state is handed to two decode kernels transposed

RealDiligent:
## Summary

The Gated-DeltaNet recurrent state is **one shared buffer** (`s.lin_state`): the batched-prefill
scan writes the post-prompt state and the decode `gdn_ar` kernel continues from it. Its memory
layout is therefore a cross-kernel contract — but only three of the five kernels that touch it
honour the same one. `gdn_ar_warpgrid_kernel` and `gdn_ar_kernel` index the buffer transposed
relative to what prefill writes, so on those paths every linear layer starts decoding from
`S^T` instead of `S`.

This is a pure correctness fix in `kernels/`. No shapes, defaults, env knobs or fast paths change,
and the currently-default configuration is untouched.

*(No proof-of-speedup section: this makes no speed claim, so there is nothing for the on-device
eval to grade. It does not touch `runtime/`.)*

## Root cause

Both prefill writers emit the **transposed `[v_head][col][row]`** layout:

```cpp
// kernels/csrc/cuda/fused/batched_prefill.cu:568   (pf_gdn_scan_kernel)
float* col = state + ((size_t)vh * HEAD_DIM + j) * HEAD_DIM;   // transposed [vh][col][row]

// kernels/csrc/cuda/fused/prefill_gdn_chunk.cu:415 (pf_gdnc_scan_kernel)
state[((size_t)h * HD + (j0 + jj)) * HD + m] = s_S[e];
```

`gdn_ar_fast_kernel` reads that layout and is correct. The other two decode kernels did not:

```cpp
// gdn_ar_warpgrid_kernel — SPARKINFER_GDN_FAST=0
float* sptr = state + (size_t)vh * HEAD_DIM * HEAD_DIM;
const float s = sptr[(size_t)i * HEAD_DIM + j];      // [vh][row][col]  <-- mismatch

// gdn_ar_kernel — head_dim != 128 fallback
float  s    = sptr[(size_t)i * head_dim + j] * g;    // [vh][row][col]  <-- mismatch
```

`S` is not symmetric, so `S^T != S` and the recurrence diverges immediately.

The mismatch was invisible because each kernel is **self-consistent in isolation**: with token-loop
prefill the same kernel both builds and consumes the state, starting from the memset-zero buffer,
and any permutation of zeros is still zeros. It only bites at the **prefill → decode handoff**,
which is exactly the default route for the hybrid models (`batched_prefill_enabled` is on by
default). The stale comment on `gdn_ar_warpgrid_kernel` — *"native `[vh][row][col]` layout (no
transposed state)"* — documented the divergence as if it were a free per-kernel choice, and the
dispatcher's note said only the fast kernel used a transposed layout.

Affected paths:

| path | trigger | before |
|---|---|---|
| `gdn_ar_warpgrid_kernel` | `SPARKINFER_GDN_FAST=0` | transposed state, wrong output |
| `gdn_ar_kernel` | `head_dim != 128` | transposed state, wrong output |
| `gdn_ar_fast_kernel` | default | correct |

`SPARKINFER_GDN_FAST=0` is the shipped opt-out and the code's own designated fallback for the
register-cached kernel — it is documented as the slower-but-equivalent path, and it is the first
thing anyone flips when bisecting a suspected GDN issue. Instead of being equivalent it silently
produces garbage for all linear layers, which sends the bisect the wrong way. `head_dim == 128`
holds for both shipped checkpoints, so that row is latent today but is a live trap for the next
linear-head-dim.

## Fix

Address the state as `[v_head][col][row]` in both kernels, matching the prefill writers and
`gdn_ar_fast_kernel`. Only the index arithmetic changes — the arithmetic, its ordering and its
rounding are byte-for-byte untouched:

```diff
-    float* sptr = state + (size_t)vh * HEAD_DIM * HEAD_DIM;
+    float* col = state + ((size_t)vh * HEAD_DIM + j) * HEAD_DIM;   // transposed [vh][col][row]
-        const float s = sptr[(size_t)i * HEAD_DIM + j];
+        const float s = col[i];
```

The dispatcher comment now states the layout once, as the cross-kernel contract it is, so a future
GDN kernel does not have to rediscover it.

## Verification

A CPU transcription of the prefill scan and all three decode kernels (same arithmetic, same
addressing), run as the runtime does it — batched prefill writes the state, then one decode step
reads it — with a random non-symmetric `S` and `ssm_a < 0`:

```
After batched prefill, first decode step (max |diff| vs GDN_FAST reference):
  main   GDN_FAST=0 (native [vh][row][col]) : 3.226229  <-- WRONG OUTPUT
  patch  GDN_FAST=0 (transposed          ) : 0.000000  OK
  resulting state  main : 8.583288   patch : 0.000000

Token-loop only (zero-init state, no prefill writer) — must be unchanged:
  main vs patch output : 0.000000000  identical
```

So the patched fallback now reproduces the default path exactly, and the configurations that were
already self-consistent are bit-identical.

## Impact and risk

- **Default configuration is unchanged.** `gdn_ar_fast_kernel` is not modified and stays the
  dispatch default, so decode/prefill throughput and the accuracy gate are unaffected.
- **Backward compatible.** The state buffer is scratch — allocated per session and memset to zero
  at `position == 0` — so there is no persisted layout to migrate.
- **Risk: low.** One file, two kernels, index arithmetic only; nothing else reads or writes
  `lin_state`. The only behaviour change is on the two paths that were already producing wrong
  results.
- Minor bonus, not claimed as a speedup: in the warpgrid kernel each warp's `NROW` rows are now
  contiguous, so its state reads and writes coalesce instead of striding by `HEAD_DIM`.

## Verified on RTX 5090

**No speed claim — this is a correctness fix, so the before/after tables below are intentionally
not filled and `needs-benchmark` is the correct outcome. It should not consume eval GPU time.**
The box is ticked because the change genuinely was built and executed on an RTX 5090; the logs are
below. Driver 580.173.02, CUDA 13.0.88, `-DCMAKE_CUDA_ARCHITECTURES=120`.

**Decode tok/s** (end-to-end): not applicable — no decode speed change is claimed.

| | decode tok/s |
|---|--:|
| before (main) | n/a — correctness fix |
| after (this PR) | n/a — correctness fix |

### Build + test suite, both refs

```text
$ nvidia-smi
NVIDIA-SMI 580.173.02   Driver Version: 580.173.02   CUDA Version: 13.0
GPU 0: NVIDIA GeForce RTX 5090

$ cmake -B build -DCMAKE_CUDA_ARCHITECTURES=120 -DCMAKE_BUILD_TYPE=Release && cmake --build build -j32
BUILD_EXIT_main=0            # main @ e5c35c8
BUILD_EXIT_patched=0         # this PR

$ ctest --test-dir build
100% tests passed, 0 tests failed out of 9      # main
100% tests passed, 0 tests failed out of 9      # this PR
```

No new compiler warnings from `qwen36.cu` on either ref.

### The bug, reproduced on device

Test does exactly what the runtime does for a hybrid layer — `launch_prefill_gdn_scan()` over a
prompt, then one `launch_qwen36_gdn_ar()` decode step continuing from that state — and prints
checksums of the decode output and the resulting state. `SPARKINFER_GDN_FAST` only selects which
decode kernel runs, so both settings must agree. `N=512` exercises the chunk-parallel scan, `N=64`
the sequential one.

**main @ e5c35c8 — the two decode kernels disagree completely:**

```text
N=512 GDN_FAST=1  decode_out sum=-2.440724 abs=139.345794   state abs=61642.8150
N=512 GDN_FAST=0  decode_out sum= 2.243543 abs= 64.015381   state abs=30024.6213
N=64  GDN_FAST=1  decode_out sum=-3.935697 abs=133.583020   state abs=86688.8714
N=64  GDN_FAST=0  decode_out sum=-1.717917 abs= 91.371400   state abs=48703.8434
```

Not drift — the output sum flips sign and the state magnitude halves. Per-element, e.g. at N=512
`out[2]` is `-0.010803` vs `+0.008789`.

**This PR — they agree exactly, at both scan paths:**

```text
N=512 GDN_FAST=1  decode_out sum=-2.440724 abs=139.345794   state abs=61642.8150
N=512 GDN_FAST=0  decode_out sum=-2.440724 abs=139.345794   state abs=61642.8150
N=64  GDN_FAST=1  decode_out sum=-3.935697 abs=133.583020   state abs=86688.8714
N=64  GDN_FAST=0  decode_out sum=-3.935697 abs=133.583020   state abs=86688.8714
```

### No regression on the default path

The patched `GDN_FAST=1` numbers are **identical to main's `GDN_FAST=1`** — same sums, same
`state abs`, same per-element values at every position printed. The shipped default path is
bit-for-bit untouched; only the fallback moves, from disagreeing to agreeing.

### DIFF
--- a/kernels/csrc/cuda/fused/qwen36.cu
+++ b/kernels/csrc/cuda/fused/qwen36.cu
@@ -208,19 +208,22 @@ __global__ void gdn_ar_kernel(const __nv_bfloat16* __restrict__ q,
     const __nv_bfloat16* qhptr = q + (size_t)qh * head_dim;
     const __nv_bfloat16* khptr = k + (size_t)qh * head_dim;
     const __nv_bfloat16* vhptr = v + (size_t)vh * head_dim;
-    float* sptr = state + (size_t)vh * head_dim * head_dim;
+    // Transposed state layout [vh][col][row] -- the ONE layout every GDN producer and consumer
+    // uses (both batched-prefill scans write it, gdn_ar_fast_kernel reads it). S[i][j] is at
+    // state[(vh*head_dim + j)*head_dim + i]; this thread owns column j, so its rows are contiguous.
+    float* col = state + ((size_t)vh * head_dim + j) * head_dim;
 
     float sk = 0.f;
     for (int i = 0; i < head_dim; i++) {
-        float s = sptr[(size_t)i * head_dim + j] * g;
-        sptr[(size_t)i * head_dim + j] = s;
+        float s = col[i] * g;
+        col[i] = s;
         sk += s * q36_to_f(khptr[i]);
     }
     const float delta = (q36_to_f(vhptr[j]) - sk) * b;
     float y = 0.f;
     for (int i = 0; i < head_dim; i++) {
-        float s = sptr[(size_t)i * head_dim + j] + q36_to_f(khptr[i]) * delta;
-        sptr[(size_t)i * head_dim + j] = s;
+        float s = col[i] + q36_to_f(khptr[i]) * delta;
+        col[i] = s;
         y += s * q36_to_f(qhptr[i]) * scale;
     }
     out[(size_t)vh * head_dim + j] = __float2bfloat16(y);
@@ -289,8 +292,11 @@ __global__ void gdn_ar_fast_kernel(const __nv_bfloat16* __restrict__ q,
     if (lane == 0) out[(size_t)vh * HEAD_DIM + j] = __float2bfloat16(y);
 }
 
-// Default GDN path: warp-per-state-column with native [vh][row][col] layout (no transposed
-// state). One warp owns column j; row slices live in registers; grid fills the GPU.
+// Fallback GDN path (SPARKINFER_GDN_FAST=0): warp-per-state-column, row slices in registers,
+// grid fills the GPU. Uses the same transposed [vh][col][row] state layout as gdn_ar_fast_kernel
+// and both batched-prefill scans -- the state is one shared buffer, so the layout is not a
+// per-kernel choice. It differs from gdn_ar_fast_kernel only in keeping the naive kernel's
+// separate decay pass, not in how the state is addressed.
 template <int WARPS_PER_BLK, int HEAD_DIM>
 __global__ void gdn_ar_warpgrid_kernel(const __nv_bfloat16* __restrict__ q,
                                        const __nv_bfloat16* __restrict__ k,
@@ -315,14 +321,14 @@ __global__ void gdn_ar_warpgrid_kernel(const __nv_bfloat16* __restrict__ q,
     const __nv_bfloat16* qhptr = q + (size_t)qh * HEAD_DIM;
     const __nv_bfloat16* khptr = k + (size_t)qh * HEAD_DIM;
     const __nv_bfloat16* vhptr = v + (size_t)vh * HEAD_DIM;
-    float* sptr = state + (size_t)vh * HEAD_DIM * HEAD_DIM;
+    float* col = state + ((size_t)vh * HEAD_DIM + j) * HEAD_DIM;   // transposed [vh][col][row]
 
     float sloc[NROW];
     float part_sk = 0.f;
     #pragma unroll
     for (int r = 0; r < NROW; r++) {
         const int i = lane + r * 32;
-        const float s = sptr[(size_t)i * HEAD_DIM + j];
+        const float s = col[i];
         sloc[r] = s;
         part_sk += s * q36_to_f(khptr[i]);
     }
@@ -333,7 +339,7 @@ __global__ void gdn_ar_warpgrid_kernel(const __nv_bfloat16* __restrict__ q,
     for (int r = 0; r < NROW; r++) {
         const int i = lane + r * 32;
         const float s = sloc[r] * g + q36_to_f(khptr[i]) * delta;
-        sptr[(size_t)i * HEAD_DIM + j] = s;
+        col[i] = s;
         part_y += s * q36_to_f(qhptr[i]) * scale;
     }
     const float y = q36_wsum(part_y);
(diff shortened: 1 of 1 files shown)


RealDiligent:
Requesting a `hold` label so this can be reviewed as a non-speedup correctness fix.

It makes no speed claim, so it should not consume RTX 5090 eval time, and I will not tick the proof-of-speedup box without having run on one. Draft is currently the only state the greenlight gate exempts (`isDraft` is checked before the 5090 gate), which is why it is open as a draft rather than ready for review — the diff itself is final. With `hold` applied I'll mark it ready immediately.

Note that draft PRs are auto-closed after 4 days in draft regardless of activity (`SPARKINFER_DRAFT_STALE_DAYS`), so this will lapse on its own if it isn't picked up.

skyrocket2026:

```bash
```

```bash
```

RealDiligent:
## Rebased onto current `main`, and the prefill→decode handoff re-reproduced against it

The branch predated #614 (which reworked the GDN chunk scan), so it is now rebased onto current `main`
(`37b3bb5`). The diff is unchanged: still `qwen36.cu` only (+24/−13). Re-ran the repro against post-#614
`main` on an RTX 5090 to confirm the mismatch is still live and the fix still closes it.

Both prefill writers still emit the transposed layout on current `main` —
`batched_prefill.cu:568` and `prefill_gdn_chunk.cu:502` — and the public header states the contract
outright:

> `launch_prefill_gdn_scan` … *leaves the recurrent state in the SPARKINFER_GDN_FAST transposed layout
> `[v_head][col][row]` the decode `gdn_ar_fast` kernel expects.*

Confirmed the buffer really is shared with no third party: `s.lin_state` is written only by
`launch_prefill_gdn_scan` (`qwen35_prefill.cpp:510`) and read/written only by `launch_qwen36_gdn_ar`
(`qwen35.cpp:650`), on the same per-layer slice.

### Repro: batched prefill, then one decode step continuing from that state

Exactly what the runtime does for a hybrid linear layer. `SPARKINFER_GDN_FAST` only selects which decode
kernel consumes the state, so both settings must agree. `N=512` exercises the chunk-parallel scan, `N=64`
the sequential one. Fixed-seed inputs, `ssm_a < 0`, `head_dim=128`, `q_heads=4`, `v_heads=8`.

```
===== N=512 (chunk-parallel prefill scan) =====
main    GDN_FAST=1 : decode_out sum=-0.061459 abs=15.391989   state abs=5231.0140
main    GDN_FAST=0 : decode_out sum=-0.364686 abs=15.302863   state abs=5302.5361   <-- disagrees
patch   GDN_FAST=1 : decode_out sum=-0.061459 abs=15.391989   state abs=5231.0140
patch   GDN_FAST=0 : decode_out sum=-0.061459 abs=15.391989   state abs=5231.0140

===== N=64 (sequential prefill scan) =====
main    GDN_FAST=1 : decode_out sum= 0.142402 abs=12.580221   state abs=6020.7208
main    GDN_FAST=0 : decode_out sum=-0.020396 abs= 9.679107   state abs=5985.3542   <-- disagrees
patch   GDN_FAST=1 : decode_out sum= 0.142402 abs=12.580221   state abs=6020.7208
patch   GDN_FAST=0 : decode_out sum= 0.142402 abs=12.580221   state abs=6020.7208
```

Byte-comparison of the full decode output + resulting state dumps:

```
                                        N=512        N=64
default path (FAST=1) main vs patch : IDENTICAL   IDENTICAL
patched  FAST=1 vs FAST=0           : IDENTICAL   IDENTICAL
main     FAST=1 vs FAST=0           : DIFFERS     DIFFERS     <-- the bug
```

Not drift: at `N=64` the fallback's output sum flips sign and `abs` drops from 12.58 to 9.68. After the
fix the two kernels are byte-identical at both scan paths, and the shipped default path is byte-identical
to `main` — so nothing about the current default configuration moves.

### Build + tests on RTX 5090

Driver 580.173.02, CUDA 13.0.88 (`nvcc V13.0.88`), `-DCMAKE_CUDA_ARCHITECTURES=120`, `sm_120`.
Each ref configured and built in its own fresh build tree (a copied CMake tree keeps the original
`CMAKE_CACHEFILE_DIR` and silently builds back into it — worth knowing if you reproduce this):

```
main @ 37b3bb5   BUILD_EXIT=0   100% tests passed, 0 tests failed out of 9
this PR          BUILD_EXIT=0   100% tests passed, 0 tests failed out of 9
```

No new warnings from `qwen36.cu`.

Still no speed claim, so `needs-benchmark` remains the right outcome and this should not consume eval GPU
time. Happy to land the harness as a proper test under `kernels/tests/` if you'd like it in-tree.

PR #617, opened 2026-07-24
test(kernels): cover the default and fused MoE router top-k selectors

kai392:
## Summary

`kernels/csrc/cuda/moe/router.cu` has **three** top-k selectors. `cpu_reference_test.cpp` models
one of them — and it is the one that does not normally run.

| selector | role | covered before |
|---|---|---|
| `moe_router_kernel` | k-pass mask-argmax | **fallback** | yes (`test_router`) |
| `moe_router_kernel2` | single-pass rank select | **default** | no |
| `si_warp_bitonic_top8` | fused decode-path bitonic top-8 | decode fast path | no |

`launch_moe_router` takes the rank-select path unless `SPARKINFER_ROUTER2=0`, so the selector that
actually picks the 8-of-256 experts for every token had no algorithm-level coverage — and neither
did the hand-written bitonic network `moe_router_fused_kernel` folds into the decode router GEMV,
which is the one place a bug would be least visible and most expensive.

This is **test-only**: no kernel, launcher, header or default is touched. It makes no speed claim,
so there is nothing here for the on-device eval to grade.

## What it adds

Both selectors are modelled against the same stable-sorted `(value desc, index asc)` reference the
existing `test_router` already uses, across three logit distributions:

- **random** — the ordinary case
- **ties** — logits drawn from a 3-value set, so exact duplicates are the norm
- **equal** — every logit identical, the degenerate extreme

The tie-heavy distributions are the substance. All three selectors document the *identical*
tie-break, and `moe_router_fused_kernel` claims its selection is "byte-identical" to
`gemv_f32 + kernel6". A divergence in any one of them silently routes tokens to different experts:
no shape changes, no assertion trips, nothing fails loudly — the model just quietly gets worse.

The rank-select model additionally asserts the rank map is a **bijection onto `[0, top_k)`** — every
slot written exactly once. That is the invariant that keeps `s_sel_id[]` / `s_sel_logit[]` fully
initialised; if two experts ever computed the same rank, one slot would go unwritten and the kernel
would emit an uninitialised expert id straight into the expert-FFN pointer arithmetic.

## Verification

All three selectors agree with the reference today — this codifies that rather than changing it:

```
$ g++ -O2 -Wall -Wextra -std=c++17 kernels/tests/cpu_reference_test.cpp -o t && ./t
  [PASS] router rank-select E256            max_err=1.938e-08 (tol=1e-06)
  [PASS] router rank-select ties            max_err=0.000e+00 (tol=1e-06)
  [PASS] router rank-select equal           max_err=0.000e+00 (tol=1e-06)
  [PASS] router bitonic8 E256               max_err=8.223e-09 (tol=1e-06)
  [PASS] router bitonic8 ties               max_err=0.000e+00 (tol=1e-06)
  [PASS] router bitonic8 equal              max_err=0.000e+00 (tol=1e-06)
ALL PASSED (0 failures)          # 24/24, warning-clean
```

To show the coverage has teeth rather than just passing, I mutated each modelled selector:

| mutation | result |
|---|---|
| rank-select tie-break `f < e` → `f > e` | `ties` + `equal` **FAIL**; bitonic unaffected |
| bitonic compare-exchange tie-break flipped | bitonic `ties` + `equal` **FAIL**; rank-select unaffected |
| bitonic merge loses its last reduction step (`off > 0` → `off > 1`) | **all three** bitonic cases FAIL |

Note that **random logits alone catch neither tie-break mutation** — only the duplicate-heavy
distributions do. That is the argument for including them.

## Risk

None to the runtime. The diff is one file, additive only, inside `kernels/tests/`; no `.cu`, no
launcher, no header, no default. Worst case it is dead weight in `ctest`; it cannot change decode
or prefill behaviour or throughput.

Minor note for whoever lands both: #613 also appends to `cpu_reference_test.cpp` (before `main()`
and in the `check(...)` list), so if that one merges first this may need a trivial textual
rebase — no semantic overlap, the two touch different tests.

### DIFF
--- a/kernels/tests/cpu_reference_test.cpp
+++ b/kernels/tests/cpu_reference_test.cpp
@@ -117,6 +117,137 @@ static double test_router(int E, int K) {
 }
 
 // ---------------------------------------------------------------------------
+// 2b/2c shared helpers. `ties` draws logits from a tiny value set so EQUAL
+//     logits are the common case rather than a rare accident — every selector
+//     below resolves ties by (value desc, index asc), and that rule is the part
+//     most easily broken by a rewrite.
+// ---------------------------------------------------------------------------
+static vector<float> router_logits(int E, int ties) {
+    vector<float> lg(E);
+    if (ties == 0)      for (auto& x : lg) x = frand();
+    else if (ties == 1) for (auto& x : lg) x = (float)(rng() % 3);   // many exact duplicates
+    else                for (auto& x : lg) x = 1.25f;                // all identical
+    return lg;
+}
+static vector<int> router_ref_topk(const vector<float>& lg, int K) {
+    vector<int> idx(lg.size()); for (size_t i = 0; i < idx.size(); i++) idx[i] = (int)i;
+    std::stable_sort(idx.begin(), idx.end(), [&](int a, int b) {
+        return lg[a] > lg[b] || (lg[a] == lg[b] && a < b); });
+    return vector<int>(idx.begin(), idx.begin() + K);
+}
+
+// ---------------------------------------------------------------------------
+// 2b. Router top-k, single-pass rank select (moe_router_kernel2) — the DEFAULT
+//     router: launch_moe_router takes this path unless SPARKINFER_ROUTER2=0, so
+//     it is what actually picks the experts for every token. test_router above
+//     models only the k-pass mask-argmax FALLBACK, leaving the live selector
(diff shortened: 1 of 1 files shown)


skyrocket2026:

To submit for review:

3. **Reopen** this PR

skyrocket2026:
closed

PR #638, opened 2026-07-29
fix: restore convert_gguf vocab fallback and align auto-merge docs

### LINKED ISSUE #637: fix(runtime): convert_gguf.py vocab fallback is dead-coded with if False
## Summary

`runtime/tools/convert_gguf.py` claims to take `vocab` from `token_embd.weight` when writing `config.txt`, but the assignment is disabled:

```python
```

The C++ GGUF path always prefers the embedding tensor (`emb->dims[1]` in `runtime/examples/qwen3_gguf_config.h` / `qwen35.cpp`). The Python converter therefore can write a wrong `vocab=` into `config.txt` when metadata is missing or disagrees with the tensor, which breaks downstream `load_weights` / LM-head sizing for this tool path.

## Expected

Prefer vocab from `token_embd.weight` shape (ggml `dims[1]` / Python `shape[1]` for 2D), matching the runtime.

## Scope

No GPU required — tooling correctness only.

kurosawareiji7007-hub:
## Summary

- Restore `convert_gguf.py` vocab fallback from `token_embd.weight` (was dead-coded with `if False`), matching the C++ GGUF config path (`emb->dims[1]`).
- Align README + dashboard with CONTRIBUTING: guarded `merge-first` auto-merge is real; drop the incorrect "bot never merges" wording.
- Update miner-guide current target to Qwen3.6 and CONTRIBUTING ctest guidance so contributors are not pointed at stale targets / a fake "5/5" gate.

Fixes #637

## Notes

Docs/tooling only — no GPU eval requested. Proof-of-speedup section intentionally omitted so the RTX 5090 opt-in CI does not auto-close this PR.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -27,7 +27,7 @@ source of the incentive loop is clear: SPARKINFER is built through **SN74 on Git
 ## Before you open a PR
 
 ```bash
-# 1. build + tests (must be 5/5)
+# 1. build + tests (all registered tests must pass; GPU tests self-skip without a device)
 cmake -B build -DCMAKE_CUDA_ARCHITECTURES=120 && cmake --build build -j && ctest --test-dir build
 
 # 2. speed — does it actually go faster?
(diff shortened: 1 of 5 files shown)


kurosawareiji7007-hub:
Note: `rtx5090-required` auto-closed this because it touches `runtime/`. Non-speed `runtime/` fixes need a maintainer `hold` label (or a real 5090 proof section). Happy to reopen if a maintainer can apply `hold` — otherwise I'll keep no-GPU fixes outside `runtime/`/`kernels/`/`moe/`.

PR #654, opened 2026-07-29
fix(dashboard): bind KPI Frontier decode to scored 128-tok frontier

### LINKED ISSUE #646: dashboard: KPI Frontier decode shows 512-ctx instead of scored 128-tok frontier
## Summary

The public dashboard KPI strip labels a number as **“Frontier decode”** but prefers Qwen3.6 **512-context** throughput (`q36ctx512.tps`) over the scored competition frontier (`q36.frontier_tps`, which matches the **128-tok** row and the SOTA card).

Against live `dashboard/data.js`:

| source | tok/s |
|---|---:|
| `qwen36.frontier_tps` / ctx `128` | **476.54** |
| ctx `512` (what KPI strip shows) | **480.91** |

So the page publishes two disagreeing “frontier” decode numbers, and the vs-llama.cpp % is computed on the wrong context.

## Expected

Bind the KPI strip to `q36.frontier_tps` / `q36.ref_tps` (128-tok competition frontier), matching the SOTA card and SN74 scored frontier identity.

## Scope

`dashboard/index.html` only. No GPU. Distinct from #639/#640 (bar colors) and #641/#642 (prefill labels).

kurosawareiji7007-hub:
## Summary

- KPI strip preferred Qwen3.6 512-ctx tok/s while labeling it Frontier decode.
- Scored frontier / SOTA card use 128-tok (q36.frontier_tps).
- Bind KPIs to frontier_tps / ref_tps so the public headline matches SN74 frontier identity.

Fixes #646
Supersedes #648

## Scope

Dashboard-only correctness fix (dashboard/index.html). Does not touch runtime/, kernels/, or moe/. No speed evaluation requested; proof section omitted on purpose.

### DIFF
--- a/dashboard/index.html
+++ b/dashboard/index.html
@@ -588,8 +588,10 @@
     </div>` : "");
   }
 
-  const kpiTps = q36ctx512?.tps || q36.frontier_tps || s.frontier_tps;
-  const kpiRef = q36ctx512?.ref_tps || q36.ref_tps || s.ref_tps;
+  // Bind the KPI strip to the scored Qwen3.6 competition frontier (128-tok), not 512-ctx.
+  // Preferring 512 here made "Frontier decode" disagree with q36.frontier_tps / the SOTA card.
+  const kpiTps = q36.frontier_tps || q36ctx128?.tps || s.frontier_tps;
+  const kpiRef = q36.ref_tps || q36ctx128?.ref_tps || s.ref_tps;
   const kpiRefName = q36.ref_name || s.ref_name;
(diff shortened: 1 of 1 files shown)


kurosawareiji7007-hub:
Kept as **draft on purpose**.

This is a no-GPU correctness PR (proof-of-speedup section omitted). Prior ready-for-review attempts (#648) were auto-closed by the eval bot RTX 5090 gate even with a clean description.

Please add the `hold` label so this can be marked ready without the gate closing it. Happy to mark ready immediately after `hold` is applied.

PR #652, opened 2026-07-29
fix(dashboard): map 32k/64k/128k decode labels to correct ctx colors

kurosawareiji7007-hub:
﻿## Summary

- Fix decode chart color mapping for `32k` / `64k` / `128k` labels.
- `parseInt("128k")` became `128`, so long-context bars used the wrong hue.
- Add explicit label-to-context mapping for decode charts.

Fixes #639

## Scope

Dashboard-only UI correctness fix (`dashboard/index.html`). No `runtime/` / `kernels/` / `moe/` changes.

PR #656, opened 2026-07-29
fix(dashboard): map 32k/64k/128k decode labels to correct ctx colors

### LINKED ISSUE #639: fix(dashboard): decode charts mis-color 32k/64k/128k via parseInt(label)
## Summary

Qwen3.5 / Qwen3.6 decode detail charts pick bar colors with `parseInt(x.label, 10)`.

That breaks any `Nk` label:

| label | `parseInt` | intended ctx |
|---|---:|---:|
| `32k` | 32 | 32768 |
| `64k` | 64 | 65536 |
| `128k` | 128 | 131072 |

So `32k`/`64k` bars get the wrong hue (and `128k` collapses onto the short-decode `128` color). Prefill charts already map labels as strings; decode should do the same.

## Expected

Shared label → ctx map (`4k`/`16k`/`32k`/`64k`/`128k`) so decode colors stay consistent across charts.

## Scope

Dashboard UI only (`dashboard/index.html`). No GPU / no `runtime`/`kernels`/`moe`.

kurosawareiji7007-hub:
## Summary

- Fix decode chart color mapping for 32k / 64k / 128k labels.
- parseInt("128k") became 128, so long-context bars used the wrong hue.
- Add explicit label-to-context mapping for decode charts.

Fixes #639
Supersedes #652

## Scope

Dashboard-only UI correctness fix (dashboard/index.html). Does not touch runtime/, kernels/, or moe/. No speed evaluation requested; proof section omitted on purpose.

### DIFF
--- a/dashboard/index.html
+++ b/dashboard/index.html
@@ -475,9 +475,13 @@
   const D = window.SPARKINFER, $ = id => document.getElementById(id);
   const s = D.status, maxPass = Math.max(...D.passes.map(p=>p.tps));
   // Shared context colors — same ctx size → same hue on every chart.
-  const CTX_COLOR = {128:"#D14D72",512:"#7B5DFF",4096:"#0E8A16",16384:"#B8860B",32768:"#6F42C1"};
-  const CTX_LABEL = {128:"128",512:"512",4096:"4k",16384:"16k",32768:"32k"};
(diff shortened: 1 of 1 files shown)


kurosawareiji7007-hub:
Kept as **draft on purpose**.

This is a no-GPU correctness PR (proof-of-speedup section omitted). Prior ready-for-review attempts (#652) were auto-closed by the eval bot RTX 5090 gate even with a clean description.

Please add the `hold` label so this can be marked ready without the gate closing it. Happy to mark ready immediately after `hold` is applied.

PR #658, opened 2026-07-29
perf(moe): dispatch pack4 MoE gate/up MMVQ (4 rows/CTA)

kurosawareiji7007-hub:
## Summary

- MoE gate/up MMVQ already had a pack4 kernel (4 FFN rows / CTA) that was never dispatched.
- Prefer pack4 over pack2 for Qwen3.6 (2048x512, 2048x768, top_k=8) and dense Qwythos (4096x12288).
- Align pack4 PDL with pack2 (signal at end). Opt out with SPARKINFER_GU_PACK4=0.

## Proof of speedup

> Tick the box only after a real RTX 5090 run of bench/scripts/bench.sh. Empty tables will not be evaluated.

| | decode tok/s |
|---|--:|
| before (main) |  |
| after (this PR) |  |

```text
# paste bench/scripts/bench.sh output here (before -> after)
```

## Notes

Draft until 5090 numbers are filled. Correctness should match pack2 (same si_vec_dot_q4_K); if accuracy regresses, set SPARKINFER_GU_PACK4=0.

### DIFF
--- a/kernels/csrc/cuda/moe/expert_ffn_q4k.cu
+++ b/kernels/csrc/cuda/moe/expert_ffn_q4k.cu
@@ -658,13 +658,14 @@ __global__ void gate_up_mmvq2_pack2_qwen_kernel(
     if (pdl) si_pdl_lc();
 }
 
+// Pack4: 4 FFN rows / CTA (16 warps). Same si_vec_dot_q4_K math as pack2; halves the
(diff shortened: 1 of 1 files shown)


kurosawareiji7007-hub:
Kept as draft until real RTX 5090 `bench/scripts/bench.sh` before/after numbers are filled and the checkbox is ticked. Opt-out: `SPARKINFER_GU_PACK4=0`.

PR #661, opened 2026-07-29
perf(qwen36): batch DFlash draft model projections and LM head across the block

widecloud:
## Summary

The DFlash draft model computed its Q/K/V/O/gate/up/down projections and LM head with one `launch_gemv` call per token in the 16-token block — 16x redundant re-reads of each weight matrix from DRAM. Replaces those loops with a batched-GEMV kernel (one warp per output row, 16-byte vectorized loads, weight read once and reused across the whole block) for every projection whose row count is exactly `block_size` (16), plus the LM head (now scored via a one-time eager bf16 dequant of the lm_head weight instead of a per-token quantized GEMV).
The variable-length context path (`fc`/`wk_ctx`/`wv_ctx`) is untouched — batching it the same way measured net-negative (padding overhead exceeds the win when true context length is well under 16) and was not shipped.

Confined to the draft model's own forward pass (`dflash_draft.cpp`, `dflash_kernels.cu`) — the target-model verify/accept-reject loop is untouched. Correctness is unaffected: SPEC_AGREE=100% speculative-decode exactness comes from the verify loop, which is independent of how the draft computes its proposals.

## Proof of speedup

**DFlash decode tok/s** (`build/runtime/qwen3_gguf_dflash_bench <target.gguf> <draft_dir> 256`, interleaved A/B, 5 reps, median-stable):

| | DFLASH_TPS |
|---|--:|
| before (main) | 311.93 |
| after (this PR) | 347.96 |

+11.55% marginal DFLASH_TPS. `qwen3_gguf_dflash_check`: `SPEC_AGREE 64/64 = 1.0000`, `MEAN_ACCEPT` unchanged vs main on every tested seed. `bench/scripts/dflash_accuracy.sh`: `VERDICT PASS`.

```text
=== rep 1: main ===
METRIC DFLASH_TPS 311.9369
=== rep 1: this PR ===
METRIC DFLASH_TPS 347.8526
=== rep 2: main ===
METRIC DFLASH_TPS 311.8391
=== rep 2: this PR ===
METRIC DFLASH_TPS 348.0370
=== rep 3: main ===
METRIC DFLASH_TPS 311.9138
=== rep 3: this PR ===
METRIC DFLASH_TPS 347.8096
```

### DIFF
--- a/runtime/csrc/cuda/dflash_kernels.cu
+++ b/runtime/csrc/cuda/dflash_kernels.cu
@@ -143,6 +143,87 @@ __global__ void k_attn(const bf16* q, const bf16* k, const bf16* v, bf16* out,
         ov[i] = f2b(acc[i] * inv);
 }
 
+// One warp per output row `n`, computing y[b][n] = sum_k x[b][k] * W[n][k] for all b in
+// [0,BATCH) at once. W stays in its native [N,K] "out,in" row-major layout (the same layout
+// the single-row GEMV path reads) -- unlike a tiled A@B GEMM, this needs no relayout. Each
+// warp reads its weight row from DRAM once and reuses it across the whole BATCH instead of
+// once per row, so BATCH separate GEMV launches collapse into one launch with ~BATCH-times
+// less weight traffic. Register cost is O(BATCH), so this is only used for the draft's
+// fixed block_size batch (16), never for the variable, potentially-large context length.
+// K is always a multiple of 8 for every weight this is used on (H=2048, qdim=4096,
+// kvdim=1024, I=6144), and cudaMalloc'd row bases land on >=16-byte boundaries, so every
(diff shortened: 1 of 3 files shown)


skyrocket2026:
## sparkinfer dflash auto-eval — `eval-dflash:XL`

| metric | value |
|---|---|
| **label** | `eval-dflash:XL` |
| PR DFlash tok/s | 236.86 |
| main DFlash tok/s | 193.78 |
| speedup vs main | **1.22×** (+22.2%) |
| PR AR tok/s | 291.21 |
| DFlash vs AR | 0.81× |
| mean accept τ | 5.200 |
| commit | `1e950f7bf` |

ok

<sub>Scored on pinned RTX 5090 vs same-box `origin/main` DFlash. AR `eval:*` labels are unchanged.</sub>

skyrocket2026:

PR #555, opened 2026-07-20
perf(qwen36): cache bf16 projection weights at load for batched prefill

widecloud:
## Summary

Qwen3.6-35B-A3B batched MoE prefill runs its Q/K/V/O, GDN, shared-expert and router projections as bf16 GEMMs (the int8 projection path is off for MoE). Every prefill pass re-dequantizes the **same static** GGUF weights to bf16 into a shared scratch buffer — `deq_q4k`/`deq_q8_0` measure ~26 ms of per-pass work at ctx 512 that is entirely N-independent (the weights don't change with prompt length). This PR dequantizes those projection weights **once at load** into persistent bf16 buffers (keyed by the quantized source pointer) and reuses them across passes, so the dequant is paid once instead of per prefill. Cached values are exactly what the per-pass dequant produced, so results are unchanged. Qwen3.6 MoE only; the dense Qwythos/Qwen3.5 path is untouched.

## Proof of speedup

> `main` (current frontier `e8339f0`, incl. #553) and this PR built into separate build trees on one idle RTX 5090, run **interleaved** (frontier, PR, frontier, PR, …) so drift hits both arms equally. Median of 3 each; scored sweep (`SPARKINFER_BENCH_SWEEP_CTXS=0,512,4096,16384,32768 SPARKINFER_BENCH_SWEEP_REPS=1`, `KV_INT8=1`), since `bench.sh` is single-context and defaults to a different model.

- [x] Tested on **RTX 5090** (`sm_120`)

**Decode tok/s** (this PR does **not** target decode; no-regression check):

| | decode tok/s |
|---|--:|
| before (main) | 492.32 |
| after (this PR) | 492.05 |

**Prefill pp tok/s** (Qwen3.6-35B-A3B-UD-Q4_K_M, best context = 512):

| | prefill pp tok/s |
|---|--:|
| before prefill (main = frontier, #553) | 3087.08 |
| after prefill (this PR) | 3661.96 |

All scored contexts, median of 3 interleaved runs:

| ctx | frontier (`e8339f0`) | this PR | marginal |
|---|--:|--:|--:|
| 512 | 3087.08 | **3661.96** | **+18.62%** |
| 4096 | 10535.97 | **11251.97** | +6.80% |
| 16384 | 13260.95 | 13509.07 | +1.87% |
| 32768 | 13690.51 | 13828.50 | +1.01% |

Decode at every scored context (flat, nowhere near the 2% guard):

| ctx | 128 | 512 | 4096 | 16384 | 32768 |
[...]

PR #673, opened 2026-07-30
fix(eval): fail closed when a failing bidir side carries no label

atoz96:
## Summary

`_public_eval_label` exists so a passing Qwen3.6 XL cannot paper over a Qwen3.5 regression (#555). Both of its selection loops require the failing side to *carry a label*, so a side with `pass: False` and no label of its own falls past them to `return res.get("label")` — which on a bidir run is the passing side's headline. Not a kernel change and claims no speedup.

> **No GPU eval needed.** Per the auto-close bot's own guidance ("If this PR does not need
> GPU eval, remove the proof-of-speedup section"), that section is omitted rather than left
> with an unticked box. This touches only `eval/` Python — the harness that grades PRs —
> changes no kernel and no runtime path, and cannot move decode or prefill. I have no
> Blackwell GPU, so I did not run `bench.sh`, `accuracy.sh` or `ctest`, and I claim no
> speedup. I am not ticking the RTX 5090 box, because that attestation would be false.
>
> (Supersedes #671, which was auto-closed for the unticked box and could not be reopened.)

## The bug

```python
if any(s.get("pass") is False for s in sides):
    for s in sides:
        if s.get("pass") is False and s.get("label") == "REJECT":
            return "REJECT"
    for s in sides:
        if s.get("pass") is False and s.get("label"):   # <- needs a label
            return s["label"]
return res.get("label")                                 # <- the passing headline
```

Running the current `_public_eval_label` over the bidir matrix:

| `score_qwen35` | `score_qwen36` | headline |
|---|---|---|
| `S` / pass | `XL` / pass | `XL` |
| `REJECT` / fail | `XL` / pass | `REJECT` |
| `none` / fail | `XL` / pass | `none` |
| **`None` / fail** | **`XL` / pass** | **`XL`** |

The last row is the papering-over the docstring names. It is reward-visible rather than cosmetic: in the SN74 registry `eval:XL` is ×4.0 and `eval:REJECT` is ×0.0, so a bidir run whose failing side lost its label pays four times instead of nothing.

`pass: False` with no label is the shape a partial record produces — the side failed before a verdict tier was assigned. That is the same class as the fail-open in the enclave scorer (#670): a stated guard defeated by incomplete evidence rather than by wrong evidence.

## The change

A failed side now falls back to `REJECT` instead of to the headline.

The existing preference order is untouched — a side carrying `REJECT` still wins, and a side carrying a real non-REJECT verdict (`none`) still headlines as that verdict — so the fallback only covers the case where no side offered one.

## Tests

Added next to the existing `test_bidir_public_label_reject_beats_passing_xl`:

- `test_bidir_failing_side_without_label_does_not_headline_as_passing` — the bug. Fails on `main`:
  ```
  AssertionError: 'XL' != 'REJECT'
  ```
- `test_bidir_failing_side_with_label_still_headlines_that_label` — pins that the fallback does not swallow a real `none` verdict
- `test_bidir_all_sides_passing_keeps_the_top_level_label` — pins that an all-passing run is unaffected

The last two pass before and after, on purpose: they bound the change rather than demonstrate it.

## Verification

```
python3 -m pytest eval/ -q     103 passed        (100 before, +3 new)
```

No new dependencies.

### DIFF
--- a/eval/pr_eval_bot.py
+++ b/eval/pr_eval_bot.py
@@ -638,6 +638,11 @@ def _public_eval_label(res):
             for s in sides:
                 if s.get("pass") is False and s.get("label"):
                     return s["label"]
+            # A side failed but carried no label of its own (partial record).
+            # Falling through here would headline the top-level label, which on
+            # a bidir run is the passing side — the exact papering-over this
+            # function exists to prevent. Fail closed instead.
+            return "REJECT"
     return res.get("label")
 
 
--- a/eval/test_pr_eval_bot.py
+++ b/eval/test_pr_eval_bot.py
@@ -896,6 +896,38 @@ class PrEvalBotPolicyTest(unittest.TestCase):
         self.assertIn("eval-qwen35:REJECT", body)
         self.assertIn("eval-qwen36:XL", body)
 
+    def test_bidir_failing_side_without_label_does_not_headline_as_passing(self):
+        """A failed side with no label of its own must not fall through to the
+        top-level label — on a bidir run that is the passing side, which is the
+        papering-over #555 established this function must prevent.
+
+        Reward-visible: `eval:XL` is x4.0 in the registry, `eval:REJECT` is x0.0.
+        """
+        res = {
+            "mode": "bidir", "label": "XL", "pass": True,
+            "score_qwen35": {"label": None, "pass": False, "tps": 0},
+            "score_qwen36": {"label": "XL", "pass": True, "tps": 3388.88},
+        }
+        self.assertEqual(bot._public_eval_label(res), "REJECT")
+
+    def test_bidir_failing_side_with_label_still_headlines_that_label(self):
+        """The fail-closed fallback must not swallow a real non-REJECT verdict."""
+        res = {
+            "mode": "bidir", "label": "XL", "pass": True,
+            "score_qwen35": {"label": "none", "pass": False},
+            "score_qwen36": {"label": "XL", "pass": True},
+        }
+        self.assertEqual(bot._public_eval_label(res), "none")
+
+    def test_bidir_all_sides_passing_keeps_the_top_level_label(self):
+        """No side failed, so the headline is unchanged."""
+        res = {
+            "mode": "bidir", "label": "XL", "pass": True,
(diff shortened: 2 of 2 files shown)


atoz96:
Heads-up: this fails `sensitive-paths-guard` — `SENSITIVE='^(eval/|bench/scripts/|\.github/|\.gittensor/|dashboard/data\.json$)'` covers everything here, and I am not a maintainer. That is the guard working as intended, so I am not going to try to route around it.

I have written both findings up in #674 so they survive independently of these PRs. Happy for a maintainer to take the fix through the maintainer path and close this, or to keep the branch if the diff is useful — your call, I will not close it unilaterally.

For the record on the earlier auto-close: I did not tick the RTX 5090 box, because these are eval-harness fixes with no speedup claim and the attestation would have been false. I removed the proof-of-speedup section per the bot's own guidance instead.

atoz96:
The failing check here is `sensitive-paths-guard`, and it is not fixable from a fork — that is the guard working as designed, not a problem with this diff:

```
PR #673 by atoz96 (NONE)
--- sensitive paths touched ---
eval/pr_eval_bot.py
eval/test_pr_eval_bot.py
::error::modifies maintainer-only paths (scoring / eval / governance). Only maintainers may change these. Blocked.
```

It passes on exactly two conditions — no changed file matching `^(eval/|bench/scripts/|\.github/|\.gittensor/|dashboard/data\.json$)`, or an author in `MAINTAINERS`/OWNER/MEMBER. This PR is entirely `eval/`, so the first is impossible without deleting the fix, and the second is not something I can satisfy. Editing the workflow would not help either: it runs under `pull_request_target` from `main`, and `.github/` is itself on the sensitive list. I am not going to try to route around any of that.

The other three checks (`rtx5090-required`, `cap-open-prs`, `copycat-guard`) are green, and `python3 -m pytest eval/ -q` passes locally with the regression test failing on `main` first.

I have posted both fixes as `git apply`-able patches on #674 so they can land without needing this PR at all. Fine by me to close this — the patch is the useful artifact, not the branch.

skyrocket2026:

To submit for review:

3. **Reopen** this PR

skyrocket2026:
closed

PR #677, opened 2026-07-31
perf(qwen36): cache static prefill weight conversions + refill the short-prompt prefill grids

fansilas:
## Summary

Four pieces of waste in the short-prompt batched prefill. None changes an arithmetic step — every one is a weight-conversion, occupancy or launch-count fix, so each output element keeps the same operands in the same order.

- **Static weight conversions ran every call.** Each dense projection re-ran `Q4_K/Q8_0 → bf16 → per-row fp8/int8` per layer per prefill — O(weight bytes), not O(prompt). Converted once at load into a 1408 MB / 250-tensor cache. Eager, not lazy, for the same reason `moe_rs_*` is: the sweep times each context exactly once, so a lazy fill lands inside the pass it is meant to speed up.
- **`pfm_moe_gemm_qi8_bm16` ran at 2 blocks/SM.** It inherited `BN=128` from `prefill_moe.cu`'s bm16 shape, putting `Bs` at 34816 B — against the BM=128 kernel above it, whose design note is that this technique *"lives on"* occupancy because it alternates a decode phase with an MMA phase. `BN=64` → 17408 B, 5 blocks/SM, identical work per thread (`threads = 2*BN` either way). **+7.3% pp.** `BN=32` measured worse (10537 vs 10711) and is not the default.
- **`ssm_alpha`/`ssm_beta` were two serialized latency chains** — same narrow bf16 projection of the same `xn`, 16 blocks each, un-prefetched K slices. One pass stages A once and cp.async double-buffers: **88.0 → 16.7 µs** per GDN layer.
- **Shared-expert gate+up were two 16-block launches** serializing on the stream. One `2·nx`-wide launch stages A once; each weight is still read exactly once, unlike shrinking `PF_BM`. **+2.1% pp.**

All default ON behind `SPARKINFER_PREFILL_WCACHE` / `_GEMM_SKINNY2` / `_MOE_QB_BN` / `_GEMM_PAIR` — one binary, A/B interleaved below. Scoped to the Qwen3.6 MoE hybrid; the cache releases its memory and disables itself if the prefill scratch arena ever fails.

## Proof of speedup

| | decode tok/s |
|---|--:|
| before (main) | 496.33 |
| after (this PR) | 495.62 |

Decode is untouched; −0.14% is session drift (every context lands ≥ 99.8% of main, guard is 98%).

**Prefill pp tok/s** (Qwythos / Qwen3.5 — fill if this PR targets prefill; use `--ctx 4096`, `32768`, `65536`, or `131072` and copy the `prefill pp` line — report your best context):

| | prefill pp tok/s |
|---|--:|
| before prefill (main) | 8611.02 |
| after prefill (this PR) | 10844.64 |

```text
qwen3_gguf_bench <gguf> 128 sweep, ctxs 128/512/4k/16k/32k, medians of 5 interleaved runs, same box

ctx     pp before     pp after      pp        TTFT cut    decode before → after
512      8611.02      10844.64    +25.9%      -20.6%       496.33 → 495.62
4096    24725.92      25917.11     +4.8%       -4.6%       476.25 → 475.83
16384   26194.95      26517.15     +1.2%       -1.2%       478.78 → 478.52
32768   27411.72      27625.94     +0.8%       -0.8%       479.19 → 478.86

512-token TTFT 59.46 -> 47.21 ms.  Scored context is 512 (best gain).
```

`ctest` 9/9. `accuracy.sh` vs llama.cpp with the flags OFF and ON is **identical to every digit** (long passes top-1 1.0000 / 0.9922 / 0.9922, KL 0.00138 / 0.04776 / 0.16069) — it scores `qwen3_gguf_score`, the token loop, which this PR does not touch. The batched-prefill H3 veto (`qwen3_gguf_prefill_check` 512/16, bars 0.80 / 0.05) reads top-1 0.875–1.000 / KL ≤ 0.0075 over 6 flags-ON runs against 0.875–0.9375 / KL ≤ 0.0072 on main over 4 — the same range (the MoE down projection scatters through float `atomicAdd`, so neither side is run-to-run reproducible).

### DIFF
--- a/kernels/csrc/cuda/fused/prefill_gemm_i8.cu
+++ b/kernels/csrc/cuda/fused/prefill_gemm_i8.cu
@@ -11,6 +11,7 @@
 #include <cuda_runtime.h>
 #include <cuda_bf16.h>
 #include <cuda_pipeline.h>
+#include <cstdlib>
 #include "sparkinfer/kernels/prefill_i8.h"
 
 #include "sparkinfer/kernels/prefill_quant_rows.h"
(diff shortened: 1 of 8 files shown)


skyrocket2026:

PR #701, opened 2026-08-07
perf(dflash): keep the Qwen3.6 block verify engaged and prebuild its replay graph

widecloud:
## Summary

Two things were costing the DFlash decode loop more than any kernel on the critical path: the block verifier dropped back to the token loop after *any* partial accept, and it built its ~810-node CUDA graph on decode step 2, inside the decode window. Fixing both, plus a K-split for the draft's projections, takes the eval prompt from 939.1 to 1066.8 tok/s (**+13.6%**) with byte-identical output and no change to mean accept length.

## Proof of speedup

- [x] Tested on **RTX 5090** (`sm_120`)

**Decode tok/s** (end-to-end, DFlash eval prompt, 2-rep alternating A/B, same box, same build flags):

| | decode tok/s |
|---|--:|
| before (main) | 939.14 |
| after (this PR) | 1066.84 |

**+13.6%** DFLASH_TPS. Mean accept length is unchanged at 5.3333 (24 steps), so this is entirely per-step cost, not acceptance.

**Prefill pp tok/s** — not targeted by this PR; prefill is untouched.

```text
# main @ 3373155 vs this PR, alternating, /tmp/dflash_eval_ids.txt (seed=fixed, 101-token prompt)
main rep1  METRIC DFLASH_TPS  940.1652   MEAN_ACCEPT 5.3333  steps=24
PR   rep1  METRIC DFLASH_TPS 1067.8550   MEAN_ACCEPT 5.3333  steps=24
main rep2  METRIC DFLASH_TPS  938.1072   MEAN_ACCEPT 5.3333  steps=24
PR   rep2  METRIC DFLASH_TPS 1065.8254   MEAN_ACCEPT 5.3333  steps=24

mean: main 939.14 -> PR 1066.84   (+13.6%)

# accuracy gate (TOK_REPO=Qwen/Qwen3.6-35B-A3B)
METRIC SPEC_AGREE 32/32 = 1.0000
VERDICT PASS

# bit-exactness of the row-batched Q4_K / GDN checkpoint path
[PASS] DFlash GDN checkpoints and compact commits equal serial state

# Qwen3.6 no-regression guard (qwen3_gguf_bench, n=64, 2 reps alternating)
main  533.07 / 532.98 tok/s
PR    533.26 / 532.88 tok/s
```

## Correctness

[...]

PR #704, opened 2026-08-07
perf(dflash): fuse row-batched router and shared-gate sigmoid in compact verification

andriypolanski:
## Summary

Reduce CUDA graph and dependent-launch overhead in Qwen3.6 DFlash compact verification:

- a row-batched router kernel that combines the BF16 split-K router projection with per-row
  bitonic top-k selection; and
- a BF16 shared-gate projection that applies sigmoid in the projection epilogue.

The fused router replaces the separate router GEMV and top-k launches in each target MoE layer.
Its completion counter resets itself after every launch, making it safe for CUDA graph replay.

## Scope and correctness

- The new dispatch is used only by DFlash compact block verification.
- Standard Qwen3.5/Qwen3.6 autoregressive decode and prefill call sites are unchanged.
- Router accumulation retains the existing row-batched split-K reduction order.
- Router selection preserves descending-logit order and the lower-expert-index tie-break.
- The shared-gate fusion preserves the original operation order:
  FP32 accumulation -> BF16 rounding -> FP32 sigmoid.
- Unsupported router and shared-gate shapes retain their existing fallback paths.
- `SPARKINFER_DFLASH_ROUTER_FUSED=0` disables the fused router for A/B testing.

## Proof of speedup

**Decode tok/s — unverified engineering estimate, not a benchmark result:**

| | DFlash decode tok/s |
|---|---:|
| before (`main`, source-derived reference) | 1,072.6 |
| after (this PR, projected) | ~1,126.2 |
| expected change | ~+5.0% |

RTX 5090 measurements are not available from the development environment. The 1,072.6 tok/s
reference is the latest relevant measured value documented by merged PR #701 for its DFlash K-split
path; PR #702 changed evaluation labels rather than runtime performance. The projected result applies
a 5% engineering estimate to that reference. It is not a same-box measurement of this commit and
must not be used to check the RTX 5090 attestation.

The optimization removes up to two dependent graph nodes per eligible MoE layer: the standalone
router top-k launch and shared-gate sigmoid launch. The fused paths retain opt-out and automatic
fallback behavior for direct comparison during maintainer evaluation.

## Validation

- [x] `git diff --check`
- CUDA build and GPU evaluation were not run locally because no CUDA/RTX environment was
  available.

## A/B commands

```bash
# Fused path (default)
build/runtime/qwen3_gguf_dflash_bench TARGET.gguf DRAFT_DIR 256 <token-ids>

# Original router path
SPARKINFER_DFLASH_ROUTER_FUSED=0 \
  build/runtime/qwen3_gguf_dflash_bench TARGET.gguf DRAFT_DIR 256 <token-ids>

# Correctness gate
bench/scripts/dflash_accuracy.sh TARGET.gguf DRAFT_DIR
```

### DIFF
--- a/kernels/csrc/cuda/gemm/gemv.cu
+++ b/kernels/csrc/cuda/gemm/gemv.cu
@@ -115,7 +115,7 @@ template __global__ void gemv_f32_sk_kernel<__nv_bfloat16, 8>(const __nv_bfloat1
 // Compact verification supplies up to four activation rows for the same matrix.
 // Preserve the split-K reduction independently for each row while sharing every
 // weight packet across those row accumulators.
-template <typename OutT, int S, int M>
+template <typename OutT, int S, int M, bool SIGMOID = false>
 __global__ void gemv_bf16_rows_sk_kernel(const __nv_bfloat16* __restrict__ x,
                                          const __nv_bfloat16* __restrict__ W,
                                          OutT* __restrict__ y, int N, int K) {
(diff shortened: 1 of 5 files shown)


skyrocket2026:
## Closed: stale — no commits in 2.0 days

This PR has had no new commits in over 2 days — closing automatically to keep the dflash eval queue clean. Reopen (or push a new commit / open a fresh PR) whenever you're ready to continue; it'll be picked back up on the next eval cycle.

PR #664, opened 2026-07-29
perf(qwen36): GQA-fused prefill attention (+8% pp @32k)

fansilas:
## Summary

GQA-fused prefill attention for Qwen3.6 (and Qwythos). The GQA-8 query heads that share a KV head now load each int8 K page / V tile **once** and feed all `RQH` heads from the resident fragment, instead of re-reading K/V per query head. New templated kernel `pf_attn_mma_gqa_kernel<HD,GROUP_BLKS,RQH>`, dispatched from `launch_prefill_attn_int8_paged`, shipped at `RQH=4` (`SPARKINFER_PREFILL_ATTN_GQA_RQH=1` restores the per-head kernel). One file, prefill-only, decode graph untouched.

Bit-identical to the per-head path: RQH=1 and RQH=4 produce the same `prefill_check` draw (integer int8 accumulation + the same online-softmax rounding).

## Proof of speedup

- [x] Tested on **RTX 5090** (`sm_120`)

**Decode tok/s** (end-to-end, `bench/scripts/bench.sh` -- decode-neutral, prefill-only diff):

| | decode tok/s |
|---|--:|
| before (main) | 427.73 |
| after (this PR) | 426.98 |

**Prefill pp tok/s** (Qwen3.6-35B-A3B, best context `--ctx 32768`):

| | prefill pp tok/s |
|---|--:|
| before prefill (main) | 25762.75 |
| after prefill (this PR) | 27774.07 |

```text
# Qwen3.6-35B-A3B (Q4_K_M), RTX 5090, bench/scripts/bench.sh --tokens 64 --ctx 32768
# main @05494a1:
decode tg    : 427.73 tok/s  (2.3 ms/token, n=64, ctx=32768, bs=1)
prefill pp   : 25762.75 tok/s  (ctx=32768, sequential KV fill)
# this PR:
decode tg    : 426.98 tok/s  (2.3 ms/token, n=64, ctx=32768, bs=1)
prefill pp   : 27774.07 tok/s  (ctx=32768, sequential KV fill)   (+7.8%)
```

PR #713, opened 2026-08-08
verify the GQA MMA smem opt-in per device and fail over instead of reporting success

tryeverything24:
## Summary

Hardening fix for the GQA prefill-attention tier added in #664: the `RQH=4` kernel
needs **76,032 B** of dynamic shared memory — the first configuration in this file past
the 48 KB default, so the `cudaFuncSetAttribute` opt-in stopped being advisory and
became load-bearing. The launcher latched it once **per process** (the attribute is
**per device**), discarded both the attribute's and the launch's error status, and
`launch_prefill_attn_mma` returned `true` unconditionally — so on any failure the
scalar windowed fallback was skipped and `attn` kept whatever it already held: stale
logits from the previous call, or garbage on the first. Silent, no diagnostic.

## Failure modes closed

1. **Second CUDA device in a process**: never receives the smem raise → the RQH=4
   launch fails with `cudaErrorInvalidValue` → prior contents of `attn` are consumed
   as this layer's attention output.
2. **Any arch/driver where `sharedMemPerBlockOptin` < 76,032 B**: the attribute call
   itself fails; same silent outcome.

Not a live miscompute on the current single-5090 eval box (99 KB opt-in) — stated
plainly. It is the failure mode the file's own fallback structure exists to prevent,
and #664 was the change that pushed the requirement past the default.

## Fix

`launch_attn_gqa` now returns whether it actually launched:
- latch keyed on the device ordinal (`cfg[dev]`),
- a refused opt-in fails the tier only where the raise is required (`sm > 48 KB`),
- the launch is checked with `cudaPeekAtLastError` (peek, not get — a pre-existing
  sticky error is not silently cleared here),
- a refusal cascades RQH=4 → RQH=2 (46,720 B, under the default) → per-q-head kernel,
  which is exactly the bool-and-fall-through contract `prefill_attn_mma.h` documents:
  `if (launch_prefill_attn_mma(...)) return;   // else fall through to the scalar kernels`

The per-head fallback's own latch is left as-is (32,064 B fits the default; advisory).
Success path unchanged instruction-for-instruction.

- ctest **10/10**.
- `qwen3_gguf_prefill_check` on Qwythos-9B Q4_K_M (hd256 — this exact path): TOP1 8/8,
  KL 0.00617, seed 68 — identical to main.
- `qwen3_gguf_bench` @ctx=4096: decode tg 164.2 tok/s, prefill pp 11,398 tok/s — in
  family with main (success path untouched).
- The multi-device repro needs a second GPU; not reproducible on this box — the fix is
  verified as no-behavior-change on the success path plus by inspection of the failure
  path. `compute-sanitizer` unavailable under WSL/WDDM.

### DIFF
--- a/kernels/csrc/cuda/fused/prefill_attn_mma.cu
+++ b/kernels/csrc/cuda/fused/prefill_attn_mma.cu
@@ -522,7 +522,7 @@ __global__ __launch_bounds__(GROUP_BLKS * 32, (RQH <= 2 ? 2 : 1)) void pf_attn_m
 }
 
 template <int HD, int GROUP_BLKS, int RQH>
-static void launch_attn_gqa(const void* q, const signed char* k_pool, const signed char* v_pool,
+static bool launch_attn_gqa(const void* q, const signed char* k_pool, const signed char* v_pool,
                             const void* k_scale, const void* v_scale, const int* block_table,
                             void* attn, int n_tokens, int n_q_heads, int n_kv_heads,
                             int block_size, int max_blocks_per_seq, float scale, int win_blocks,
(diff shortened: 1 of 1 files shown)


skyrocket2026:
## Closed: stale — no commits in 2.0 days

PR #724, opened 2026-08-08
check the windowed-attention smem opt-in per device and honor the fallback contract

rsnetworkinginc:
## Summary

The lane-parallel windowed prefill kernels request **98,816 B** of dynamic shared
memory — well past the 48 KB default — so their `cudaFuncSetAttribute` opt-in is
load-bearing, not advisory. `launch_prefill_attn_windowed` took that opt-in through a
**per-process** do-once latch (the attribute is a **per-device** setting), discarded
its status, discarded every launch's status, and returned `true` unconditionally. On
any failure, the caller — which this function's own tail documents as *"caller runs
full attention"* when the launcher declines — was told the work was done, skipped its
fallback, and consumed whatever `attn` already held: stale logits from the previous
layer, or garbage on the first. Silent, no diagnostic. Lane-parallel is the **default**
path here (`SPARKINFER_PREFILL_ATTN_LANEPAR` and `_WINDOW` both default on), so this is
the launcher's mainline.

## Failure modes closed

1. **Second CUDA device in a process**: never receives the smem raise → the ~97 KB
   launch fails with `cudaErrorInvalidValue` → reported as success → stale `attn`
   consumed as this layer's attention output.
2. **Any arch/driver with `sharedMemPerBlockOptin` < 98,816 B**: the attribute call
   itself fails; same silent outcome.

Not a live miscompute on a single sm_120 (opt-in ceiling ≈ 99 KB) — stated plainly.
It is exactly the failure mode this launcher's bool contract exists to prevent.

## Fix (follows the contract the function already defines)

- All four do-once latches keyed on the device ordinal (`cfg[dev]`).
- The attribute result is checked and is fatal only where the raise is **required**
  (`sm > 48 KB`); the 32 KB scalar-windowed and tiled kernels keep it advisory.
- Every launch checked with `cudaPeekAtLastError` (peek, not get — a pre-existing
  sticky error is not silently cleared here).
- On refusal: the windowed branch **degrades to the 32 KB scalar windowed kernel**;
  the unwindowed branch falls through to tiled / `return false` — the same paths the
  caller and this function already own (`return false; // window + tiling both off ->
  caller runs full attention`).

Success paths are unchanged instruction-for-instruction.

- ctest **10/10**.
- With the MMA path disabled so THIS launcher carries the entire hd256 prefill
  (`SPARKINFER_PREFILL_ATTN_MMA=0`): `qwen3_gguf_prefill_check` on Qwythos-9B Q4_K_M —
  **identical** to an unpatched build at both lengths: P=320 TOP1 8/8 / KL 0.00893 /
  seed 68; P=1031 TOP1 8/8 / KL 0.00580 / seed 68.
- Default path unchanged: prefill_check TOP1 8/8 / KL 0.00713 / seed 82;
  `qwen3_gguf_bench` decode 175.8 tok/s, pp 12,006 tok/s @ctx=4096 (in family).
- The multi-device repro needs a second GPU; not reproducible on this box — verified
  as no-behavior-change on the success path plus by inspection of the failure path.
  `compute-sanitizer` unavailable under WSL/WDDM.

### DIFF
--- a/kernels/csrc/cuda/fused/prefill_attn_window.cu
+++ b/kernels/csrc/cuda/fused/prefill_attn_window.cu
@@ -466,6 +466,25 @@ bool launch_prefill_attn_windowed(
     const size_t sm_lp = ((size_t)TK * (HD + 4) + (size_t)TK * HD) * sizeof(float)
                        + (size_t)TQ_LP * HD * sizeof(__nv_bfloat16);
 
+    // The lane-parallel tiles need 98,816 B of dynamic shared memory — well past the
+    // 48 KB default — so the cudaFuncSetAttribute opt-in below is REQUIRED for those
+    // launches to be valid, and both it and the launch itself have to be checked: a
+    // discarded failure used to report success to the caller, which then skipped its
+    // full-attention fallback and consumed whatever `attn` already held — silently
+    // wrong logits, no diagnostic. The attribute is also a PER-DEVICE setting, so the
+    // do-once latches are keyed on the device ordinal, not the process (a process-wide
+    // latch leaves every device but the first unconfigured, and the launch then fails
+    // with cudaErrorInvalidValue on exactly the path that needs the raise). On a
+    // refusal the windowed branch degrades to the 32 KB scalar windowed kernel, and
+    // the full branch to `return false` — the contract the tail of this function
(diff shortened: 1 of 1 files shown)


skyrocket2026:
## Closed: stale — no commits in 1.4 days

PR #728, opened 2026-08-09
perf(dflash): run the row-batched verify at long context (1.24x decode @4k)

### LINKED ISSUE #712: DFlash: compact-verify (dflash_verify_short_run) has an independent numerical gap from AR, off by default pending root cause
## Summary
`main` currently has a verified DFlash correctness regression at ~4096-token context, introduced by #707 ("Perf/dflash q5down nsplit fix"). DFlash diverges from greedy AR almost immediately at this context — not a minor accuracy wobble, a real correctness break.

## Evidence
Rebuilt both commits and ran `qwen3_gguf_dflash_check` (the `SPEC_AGREE` accuracy checker) at the exact same held-out 4096-token prompt on both:

| build | commit | SPEC_AGREE | verdict |
|---|---|---|---|
| pre-#707 | `c02297a` | 80/80 = 1.000 | PASS (clean) |
| post-#707 (current `main`) | `fc9dca7` | 5/80 = 0.0625 | **FAIL** — diverges from AR after ~5 tokens |

For contrast, the same test at 512-ctx:

| build | commit | SPEC_AGREE | verdict |
|---|---|---|---|
| pre-#707 | `c02297a` | 6/80 = 0.075 | FAIL — repeated `illegal memory access` CUDA errors, then degenerates into repeating one token |
| post-#707 (current `main`) | `fc9dca7` | 16/80 = 0.200 | FAIL — no more crash, but still doesn't match AR |

So #707 fixed a real crash at 512-ctx (though 512-ctx correctness is *still* broken, just not crashing), but appears to have introduced a *new* regression at 4k that didn't exist before.

## Likely root cause
#707's fix in `runtime/src/models/qwen35.cpp`:
```cpp
-    if (s.adaptive_splits) {
+    if (s.adaptive_splits && !s.dflash_capture) {
         int want = 32;
         if ((long)seqlen > 2L * s.split_chunk) want = 128;
         if ((long)seqlen > 28L * s.split_chunk && (long)seqlen <= 48L * s.split_chunk)
             want = MAX_NSPLITS;
         ...
```
freezes `n_splits` entirely during DFlash graph capture (`!s.dflash_capture`), to stop the adaptive-split logic from invalidating and re-capturing the DFlash CUDA graph mid-stream (the cause of the 512-ctx crash). But freezing means whatever `n_splits` was set to *before* capture began is used for the entire capture — if capture happens to lock in a split count appropriate for a short/different context (e.g. defaults to the `want = 32` short-context value) rather than the correct value for the actual ~4096-token context (`want = 128` per the un-frozen formula), decode at 4k would run with a mismatched split count. That's a plausible mechanism for silently wrong numerics rather than a crash — consistent with what's observed (no error, just wrong output).

This needs verifying against the actual capture-time code path, not just the diff.

## Impact
- `eval-dflash:*` auto-evaluation is **paused** (cron disabled) until this is resolved, so no more PRs get scored/merged against a baseline with unverified DFlash correctness at 4k.
- Any DFlash speed comparison at ~4k context on current `main` is unreliable until fixed — a PR could look faster while actually computing wrong output, the same failure mode documented here.

## Test plan for the fix
- `qwen3_gguf_dflash_check` at the 4096-token held-out prompt must show `SPEC_AGREE` back to a clean `VERDICT PASS` (ideally 1.0, matching pre-#707).
- Also verify 512-ctx doesn't regress back into the original crash.
- Ideally also check a context that exercises the `28*split_chunk`/`48*split_chunk`/`MAX_NSPLITS` thresholds, since the same class of bug (correct split count not reaching DFlash capture) could affect those bands too.

JSONbored:
## Summary

Compact block verification was capped at a 384-token sequence bound because the row-batched path
diverged from autoregressive decode past it (#712). This fixes both causes of that divergence and
lifts the bound, so the batched verify runs at long context: **+23.2% decode at 4k**, with 128-ctx
and 512-ctx behaviour bit-for-bit unchanged.

Also fixes a shared-memory reuse hazard in the batched GDN causal-conv kernels (#705).

## Why this closes #712

#712 reports that `dflash_verify_short_run` carries a numerical gap from AR that is independent of
the `n_splits` freeze, and asks for a root cause before the path is re-enabled. Two causes were
found and fixed:

1. the batched GDN causal conv accumulated its taps in a different order than the single-token
   decode conv, so every `k`/`v` differed from AR's by ~1 ulp and the difference compounded through
   the GDN recurrence `S = S*g + k*delta` until it flipped an argmax; and
2. above the bound, the verify drove the MoE with a single batched call over all N rows, which does
   not reproduce the `num_tokens=1` call autoregressive decode makes.

With the conv matching decode's order, and the verify issuing the same MoE call AR issues once per
row above the bound, the batched path reproduces AR exactly by construction. Verified at
**12 seeds × 3 contexts, 128 generated tokens each — 36/36 `SPEC_AGREE 128/128`**, four times the
length of the standard accuracy check.

## Proof of speedup

**Decode tok/s** (end-to-end DFlash decode, 128 generated tokens, alternating same-box A/B,
shipped defaults with no environment overrides — best of the three scored contexts is 4k):

| | decode tok/s |
|---|--:|
| before (`main` at `922fff0`) | 395.55 |
| after (this PR) | 487.38 |

**Prefill pp tok/s** — not applicable; this PR changes only the DFlash decode path and touches no
prefill entry point.

| | prefill pp tok/s |
|---|--:|

```text
GPU: NVIDIA GeForce RTX 5090, sm_120, CUDA 12.8
Model: Qwen3.6-35B-A3B-UD-Q4_K_M + z-lab/Qwen3.6-35B-A3B-DFlash
Alternating same-box A/B (main, PR, PR, main), shipped defaults, 0 SPARKINFER_* overrides.

  ctx=base main rep1 METRIC DFLASH_TPS 789.0036  MEAN_ACCEPT 5.3333
  ctx=base pr   rep1 METRIC DFLASH_TPS 785.1571  MEAN_ACCEPT 5.3333
  ctx=base pr   rep2 METRIC DFLASH_TPS 785.8258  MEAN_ACCEPT 5.3333
  ctx=base main rep2 METRIC DFLASH_TPS 784.6005  MEAN_ACCEPT 5.3333
  => base   main 786.80 -> PR 785.49   -0.2%

  ctx=512  main rep1 METRIC DFLASH_TPS 413.8721  MEAN_ACCEPT 2.1864
  ctx=512  pr   rep1 METRIC DFLASH_TPS 414.2402  MEAN_ACCEPT 2.1864
  ctx=512  pr   rep2 METRIC DFLASH_TPS 416.4665  MEAN_ACCEPT 2.1864
  ctx=512  main rep2 METRIC DFLASH_TPS 414.5539  MEAN_ACCEPT 2.1864
  => 512    main 414.21 -> PR 415.35   +0.3%

  ctx=4096 main rep1 METRIC DFLASH_TPS 395.6082  MEAN_ACCEPT 3.9091
  ctx=4096 pr   rep1 METRIC DFLASH_TPS 488.4439  MEAN_ACCEPT 3.9091
  ctx=4096 pr   rep2 METRIC DFLASH_TPS 486.3136  MEAN_ACCEPT 3.9091
  ctx=4096 main rep2 METRIC DFLASH_TPS 395.4824  MEAN_ACCEPT 3.9091
  => 4096   main 395.55 -> PR 487.38   +23.2%

Mean accepted length is identical between arms at every context, i.e. the committed token
stream is unchanged — the gain is fewer target forwards per step, not different tokens.
```

## Scope

- 128-ctx and 512-ctx take the same gate and the same single batched MoE call as `main`, so those
  contexts are unchanged by construction; the measurements above confirm it (-0.2% / +0.3%).
- Standard Qwen3.5/Qwen3.6 autoregressive decode and prefill call sites are untouched.
- `SPARKINFER_DFLASH_COMPACT_MAX_SEQ`, `SPARKINFER_DFLASH_ENGAGE_KEEP`,
  `SPARKINFER_DFLASH_ENGAGE_MINSEQ`, `SPARKINFER_DFLASH_MOE_ROWWISE` and
  `SPARKINFER_DFLASH_MOE_ROW_FANOUT` are available for A/B.

Fixes #712
Fixes #705

### DIFF
--- a/kernels/csrc/cuda/fused/batched_prefill.cu
+++ b/kernels/csrc/cuda/fused/batched_prefill.cu
@@ -351,6 +351,11 @@ __global__ void pf_gdn_conv_kernel(const __nv_bfloat16* __restrict__ qkv,
             }
             __syncthreads();
             cval *= sw[0];
+            // sw[] is reused by the next token's reduction, so without a barrier here a warp
+            // that has already advanced can overwrite sw[t>>5] while a slower warp is still
+            // reading sw[0] for THIS token (#705). do_norm is block-uniform, so the barrier is
+            // reached by every thread in the block.
+            __syncthreads();
         }
         out[(size_t)tok * out_dim + hh * head_dim + t] = __float2bfloat16(cval);
     }
@@ -492,6 +497,11 @@ __global__ void pf_gdn_conv_tile_kernel(const __nv_bfloat16* __restrict__ qkv,
             }
             __syncthreads();
             cval *= sw[0];
+            // sw[] is reused by the next token's reduction, so without a barrier here a warp
+            // that has already advanced can overwrite sw[t>>5] while a slower warp is still
+            // reading sw[0] for THIS token (#705). do_norm is block-uniform, so the barrier is
+            // reached by every thread in the block.
+            __syncthreads();
         }
         out[(size_t)tok * out_dim + hh * head_dim + t] = __float2bfloat16(cval);
     }
@@ -720,9 +730,18 @@ __global__ void df_gdn_conv_checkpoint_kernel(
     const size_t state_elems = (size_t)(conv_kernel - 1) * qkv_dim;
     for (int tok = 0; tok < n_tokens; tok++) {
         const float cur = pf_to_f(qkv[(size_t)tok * qkv_dim + d]);
-        float y = cur * w[conv_kernel - 1];
+        // Accumulate in EXACTLY the order the single-token decode conv uses
+        // (conv_split_l2norm_fused_kernel in qwen36.cu): every history tap in ascending order
+        // first, then the current token's tap last. Float addition is not associative, so summing
+        // `cur` first -- as this kernel used to -- lands a different low bit in q/k/v than decode
+        // does for identical inputs. That difference is ~1 ulp, but k and v feed the GDN
+        // recurrence (S = S*g + k*delta), whose state persists across decode steps, so it
+        // compounds until it flips an argmax and DFlash stops reproducing AR (#712). Matching the
+        // order makes the batched and single-token paths bit-identical.
+        float y = 0.f;
         #pragma unroll
         for (int c = 0; c < 7; c++) if (c < conv_kernel - 1) y += hist[c] * w[c];
+        y += cur * w[conv_kernel - 1];
         #pragma unroll
         for (int c = 0; c < 6; c++) if (c < conv_kernel - 2) hist[c] = hist[c + 1];
         if (conv_kernel >= 2) hist[conv_kernel - 2] = cur;
@@ -738,6 +757,11 @@ __global__ void df_gdn_conv_checkpoint_kernel(
             }
             __syncthreads();
             cval *= sw[0];
+            // Same sw[] reuse hazard as the prefill convs above (#705). This kernel is the one
+            // the DFlash compact verify runs, and a non-deterministic q/k norm here feeds k and v
+            // straight into the GDN recurrence, whose state persists across decode steps -- so the
+            // race compounds until it flips an argmax and the batched path stops reproducing AR.
+            __syncthreads();
         }
         out[(size_t)tok * out_dim + hh * head_dim + t] = __float2bfloat16(cval);
         if constexpr (WRITE_CHECKPOINT) {
--- a/runtime/src/models/qwen35.cpp
+++ b/runtime/src/models/qwen35.cpp
@@ -1986,7 +1986,49 @@ std::vector<int> Qwen35Model::dflash_generate(const std::vector<int>& prompt, in
         int v = e ? atoi(e) : 3;
         return v < 1 ? 1 : v;
     }();
+    // ...but "full block accepted" is a proxy, and a lossy one. What actually decides whether the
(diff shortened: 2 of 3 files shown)


skyrocket2026:
## sparkinfer dflash auto-eval — `eval-dflash:XL`

| metric | value |
|---|---|
| **label** | `eval-dflash:XL` |
| scored at context | 4k (best of 128/512/4k; regression at ANY rejects) |
| PR DFlash tok/s | 502.68 |
| main DFlash tok/s | 406.44 |
| speedup vs main | **1.24×** (+23.7%) |
| PR AR tok/s | 16.64 |
| DFlash vs AR | 30.21× |
| mean accept τ | 5.333 |
| DFlash @128 tok/s | PR 802.36 vs main 806.11 (-0.5%) |
| DFlash @512 tok/s | PR 425.88 vs main 427.06 (-0.3%) |
| DFlash @4k tok/s ⭐ *(scored)* | PR 502.68 vs main 406.44 (+23.7%) |
| Polaris receipt | `b09b590ec371284a…` — TDX (Intel hardware attestation) |
| commit | `729c0aa4b` |

ok (best at 4k context)

skyrocket2026:

PR #732, opened 2026-08-09
docs(server): fix broken PRO 6000 changelog link

marktech0813:
## Summary

`server/README.md` linked to `../changelog-pro6000.md`, which does not exist in the repo, so the PRO 6000 deploy section’s “full migration notes and benchmark table” reference was a dead link. Retarget it to the existing [`bench/results/qwen3-30b-a3b_q4km_pro6000.md`](../blob/main/bench/results/qwen3-30b-a3b_q4km_pro6000.md).

## Kind of change

- [x] Docs fix
- [ ] Perf / scored decode or prefill change

## Verification

Relative link from `server/README.md` resolves to a tracked file. No GPU / no `runtime/`.

PR #733, opened 2026-08-09
fix(server): make bench_api_vs_native.sh invoke the Python harness

marktech0813:
## Summary

`server/scripts/bench_api_vs_native.sh` was non-runnable: the `python3 <<'PY' ... PY` heredoc closed before the argv line, so Python saw `sys.argv=['']` and bash then tried to execute the repo root as a command. The embedded snippet also used empty chat special-tokens and killed the API server before calling it.

Delegate to the working `bench_api_vs_native.py` (optional gguf as `argv[3]`). Also correct `server/README.md`: build on `main` (not `dev`), default tokenizer is Qwen3.6-35B-A3B, and the PRO 6000 notes link targets `bench/results/qwen3-30b-a3b_q4km_pro6000.md` (supersedes #732).

## Kind of change

- [x] Bug fix (server tooling / docs)

## Verification

```bash
# argv wiring (no model required):
bash -c 'ROOT=/tmp/sparkinfer; API=http://127.0.0.1:8080; GGUF=/x.gguf; python3 -c "import sys; print(sys.argv)" "$ROOT" "$API" "$GGUF"'
# wrapper syntax:
bash -n server/scripts/bench_api_vs_native.sh
```

No GPU / RTX 5090 required.

### DIFF
--- a/server/README.md
+++ b/server/README.md
@@ -2,7 +2,7 @@
 
 OpenAI-compatible HTTP API for local GGUF inference — backend for sparkinfer.com.
 
-Enable with `-DBUILD_SERVER=ON` when building this repo (`dev` branch).
+Enable with `-DBUILD_SERVER=ON` when building this repo (`main`).
 
 ## Build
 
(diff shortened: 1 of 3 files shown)


marktech0813:
Closing: docs/tooling-only — does not target scored eval:* speedup under current master_repositories.json (sparkinfer emissions are eval-tier only). Will replace with a measured runtime/kernels perf PR.

PR #734, opened 2026-08-09
fix(quality): point sparkinfer --bin at build/runtime/

marktech0813:
## Summary

The quality harness docs still told users to pass `--bin ./build/qwen3_gguf_generate`. The superbuild installs that binary under `build/runtime/` (same path `bench/scripts/_common.sh` uses), so copy-paste runs failed with a missing binary.

## Kind of change

- [x] Bug fix (docs / quality tooling)

## Verification

```bash
grep -n 'build/runtime/qwen3_gguf_generate' bench/quality/README.md bench/quality/run_quality.py
```

No GPU required. Does not touch `bench/scripts/`.


marktech0813:
Reopened now that #732 is closed and we are back under the 5-PR author cap. Docs PR #735 stays closed until another slot frees; happy to reopen it then.

PR #736, opened 2026-08-09
fix(quality): point sparkinfer --bin at build/runtime/

marktech0813:
## Summary

The quality harness docs still told users to pass `--bin ./build/qwen3_gguf_generate`. The superbuild installs that binary under `build/runtime/` (same path `bench/scripts/_common.sh` uses), so copy-paste runs failed with a missing binary.

Replaces #734 (auto-closed by the 5-PR author cap during a burst open).

## Kind of change

- [x] Bug fix (docs / quality tooling)

## Verification

```bash
grep -n 'build/runtime/qwen3_gguf_generate' bench/quality/README.md bench/quality/run_quality.py
```

No GPU required. Does not touch `bench/scripts/`.

### DIFF
--- a/bench/quality/README.md
+++ b/bench/quality/README.md
@@ -43,7 +43,7 @@ python3 run_quality.py --backend mock    # ~0%  - floor
 # 2. score sparkinfer on a GPU box:
 python3 run_quality.py --backend sparkinfer \
     --model /workspace/models/Qwen3-30B-A3B-Q4_K_M.gguf \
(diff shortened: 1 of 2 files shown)


marktech0813:
Closing: docs/tooling-only — does not target scored eval:* speedup under current master_repositories.json (sparkinfer emissions are eval-tier only). Will replace with a measured runtime/kernels perf PR.

PR #741, opened 2026-08-09
fix(prefill): GDN chunk scan reports success when smem opt-in/launch fails

kai392:
## Summary

`launch_prefill_gdn_chunk` needs ~68 KB dynamic shared memory for the scan kernel (past the 48 KB default). On opt-in or launch failure it still returned `true`, so `launch_prefill_gdn_scan` skipped the sequential `pf_gdn_scan_kernel` fallback and left GDN state/out stale — silent wrong recurrence into decode.

Same failure class as the open attn hardenings (#713 / #724), but in `prefill_gdn_chunk.cu` (no open PR touches this file).

## Root cause

1. Process-wide `static int cfg` latch for a **per-device** `cudaFuncSetAttribute`.
2. Unconditional `return true` after launch with no `cudaPeekAtLastError` check.

## Fix approach

- Per-device opt-in latch (`cfg[dev]`).
- Refuse (`return false`) when the raise is required (`sm > 48KB`) and `cudaFuncSetAttribute` fails.
- `cudaPeekAtLastError()` after launch so a rejected enqueue falls through to the sequential scan.

Kernels-only. Not a speedup claim — no RTX 5090 eval requested.

## Impact

When chunk GDN cannot legally launch, the caller gets `false` and runs the correct sequential fallback instead of decoding from garbage state.

## Risk / tradeoffs

- On a device that cannot raise smem, prefill uses the slower sequential GDN scan (correct, previously silently wrong).
- Single-GPU RTX 5090 default path unchanged when opt-in succeeds (normal case).

## Test plan

- [x] `cmake --build build -j --target si_fused sparkinfer_runtime`
- [x] `ctest --test-dir build` — 10/10 passed
- [ ] Multi-GPU / constrained-smem box: confirm fallback engages instead of false success

### DIFF
--- a/kernels/csrc/cuda/fused/prefill_gdn_chunk.cu
+++ b/kernels/csrc/cuda/fused/prefill_gdn_chunk.cu
@@ -577,15 +577,29 @@ bool launch_prefill_gdn_chunk(const void* q, const void* k, const void* v,
                          + (size_t)HD * (JC + PAD) * sizeof(__nv_bfloat16)      // s_Sb
                          + (size_t)C * (JC + PAD) * sizeof(__nv_bfloat16);      // s_Ub
 
-    static int cfg = 0;
-    if (!cfg) {
-        if (cudaFuncSetAttribute(pf_gdnc_prep_kernel<C, HD>,
-                                 cudaFuncAttributeMaxDynamicSharedMemorySize, (int)sm_prep) != cudaSuccess)
-            return false;
-        if (cudaFuncSetAttribute(pf_gdnc_scan_kernel<C, HD, JC>,
-                                 cudaFuncAttributeMaxDynamicSharedMemorySize, (int)sm_scan) != cudaSuccess)
-            return false;
-        cfg = 1;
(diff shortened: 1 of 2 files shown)


kai392:
CI note: **CUDA build/package succeeded**. `build-linux` only fails at `Generate artifact attestation` with `missing "id-token" permission` — the known fork-PR limitation tracked in #612 (same as other open fork PRs that touch `kernels/`).

Verified locally: `ctest --test-dir build` → 10/10 passed after rebuilding `si_fused`.

kai392:
## CI fix

`build-windows` was a real compile failure (not attestation): `launch_gemv_rows2` instantiated `gemv_bf16_rows_sk2_kernel`, which is `#ifndef _MSC_VER`-only.

Pushed `89b7e63`: on MSVC, `launch_gemv_rows2` falls back to two `launch_gemv_rows` calls (bit-identical).

`build-linux` Build/Package already succeeded; remaining attestation `id-token` failure is the known fork issue (#612).

PR #743, opened 2026-08-09
fix(server/cb): prefix cache skipped after ContinuousBatch frees session 0

RealDiligent:
## Summary
- With `SPARKINFER_SERVER_PREFIX_TOKEN_*`, the server warms session 0 via `cache_prefix()` and sets `use_prefix_session`.
- `ContinuousBatchEngine` freed session-0 KV on completion but left `prefix_active == true`.
- The next exclusive prefix hit then saw `prefix_cached_len() == prefix.size()`, **skipped** `cache_prefix()`, and prefilling only the suffix against an empty block table → wrong tokens / missing GDN state.
- `generate()` already soft-invalidates (`prefix_active = false`, keep token fingerprint). CB/server did not.
- Add `Qwen35Model::release_prefix_session()` and call it wherever CB (and `generate`) frees session 0 after a prefix reuse. Also fix the destructor so in-flight `seq_id == 0` jobs are not skipped via truthiness.

## Test plan
- [x] Tested on RTX 5090
- [ ] With `SPARKINFER_SERVER_PREFIX_TOKEN_IDS` set, run two sequential chat completions that share the prefix; second request must re-warm (log / `cache_prefix`) and match first-request quality
- [ ] Without prefix env, CB path unchanged
- [ ] CI policy gates (`cap` / `gate` / `guard`)

### DIFF
--- a/kernels/csrc/cuda/gemm/gemv.cu
+++ b/kernels/csrc/cuda/gemm/gemv.cu
@@ -1842,6 +1842,12 @@ bool launch_gemv_rows(const void* x, const void* W, void* y,
 bool launch_gemv_rows2(const void* x, const void* W0, const void* W1, void* y0, void* y1,
                        int M, int N0, int N1, int K, cudaStream_t stream) {
     if (M < 1 || M > 8 || N0 < 1 || N1 < 1 || (K & 7)) return false;
+#ifdef _MSC_VER
+    // gemv_bf16_rows_sk2_kernel is compiled out under MSVC (see #ifndef _MSC_VER above).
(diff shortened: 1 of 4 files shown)


RealDiligent:
## CI fix

`build-windows` was a real compile failure: `launch_gemv_rows2` instantiated `gemv_bf16_rows_sk2_kernel`, which is `#ifndef _MSC_VER`-only (pre-existing on main; exposed by the full attested build).

Pushed an MSVC fallback: two `launch_gemv_rows` calls (bit-identical).

`build-linux` Build/Package already succeeded; remaining attestation `id-token` failure is the known fork issue (#612).

PR #745, opened 2026-08-09
perf(dflash): engage the batched verify on the evidence that pays (1.09x @128, 1.08x @4k)

FranDev132:
## Summary

The row-batched block verify is engaged by two gates — a full-block-accept score below a sequence
bound, an acceptance EMA above it. Both are slow to arm, and both spend that caution on exactly the
prompts the draft handles best. This tunes when each arms, and removes the row-count dependence that
kept the batched MoE from being driven in one call. Nothing here changes what the verify computes:
greedy DFlash still reproduces AR exactly (`SPEC_AGREE 32/32`, mean accepted length unchanged, at
every context measured).

## Proof of speedup

**Decode tok/s** (end-to-end, `qwen3_gguf_dflash_bench`, held-out prompt, 128 new tokens):

| | decode tok/s |
|---|--:|
| before (main) | 806.94 |
| after (this PR) | 879.59 |

**Prefill pp tok/s** (Qwythos / Qwen3.5 — not targeted by this PR):

| | prefill pp tok/s |
|---|--:|
| before prefill (main) | |
| after prefill (this PR) | |

All three DFlash contexts, same box, same build, paired runs — mean accepted length (τ) is identical
in every pair, so these are throughput changes and not a different continuation:

| context | main | this PR | change | τ | SPEC_AGREE |
|---|--:|--:|--:|--:|---|
| held-out primary (101 tok) | 806.94 | 879.59 | **+9.0%** | 5.3333 | 32/32 PASS |
| 512 | 427.79 | 426.68 | −0.26% | 2.1864 | 32/32 PASS |
| 4096 | 501.93 | 542.98 | **+8.2%** | 3.9091 | 32/32 PASS |

### What changed

**1. Arm on one full block, not a run of three.** A full-block accept is already evidence the draft
is tracking the target, and the decay still needs two non-full blocks to disarm, so the run length
only sets how long that evidence takes to accumulate — a wait paid on every prompt the draft handles
well. On the held-out short prompt (τ 5.33) that wait is the first steps of a ~24-step generation.
A prompt the draft handles badly is untouched, because it never lands the full block that arms it at
all: measured at τ 2.08, 437.2 → 434.4 tok/s, against −15% for forcing the batched path on there
unconditionally. This path only governs sequences below the short-context bound, so 512 and 4k are
unaffected by it.

**2. Seed the acceptance EMA past the sequence-length floor.** It ramps from zero at α = 1/8, so it
needs 8–10 steps to cross the threshold even on a stream that clears it comfortably — at 4k the
batched path sat idle for 13 of 33 steps, all warmup rather than genuine low acceptance. The ramp
exists so a transient run of full blocks cannot arm the batched path at short context; the floor
already excludes that case, so the seed is scoped above the floor only.

**3. Lower the engage threshold above that floor.** The acceptance that makes batching pay is not a
constant. One batched pass replaces `keep` sequential target forwards, and each of those forwards
re-reads the whole KV cache, so the token loop gets steadily more expensive with sequence length
while the N-row pass barely moves. A threshold calibrated where the token loop is cheap therefore
sits too high once the KV is long. At 4k: 529.2 → 542.1 tok/s.

**4. Pin the Q5_K down split count so the batched MoE reproduces AR exactly.** The MoE down
projection picks its split count from `num_tokens` — 8 at one row, 1 above — and the split count
fixes the reduction order, so a batched call returned different roundings than the `num_tokens = 1`
call AR decode makes, from identical inputs, expert ids and weights. Pinning S to AR's choice removes
the dependence: the down kernel then grids as `dim3(num_tokens, ...)`, one block column per token,
and the gate/up and quantize kernels grid per (token, expert, column), so every row computes exactly
what it would as a one-row call, for any N. The verify then takes one batched launch instead of one
per row. Scoped to the same bound, so short context stays byte-identical.

The new `ar_exact_splitk` argument defaults to `false`, so every other caller of
`launch_moe_expert_ffn_q4k` keeps its current behaviour.

```text
--- BEFORE (main) primary ---
AR            : 272.54 tok/s  (wall 0.470s)
DFlash        : 806.94 tok/s  (decode 0.159s, ttft 0.258s, wall 0.425s)
METRIC AR_TPS 272.5355
METRIC DFLASH_TPS 806.9380
METRIC MEAN_ACCEPT 5.3333

--- AFTER (this PR) primary ---
AR            : 272.78 tok/s  (wall 0.469s)
DFlash        : 879.59 tok/s  (decode 0.146s, ttft 0.258s, wall 0.412s)
METRIC AR_TPS 272.7815
METRIC DFLASH_TPS 879.5851
METRIC MEAN_ACCEPT 5.3333

--- BEFORE (main) 4096 ---
AR            : 16.87 tok/s  (wall 7.589s)
DFlash        : 501.93 tok/s  (decode 0.255s, ttft 10.953s, wall 11.221s)
METRIC DFLASH_TPS 501.9271
METRIC MEAN_ACCEPT 3.9091

--- AFTER (this PR) 4096 ---
AR            : 16.62 tok/s  (wall 7.700s)
DFlash        : 542.98 tok/s  (decode 0.236s, ttft 10.851s, wall 11.096s)
METRIC DFLASH_TPS 542.9784
METRIC MEAN_ACCEPT 3.9091

--- BEFORE (main) 512 ---
DFlash        : 427.79 tok/s  (decode 0.299s, ttft 1.308s, wall 1.620s)
METRIC DFLASH_TPS 427.7938
METRIC MEAN_ACCEPT 2.1864

--- AFTER (this PR) 512 ---
DFlash        : 426.68 tok/s  (decode 0.300s, ttft 1.308s, wall 1.617s)
METRIC DFLASH_TPS 426.6773
METRIC MEAN_ACCEPT 2.1864

--- accuracy gate, this PR (qwen3_gguf_dflash_check, 32 new tokens) ---
primary : METRIC SPEC_AGREE 32/32 = 1.0000   VERDICT PASS
512     : METRIC SPEC_AGREE 32/32 = 1.0000   VERDICT PASS
4096    : METRIC SPEC_AGREE 32/32 = 1.0000   VERDICT PASS
```

### DIFF
--- a/kernels/csrc/cuda/moe/expert_ffn_q4k.cu
+++ b/kernels/csrc/cuda/moe/expert_ffn_q4k.cu
@@ -1352,7 +1352,8 @@ void launch_moe_expert_ffn_q4k(
     int gate_type, int up_type, int down_type,
     const int* expert_ids, const float* expert_weights, void* output,
     float* h_scratch, float* out_scratch,
-    int num_tokens, int top_k, int hidden, int ffn, const void* input_q8, cudaStream_t stream
+    int num_tokens, int top_k, int hidden, int ffn, const void* input_q8, cudaStream_t stream,
+    bool ar_exact_splitk
 ) {
     // Qwythos dense hybrid fast path: pack2 gate/up without expert lookup + PDL-chained
     // quant/down. SPARKINFER_DENSE_FFN_FUSE=0 restores the generic MoE dispatch below.
@@ -1567,8 +1568,16 @@ void launch_moe_expert_ffn_q4k(
         // Row-count-aware split-K: S=8 (July-2026 sweep) is tuned for 1-row AR decode; the DFlash
         // compact verify runs num_tokens=6 rows -> already high occupancy, so split-K reduction
         // overhead dominates and S=1 wins (+2.7% DFlash decode @2550, bit-exact @128, SPEC 32/32).
+        //
+        // S also fixes the reduction order, so it is the one thing here that is not row-count
+        // invariant: at S=1 each output row is one warp's serial sum over ffn, at S=8 it is eight
+        // partial sums combined. Both are correct, but they round differently, which is why a
+        // batched call did not reproduce what AR computes one token at a time. A caller that needs
+        // that reproduction passes ar_exact_splitk and gets AR's split count for any num_tokens;
+        // with S pinned the kernel's grid is dim3(num_tokens, ...), one block column per token, so
+        // every row's arithmetic is identical to the num_tokens == 1 call.
         const char* s5env = getenv("SPARKINFER_DOWN_SPLITK_S_Q5");
-        const int Sbase = (num_tokens > 1 && !s5env) ? 1 : down_splitk_s_q5();
+        const int Sbase = (num_tokens > 1 && !s5env && !ar_exact_splitk) ? 1 : down_splitk_s_q5();
         const int S = dense_top1_down_splitk(Sbase, top_k, "SPARKINFER_DOWN_SPLITK_S_Q5");
         if (S > 1) {
             const int RPB = WPB / S;
--- a/kernels/include/sparkinfer/kernels/moe.h
+++ b/kernels/include/sparkinfer/kernels/moe.h
@@ -75,7 +75,12 @@ void launch_moe_expert_ffn_q4k(
     float* h_scratch, float* out_scratch,
     int num_tokens, int top_k, int hidden, int ffn,
     const void* input_q8 = nullptr,   // pre-quantized Q8_1(input) from the fused norm; nullptr = quantize internally
-    cudaStream_t stream = nullptr);
+    cudaStream_t stream = nullptr,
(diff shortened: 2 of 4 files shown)


skyrocket2026:

| metric | value |
|---|---|
| scored at context | 128 (best of 128/512/4k; regression at ANY rejects) |
| PR DFlash tok/s | 880.39 |
| main DFlash tok/s | 806.71 |
| speedup vs main | **1.09×** (+9.1%) |
| PR AR tok/s | 272.60 |
| DFlash vs AR | 3.23× |
| mean accept τ | 5.333 |
| DFlash @128 tok/s ⭐ *(scored)* | PR 880.39 vs main 806.71 (+9.1%) |
| DFlash @512 tok/s | PR 426.63 vs main 425.64 (+0.2%) |
| DFlash @4k tok/s | PR 543.18 vs main 501.96 (+8.2%) |
| Polaris receipt | `8fcba19706b05f1d…` — TDX (Intel hardware attestation) |
| commit | `24b52c371` |

ok (best at 128 context)

skyrocket2026:

PR #752, opened 2026-08-10
perf(dflash): skip context rows a windowed draft layer can never read (1.02x decode @16k)

FranDev132:
## Summary

The draft ingests the whole prompt on the first block of a generation, projecting every context row
into K/V for all six of its layers. Five of those layers are `sliding_attention` with
`sliding_window=4096`, and the split attention path now starts its key partition above the window
rather than masking inside the loop — so on a long prompt those leading rows are computed, written,
and then never read by anything. Skip producing them.

The bound is not recomputed here: it comes from the same function the launcher narrows with, so the
rows the producer skips are exactly the rows the consumer skips. Full-attention layers, and every
steady-state block (where the window still covers the cache), get a skip of zero and are unchanged.

## Proof of speedup

**Decode tok/s** (end-to-end, `qwen3_gguf_dflash_bench`, held-out 16k prompt, 128 new tokens):

| | decode tok/s |
|---|--:|
| before (main) | 580.11 |
| after (this PR) | 592.25 |

**Prefill pp tok/s** (Qwythos / Qwen3.5 — not targeted by this PR):

| | prefill pp tok/s |
|---|--:|
| before prefill (main) | |
| after prefill (this PR) | |

Every scored DFlash context, one binary, three alternating pairs each, medians reported. Mean
accepted length is identical in all twelve pairs, so these are step-cost changes and not a
different continuation:

| context | main | this PR | change | τ | SPEC_AGREE |
|---|--:|--:|--:|--:|---|
| held-out primary (101 tok) | 865.23 | 865.44 | +0.02% | 5.3333 | 32/32 PASS |
| 512 | 503.56 | 503.63 | +0.01% | 1.0000 | 32/32 PASS |
| 4096 | 534.95 | 534.32 | −0.12% | 3.9091 | 32/32 PASS |
| 16384 | 580.11 | 592.25 | **+2.09%** | 5.3750 | 32/32 PASS |

### What changed

**Skip the context rows a windowed layer can never read.** A host-timed split of the decode loop at
a 16k context shows the first block costs the draft 17.28 ms against a 1.83 ms steady state — that
one block is 6.9% of a 128-token decode (at 4k the same block is 1.9%, which is why this is a
long-context effect and not a general one). An `nsys` profile of the decode loop attributes it to 13
`pf_gemm` launches: one `fc` projection over the whole prompt, then a K and a V projection per layer.
Ten of those twelve per-layer projections belong to windowed layers, and three quarters of each was
dead on arrival.

The window predicate is `(q_pos - k_pos) >= window`, and the smallest query position in a block is
`q_pos0`, so nothing at or below `q_pos0 - k_pos0 - window` can contribute for any row. That is the
bound the split path already applies, so this factors it out of the launcher into
`dflash_kernels::attn_gqa_kv_lo` and calls it from the ingestion. Deriving it in one place is the
point: a row the ingestion skipped but the kernel still read would be uninitialized memory, and the
two must agree by construction rather than by matching two copies of a predicate. The launcher's
behaviour is unchanged — same value, same call site, same guards.

Evaluating the bound at the current block is sufficient for the whole generation. `q_pos0` grows by a
token per step while the one-split-worth guard the bound is subject to grows by a fraction of one, so
a row that is dead for this block stays dead for every later block.

The norm+RoPE over the K slice follows the same range. Row `i` of the surviving slice keeps its own
absolute position, so the rows that remain get exactly the angles they got before.

Result at 16k: the first block's draft cost falls 17.28 → 12.94 ms, the steady-state block is
unchanged at 1.83 ms, and mean accepted length is identical. `SPARKINFER_DFLASH_CTX_TRIM=0` restores
the previous behaviour.

```text
--- BEFORE (main) 16384 ---
DFlash        : 580.11 tok/s
METRIC DFLASH_TPS 580.1050
METRIC MEAN_ACCEPT 5.3750
[step split] step0 draft=17.28ms | steady draft=1.84ms/step verify=6.83ms/step

--- AFTER (this PR) 16384 ---
DFlash        : 592.25 tok/s
METRIC DFLASH_TPS 592.2451
METRIC MEAN_ACCEPT 5.3750
[step split] step0 draft=12.94ms | steady draft=1.83ms/step verify=6.82ms/step

--- alternating pairs, one binary (SPARKINFER_DFLASH_CTX_TRIM=0/1) ---
ctx=0      rep1 off 865.4444  on 864.2438   tau 5.3333 / 5.3333
ctx=0      rep2 off 865.2343  on 865.5351   tau 5.3333 / 5.3333
ctx=0      rep3 off 865.0554  on 865.4401   tau 5.3333 / 5.3333
ctx=512    rep1 off 503.9168  on 504.4901   tau 1.0000 / 1.0000
ctx=512    rep2 off 503.4232  on 503.1110   tau 1.0000 / 1.0000
ctx=512    rep3 off 503.5648  on 503.6300   tau 1.0000 / 1.0000
ctx=4096   rep1 off 535.2444  on 534.9691   tau 3.9091 / 3.9091
ctx=4096   rep2 off 533.3651  on 534.3174   tau 3.9091 / 3.9091
ctx=4096   rep3 off 534.9508  on 534.1508   tau 3.9091 / 3.9091
ctx=16384  rep1 off 580.9451  on 590.9113   tau 5.3750 / 5.3750
ctx=16384  rep2 off 580.1050  on 592.3017   tau 5.3750 / 5.3750
ctx=16384  rep3 off 579.3410  on 592.2451   tau 5.3750 / 5.3750

--- accuracy gate, this PR (qwen3_gguf_dflash_check, 32 new tokens) ---
primary : METRIC SPEC_AGREE 32/32 = 1.0000   VERDICT PASS
512     : METRIC SPEC_AGREE 32/32 = 1.0000   VERDICT PASS
4096    : METRIC SPEC_AGREE 32/32 = 1.0000   VERDICT PASS
16384   : METRIC SPEC_AGREE 32/32 = 1.0000   VERDICT PASS
```

### DIFF
--- a/runtime/csrc/cuda/dflash_kernels.cu
+++ b/runtime/csrc/cuda/dflash_kernels.cu
@@ -1184,6 +1184,30 @@ int attn_gqa_splits(int kv_len) {
     return s;
 }
 
+static inline int attn_gqa_split_path_ok(int q_len, int kv_len, int n_q, int n_kv, int d) {
+    static const int kSplitAttn = []{ const char* e = getenv("SPARKINFER_DFLASH_SPLIT_ATTN");
+                                      return (e && e[0] == '0') ? 0 : 1; }();
+    return kSplitAttn && d == 128 && n_kv > 0 && (n_q % n_kv) == 0 &&
+           q_len <= kDFlashAttnMaxRows && kv_len >= kDFlashAttnMinKv;
+}
+
+int attn_gqa_kv_lo(int q_len, int kv_len, int n_q, int n_kv, int d,
+                   int q_pos0, int k_pos0, int window) {
+    if (window <= 0 || !attn_gqa_split_path_ok(q_len, kv_len, n_q, n_kv, d)) return 0;
+    static const int kWinSpan = []{ const char* e = getenv("SPARKINFER_DFLASH_WINSPAN");
+                                    return (e && e[0] == '0') ? 0 : 1; }();
+    if (!kWinSpan) return 0;
+    // Narrow only when at least one whole split's worth of keys goes away: while the window still
+    // covers the cache this would trim a handful of keys, and re-partitioning the key range changes
+    // the order the online-softmax partials merge in -- not worth spending the draft's acceptance
+    // on for no traffic saved. The kv_len bound is belt-and-braces; the newest key is always inside
+    // the window, so lo < kv_len holds for any window >= 1.
+    const int n_splits_full = attn_gqa_splits(kv_len);
+    const long lo = (long)q_pos0 - (long)k_pos0 - (long)window + 1;
+    if (lo >= (kv_len + n_splits_full - 1) / n_splits_full && lo < (long)kv_len) return (int)lo;
+    return 0;
+}
+
 void launch_attn_gqa(const void* q, const void* k, const void* v, void* out,
                      int q_len, int kv_len, int n_q, int n_kv, int d,
                      int q_pos0, int k_pos0, int window, bool causal, float scale,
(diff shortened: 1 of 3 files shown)


skyrocket2026:
## sparkinfer dflash auto-eval — `eval-dflash:XS`

| metric | value |
|---|---|
| **label** | `eval-dflash:XS` |
| scored at context | 16k (best of 128/512/4k; regression at ANY rejects) |
| PR DFlash tok/s | 606.22 |
| main DFlash tok/s | 593.71 |
| speedup vs main | **1.02×** (+2.1%) |
| PR AR tok/s | 4.05 |
| DFlash vs AR | 149.67× |
| mean accept τ | 5.333 |
| DFlash @128 tok/s | PR 893.60 vs main 897.59 (-0.4%) |
| DFlash @512 tok/s | PR 522.03 vs main 521.57 (+0.1%) |
| DFlash @4k tok/s | PR 556.50 vs main 554.46 (+0.4%) |
| DFlash @16k tok/s ⭐ *(scored)* | PR 606.22 vs main 593.71 (+2.1%) |
| Polaris receipt | `0985b3e2e72154cf…` — TDX (Intel hardware attestation) |
| commit | `d37280ece` |

ok (best at 16k context)

skyrocket2026:

PR #763, opened 2026-08-10
perf(dflash): share the verify's KV tile across folded rows (1.08x decode @32k)

inference2026:
The batched verifier answers each candidate row with its own CTA, so the flash-decode split
kernel stages the same K/V tile into shared memory once per row and every row re-streams the
whole KV cache. Measured per launch at 4k, the split costs `6.3 + 7.83*rows` us — at the six
rows a block verifies, 88% of it is per-row, and at 32k that one kernel is 26% of the decode
step.

Fold three rows into one CTA so the staged tile is read once for all of them.

**Bit-exact by construction.** The folded rows end at different positions, so each keeps its
own chunk/start/end — exactly the range it gets today — and the CTA stages the union of those
ranges while every row masks back to its own bounds. Each row therefore accumulates the same
keys in the same order as before. Verified against the unfolded kernel on random Q/K/V at 32k:
max relative difference **0**, and mean accept is identical at every context.

**Scope.** Only the batched verify folds: gated on a bf16 KV cache, `head_dim` 256, GQA-8 and
`num_seqs > 1`, so single-token decode keeps one row per CTA and is untouched. The fold also
requires the folded rows to share a block table, which the verify already guarantees.

**Why 3.** The fold is limited by the extra accumulator set it keeps live per row. Isolated at
32k the split costs 1386 us unfolded, 952 us folded by 2, 944 us by 3 and 1548 us by 6 — wider
than 3 is worse than not folding at all. `SPARKINFER_FA_SEQFOLD=1` restores the unfolded path.

### Measured (RTX 5090, clocks locked 2550, mean accept identical in every pair)

| | decode tok/s |
|---|--:|
| before (main) | 460.0 |
| after (this PR) | 491.9 |

512 is unaffected — that band takes the autoregressive path, which never folds.

### Checks

- `ctest` 10/10
- `SPEC_AGREE 32/32 = 1.0000`, `VERDICT PASS` at 128 / 512 / 4k / 16k / 32k
- folded vs unfolded kernel output bit-identical (standalone, random inputs)

### DIFF
--- a/kernels/csrc/cuda/attention/flash_decode_split.cu
+++ b/kernels/csrc/cuda/attention/flash_decode_split.cu
@@ -13,6 +13,7 @@
 
 #include <cuda_bf16.h>
 #include <cuda_fp16.h>
+#include <climits>
 #ifndef SPARKINFER_NVRTC_DEVICE_ONLY
 #include <cuda_runtime.h>
 #endif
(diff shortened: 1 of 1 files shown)


skyrocket2026:

| metric | value |
|---|---|
| scored at context | 32k (best of 128/512/4k; regression at ANY rejects) |
| PR DFlash tok/s | 556.71 |
| main DFlash tok/s | 517.79 |
| speedup vs main | **1.07×** (+7.5%) |
| PR AR tok/s | 382.80 |
| DFlash vs AR | 1.45× |
| mean accept τ | 5.333 |
| DFlash @128 tok/s | PR 887.05 vs main 883.65 (+0.4%) |
| DFlash @512 tok/s | PR 518.59 vs main 519.57 (-0.2%) |
| DFlash @4k tok/s | PR 554.94 vs main 549.61 (+1.0%) |
| DFlash @16k tok/s | PR 627.35 vs main 602.12 (+4.2%) |
| DFlash @32k tok/s ⭐ *(scored)* | PR 556.71 vs main 517.79 (+7.5%) |
| Polaris receipt | `666e49d51143c448…` — TDX (Intel hardware attestation) |
| commit | `db3ccf761` |

ok (best at 32k context)

skyrocket2026:

PR #764, opened 2026-08-11
perf(dflash): ask the draft for two more tokens once the KV is long (1.04x decode @32k)

inference2026:
At 32k the mean accept length is **5.6087 out of a maximum of 6**. The proposal depth is 5, so
roughly 93% of steps end because the block ran out of proposals rather than because the target
rejected one — the block is being truncated by how much we ask for, not by how much the draft
gets right.

Raise the depth to 7 for sequences at or past 12288 and leave everything shorter exactly as it is.

**Why 7.** It is the widest depth that keeps every batched path: the compact verify itself, the
row-batched Q4_K / Q6_K / Q8_0 GEMVs (`MMAX=8` is the widest instantiation) and the batched MoE
all stop at 8 rows. It also costs the draft nothing — a width of `depth+1` rounds up to the same
8-wide block at depth 5 and 7, so the draft already computed those two rows and discarded them.
Depth 11 and 15 fall off all of the above and produce no tokens at all.

**Why gate it.** An extra verify row costs about one more set of routed experts — each row picks
its own 8 of 256 and they barely overlap — which is a flat price per row. What it buys is fewer
decode steps, and every step re-reads the whole KV cache. Step time at 32k is `3.08 ms fixed +
1.365 ms per row`, so short context, with no fixed cost to amortise, loses on the deal and long
context wins it.

The crossover sits between 8k and 12k, so the floor is 12288. Acceptance follows the text as much
as the length — the 8k prompt accepts 3.66 where the 4k one accepts 3.91 — so the floor is set
past the last length measured to lose rather than at it. `SPARKINFER_DFLASH_DEEP_MIN_SEQ` moves it
and `SPARKINFER_DFLASH_PROPOSALS` pins a depth outright.

**Fold.** An 8-row verify needs a fold that divides 8, so the split kernel gains a 4-row
instantiation and its launcher now defaults to the widest fold that divides the row count exactly
(8 → 4, 6 → 3) instead of a fixed 3, which would have switched folding off on `8 % 3 != 0`.
Measured on the 8-row verify at 32k: fold 4 = 518.6 tok/s against fold 2 = 512.4.

### Measured (RTX 5090, clocks locked 2550, 2 reps per arm)

| | decode tok/s |
|---|--:|
| before (main) | 498.1  |
| after (this PR) | 519.6 |

128 / 512 / 4k are below the floor and run the same code they do today; the differences there are
run-to-run noise. 512 takes the autoregressive path regardless.

### Checks

- `ctest` 10/10
- `SPEC_AGREE 32/32 = 1.0000`, `VERDICT PASS` at 128 / 512 / 4k / 16k / 32k

### DIFF
--- a/kernels/csrc/cuda/attention/flash_decode_split.cu
+++ b/kernels/csrc/cuda/attention/flash_decode_split.cu
@@ -839,21 +839,31 @@ void launch_flash_decode_split(
             // Fold S verify rows into one CTA so the staged K/V tile is read once for all of them
             // instead of once per row. Only when S divides num_seqs exactly, so no row is padded.
             static int seqfold = -1;
-            // 3 measured best on RTX 5090 (32k: 459.1 -> 497.0 tok/s). The fold is limited by the
-            // extra per-row accumulators it keeps live: 2 and 3 are within noise of each other in
-            // isolation (951.6 vs 943.8 us at 32k) while 6 regresses to 1548.0 us, worse than not
-            // folding at all. SPARKINFER_FA_SEQFOLD=1 restores one row per CTA.
-            if (seqfold < 0) { const char* e = getenv("SPARKINFER_FA_SEQFOLD"); seqfold = e ? atoi(e) : 3; }
-            if (!int8_kv && seqfold > 1 && num_seqs > 1 && (num_seqs % seqfold) == 0) {
+            // The fold is limited by the extra per-row accumulators it keeps live: 2, 3 and 4 are
+            // close in isolation (951.6 / 943.8 us at 32k for 2 and 3) while 6 regresses to
+            // 1548.0 us, worse than not folding at all. Since a fold only runs when it divides the
+            // row count exactly, the default picks the widest instantiation that does: a 6-row
+            // verify folds by 3 and an 8-row verify by 4, each leaving 2 CTAs per (kv head, split).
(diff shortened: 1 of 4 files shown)


skyrocket2026:
## sparkinfer dflash auto-eval — `eval-dflash:S`

| metric | value |
|---|---|
| **label** | `eval-dflash:S` |
| scored at context | 32k (best of 128/512/4k; regression at ANY rejects) |
| PR DFlash tok/s | 578.97 |
| main DFlash tok/s | 555.91 |
| speedup vs main | **1.04×** (+4.1%) |
| PR AR tok/s | 382.41 |
| DFlash vs AR | 1.51× |
| mean accept τ | 5.333 |
| DFlash @128 tok/s | PR 879.97 vs main 880.58 (-0.1%) |
| DFlash @512 tok/s | PR 518.79 vs main 518.41 (+0.1%) |
| DFlash @4k tok/s | PR 555.33 vs main 555.05 (+0.1%) |
| DFlash @16k tok/s | PR 632.94 vs main 627.53 (+0.9%) |
| DFlash @32k tok/s ⭐ *(scored)* | PR 578.97 vs main 555.91 (+4.1%) |
| Polaris receipt | `ffef368fad0dd382…` — TDX (Intel hardware attestation) |
| commit | `e8372ba65` |

ok (best at 32k context)

skyrocket2026:

PR #769, opened 2026-08-11
perf(muse-glimmer): keep attn_q/attn_gate quantized instead of dequantizing at load (1.19x decode @128)

FranDev132:
## Summary

`attn_q.weight` and `attn_gate.weight` both ship **Q4_K**, but the loader dequantized both to bf16
and interleaved them into a single `[q|gate]` tensor so it could reuse `split_q_gate_kernel` — which
expects the pre-fused layout Qwen3.6's GGUF writes. Muse Glimmer ships them as two separate tensors,
so the fuse existed only to satisfy the consumer.

That cost **109 MB per layer against 31 MB kept quantized — 4.08 GB of extra weight reads on every
decode token** across 52 layers, plus a resident bf16 copy of both tensors. In the 128-token decode
profile it made the dense bf16 GEMV the single largest kernel:

| ms | % of GPU busy | n | us/call | kernel |
|--:|--:|--:|--:|---|
| 712.96 | **29.8** | 7208 | 98.91 | `gemv_f32_sk` (~1 per layer) |
| 686.90 | 28.7 | 7072 | 97.13 | `gate_up_mmvq2` |
| 341.48 | 14.3 | 7072 | 48.29 | `down_q4k_mmvq_splitk` |

Its achieved bandwidth was already healthy (109 MB / 98.91 us = 1102 GB/s) — it was simply moving
3.6x the necessary bytes.

Both tensors now stay quantized (`attn_w`, same helper `wk`/`wv`/`wo` already use) and each is
projected straight into its own destination — Q into `s.q`, gate into `s.qgate` — through the
existing quantized `proj_xn` dispatch (`t==12` -> `launch_mmvq_q4k`). **No new kernel.** The
interleave buffer, its `cudaMalloc` + stream sync, and the split step all go away. The fused
`attn_qkv` path is excluded on this route because it writes one contiguous Q of width `nq` and has
no notion of a separate gate tensor.

## Proof of speedup

**Decode tok/s** (128 tokens, ctx=0 — the configuration the Muse Glimmer bot scores):

| | decode tok/s |
|---|--:|
| before (main) | 66.27 |
| after (this PR) | 78.97 |

Both arms come from **one binary**, switched by `SPARKINFER_MUSE_QGATE_Q` (`=0` restores the
previous load path exactly), so the comparison cannot be contaminated by a build difference.

```text
--- SPARKINFER_MUSE_QGATE_Q=0 (previous behaviour) ---
VRAM used    : 30.3 GB
decode tg    : 66.27 tok/s  (15.1 ms/token, n=128, ctx=0, bs=1)

--- SPARKINFER_MUSE_QGATE_Q=1 (this PR) ---
VRAM used    : 20.7 GB
decode tg    : 78.97 tok/s  (12.7 ms/token, n=128, ctx=0, bs=1)

+19.16% decode, -9.6 GB VRAM
```

VRAM falls by more than the removed interleave buffer alone because the dequantized copies of both
tensors are no longer resident.

### Accuracy

Teacher-forced `qwen3_gguf_score` over 19 positions, both arms, same binary. **Every argmax matches
and the top-5 candidate sets match**, with only near-tied reordering and third-decimal logprob
drift:

```text
arm0 (previous)  S i=2 tgt=382 am=264 lp=-16.718046 top=264:-1.554817,341:-2.366933,342:-3.180102
arm1 (this PR)   S i=2 tgt=382 am=264 lp=-16.750982 top=264:-1.582518,341:-2.406661,342:-3.272139
arm0 (previous)  S i=6 tgt=15334 am=290 lp=-14.555276 top=290:-2.964830,198:-3.650599,9264:-3.754043
arm1 (this PR)   S i=6 tgt=15334 am=290 lp=-14.542180 top=290:-2.945256,198:-3.697701,446:-3.780009
```

Bit-identical output was never possible here: moving from a dequantized-bf16 GEMV to Q4_K MMVQ
changes accumulation order and dequant rounding. What matters is that the distribution is unchanged
— and the baseline these arms are compared against is the one measured at top1=0.980 / KL=0.0029
against llama.cpp.

### Scope

Entirely inside the `c.muse_glimmer` load branch and the `w.wgate != nullptr` projection path;
`w.wgate` is only ever set for Muse Glimmer, so every other model keeps its existing fused
`s.qraw` + `split_q_gate` route unchanged.

### DIFF
--- a/runtime/include/sparkinfer/models/qwen35.h
+++ b/runtime/include/sparkinfer/models/qwen35.h
@@ -20,6 +20,10 @@ struct Qwen35LayerWeights {
     bool swa = false;
     const void* input_norm   = nullptr;  // [hidden]
     const void* wq = nullptr;            // [hidden, n_q_heads*head_dim]
+    // Muse Glimmer only: attn_gate ships as its OWN [hidden, n_q_heads*head_dim] tensor rather
+    // than pre-fused into wq the way Qwen3.6's GGUF writes it, so it is kept quantized and
+    // projected separately instead of being dequantized and interleaved into wq at load.
+    const void* wgate = nullptr;         // [hidden, n_q_heads*head_dim]
     const void* wk = nullptr;            // [hidden, n_kv_heads*head_dim]
     const void* wv = nullptr;            // [hidden, n_kv_heads*head_dim]
     const void* wo = nullptr;            // [n_q_heads*head_dim, hidden]
@@ -73,7 +77,7 @@ struct Qwen35LayerWeights {
     int gate_qtype = 0, up_qtype = 0, down_qtype = 0;
     // attention projections: 0 = bf16 dense (default); else ggml type id (12=Q4_K,
     // 14=Q6_K) -> weights kept quantized in VRAM, decoded on-read by launch_gemv_q.
-    int wq_type = 0, wk_type = 0, wv_type = 0, wo_type = 0;
+    int wq_type = 0, wgate_type = 0, wk_type = 0, wv_type = 0, wo_type = 0;
     int wqkv_type = 0, wqkv_gate_type = 0, ssm_beta_type = 0, ssm_alpha_type = 0, ssm_out_type = 0;
     int shared_gate_inp_type = 0;
 };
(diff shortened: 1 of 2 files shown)


skyrocket2026:

| metric | value |
|---|---|
| PR decode tok/s | 81.98 |
| main decode tok/s | 68.72 |
| speedup vs main | **1.19×** (+19.3%) |
| accuracy gate | ✅ top1=0.980 (bar >=0.9) · KL=0.0040 (bar <=0.2) |
| PPL sparkinfer / llama.cpp | 2.7315 / 3.0876 |
| Polaris receipt | `29ae0c90f4f136f3…` — TDX (Intel hardware attestation) |
| commit | `c02366bcf` |

ok

PR #770, opened 2026-08-11
perf(muse-glimmer): stop dequantizing the Q5_K LM head, and stop running hot decode kernels on one CTA (1.14x decode @128)

inference2026:
## Summary

Four independent wins on the 128-token decode path. The big one: `output.weight` is the **only
Q5_K tensor in the file**, and `lm_w()` kept a head quantized only for ggml types 12/14/8 — so
Q5_K fell through to `dense()` and the LM head was dequantized to bf16 at load, making decode read
**2.69 GB of head every token** through `gemv_f32_sk`. The rest are launch-latency: three of the
hot kernels here run as a *single CTA* (one SM of 170) because `rows == 1` at decode.

## Proof of speedup

**Decode tok/s** (`qwen3_gguf_bench <gguf> 128 0` — the configuration the Muse Glimmer bot scores):

| | decode tok/s |
|---|--:|
| before (main) | 78.97 |
| after (this PR) | 90.21 |

```text
=== BEFORE (main f7877a8) ===
main rep=1 decode tg    : 78.98 tok/s  (12.7 ms/token, n=128, ctx=0, bs=1)
main rep=2 decode tg    : 78.96 tok/s  (12.7 ms/token, n=128, ctx=0, bs=1)

=== AFTER (main + this PR) ===
patched rep=1 decode tg    : 90.19 tok/s  (11.1 ms/token, n=128, ctx=0, bs=1)
patched rep=2 decode tg    : 90.23 tok/s  (11.1 ms/token, n=128, ctx=0, bs=1)

+14.24% decode, VRAM 20.7 -> 18.8 GB
```

Two **separate builds**, clocks locked at 2550, 2 reps each and deterministic to the second decimal.

### What changed, largest first

**1. LM head Q5_K → Q4_K at load (+10.3%).** The head was the largest single kernel left on the
step at 1.60 ms of 12.4 ms. Requantized at load it reads 0.76 GB instead of 2.69 GB and runs on the
existing Q4_K MMVQ path. Accepting Q5_K is **opt-in per call site** (`allow_q5k`) rather than a
widening of `dev_quant_requant_q4k`'s accepted set — `dev_quant_down()` also routes through that
helper with requant on by default, and would otherwise start requantizing any other model's Q5_K
dense-FFN down tensor. `SPARKINFER_MUSE_LMHEAD_Q5K=0` restores the previous load path.

**2. Logit softcap grid (+2.32%, bit-identical).** `launch_logit_softcap` launched
`<<<n_rows, 256>>>`; at decode `n_rows == 1`, so a single 256-thread block walked all 202048
logits — one SM out of 170, 283 µs to touch 1.6 MB (5.7 GB/s). The vocab now spreads across
`blockIdx.x` with the row on `blockIdx.y`. Each element is computed by the same expression, so the
result is unchanged bit for bit.

**3. `si_mmvq_q4k_kfixed<26>` for K=6656 (+0.19%, bit-identical).** There were K-specialized
kernels for 2048 and 4096 and a runtime-K kernel otherwise, so Muse Glimmer's 6656 projections took
the generic path. Same loop and same reduction, only the trip count becomes constant. Small,
because that q/gate/k/v mix turns out to be launch-bound rather than instruction-bound — k and v
are only 256 rows.

**4. Fused sandwich-norm tail (+1.03%, byte-identical output).** Muse runs
`launch_norm_then_add` + `launch_rmsnorm` twice per layer; both are `<<<1, 256>>>` at decode, so
each is one CTA touching 13 KB in ~3.2 µs — 208 launches per token of pure latency. Fused into one
kernel per pair, keeping each stage's exact loop order and reduction tree and re-reading the
bf16-rounded intermediate from memory. The `qwen3_gguf_score` dump is byte-for-byte identical with
the fusion on and off. `SPARKINFER_MUSE_FUSE_TAIL=0` restores the pair.

**Measured and deliberately not shipped:** pairing `attn_q` + `attn_gate` into one 8192-CTA launch
is worth nothing (90.21 → 90.22, noise) — those 4096-CTA launches are not latency-starved and k/v
already ride separate streams.

**Deliberately not touched:** `gate_up_mmvq2` (5.07 ms/token) and `down_q4k_mmvq_splitk`
(2.49 ms/token) already run at 1534 and 1562 GB/s. They are bandwidth-saturated and shape-
specializing them wins nothing.

### Accuracy

`ctest` 11/11. Teacher-forced over 98 positions, both arms from one binary
(`SPARKINFER_MUSE_LMHEAD_Q5K`):

```text
PPL              2.70678  ->  2.71124   (+0.16%)
argmax agreement 97/98 = 0.9898
mean |dlogprob|  0.029414   max 0.234658
```

Bit-identical output is not possible for change 1 — requantizing the head changes dequant rounding.
The baseline arm here is the path validated at top1 0.980 / KL 0.0029 against llama.cpp in
`6d911d4`, and the distribution barely moves off it.

### Scope

Changes 1 and 3 are reachable only for Muse Glimmer (`c.muse_glimmer` + a Q5_K head; K=6656 is this
model's hidden size). Change 2 is shared but computes the same values for every caller.

### DIFF
--- a/kernels/csrc/cuda/fused/model_ops.cu
+++ b/kernels/csrc/cuda/fused/model_ops.cu
@@ -148,11 +148,15 @@ __global__ void decode_feedback_kernel(int* __restrict__ scalars,
 
 // Gemma2-style final-logit softcap, in place: logits[v] = tanh(logits[v] * scale / cap) * cap.
 // One block per row, grid-stride over vocab.
+// One CTA per row left a 202k-element vocab to a single 256-thread block at decode (n_rows == 1),
+// i.e. one SM of the device: 283 us to touch 1.6 MB, about 5.7 GB/s. Spread the vocab across
+// blockIdx.x and keep the row on blockIdx.y. Every element is still computed by exactly the same
+// expression, so the result is bit-identical -- only which thread evaluates it changes.
 __global__ void logit_softcap_kernel(float* __restrict__ logits, int vocab,
                                      float scale, float cap) {
-    float* L = logits + (size_t)blockIdx.x * vocab;
+    float* L = logits + (size_t)blockIdx.y * vocab;
     const float inv_cap = 1.f / cap;
-    for (int v = threadIdx.x; v < vocab; v += blockDim.x)
+    for (int v = blockIdx.x * blockDim.x + threadIdx.x; v < vocab; v += gridDim.x * blockDim.x)
         L[v] = tanhf(L[v] * scale * inv_cap) * cap;
 }
 
@@ -241,7 +245,10 @@ void launch_decode_feedback(int* scalars, const int* out_id, cudaStream_t stream
 
 void launch_logit_softcap(float* logits, int n_rows, int vocab, float scale, float cap,
                           cudaStream_t stream) {
-    logit_softcap_kernel<<<n_rows, 256, 0, stream>>>(logits, vocab, scale, cap);
+    // Cap the x-grid so a huge vocab does not spawn more blocks than the device can use; the
+    // kernel's grid-stride loop covers whatever is left over.
+    const int bx = (vocab + 255) / 256 > 1024 ? 1024 : (vocab + 255) / 256;
+    logit_softcap_kernel<<<dim3(bx < 1 ? 1 : bx, n_rows), 256, 0, stream>>>(logits, vocab, scale, cap);
 }
 #endif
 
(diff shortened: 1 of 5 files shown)


skyrocket2026:

| metric | value |
|---|---|
| PR decode tok/s | 93.41 |
| main decode tok/s | 81.97 |
| speedup vs main | **1.14×** (+14.0%) |
| accuracy gate | ✅ top1=0.980 (bar >=0.9) · KL=0.0103 (bar <=0.2) |
| PPL sparkinfer / llama.cpp | 2.7373 / 3.0876 |
| Polaris receipt | `090f232adcf19581…` — TDX (Intel hardware attestation) |
| commit | `6f8ade79e` |

ok

skyrocket2026:

PR #772, opened 2026-08-11
perf(muse-glimmer): shape-specialize the dense-FFN GEMVs and widen the sandwich-norm tail (1.04x decode @128)

Paral1995:
## Summary

Three changes on the Muse Glimmer 128-decode path, found by an `nsys --cuda-graph-trace=node`
trace of the decode graph (a plain `--trace=cuda` run shows none of these — the graph is collapsed
to a single entity).

**1. Dense-FFN gate/up shape arm (+1.42%).** `gate_up_mmvq2` was the largest kernel on the step at
5.07 ms of a 12.7 ms token, and `(6656, 19968, 1)` was the only shape still landing on the
runtime-parameterised kernel; every other model in the tree already has a compile-time arm. Adds
`gate_up_mmvq2_qwen_kernel<6656, 19968, 1>`.

**2. Dense-FFN down shape arm (+0.38%).** The same story for the down projection at 2.49 ms —
`down_q4k_mmvq_splitk_qwen_kernel<S, 78, 1>`. The split count was swept before specializing rather
than assumed: `S=8` is already optimal here (`S=8` 91.48, `S=4` 90.62, `S=2` 90.12), so only the
shape arm changed.

Both are **bit-identical**: same `si_vec_dot_q4_K` terms, same `kbx`/`kqs` mapping, same reduction
order as the generic kernels. Verified by a byte-for-byte `cmp` of the teacher-forced logit dump.

**No pack2 arm for gate/up**, unlike the other three shapes. pack2 pairs two output rows into an
8-warp CTA to win occupancy when `top_k*ffn` is only 4096; `F = 19968` already launches five times
that many blocks, so pairing only coarsens scheduling — measured, it gives back most of what the
plain specialization wins. Deliberately absent rather than merely unwritten.

**3. Sandwich-norm tail block width (+1.79%).** The fused tail runs one CTA of **256 threads** over
a 6656-wide row, twice per layer: `npack = 832` packs walked 3-4 deep per thread through two
dependent sum-of-squares reductions, on one SM of 170. That is 568 µs/token for ~13 KB of traffic
per call. One thread per pack collapses the chain; 1024 is the first width covering all 832 in a
single pass.

## Proof of speedup

**Decode tok/s** (`qwen3_gguf_bench <gguf> 128 0` — the configuration the Muse Glimmer bot scores),
two **separate builds**, clocks locked at 2550, two reps each:

| | decode tok/s |
|---|--:|
| before (main) | 90.20 |
| after (this PR) | 93.49 |

```text
=== BUILD A: main ===
decode tg    : 90.20 tok/s  (11.1 ms/token, n=128, ctx=0, bs=1)
decode tg    : 90.19 tok/s  (11.1 ms/token, n=128, ctx=0, bs=1)

=== BUILD B: main + this PR ===
decode tg    : 93.49 tok/s  (10.7 ms/token, n=128, ctx=0, bs=1)
decode tg    : 93.50 tok/s  (10.7 ms/token, n=128, ctx=0, bs=1)

+3.65% decode
```

Per-change attribution, so each carries its own number:

| | decode tok/s |
|---|--:|
| main | 90.20 |
| + gate/up arm | 91.48 |
| + down arm | 91.81 |
| + tail block width | 93.49 |

Tail width sweep on one binary (`SPARKINFER_MUSE_TAIL_W`): 256 → 91.80, 512 → 92.87,
768 → 92.69, **1024 → 93.49**.

### Correctness

`ctest` 11/11.

Changes 1 and 2 are bit-identical, and are checked as such in isolation:

```text
cmp of qwen3_gguf_score dump, generic vs specialized arms:   IDENTICAL (byte for byte)
```

Change 3 is **not** bit-identical and is not claimed to be: the block width sets how the
sum-of-squares partials are grouped, so fp32 reassociates. It is a reassociation rather than a
loss of quality. Teacher-forced over 191 in-distribution positions (a pattern the model predicts
at PPL ~2.1, so the distribution is sharp rather than degenerate):

```text
                              main        this PR
perplexity                    2.12623  ->  2.11524
agreement w/ true next token  176/191  ->  175/191

this PR vs main, same positions:  argmax agree 189/191,  mean |dlogprob| 1.1e-2
control (same build twice):       argmax agree 191/191,  |dlogprob| exactly 0
```

Perplexity moves slightly in the better direction and agreement with the true next token is
unchanged; the inter-arm flips are near-ties.

### Scope

Changes 1 and 2 are reachable only for `hidden == 6656 && ffn == 19968 && top_k == 1`. Change 3 is
reachable only through the Muse Glimmer sandwich-norm path. Other models match earlier arms and
take exactly the path they took before.

`SPARKINFER_GU_SPEC=0`, `SPARKINFER_DOWN_SPEC=0` and `SPARKINFER_MUSE_TAIL_W=256` restore the
previous behaviour individually.

### DIFF
--- a/kernels/csrc/cuda/fused/rmsnorm.cu
+++ b/kernels/csrc/cuda/fused/rmsnorm.cu
@@ -554,6 +554,7 @@ __global__ void rmsnorm_qk_kernel(__nv_bfloat16* __restrict__ q, __nv_bfloat16* 
 #ifndef SPARKINFER_NVRTC_DEVICE_ONLY
 #include "sparkinfer/kernels/fused.h"
 #include <cassert>
+#include <cstdlib>
 
 void launch_rmsnorm_qk(void* q, void* k, const void* q_w, const void* k_w,
                        int n_q_heads, int n_kv_heads, int head_dim, float eps, cudaStream_t stream) {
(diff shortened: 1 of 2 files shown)


skyrocket2026:

| metric | value |
|---|---|
| PR decode tok/s | 96.67 |
| main decode tok/s | 93.33 |
| speedup vs main | **1.04×** (+3.6%) |
| accuracy gate | ✅ top1=0.960 (bar >=0.9) · KL=0.0114 (bar <=0.2) |
| PPL sparkinfer / llama.cpp | 2.7616 / 3.0876 |
| Polaris receipt | `84e27e521f1cc340…` — TDX (Intel hardware attestation) |
| commit | `28594bdfb` |

ok

skyrocket2026:

PR #783, opened 2026-08-12
perf(muse-glimmer): windowed batched-GEMM prefill — 10.1x prefill@128 (opt-in)

James-CUDA:
## Summary

Windowed **batched-GEMM prefill for Muse Glimmer** — directly targeting the newly-scored
`prefill@128` dimension. Muse was the one hybrid arch excluded from `prefill_batched_run`
(`if (cfg.muse_glimmer) return false`) because that path's full-attention kernels apply RoPE over
the whole sequence with no per-layer sliding-window/NoPE hook — so Muse prompts fell back to the
sequential token loop (one memory-bound GEMV per position, re-reading all ~17 GB of weights every
token, flat ~112 pp regardless of context). This PR adds the missing muse-specific kernels so
prefill runs as a single compute-bound pass:

- `pf_qknorm_ropenorm_kv_bf16_kernel` — bf16-KV QK-norm + NORMAL-RoPE (SWA layers) / NoPE (global
  layers) KV-write, all N tokens at once;
- `win_prefill_pure_bf16_kernel` — pure rolling-window bf16 attention (`win_blocks>0` = SWA,
  `≤0` = global full-causal), one kernel for both Muse layer types;
- sandwich-norm / embedding-norm / logit-softcap parity with the decode path.

Numerically equivalent to the validated token-loop, so a subsequent decode reads a consistent KV
cache. **Opt-in** behind `SPARKINFER_MUSE_BATCHED=1` (default routes the token loop, unchanged).
Reaches only Muse Glimmer (`cfg.muse_glimmer`, hd128 dense, bf16 KV); every other model takes
exactly the arm it took before. Decode is untouched (no regression).

## Proof of speedup

**Decode tok/s** (Muse Glimmer, ctx=0 — unchanged; this PR targets prefill, no decode regression):

| | decode tok/s |
|---|--:|
| before (main) | 99.2 |
| after (this PR) | 99.7 |

**Prefill pp tok/s** (Muse Glimmer, `prefill@128` — the newly-scored dimension; clocks 2550, bf16 KV):

| ctx | before (token loop) | after (`SPARKINFER_MUSE_BATCHED=1`) | speedup |
|---|--:|--:|--:|
| **128** | 112.2 | 1134.5 | **10.1×** |
| 512 | 111.5 | 3253.0 | 29.2× |

## Correctness

Accuracy parity vs the sequential token-loop prefill (== the validated decode path), scored with a
128-token cached prefix over a 400-token sequence (271 scored positions), bf16 KV:

| | PPL | argmax-agree |
|---|--:|--:|
| sequential prefill (token loop) | 1.23312 | 267/271 |
| **batched prefill (this PR)** | **1.23191** | **267/271** |

0.1% PPL (batched marginally lower), identical argmax — the expected non-bit-identical GEMM-vs-GEMV
reduction-order difference. `compute-sanitizer memcheck`: **0 errors**. Qwen3.5/3.6 cross-model
guard: **no regression** — same-box A/B shows Qwen3.6 decode+prefill within noise and bit-identical
accuracy (the muse arms are gated behind `cfg.muse_glimmer`).

```text
# qwen3_gguf_bench muse-glimmer-30B-kquant-17gb.gguf 8 128   (before -> after, RTX 5090, clocks 2550)
before (SPARKINFER_MUSE_BATCHED unset): prefill pp   : 112.21 tok/s  (ctx=128, sequential KV fill)
after  (SPARKINFER_MUSE_BATCHED=1)    : prefill pp   : 1134.47 tok/s (ctx=128)
# ctx=512: 111.49 -> 3252.95 tok/s ; decode tg unchanged: 99.2 -> 99.7 tok/s
```

### DIFF
--- a/kernels/csrc/cuda/fused/batched_prefill.cu
+++ b/kernels/csrc/cuda/fused/batched_prefill.cu
@@ -923,6 +923,84 @@ __global__ void pf_qknorm_rope_kv_int8_kernel(
 }
 
 // ============================================================================
+// Muse Glimmer full-attention prefill: QK-norm + NORMAL RoPE (consecutive-pair,
+// LLAMA_ROPE_TYPE_NORM) + BF16 KV write into the single-sequence paged pool. The
+// bf16 counterpart of pf_qknorm_rope_kv_int8_kernel: same prefill block-table
+// layout (pos==tok, block_table[blk]), but writes bf16 K/V (no int8 quant/scale)
+// and rotates consecutive pairs (2i,2i+1) instead of NeoX (i,i+half) -- Muse
+// Glimmer is LLAMA_ROPE_TYPE_NORM in llama.cpp (see rope_kv_append_normal_kernel,
+// rope.cu). rotary_dim>0 => rope (SWA layers, rotary_dim=head_dim); rotary_dim==0
+// => NoPE (global layers, QK-norm then append as-is). Bit-for-bit the same KV a
+// forward_token decode step writes via launch_rmsnorm + launch_rope_kv_append_normal
+// / launch_kv_append, so a subsequent decode reads a consistent cache.
+// grid = (N, n_q_heads + 2*n_kv_heads); blockDim = head_dim.
+// ============================================================================
+__global__ void pf_qknorm_ropenorm_kv_bf16_kernel(
+    __nv_bfloat16* __restrict__ q, __nv_bfloat16* __restrict__ k, const __nv_bfloat16* __restrict__ v,
+    const __nv_bfloat16* __restrict__ q_w, const __nv_bfloat16* __restrict__ k_w,
+    __nv_bfloat16* __restrict__ k_pool, __nv_bfloat16* __restrict__ v_pool,
+    const int* __restrict__ block_table,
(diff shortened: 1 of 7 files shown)


skyrocket2026:

| metric | value |
|---|---|
| PR decode tok/s | 99.41 |
| main decode tok/s | 99.36 |
| decode speedup vs main | **1.00×** (+0.1%) |
| PR prefill@128 pp tok/s | 112.16 |
| main prefill@128 pp tok/s | 111.85 |
| prefill speedup vs main | **1.00×** (+0.3%) |
| Polaris receipt | `383491e87692368c…` — TDX (Intel hardware attestation) |
| commit | `16bd3193a` |

within significance gate — not a verified decode improvement

skyrocket2026:

This PR's Muse Glimmer 128-decode/prefill speed measured **0.1%** vs main, with no verified improvement on both decode and prefill — closing automatically. This bot evaluates every eligible PR in the repo against Muse Glimmer's decode AND prefill@128 speed specifically, regardless of what the PR is actually about — a close here isn't a judgment on the PR's purpose, just that it didn't move these particular metrics. Reopen (or open a fresh PR) if you have a fix or a different approach.

PR #787, opened 2026-08-12
perf(muse-glimmer): windowed batched-GEMM prefill — default on (10x prefill@128)

James-CUDA:
## Summary

Windowed **batched-GEMM prefill for Muse Glimmer** — directly moving the `prefill@128` score.
Muse was the one hybrid arch excluded from `prefill_batched_run` (`if (cfg.muse_glimmer) return
false`) because that path's full-attention kernels apply RoPE over the whole sequence with no
per-layer sliding-window/NoPE hook — so Muse prompts fell back to the sequential token loop (one
memory-bound GEMV per position, re-reading all ~17 GB of weights every token, flat ~112 pp). This
PR adds the missing muse-specific kernels so prefill runs as a single compute-bound pass:

- `pf_qknorm_ropenorm_kv_bf16_kernel` — bf16-KV QK-norm + NORMAL-RoPE (SWA layers) / NoPE (global
  layers) KV-write, all N tokens at once;
- `win_prefill_pure_bf16_kernel` — pure rolling-window bf16 attention (`win_blocks>0` = SWA,
  `≤0` = global full-causal), one kernel for both Muse layer types;
- sandwich-norm / embedding-norm / logit-softcap parity with the decode path.

**Default on** (numerically equivalent to the token loop; `SPARKINFER_MUSE_BATCHED=0` restores the
sequential path as an escape hatch / for A/B). Reaches only Muse Glimmer (`cfg.muse_glimmer`, hd128
dense, bf16 KV); every other model takes exactly the arm it took before. Decode is untouched.

## Proof of speedup

**Prefill pp tok/s** (Muse Glimmer `prefill@128`, `qwen3_gguf_bench <gguf> 8 128`, RTX 5090, clocks 2550):

| ctx | before (main, token loop) | after (this PR, default) | speedup |
|---|--:|--:|--:|
| **128** | 112.2 | 1123.0 | **10.0×** |
| 512 | 111.5 | 3253.0 | 29.2× |

**Decode tok/s** (Muse Glimmer, ctx=0 — unchanged; this PR targets prefill, no regression):

| | decode tok/s |
|---|--:|
| before (main) | 102.4 |
| after (this PR) | 103.1 |

## Correctness

Batched vs token-loop prefill (== the validated decode path), bf16 KV:

| test | batched | token-loop |
|---|--:|--:|
| ctx=128 (no clip) — PPL / argmax | 1.23191 / 267·271 | 1.23312 / 267·271 |
| ctx=3072 (SWA window clips, win=2048) — PPL / argmax | 1.28077 / 321·327 | 1.27704 / 321·327 |

0.1% PPL, identical argmax — the expected non-bit-identical GEMM-vs-GEMV reduction-order difference.
`compute-sanitizer memcheck`: **0 errors**. Qwen3.5/3.6 cross-model guard: **no regression** (same-box
A/B, decode+prefill within noise, bit-identical accuracy; the muse arms are gated behind
`cfg.muse_glimmer`).

```text
# qwen3_gguf_bench muse-glimmer-30B-kquant-17gb.gguf 8 128   (main -> this PR, RTX 5090, clocks 2550)
main (token loop)      : prefill pp   : 112.21 tok/s  (ctx=128)
this PR (default)      : prefill pp   : 1122.97 tok/s (ctx=128)      # SPARKINFER_MUSE_BATCHED=0 -> 115.86
# ctx=512: 111.49 -> 3252.95 tok/s ; decode tg unchanged: 102.4 -> 103.1 tok/s
```

### DIFF
--- a/kernels/csrc/cuda/fused/batched_prefill.cu
+++ b/kernels/csrc/cuda/fused/batched_prefill.cu
@@ -923,6 +923,84 @@ __global__ void pf_qknorm_rope_kv_int8_kernel(
 }
 
 // ============================================================================
+// Muse Glimmer full-attention prefill: QK-norm + NORMAL RoPE (consecutive-pair,
+// LLAMA_ROPE_TYPE_NORM) + BF16 KV write into the single-sequence paged pool. The
+// bf16 counterpart of pf_qknorm_rope_kv_int8_kernel: same prefill block-table
+// layout (pos==tok, block_table[blk]), but writes bf16 K/V (no int8 quant/scale)
+// and rotates consecutive pairs (2i,2i+1) instead of NeoX (i,i+half) -- Muse
+// Glimmer is LLAMA_ROPE_TYPE_NORM in llama.cpp (see rope_kv_append_normal_kernel,
+// rope.cu). rotary_dim>0 => rope (SWA layers, rotary_dim=head_dim); rotary_dim==0
+// => NoPE (global layers, QK-norm then append as-is). Bit-for-bit the same KV a
+// forward_token decode step writes via launch_rmsnorm + launch_rope_kv_append_normal
+// / launch_kv_append, so a subsequent decode reads a consistent cache.
+// grid = (N, n_q_heads + 2*n_kv_heads); blockDim = head_dim.
+// ============================================================================
+__global__ void pf_qknorm_ropenorm_kv_bf16_kernel(
+    __nv_bfloat16* __restrict__ q, __nv_bfloat16* __restrict__ k, const __nv_bfloat16* __restrict__ v,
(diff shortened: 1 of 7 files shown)


skyrocket2026:

| metric | value |
|---|---|
| PR decode tok/s | 102.86 |
| main decode tok/s | 102.85 |
| PR prefill@128 pp tok/s | 1107.71 |
| main prefill@128 pp tok/s | 116.23 |
| prefill speedup vs main | **9.53×** (+853.0%) |
| Polaris receipt | `451eaf11ef743088…` — TDX (Intel hardware attestation) |
| commit | `bd47f9310` |

ok

PR #774, opened 2026-08-11
perf(muse-glimmer): skip gated-off SwiGLU neurons' up-projection reads (contextual-sparsity FFN, 1.03x decode @128)

James-CUDA:
## Summary

Muse Glimmer's dense FFN is strongly **contextually sparse** per token: SwiGLU gates every
neuron by `silu(gate·x)`, and a neuron whose `|silu(gate·x)| ≈ 0` contributes ≈nothing to
`h = silu(gate·x)·up·x` (and thus to the down projection). `gate_up_mmvq2_qwen_sparse_kernel`
computes the **gate** projection first and, for a gated-off neuron (`|silu(gate·x)| < tau`),
writes `h = 0` and **skips that neuron's up-projection Q4_K weight read entirely** — reading
fewer weight bytes per token, which is the memory-bandwidth bottleneck dense 128-decode is
bound by (the gate/up GEMV is the largest kernel on the step).

The mask is **uniform across the CTA** (one CTA per neuron `f`), so there is no warp divergence,
and the launch grid is **unchanged** — the captured decode graph is unaffected. `tau` is read
once from the environment (default `0.06`); `SPARKINFER_MG_SPARSE_FFN=0` restores the dense
gate/up arm and `SPARKINFER_MG_SPARSE_TAU` overrides `tau`, both for a clean A/B on one binary.

Reaches only `hidden == 6656 && ffn == 19968 && top_k == 1` (Muse Glimmer's dense FFN); every
other model takes exactly the arm it took before. One file, no header/runtime changes.

## Proof of speedup

- [x] Tested on **RTX 5090** (`sm_120`)

**Decode tok/s** (`qwen3_gguf_bench <muse-glimmer-30B-kquant-17gb.gguf> 128 0` — the configuration
the Muse Glimmer bot scores, clocks locked 2550):

| | decode tok/s |
|---|--:|
| before (dense, `SPARKINFER_MG_SPARSE_FFN=0`) | 96.60 |
| after (this PR, `tau=0.06`) | 99.76 |

`+3.27%`. `SPARKINFER_MG_SPARSE_FFN=0` reproduces the dense arm on the same binary, so this
before/after is a clean A/B, not a rebuild delta.

## Correctness

Accuracy gated against a live **llama.cpp** reference on the same GGUF, teacher-forced over the
`bench/scripts/eval_text.txt` corpus (llama-tokenize, top-128), exactly as the bot scores it:

| | top1 vs llama | KL(llama‖spark) |
|---|--:|--:|
[...]

PR #788, opened 2026-08-12
perf(muse-glimmer): skip zeroed SwiGLU neurons' down-projection reads (1.04x decode @128)

petermilord:
## Summary
#774 already skips gated-off neurons' **up** Q4_K reads and writes `h = 0`. Those zeros
quantize to Q8_1 with `d == 0`, but Muse's shape-specialized Q4_K **down** still read every
weight block. This extends the same mask into the down GEMV (`F=19968`, `SKIP_ZERO_HQ8`):
if both activation scales for a vec-dot are 0, skip the weight load. Grid unchanged
(CUDA-graph safe). Other models untouched.

`SPARKINFER_MG_SPARSE_DOWN=0` keeps sparse gate/up (#774) but restores dense down for A/B.
`SPARKINFER_MG_SPARSE_FFN=0` disables both halves.

One file: `kernels/csrc/cuda/moe/expert_ffn_q4k.cu`. No L2 prefetch / QK-fuse / flash-decode
changes (#782 already landed that work).

| | decode tok/s |
|---|--:|
| before (main) | 99.56 |
| after (this PR) | 104.4 |

A/B on one binary: `SPARKINFER_MG_SPARSE_DOWN=0` (before) vs unset (after),
`qwen3_gguf_bench <muse-glimmer-….gguf> 128 0`, clocks locked 2550.

## Checks
- Muse accuracy: top1 ≥ 0.90, KL ≤ 0.10 vs live llama.cpp
- Qwen3.6 guard: decode+prefill ≥ 98% of main (shared MoE TU — verify)
- `SPARKINFER_MG_SPARSE_DOWN=0` matches main's dense down

### DIFF
--- a/kernels/csrc/cuda/moe/expert_ffn_q4k.cu
+++ b/kernels/csrc/cuda/moe/expert_ffn_q4k.cu
@@ -1126,7 +1126,7 @@ __global__ void down_q6k_mmvq_splitk_qwen_kernel(
     }
 }
 
-template <int S, int NBLK, int TOPK>
+template <int S, int NBLK, int TOPK, bool SKIP_ZERO_HQ8 = false>
 __global__ void down_q4k_mmvq_splitk_qwen_kernel(
     const unsigned char* __restrict__ down_q, const int* __restrict__ expert_ids,
     const float* __restrict__ expert_weights, const si_block_q8_1* __restrict__ hq8,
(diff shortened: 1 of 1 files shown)


skyrocket2026:
## 🚩 Flagged: eval-gaming

This PR involves an account blocked for gaming the SN74 emission mechanism (sybil / coordinated duplicate farming): `petermilord`.

PR #790, opened 2026-08-12
perf(muse-glimmer): admit Muse's weight widths to the vector-store i8 dequant (1.27x prefill@128)

FranDev132:
## Summary

The batched prefill converts every Q4_K/Q6_K weight row to int8 before the i8 GEMM, and
`launch_gguf_dequant_rows_i8_fast` admitted only 1024-aligned rows:

```cpp
if (!enabled || rows <= 0 || cols <= 0 || (cols % (DQR_BLOCK * DQR_VEC)) != 0) return false;
```

Muse Glimmer is **6656** wide (attn q/gate/k/v, ffn gate/up) and **19968** wide (ffn_down). Both
are super-block aligned but leave 512 modulo 1024, so every one of them missed the vector-store
kernel and fell back to the byte-store two-pass kernel in `dequant_gguf.cu`. Only `attn_output`
(4096) qualified. In an nsys graph-node trace of a 128-token prefill that made the conversion cost
**more than the GEMM it feeds**:

| ms | % of prefill | kernel |
|--:|--:|---|
| 57.79 | **54** | `deq_rows_i8_twopass<12>` (Q4_K -> int8) |
| 42.44 | 40 | `pf_gemm_i8_kernel` |
| 2.38 | 2 | `win_prefill_pure_bf16` |

It ran at ~640 GB/s where the same fast kernel reaches ~1350 GB/s on the one width it did accept.

The 1024 requirement was a throughput guard, not a correctness one — every group in the kernel is
already guarded by `base < cols`, and a thread's 4 values are 4-aligned so they always land inside
one super-block. All the kernel actually needs is `cols % 256 == 0` so `nsb = cols>>8` is exact.
Admit that, restricted to widths with **at least one full group** so a row is never dominated by
its partial tail; everything below 1024 keeps exactly the path it takes today.

`ffn_down` needed a second fix. At 19968 it wants `NG=20`, i.e. `v[20][4]` = 80 floats/thread, and
that register pressure cost more occupancy than the extra threads were worth — 776 GB/s against
1153 for the same-traffic `ffn_gate`/`up`. The block width is now a template parameter, so a
1024-thread block covers that row in `NG=5`. The value->thread mapping is unchanged, so it is
purely an occupancy trade.

## Proof of speedup

- [x] Tested on **RTX 5090** (`sm_120`)

**Decode tok/s** (128-token decode, ctx=0):

| | decode tok/s |
|---|--:|
| before (main) | 102.93 |
| after (this PR) | 102.92 |

[...]

PR #793, opened 2026-08-12
perf(muse-glimmer): fused Q4_K-direct dense prefill GEMM — skip the int8 materialize (1.22x prefill@128)

James-CUDA:
## Summary

Fused **Q4_K-direct** dense prefill GEMM for Muse Glimmer — moves `prefill@128` by decoding each
native Q4_K weight to int8 **inside** the GEMM's B-stage, so the per-layer int8 *materialize*
(dequant → write `W_i8` → read `W_i8` back) never happens. At prefill's M=128 that materialize
round-trip is the dominant weight-bandwidth cost and pure overhead — the 128×128 GEMM tile is
compute-bound whatever M is — so removing it is the win. This **composes with** #790 (which speeds
up the *materialize's* dequant): the Q4_K attn + FFN gate/up weights skip the materialize entirely
here, while the Q6_K `ffn_down` this PR still materializes rides #790's faster dequant.

The mechanism already exists in-tree for the **routed MoE** path (`pfm_moe_gemm_qi8_kernel`,
`prefill_moe_q.cu`, fed by the eager `moe_rs_*` scales); this PR brings it to the **dense** path
(Muse's attn q/gate/k/v/o + FFN gate/up, all Q4_K) which had no equivalent and still materialized.
`pf_dense_gemm_qi8_kernel` strips the expert/pair routing from that kernel and reuses its
`qm_decode_j64` B-stage verbatim. The per-output-row int8 scale is precomputed once at load with the
**same** kernel the materialize path uses (`launch_gguf_dequant_rows_i8`, keeping only its `scale`),
so the fused GEMM emits the identical int8 bytes — and, int32 accumulation being order-independent,
the identical result — by construction.

**Default on**; `SPARKINFER_MUSE_PREFILL_QB=0` restores the materialize path (A/B). Reaches **only**
Muse Glimmer (`c.muse_glimmer`); every other model's `proj()` is byte-for-byte unchanged, and decode
is untouched.

## Proof of speedup

**Prefill pp tok/s** (Muse Glimmer `prefill@128`, `qwen3_gguf_bench <gguf> 8 128`, RTX 5090, medians of 5):

| | prefill pp tok/s |
|---|--:|
| before prefill (main, incl. #790) | 1263.6 |
| after prefill (this PR, fused Q4_K) | 1540.5 |

**+22.0% (1.22×)** at the scored context, on top of the already-merged #790 dequant-width path.

**Decode tok/s** (Muse Glimmer, ctx=0 — unchanged; this PR targets prefill only):

| | decode tok/s |
|---|--:|
| before (main) | 100.71 |
| after (this PR) | 100.71 |

```text
# qwen3_gguf_bench muse-glimmer-30B-kquant-17gb.gguf 8 128   (RTX 5090, sm_120, on current main b13d69e)
main (incl #790, materialize) : prefill pp : 1263.6 1259.9 1264.7 1262.3 1264.9  (median 1263.6)
this PR (fused, default)      : prefill pp : 1540.0 1537.9 1547.0 1540.5 1546.4  (median 1540.5)
# SPARKINFER_MUSE_PREFILL_QB=0 reproduces the main number; decode tg ctx=0 unchanged 100.71 -> 100.71
```

## Correctness

Fused (`SPARKINFER_MUSE_PREFILL_QB=1`, default) vs materialize (`=0`), `qwen3_gguf_prefill_check`
(batched vs the validated token-loop), bf16 KV — **identical at every context**:

| ctx | materialize (QB=0) — TOP1 / KL | fused (QB=1) — TOP1 / KL |
|---|--:|--:|
| 128  | 16/16 / 0.05244 | 16/16 / 0.05244 |
| 512  | 16/16 / 0.00987 | 16/16 / 0.00987 |
| 3072 (SWA window clips) | 15/16 / 0.00755 | 15/16 / 0.00755 |

Bit-for-bit equal to the materialize path — the fused decode uses the same per-row scale and int8
bytes, and int32 tensor-core accumulation is order-independent. `compute-sanitizer memcheck`:
**0 errors**. `ctest`: **13/13 passed**. Qwen3.6 / Qwythos: the change is gated behind
`c.muse_glimmer`, so their `proj()` path is unchanged (no-regression guard unaffected).

### DIFF
--- a/kernels/csrc/cuda/fused/prefill_moe_q.cu
+++ b/kernels/csrc/cuda/fused/prefill_moe_q.cu
@@ -458,6 +458,131 @@ void dispatch_qi8_bm16(const signed char* A_i8, const float* sx, const void* W_q
             A_i8, sx, W, row_scale, pair_tok, pair_w, offsets, tilemap, d_ntiles, C, out_f32, n_out, K);
 }
 
+// ---------------------------------------------------------------------------
+// Dense (non-routed) fused-decode int8 GEMM: C[M,N] = A_i8[M,K] @ dequant(W_q[N,K])^T.
+// The same qm_decode_j64 B-stage + int8 WMMA as pfm_moe_gemm_qi8_kernel, with the expert /
+// pair-routing indirection stripped away: one dense weight [N,K], M contiguous activation rows,
+// C written directly (no gather, no scatter). Muse Glimmer's dense prefill uses this to consume
+// its native Q4_K/Q5_K attn + FFN gate/up weights straight from VRAM, skipping the per-layer int8
+// materialize (dequant -> write W_i8 -> read W_i8 back in the GEMM) that launch_prefill_gemm_i8
+// pays. The per-output-row scale is the one launch_gguf_dequant_rows_i8 itself produced,
+// precomputed once at load (Qwen35LayerWeights::*_rs), so every int8 byte -- and thus the whole
+// accumulation -- matches the materialize path by construction. BM=128 so each weight row is
+// decoded once per M-tile: at prefill's M=128 that is exactly once (one M-tile).
+template <int QT>
+__global__ __launch_bounds__(256, 3) void pf_dense_gemm_qi8_kernel(
+        const signed char* __restrict__ A_i8, const float* __restrict__ sx,
+        const unsigned char* __restrict__ W_q, const float* __restrict__ row_scale,
+        __nv_bfloat16* __restrict__ C, int Mtot, int N, int K) {
+    using namespace nvcuda;
+    constexpr int BS = qm_bs<QT>();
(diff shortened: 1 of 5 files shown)


skyrocket2026:

| metric | value |
|---|---|
| PR decode tok/s | 102.80 |
| main decode tok/s | 102.78 |
| PR prefill@128 pp tok/s | 1709.26 |
| main prefill@128 pp tok/s | 1428.55 |
| prefill speedup vs main | **1.20×** (+19.6%) |
| Polaris receipt | `2f5ed414f04c5194…` — TDX (Intel hardware attestation) |
| commit | `a0a3078de` |

ok

skyrocket2026:

PR #796, opened 2026-08-13
perf(muse-glimmer): split-K the fused Q4_K prefill GEMM, and fuse the Q4_K ffn_down (1.29x prefill@128)

Paral1995:
## Summary

At prefill's `M=128` the fused quantized-B dense GEMM's `grid.y` collapses to 1, so its grid is only
`n_out/64` blocks — 104 for the 6656-wide o/down, 312 for the 19968-wide FFN pair — against ~510
block slots on a 5090. `ncu` on it: **0.13 waves/SM**, 16.7% achieved occupancy, DRAM 6.0%, SM
throughput 13.1%. It is neither bandwidth- nor compute-bound; it simply is not on the machine.

`blockIdx.z` now owns a slice of the K super-blocks and writes its raw int32 tile to a partials
buffer; a reduce pass sums the slices and applies `sx * row_scale` once. int32 accumulation is exact
and associative, so the summed result is **bit-identical** to the unsplit launch — only the block
count changes.

The split count is derived from the slice size, so every launched slice owns at least one
super-block. Deriving it the other way leaves trailing slices that return before writing, and the
reduce then sums uninitialised memory — that produced correct-looking throughput with TOP1 0.92 →
0.67. Only the ungrouped launches split; a grouped launch resolves a different `n_out`/`C` per
output tile and would need a per-group partials offset table.

Separately, `ffn_down` was reserved no slot in the dense row-scale pool, on the assumption that down
is always Q6_K. In this GGUF a large share of the down tensors are Q4_K — fusable — and were falling
back to the int8 materialize purely for lack of a scale. Adding the slot lets them take the fused
path; a genuinely Q6_K down still materializes.

`SPARKINFER_MUSE_QB_SPLITK=0` restores the unsplit launch, `>0` pins the slice count.

## Proof of speedup

**Decode tok/s** (ctx=0, unchanged — this PR does not touch the decode path):

| | decode tok/s |
|---|--:|
| before (main) | 100.57 |
| after (this PR) | 100.57 |

**Prefill pp tok/s** (Muse Glimmer, ctx=128 — the context this PR targets):

| | prefill pp tok/s |
|---|--:|
| before prefill (main) | 2073.69 |
| after prefill (this PR) | 2666.93 |

**+28.61%** prefill@128. Median-of-5, same binary, toggled with `SPARKINFER_MUSE_QB_SPLITK`, clocks
locked at 2550 MHz.

```text
### prefill@128 median-of-5, same binary ###
  split off (== main behaviour) : 2073.69
  split on  (this PR)           : 2666.93
  DELTA = +28.61%

### decode@ctx0 ###
  rep1 decode=100.57
  rep2 decode=100.57
  rep3 decode=100.56

### slice-count sweep (prefill@128) ###
  off 2077 | 4 -> 2557 | 6 -> 2566 | 8 -> 2586 | auto 2681
```

### Correctness

Bit-identity is the claim, so it is checked rather than argued — `qwen3_gguf_prefill_check` fills the
KV both ways over the same prompt and teacher-forces the same continuation (64 positions, bf16 KV):

```text
  [split off ] prefix=128  59/64 0.9219  KL 0.03558
  [split auto] prefix=128  59/64 0.9219  KL 0.03558
  [split =2  ] prefix=128  59/64 0.9219  KL 0.03558
  [split =8  ] prefix=128  59/64 0.9219  KL 0.03558
  [split on  ] prefix=512  60/64 0.9375  KL 0.00906

ctest: 100% tests passed, 11/11
```

Every slice count agrees to the digit with the unsplit launch, as the int32 argument requires.

### Scope

Reachable only from Muse Glimmer's batched prefill (`c.muse_glimmer`); the partials buffer is
allocated only there, and only while the prompt is a single M-tile. The routed-MoE kernels in the
same file are untouched, and `launch_prefill_gemm_qi8_dense` is not on any other model's path.

### Qwen3.6 no-regression

This edits `prefill_moe_q.cu`, which the routed-MoE prefill shares, so the cross-model guard was run
rather than argued — same sweep the eval uses (`ctx 0/512/4k/16k/32k`, median-of-5), two arms each
built and run in place on the same box:

```text
       0     481.20 ->   481.17        0.00 ->      0.00
     512     475.64 ->   475.51     8066.82 ->   8062.83
    4096     455.32 ->   455.29    22590.14 ->  22632.83
   16384     457.32 ->   457.37    24804.65 ->  24799.50
   32768     457.64 ->   457.51    26170.70 ->  26165.89
```

Max deviation ~0.03%, i.e. run-to-run noise. Expected: the routed-MoE kernels are untouched, and the
dense fused GEMM is gated behind `c.muse_glimmer`, so Qwen3.6 never reaches it.

### DIFF
--- a/kernels/csrc/cuda/fused/prefill_moe_q.cu
+++ b/kernels/csrc/cuda/fused/prefill_moe_q.cu
@@ -45,6 +45,7 @@
 #include <cuda_runtime.h>
 #include <cuda_fp16.h>
 #include <cuda_bf16.h>
+#include <cstdlib>
 #include <cuda_pipeline.h>
 #include <mma.h>
 
@@ -56,6 +57,10 @@ namespace {
 enum { QMQ_Q4_K = 12, QMQ_Q5_K = 13 };
 
 constexpr int PF_QGROUP_MAX = 4;   // projections fusable into one grouped launch
+// Block-slot budget the split-K fan-out aims at: ~3 passes over a 5090's 170 SMs x 3 blocks/SM.
+// Measured: aiming at ONE pass picks 2 splits for the 312-tile FFN and loses to 4 (2211 vs 2550
+// pp) -- oversubscribing hides latency and shortens the tail. A target, never a cap.
+constexpr int QM_TARGET_BLOCKS = 1536;
 constexpr int QM_BM = 128;    // pair rows per tile (must match the caller's tilemap)
 constexpr int QM_BN = 64;     // weight rows (output channels) per block
 constexpr int QM_SB = 256;    // values per GGUF super-block == the B-stage K depth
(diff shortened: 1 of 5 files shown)


skyrocket2026:

| metric | value |
|---|---|
| PR decode tok/s | 102.81 |
| main decode tok/s | 102.81 |
| PR prefill@128 pp tok/s | 3036.06 |
| main prefill@128 pp tok/s | 2232.66 |
| prefill speedup vs main | **1.36×** (+36.0%) |
| Polaris receipt | `aadbb7b51e2f80d3…` — TDX (Intel hardware attestation) |
| commit | `63d40ef64` |

ok

skyrocket2026:

PR #799, opened 2026-08-13
perf(muse-glimmer): atomic split-K accumulation and a table-driven Q4_K decode

widecloud:
## Summary

Two changes to the fused Q4_K prefill GEMM (`pf_dense_gemm_qi8_kernel_g`), which nsys puts at 80% of Muse Glimmer's 35.9 ms prefill@128. Both are bit-identical to main by construction.

`launch_prefill_gemm_qi8_dense{,_group}` are reached only from `c.muse_glimmer` call sites, so (1) is Muse-only. (2) lives in the shared `qm_decode_j64`, so the routed-MoE prefill gets it too — the Qwen3.6 guard below shows that as a gain, not a risk.

## Proof of speedup

- [x] Tested on **RTX 5090** (`sm_120`)

Real `origin/main` build in a separate tree, alternated with the PR tree, default env, 5 reps (`SPARKINFER_BENCH_SWEEP_CTXS=0,128 SPARKINFER_BENCH_SWEEP_REPS=5`), Muse Glimmer `muse-glimmer-30B-kquant-17gb.gguf`.

**Decode tok/s** (Muse Glimmer, ctx 0 — unchanged, this PR targets prefill):

| | decode tok/s |
|---|--:|
| before (main `7fb6a20`) | 103.58 |
| after (this PR) | 103.57 |

**Prefill pp tok/s** (Muse Glimmer, ctx 128 — the scored dimension):

| | prefill pp tok/s |
|---|--:|
| before prefill (main `7fb6a20`) | 3559.34 |
| after prefill (this PR) | **3723.58 (+4.61%)** |

```text
--- Muse Glimmer, 5-rep sweep, alternating trees, default env ---
== MAIN :: 7fb6a20 fix(runtime): stop leaking hybrid recurrent state on every session close (#779)
SWEEP_JSON {"0":{"decode_tps":103.5897,"prefill_pp":0.0000},"128":{"decode_tps":102.6896,"prefill_pp":3560.4671}}
== PR
SWEEP_JSON {"0":{"decode_tps":103.5498,"prefill_pp":0.0000},"128":{"decode_tps":102.6515,"prefill_pp":3740.7875}}
== MAIN
SWEEP_JSON {"0":{"decode_tps":103.5736,"prefill_pp":0.0000},"128":{"decode_tps":102.6716,"prefill_pp":3558.2218}}
== PR
[...]

PR #798, opened 2026-08-13
perf(muse-glimmer): compress dense FFN gate/up with accuracy-safe Q3_A

andriypolanski:
## Summary

Reintroduces the calibration-selected Muse Glimmer Q3_A dense-FFN gate/up decode path with a fixed-shape single-token kernel and repairs batched prefill. Selected decode weights use compact Q3_A while batched prefill retains and reads the native GGUF weights, preventing the illegal CUDA access that invalidated PR #785 prefill score.

On RTX 5090 against current main (`7fb6a20`), median-of-5 short decode improves from 104.38 to 110.78 tok/s (+6.13%, M). At context 128, decode improves from 103.45 to 109.71 tok/s (+6.05%) while prefill remains flat at 3066.81 to 3084.73 pp tok/s (+0.58%).

## Proof of speedup

- [x] Tested on **RTX 5090** (`sm_120`)

**Decode tok/s** (128-token autoregressive decode, median of five alternating same-binary runs):

| | decode tok/s |
|---|--:|
| before (current main `7fb6a20`) | 104.38 |
| after (this PR) | 110.78 |

```text
run      before    after
1        104.38    110.80
2        104.37    110.78
3        104.42    110.78
4        104.39    110.73
5        104.36    110.75
median   104.38    110.78 tok/s
delta              +6.13%
```

**Prefill@128 pp tok/s** (median of five alternating same-binary runs):

| | prefill pp tok/s |
|---|--:|
| before (current main; native Q4_K) | 3066.81 |
| after (this PR; native prefill weights retained) | 3084.73 |

```text
run      before     after
1        3066.81    3085.94
2        3060.54    3084.25
3        3067.46    3084.73
4        3066.22    3086.52
5        3072.02    3083.69
median   3066.81    3084.73 pp tok/s
delta               +0.58%
```

No run reported a CUDA error. The invalid approximately 3.4k pp tok/s result from #785 is not used or claimed.

**Additional guarded-context check** (single isolated process per build):

| context | metric | before | after | delta |
|---:|---|---:|---:|---:|
| 128 | decode tok/s | 103.45 | 109.71 | +6.05% |
| 512 | decode tok/s | 102.38 | 108.40 | +5.88% |
| 512 | prefill pp tok/s | 3719.83 | 3690.44 | -0.79% |

## Design

- Forty-two calibrated layers use internal Q3_A gate/up weights for decode; ten sensitive layers remain native Q4_K.
[...]

PR #789, opened 2026-08-12
perf(muse-glimmer): split-K the skinny prefill projections and stop re-allocating the scratch (1.17x prefill@128)

widecloud:
## Summary

Muse Glimmer's batched prefill spends a large share of its GEMM time in launches whose grid is too small to fill the GPU, and rebuilds its entire scratch set on every call. Two independent fixes, both on by default with an env off-switch, both scoped so no other model's numbers move.

`SPARKINFER_PREFILL_GEMM_SPLITK=0` and `SPARKINFER_PREFILL_ARENA_REUSE=0` restore the previous behaviour, so both arms come out of one binary. Split-K is additionally gated on `cfg.muse_glimmer`, so no other model reaches the new kernel at all.

## Proof of speedup

- [x] Tested on **RTX 5090** (`sm_120`)

**Decode tok/s** (Muse Glimmer, ctx 0 — this PR does not touch decode):

| | decode tok/s |
|---|--:|
| before (main `9d68aed`) | 102.96 |
| after (this PR) | 102.91 |

**Prefill pp tok/s** (Muse Glimmer, ctx 128 — the scored context):

| | prefill pp tok/s |
|---|--:|
| before prefill (main `9d68aed`) | 1127.13 |
| after prefill (this PR) | **1321.78** |

**+17.3%** at `SPARKINFER_BENCH_SWEEP_REPS=1`, **+19.6%** at 3 reps. The single-rep figure is the conservative one and is what the harness measures: the first prefill in a process still pays the one-time arena allocation, and the sweep benches ctx=0 first, which does no prefill.

```text
# main 9d68aed
reps=1  SWEEP_JSON {"0":{"decode_tps":102.9636,...},"128":{"decode_tps":102.1494,"prefill_pp":1127.1260}}
reps=3  SWEEP_JSON {"0":{"decode_tps":102.9113,...},"128":{"decode_tps":102.0731,"prefill_pp":1125.7394}}

# this PR
reps=1  SWEEP_JSON {"0":{"decode_tps":102.9106,...},"128":{"decode_tps":102.0194,"prefill_pp":1321.7827}}
reps=3  SWEEP_JSON {"0":{"decode_tps":102.8381,...},"128":{"decode_tps":101.9451,"prefill_pp":1345.7852}}
```

Not an artifact of that env knob: with the Q3_A prefill path repaired locally, the same two changes
measure 1096.29 -> 1310.01 pp (+19.5%, 3 reps) on the same commit at default settings.

[...]

PR #800, opened 2026-08-13
perf(muse-glimmer): CUDA-graph capture for the batched prefill (1.07x prefill@128)

JSONbored:
## Summary

Muse Glimmer's **batched prefill is the only hot path in this model that is not CUDA-graph captured**. Decode gets a graph (`cu_graph`/`cu_exec`); `prefill_batched_run` issues its work eagerly — ~1088 kernel launches per prefill@128, 20.9 per layer across 52 layers, with host dispatch between every one.

This captures that pass into a graph and replays it. **Bit-identical by construction**: the graph records the same kernels, in the same order, against the same pointers — only the dispatch mechanism changes.

## Where the 1088 launches go, and what they cost

`nsys profile --trace=cuda --cuda-graph-trace=node`, then `cuda_gpu_trace`, on a real RTX 5090 at `origin/main` `7fb6a20`. The prefill window is delimited by `pf_dense_gemm_qi8_kernel_g`, which runs **only** during prefill (decode uses `si_mmvq_*`), so it separates cleanly from the 128 decode steps in the same sweep:

```
prefill window : 38.18 ms wall over 1088 kernel launches (20.9 per layer)
GPU busy       : 36.17 ms   (union of kernel intervals)
GPU IDLE       :  2.01 ms   = 5.3% of the window, 1.9 us mean per launch
```

The idle is **launch overhead, not dependency stalls** — 1076 of the 1087 gaps are under 5 us, and only 8 exceed 10 us (0.24 ms total):

```
gaps < 5us  : 1.75 ms across 1076 gaps
gaps > 10us : 0.24 ms across    8 gaps
```

Per-kernel breakdown of the same window, for context on why nothing else here is the target:

| ms | calls | /layer | kernel |
|--:|--:|--:|---|
| 28.67 | 208 | 4.0 | `pf_dense_gemm_qi8_kernel_g` |
| 2.50 | 52 | 1.0 | `win_prefill_pure_bf16_kernel` |
| 2.04 | 104 | 2.0 | `pf_dense_splitk_reduce_group_kernel` |
| 0.65 | 103 | 2.0 | `pf_dense_splitk_reduce_kernel` |
| 1.10 | ~440 | 8.5 | norms, RoPE, SwiGLU, quant, memset |

## Proof of speedup

Separate `origin/main` build in its own tree, **shipped defaults — no environment variables set for either side**, `SPARKINFER_BENCH_SWEEP_CTXS=0,128`, `SPARKINFER_BENCH_SWEEP_REPS=5`, `muse-glimmer-30B-kquant-17gb.gguf`.

`main` is **interleaved between every PR sample** rather than bracketing the run: this box drifts 3.6–4.5% *within* a single session, which is larger than the effect, so each PR sample is compared against the mean of the two `main` samples immediately around it.

| | prefill pp tok/s |
|---|--:|
| before (main `7fb6a20`) | 3214.1 / 3188.0 / 3169.4 / 3171.1 / 3171.7 |
| after (this PR) | 3316.3 / 3287.4 / 3272.3 / 3272.3 |

```text
main_1 3214.1 | pr_1 3316.3 | main_2 3188.0 | pr_2 3287.4 | main_3 3169.4
pr_3   3272.3 | main_4 3171.1 | pr_4 3272.3 | main_5 3171.7

pairwise vs bracketing main: +3.60%, +3.42%, +3.22%, +3.18%
MEAN +3.35%   MEDIAN +3.32%   n=4
```

**Decode is unchanged** (this PR touches only the batched prefill path): 102.65 → 102.07 across the run on the PR side, tracking `main`'s own drift over the same samples.

## Correctness

Bit-identical, not merely within tolerance — the graph replays the same kernels in the same order:

```text
qwen3_gguf_score <gguf> 5 <64 ids>, PR vs main:   IDENTICAL (byte-for-byte)

SPARKINFER_KV_INT8=0 qwen3_gguf_prefill_check <gguf> 128 16
  TOP1 16/16 1.0000 | KL 0.05244
SPARKINFER_KV_INT8=0 qwen3_gguf_prefill_check <gguf> 512 16
  TOP1 16/16 1.0000 | KL 0.00987
```

## Why capture is legal here — three preconditions, verified not assumed

1. **No mid-body stream syncs on Muse's path.** The four `cudaStreamSynchronize` calls inside `prefill_batched_run` are all on the routed-MoE branch (`moe serial sync`, `moe sa join`, `moe sk/sv sync`); Muse Glimmer's FFN is dense, so none are reachable.
2. **The arena must be warm.** `Arena::alloc` reuses a buffer whenever it is already big enough, so the **second** call for a given `N` performs no `cudaMalloc` and hands out exactly the pointers the capture recorded. Hence capture on the second sighting of an `N`, replay thereafter.
3. **That same rule makes the one real early return inside the captured region (`!kv8`) unreachable at capture time**, because `g_pfb_warm_n == N` is only set after a call with that `N` has already completed successfully.

Prompt ids are staged through a **pinned** buffer so the captured H2D has a stable source address and replay picks up new ids. Capture closes *before* the seed D2H + sync (a synchronize cannot be recorded), and the graph is launched immediately after `cudaGraphInstantiate` because capturing records without executing. A failed capture or instantiate returns `-1` so the caller falls back to the token loop rather than returning a seed from uninitialised memory. A changed `N` destroys and re-captures.

`SPARKINFER_MUSE_PREFILL_GRAPH=0` restores eager dispatch for A/B.

## Scope and prior art

The diff is **76 lines in one file** and is gated on `c.muse_glimmer`, so no other model's path changes.

- Builds directly on **#787** (windowed batched-GEMM prefill), which created the `prefill_batched_run` path this captures, and on **#789/#791/#793/#795/#796/#797**, which shaped the kernel sequence inside it. The launch *count* those PRs produced is exactly what makes graph capture worth doing.
- **Deliberately disjoint from the open #799** (atomic split-K accumulation + table-driven Q4_K decode) and **#798** (Q3_A gate/up compression). This PR touches neither the decode path nor the split-K accumulation, so it should compose with either.
- I also measured atomic split-K independently on top of this change and it was **not** a win here (+3.21% vs +3.67% for the default in the same session), and re-swept split-K slice counts with the graph active — the shipped heuristic remains optimal. Both are reported for #799's benefit rather than claimed.

### DIFF
--- a/runtime/src/models/qwen35_prefill.cpp
+++ b/runtime/src/models/qwen35_prefill.cpp
@@ -158,6 +158,43 @@ int prefill_batched_run(const Qwen35PrefillCtx& s, const int* prompt_ids, int n)
     Arena once_a, once_a8, once_am, once_aw;                       // per-call otherwise
     if (arena_reuse) { keep_a.rewind(); keep_a8.rewind(); keep_am.rewind(); keep_aw.rewind(); }
     Arena& a = arena_reuse ? keep_a : once_a;
+
+    // ---- CUDA-graph replay of the whole batched prefill -------------------------------------
+    // Muse's batched prefill is the only hot path here that is NOT graph-captured: decode gets a
+    // graph, this issues ~1088 kernel launches per prefill@128 (20.9 per layer) with host dispatch
+    // between every one. nsys puts the resulting GPU idle at 2.01 ms of a 38.18 ms window, and
+    // 1076 of those 1087 gaps are under 5 us -- launch overhead, not dependency stalls.
+    //
+    // Capture is only legal once the arena is WARM: Arena::alloc reuses a buffer whenever it is
+    // already big enough, so the second call for a given N does no cudaMalloc and hands out the
+    // SAME pointers the capture recorded. Hence capture on the second sighting of an N, replay
+    // after that. The prompt ids are staged through a PINNED buffer so the captured H2D has a
+    // stable source address and replay picks up new ids.
+    static cudaGraph_t     g_pfb_graph = nullptr;
+    static cudaGraphExec_t g_pfb_exec  = nullptr;
+    static int  g_pfb_n = -1, g_pfb_warm_n = -1, g_pfb_pin_cap = 0;
+    static int* g_pfb_pin = nullptr;
+    // DEFAULT ON: measured +5.91% on prefill@128, byte-identical output (SCORE_EQ IDENTICAL,
+    // TOP1 16/16). SPARKINFER_MUSE_PREFILL_GRAPH=0 restores eager dispatch.
+    static const bool graph_env = [] {
+        const char* e = getenv("SPARKINFER_MUSE_PREFILL_GRAPH");
+        return !(e && e[0] == '0');
+    }();
+    const bool graph_on = graph_env && arena_reuse && c.muse_glimmer;
+    if (graph_on && N > g_pfb_pin_cap) {
+        if (g_pfb_pin) cudaFreeHost(g_pfb_pin);
+        g_pfb_pin = nullptr; g_pfb_pin_cap = 0;
+        if (cudaMallocHost(&g_pfb_pin, (size_t)N * sizeof(int)) == cudaSuccess) g_pfb_pin_cap = N;
+    }
+    const bool graph_ok = graph_on && g_pfb_pin && g_pfb_pin_cap >= N;
+    if (graph_ok) memcpy(g_pfb_pin, prompt_ids, (size_t)N * sizeof(int));
+    if (graph_ok && g_pfb_exec && g_pfb_n == N) {
+        pf_cu(cudaGraphLaunch(g_pfb_exec, st), "pfb graph launch");
+        pf_cu(cudaMemcpyAsync(s.h_out_id, s.d_out_id, sizeof(int), cudaMemcpyDeviceToHost, st),
+              "pfb graph seed");
+        pf_cu(cudaStreamSynchronize(st), "pfb graph sync");
+        return *s.h_out_id;
+    }
     bf16* x    = a.alloc<bf16>((size_t)N * H);
     bf16* xn   = a.alloc<bf16>((size_t)N * H);
     bf16* hn   = a.alloc<bf16>((size_t)N * H);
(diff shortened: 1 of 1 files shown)


skyrocket2026:

| metric | value |
|---|---|
| PR decode tok/s | 101.00 |
| main decode tok/s | 100.71 |
| PR prefill@128 pp tok/s | 3172.64 |
| main prefill@128 pp tok/s | 3072.90 |
| prefill speedup vs main | **1.03×** (+3.2%) |
| Polaris receipt | `02c7aecec64ef414…` — TDX (Intel hardware attestation) |
| commit | `59d5e8c8d` |

ok

skyrocket2026:

| metric | value |
|---|---|
| PR decode tok/s | 104.35 |
| main decode tok/s | 102.91 |
| decode speedup vs main | **1.01×** (+1.4%) |
| PR prefill@128 pp tok/s | 3763.92 |
| main prefill@128 pp tok/s | 3612.86 |
| prefill speedup vs main | **1.04×** (+4.2%) |
| Polaris receipt | `25f6be4463d70fa8…` — TDX (Intel hardware attestation) |
| commit | `1f3120471` |

ok

JSONbored:
Rebased onto `9dd350e` (post-#805). Re-validated from the pushed commit (`7e81977`) on an RTX 5090, two independent clones/build dirs, main re-sampled between every PR sample:

| | prefill@128 pp |
|---|--:|
| main `9dd350e` | 4408.2, 4365.5, 4317.6, 4324.8, 4295.7, 4305.5 |
| this PR | 4672.6, 4642.7, 4606.4, 4613.9, 4550.7 |

Per-pair vs bracketing main samples: **+6.51% +6.94% +6.60% +7.04% +5.82% → mean +6.58%, median +6.60%** (n=5).

Graph capture reclaims a roughly constant ~1.3 ms of launch gaps, so each prefill speedup that merges under this PR raises its relative value — #805 (+13%) moved it from the +4.2% last scored. `qwen3_gguf_score` output remains byte-identical to main (same kernels, same order — only the dispatch mechanism changes), decode@128 flat (105.03 → 105.02).

skyrocket2026:

| metric | value |
|---|---|
| PR decode tok/s | 102.85 |
| main decode tok/s | 102.60 |
| PR prefill@128 pp tok/s | 4268.83 |
| main prefill@128 pp tok/s | 4099.91 |
| prefill speedup vs main | **1.04×** (+4.1%) |
| Polaris receipt | `68df996bce8ffada…` — TDX (Intel hardware attestation) |
| commit | `7e8197785` |

ok

skyrocket2026:
## Eval score invalidated: `eval-museglimmer:S` → `eval-museglimmer:REJECT`

This PR's captured decode graph for the batched prefill (`prefill_batched_run`) introduced
a use-after-free: the arena holding the graph's scratch buffers gets released whenever a
call's footprint exceeds the keep-resident budget, but the CUDA graph captured against those
addresses wasn't torn down at the same time. Replaying that graph on a later call at the same
prefill N launches kernels against freed device memory — a segfault (#809), reproducible with
`SPARKINFER_BENCH_SWEEP_CTXS="512,4096" SPARKINFER_BENCH_SWEEP_REPS=5`.

Fixed on `main` in 8a10f95. Per the maintainer's decision, this PR's original `S` speedup score
and its `museglimmer-merge-first` round win are voided (labels updated, `museglimmer-merge-first`
removed, local scores ledger entry deleted) — the measured throughput gain was real, but it came
attached to a memory-safety bug that shouldn't have scored as a clean pass. The commit stays
merged on `main`; only the score/label decision is being corrected.

PR #801, opened 2026-08-13
perf(muse-glimmer): lane-parallel bf16 prefill attention (1.03x prefill@128)

kai392:
## Summary

Muse Glimmer's batched prefill attention (`win_prefill_pure_bf16_kernel`) walked keys one at a time per query-warp: every key paid a 5-shuffle warp reduction plus two dependent `expf`s, so the warp was latency-bound. Qwythos already had the lane-parallel rewrite for int8 KV; the hd128 bf16 Muse path never got it.

This PR gives each of the 32 lanes one key of the TK=32 tile (full 128-dim score in parallel, one max/sum reduce per tile). Window/causal predicates are unchanged (pure SWA, no sink). Online-softmax association order differs, so the output is not bit-identical; the recurrence is the same. Muse-only launcher — Qwen3.6 does not take this path.

`SPARKINFER_MUSE_PREFILL_ATTN_LANEPAR=0` restores the old kernel.

## Proof of speedup

Real `origin/main` build (`/tmp/qwen3_gguf_bench-main` @ `7fb6a20`) alternated with this PR tree, default env, Muse Glimmer `muse-glimmer-30B-kquant-17gb.gguf`. GPU clocks not locked (no permission); SM clock ~2392 MHz.

**Decode tok/s** (Muse Glimmer, ctx 0 — unchanged, this PR targets prefill):

| | decode tok/s |
|---|--:|
| before (main `7fb6a20`) | 103.65 |
| after (this PR) | 103.68 |

**Prefill pp tok/s** (Muse Glimmer, ctx 128 — the scored dimension):

| | prefill pp tok/s |
|---|--:|
| before prefill (main) | 3110.05 |
| after prefill (this PR) | **3196.11 (+2.77%)** |

```text
--- Muse Glimmer, interleaved main vs PR, default env ---
MAIN decode ctx=0 : 103.65, 103.62, 103.85
PR   decode ctx=0 : 103.66, 103.68, 103.85

MAIN prefill@128  : 3104.49, 3111.70, 3109.56, 3114.43
PR   prefill@128  : 3199.37, 3197.99, 3199.71, 3187.35

decode @ ctx=128 stayed flat (102.80 -> 102.85). Worst PR prefill (3187) vs best main (3114) is still +2.3%.
```

## Root cause

`launch_prefill_attn_swa_pure_bf16` launched a one-query-per-warp kernel that serialized the K loop: score, warp-sum, online-softmax, AV, repeat. At prefill M=128 that is the entire attention working set, and the fp32 units sat idle behind the shuffle/`expf` chain.

## Fix approach

Port the already-shipped int8 lane-parallel schedule (`win_prefill_lanepar_kernel`) onto Muse's bf16 KV pool, keeping Muse's pure-window `win_start` (no StreamingLLM sink). QPW=1 preserves the 256-CTA grid at ctx=128 so occupancy does not collapse.

## Impact

Verified ~2.8% prefill@128 on RTX 5090; decode flat (no regression). Single file, Muse-only path, env opt-out.

## Risk / tradeoffs

- Softmax reduction order changes (tile-wide vs per-key). Same class of change as the merged int8 lane-parallel prefill attention. Accuracy gate (top1 ≥ 0.90, KL ≤ 0.10) is the check.
- Does not touch Qwen3.6 / DFlash kernels.
- Does not overlap open PRs #798 (Q3_A FFN), #799 (fused Q4_K GEMM atomic split-K + LUT), #800 (prefill CUDA graph).

### DIFF
--- a/kernels/csrc/cuda/fused/prefill_attn_window.cu
+++ b/kernels/csrc/cuda/fused/prefill_attn_window.cu
@@ -527,6 +527,186 @@ __global__ void win_prefill_pure_bf16_kernel(
     }
 }
 
+// ----------------------------------------------------------------------------
+// Lane-parallel schedule of the Muse Glimmer bf16 kernel above.
+//
+// win_prefill_pure_bf16_kernel walks keys one at a time per query-warp: every
+// key pays a 5-shuffle warp reduction plus two dependent expf's, so the warp is
+// latency-bound and the fp32 units sit idle. That is the same bottleneck the
+// int8-KV lane-parallel kernel (win_prefill_lanepar_kernel) already removed for
+// Qwythos; Muse Glimmer's hd128 bf16 path never got the rewrite.
+//
+// Each of the 32 lanes owns ONE key of the TK=32 tile:
+//   - scoring: lane j computes the full 128-dim dot q.K[j] from smem (K rows
+//     padded to 132 floats -> lane-strided reads hit 32 distinct banks);
+//   - softmax: ONE max-reduce + ONE sum-reduce per tile;
+//   - AV: each lane owns 4 output dims; p_j broadcast by __shfl_sync.
+//
+// Window/causal predicate is Muse's pure rolling window (NO attention sink) --
+// identical to win_prefill_pure_bf16_kernel -- so SWA vs global is unchanged.
+// Online-softmax association order differs (tile-wide vs per-key), so the
+// output is not bit-identical; the values stay in the same fp32 online-softmax
+// recurrence. SPARKINFER_MUSE_PREFILL_ATTN_LANEPAR=0 restores the old kernel.
+// ----------------------------------------------------------------------------
+template <int HEAD_DIM, int QPW, int TK>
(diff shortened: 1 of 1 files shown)


skyrocket2026:

| metric | value |
|---|---|
| PR decode tok/s | 101.05 |
| main decode tok/s | 100.71 |
| PR prefill@128 pp tok/s | 3156.72 |
| main prefill@128 pp tok/s | 3072.90 |
| prefill speedup vs main | **1.03×** (+2.7%) |
| Polaris receipt | `73401533fb66932c…` — TDX (Intel hardware attestation) |
| commit | `cc919ee7a` |

ok

skyrocket2026:
## Closed: stale — no commits in 1.1 days

This PR has had no new commits in over 1 days — closing automatically to keep the Muse Glimmer eval queue clean. Reopen (or push a new commit / open a fresh PR) whenever you're ready to continue; it'll be picked back up on the next eval cycle.

PR #802, opened 2026-08-13
perf(muse-glimmer): nibble-table Q4_K decode in the fused prefill GEMM (1.03x prefill@128)

kai392:
## Summary

The fused Q4_K prefill GEMM (`pf_dense_gemm_qi8_kernel_g`) is decode-bound: ncu on this kernel shows DRAM 6% / SM 13%, and the B-stage `qm_decode_j64` was paying `roundf((ds * n - dm) * inv)` — an I2F plus an F2I — **per nibble**, 64 times per super-block. A Q4_K nibble only has 16 possible values, so 48 of those conversions were duplicates.

This PR tabulates the 16 results once per sub-block with `n` a compile-time constant (the I2F on `(float)n` folds) and gathers four packed int8 bytes with `prmt`. **Bit-identical**: entry `n` is the exact expression the per-value path evaluated. Dense GEMM only — routed MoE / Q5_K keep the original loop. Qwen3.6 is unchanged.

`SPARKINFER_MUSE_Q4K_LUT=0` restores the per-value decode.

This overlaps the same file as open #799 but is a different change: no atomic split-K, no SwiGLU epilogue fuse, no grid reorder. Does not touch `qwen35_prefill.cpp` (open #800 CUDA-graph capture) and does not touch DFlash.

## Proof of speedup

Interleaved `SPARKINFER_MUSE_Q4K_LUT=0` (this tree, old decode) vs default (this PR) on one binary, Muse Glimmer `muse-glimmer-30B-kquant-17gb.gguf`. GPU clocks not locked (no permission); SM clock ~2392 MHz.

**Decode tok/s** (Muse Glimmer, ctx 0 — unchanged, this PR targets prefill):

| | decode tok/s |
|---|--:|
| before (per-value decode) | 103.84 |
| after (this PR) | 103.84 |

**Prefill pp tok/s** (Muse Glimmer, ctx 128 — the scored dimension):

| | prefill pp tok/s |
|---|--:|
| before prefill (per-value decode) | 3117.85 |
| after prefill (this PR) | **3212.78 (+3.04%)** |

```text
--- Muse Glimmer, interleaved LUT=0 vs default, same binary ---
MAIN decode ctx=0 : 103.86, 103.85, 103.83, 103.83
PR   decode ctx=0 : 103.84, 103.83, 103.84, 103.81

MAIN prefill@128  : 3123.04, 3117.85, 3110.80, 3118.96, 3107.43
PR   prefill@128  : 3210.96, 3211.95, 3219.76, 3212.78, 3213.28

decode @ ctx=128 stayed flat (102.95 -> 102.96). Worst PR prefill (3210.96) vs best main (3123.04) is still +2.82%.
```

## Root cause

`qm_decode_j64` is the B-stage of the fused Q4_K GEMM that is ~80% of Muse prefill@128. Format conversion is quarter-rate on sm_120, and the per-value loop issued 64 I2F + 64 F2I for 16 distinct results.

## Fix approach

Evaluate the 16 nibble results once (compile-time `n`, same `roundf` / truncating cast), pack them 4-per-register, and gather with `prmt` so the table never touches local memory. An indexed `t[n]` array and a ternary mux of the same table both lost prefill (measured). Dense launcher only; Q5_K and the routed MoE kernels keep the original path.

## Impact

Verified ~3.0% prefill@128 on RTX 5090; decode flat (no regression). Single file, bit-identical, env opt-out.

## Risk / tradeoffs

- Bit-identical by construction (same `roundf` expression, same packed stores). Accuracy gate (top1 ≥ 0.90, KL ≤ 0.10) is the check.
- Does not touch Qwen3.6 / DFlash kernels.
- Does not implement CUDA-graph capture of batched prefill (open #800) and does not implement atomic split-K / SwiGLU fusion (open #799).

### DIFF
--- a/kernels/csrc/cuda/fused/prefill_moe_q.cu
+++ b/kernels/csrc/cuda/fused/prefill_moe_q.cu
@@ -97,7 +97,14 @@ __device__ __forceinline__ void qm_cp16(void* dst, const void* src, bool pred) {
 // Everything stays in registers: the 32 quant bytes come in as two 16-byte loads and the int8
 // results are packed into words before being stored, so no byte array is ever addressed
 // dynamically (a dynamically indexed local array would land in local memory).
-template <int QT>
+//
+// LUT=true (Q4_K only): a nibble has 16 possible values, so roundf((ds*n-dm)*inv) is
+// tabulated once with n a compile-time constant -- the I2F on (float)n folds away -- then
+// each group of four input nibbles is four table bytes via prmt (integer-only, no local
+// memory). The per-value path paid 64 I2F+F2I per j64; this pays 16. Bit-identical: entry n
+// is the exact expression the per-value path evaluated. Dense GEMM only;
+// SPARKINFER_MUSE_Q4K_LUT=0 restores the per-value decode.
+template <int QT, bool LUT = false>
 __device__ __forceinline__ void qm_decode_j64(const unsigned char* __restrict__ blk, int j64,
                                               float inv, signed char* __restrict__ dst) {
     const float bd = qm_h2f(blk), bdmin = qm_h2f(blk + 2);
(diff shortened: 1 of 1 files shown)


kai392:
Closing: #799 merged first with a Q4_K nibble table on this same kernel (plus atomic split-K). This PR now conflicts with main and would be redundant.

PR #807, opened 2026-08-13
perf(muse-glimmer): k-tiled activation copy for the dense prefill GEMM (1.07x prefill@128)

FranDev132:
## Summary

`pf_dense_gemm_qi8_kernel_g` (kernels/csrc/cuda/fused/prefill_moe_q.cu) stages a 32-byte K slice of all 128 rows per pipeline step. Out of the row-major activation those 128 chunks are K bytes apart, so a warp's 16 rows issue **16 separate sector requests** to move 512 B.

This emits a second, k-tiled `[k/32][row][32]` copy of the same int8 activation from the kernels that already produce it, and points the GEMM's A stage at it. One stage then becomes a contiguous 4 KB block — **4 line requests for the same 512 B**. Same bytes, same `cp.async` count, same shared-memory destination; only the addresses change, so the int8 the MMA consumes is unchanged and the K loop still carries one add per stage (`a_step` is a constant in both layouts).

I priced the axis before building it, by feeding the kernel a contiguous address for the same bytes (numerically wrong, timing only): **prefill@128 30.65 ms vs 32.21 ms**, so the scatter alone was 5% of the prefill.

The copy is written *alongside* the row-major output, never instead of it, so `proj()`'s materialize fallback and every non-Muse consumer are untouched. All five kernels that write that activation emit both (`pf_quant_rows_fast`, `pf_gate_quant_rows`, `rmsnorm_quant_i8`, `pf_swiglu_quant_i8_reg`, `pf_swiglu_quant_i8_acc`), and a validity flag in the prefill tracks which one wrote last so a stale copy cannot reach the GEMM — the pre-FFN norm's fused quantize is one of those writers and is followed by a memo that makes the normal quantize a no-op, which is exactly the case a flag is needed for.

Cost is one extra store in the quantizer — scattered, because a block owns one row whose packed bytes are 32 B chunks `rows*32` apart — against a read the GEMM repeats once per output tile, 312 times for the 19968-wide FFN pair. In the profile that is 0.48 ms of a 29 ms prefill (`rmsnorm_quant_i8` 0.242 + `pf_quant_rows_fast` 0.121 + `pf_gate_quant_rows` 0.119) against ~2.3 ms saved.

`SPARKINFER_MUSE_QB_APACK=0` restores row-major staging (A/B in one binary).

## Proof of speedup

Two separate build trees (`build_main` from `origin/main`, `build` with this change) run alternately on the same box — a CUDA binary cannot be copied out of its build tree.

**Decode tok/s** (Muse Glimmer, ctx=0, median of 5):

| | decode tok/s |
|---|--:|
| before (main) | 103.81 |
| after (this PR) | 103.80 |

**Prefill pp tok/s** (Muse Glimmer, ctx=128, median of 5):

| | prefill pp tok/s |
|---|--:|
| before prefill (main) | 4353.21 |
| after prefill (this PR) | 4675.58 |

**1.07x prefill@128, decode flat.** Three alternating PR-vs-main pairs on the same box:

| pair | main | this PR | ratio |
|---|--:|--:|--:|
| 1 | 4363.36 | 4682.82 | 1.073 |
| 2 | 4327.73 | 4675.58 | 1.080 |
| 3 | 4353.21 | 4675.51 | 1.074 |

**Accuracy — unchanged**, `SPARKINFER_KV_INT8=0 qwen3_gguf_prefill_check` (batched prefill vs the token-by-token path):

| build | N=128 TOP1 | N=128 KL | N=512 TOP1 | N=512 KL |
|---|--:|--:|--:|--:|
| main | 15/16 0.9375 | 0.03630 | 16/16 1.0000 | 0.01333 |
| this PR | 15/16 0.9375 | 0.03630 | 16/16 1.0000 | 0.01333 |

Same seed (194 / 220) in every row, and identical again with `SPARKINFER_MUSE_PREFILL_GRAPH=0`, so the result does not depend on whether the batched prefill is replayed from a graph. The same check with `SPARKINFER_MUSE_QB_APACK=0` vs `=1` inside this build also matches exactly at N=128/512/1024 (1024: 15/16 0.9375, KL 0.01730, seed 220 both ways).

**Muse Glimmer multi-context** (`SPARKINFER_KV_INT8=0 SPARKINFER_MUSE_PREFILL_GRAPH=0`, median of 5) — the longer prompts gain more, because past one M-tile there is no split-K and every M-tile re-streams its own activation rows:

| ctx | decode ratio | prefill ratio |
|---|--:|--:|
| 0 | 0.9998 | — |
| 128 | 0.9998 | 1.061 |
| 512 | 0.9997 | 1.216 |
| 4096 | 0.9999 | 1.112 |

Both flags are needed to take this sweep at all, on main as well as here, and neither is related to this change: int8 KV is broken on Muse at ctx >= 4k, and a sweep that prefills at several different N in one process segfaults with prefill graph capture enabled (main at `98af9ec` exits 139 on `0,128,512,4096`; the same sweep with `SPARKINFER_MUSE_PREFILL_GRAPH=0` completes). The scored ctx-0/128 sweep above uses stock defaults.

**Qwen3.6 no-regression guard** (`Qwen3.6-35B-A3B-UD-Q4_K_M`, ctx 0/512/4k/16k/32k, median of 5), PR/main ratio — floor is 0.98:

| ctx | decode | prefill |
|---|--:|--:|
| 0 | 1.0000 | — |
| 512 | 1.0003 | 1.0046 |
| 4096 | 1.0002 | 1.0019 |
| 16384 | 1.0004 | 0.9997 |
| 32768 | 0.9996 | 1.0002 |

```text
########## ACCURACY FIRST, graph ON (default) ##########
  N=128 main: TOP1  15/16 0.9375 KL    0.03630 (mean over 16 positions)
  N=128  PR : TOP1  15/16 0.9375 KL    0.03630 (mean over 16 positions)
  N=512 main: TOP1  16/16 1.0000 KL    0.01333 (mean over 16 positions)
  N=512  PR : TOP1  16/16 1.0000 KL    0.01333 (mean over 16 positions)
########## ACCURACY, graph OFF ##########
  N=128  PR (GRAPH=0): TOP1  15/16 0.9375 KL    0.03630 (mean over 16 positions)
  N=512  PR (GRAPH=0): TOP1  16/16 1.0000 KL    0.01333 (mean over 16 positions)
########## SPEED, alternating pairs (stock defaults) ##########
  pair1 main: ctx0 d=103.82 p=0.00  ctx128 d=102.84 p=4363.36
  pair1  PR : ctx0 d=103.80 p=0.00  ctx128 d=102.81 p=4682.82
  pair2 main: ctx0 d=103.81 p=0.00  ctx128 d=102.82 p=4327.73
  pair2  PR : ctx0 d=103.79 p=0.00  ctx128 d=102.79 p=4675.58
  pair3 main: ctx0 d=103.80 p=0.00  ctx128 d=102.81 p=4353.21
  pair3  PR : ctx0 d=103.81 p=0.00  ctx128 d=102.80 p=4675.51
########## Muse multi-ctx (KV_INT8=0, GRAPH=0) ##########
  main: ctx0 d=103.48 p=0.00  ctx128 d=102.52 p=4025.98  ctx512 d=101.04 p=4036.21  ctx4096 d=94.25 p=2905.09
  PR  : ctx0 d=103.46 p=0.00  ctx128 d=102.50 p=4269.74  ctx512 d=101.01 p=4909.33  ctx4096 d=94.24 p=3230.36
########## Qwen3.6 guard ##########
  main: ctx0 d=453.94 p=0.00  ctx512 d=447.50 p=8308.20  ctx4096 d=429.94 p=22874.25  ctx16384 d=431.86 p=24843.26  ctx32768 d=431.89 p=26187.05
  PR  : ctx0 d=453.96 p=0.00  ctx512 d=447.65 p=8346.76  ctx4096 d=430.04 p=22916.82  ctx16384 d=432.05 p=24835.48  ctx32768 d=431.72 p=26192.42
```

Profile after the change (nsys, prefill@128 window, 29.0 ms): the dense GEMM is 22.0 ms (75.6%), attention 1.66 ms, GPU idle 1.84 ms, and the three activation quantizers together 0.48 ms.

### DIFF
--- a/kernels/csrc/cuda/fused/prefill_gemm_i8.cu
+++ b/kernels/csrc/cuda/fused/prefill_gemm_i8.cu
@@ -304,13 +304,15 @@ bool launch_prefill_gemm_i8_splitk(const signed char* A, const signed char* W,
     return true;
 }
 
-void launch_prefill_quantize_rows_i8(const void* x_bf16, signed char* q, float* scale,
-                                     int rows, int cols, cudaStream_t stream) {
+bool launch_prefill_quantize_rows_i8(const void* x_bf16, signed char* q, float* scale,
+                                     int rows, int cols, cudaStream_t stream, signed char* qp) {
     // Block-parallel single-pass path (one block per row, row held in registers; bit-identical).
     // SPARKINFER_PREFILL_QUANT_ROWS=0 restores the warp-per-row kernel below.
-    if (launch_prefill_quant_rows_fast(x_bf16, q, scale, rows, cols, stream)) return;
+    if (launch_prefill_quant_rows_fast(x_bf16, q, scale, rows, cols, stream, qp)) return true;
     pf_quantize_rows_i8<<<rows, 32, 0, stream>>>(
         reinterpret_cast<const __nv_bfloat16*>(x_bf16), q, scale, rows, cols);
+    // The warp-per-row fallback has no k-tiled output, so tell the caller its packed copy is stale.
+    return qp == nullptr;
 }
 
 void launch_prefill_gemm_i8(const signed char* A, const signed char* W,
--- a/kernels/csrc/cuda/fused/prefill_moe_q.cu
+++ b/kernels/csrc/cuda/fused/prefill_moe_q.cu
@@ -813,7 +813,8 @@ __global__ __launch_bounds__(256, 2) void pf_dense_gemm_qi8_kernel_g(
         const signed char* __restrict__ A_i8, const float* __restrict__ sx,
         const unsigned char* __restrict__ W_q_arg, const float* __restrict__ row_scale_arg,
         __nv_bfloat16* __restrict__ C_arg, int Mtot, int N_arg, int K, PfQGroup gd,
-        int* __restrict__ partials = nullptr, int sb_per_split = 0, int atomic_acc = 0) {
+        int* __restrict__ partials = nullptr, int sb_per_split = 0, int atomic_acc = 0,
+        const signed char* __restrict__ A_pack = nullptr) {
     using namespace nvcuda;
     constexpr int BS = qm_bs<QT>();
     // Split-K takes blockIdx.x for the K slice and shifts the tile/M indices up one dimension.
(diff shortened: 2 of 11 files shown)


skyrocket2026:

| metric | value |
|---|---|
| PR decode tok/s | 103.58 |
| main decode tok/s | 102.98 |
| decode speedup vs main | **1.01×** (+0.6%) |
| PR prefill@128 pp tok/s | 4701.19 |
| main prefill@128 pp tok/s | 4271.86 |
| prefill speedup vs main | **1.10×** (+10.1%) |
| Polaris receipt | `8667d2f1d0978106…` — TDX (Intel hardware attestation) |
| commit | `6d2f40b41` |

ok

skyrocket2026:

PR #811, opened 2026-08-14
perf(muse-glimmer): build the SM120 FP4 prefill path by default (1.14x prefill@128)

inference2026:
## Summary

The native SM120 block-scaled FP4 gate/up prefill projections cannot be selected by a default
build. This makes them buildable and enables them, which needs an architecture fix: FP4's
block-scaled MMA is arch-specific, so the whole project must target `sm_120a` rather than one
library doing so on its own.

## Proof of speedup

**Decode tok/s** (end-to-end — this PR targets prefill; included to show no regression):

| | decode tok/s |
|---|--:|
| before (main) | 102.55 |
| after (this PR) | 102.54 |

**Prefill pp tok/s** (Muse Glimmer, `ctx=128` — the scored prefill shape for this model):

| | prefill pp tok/s |
|---|--:|
| before prefill (main) | 4895.0 |
| after prefill (this PR) | 5598.8 |

**+14.4% prefill@128.** Median-of-5, the two builds interleaved in one session on an idle card
with clocks locked at 2550 MHz. `ctest` 19/19.

```text
# main @ 3e34dc9, default build: cmake -S . -B build -DCMAKE_BUILD_TYPE=Release
main {"128":{"decode_tps":102.4023,"prefill_pp":4901.0084}}
PR   {"128":{"decode_tps":102.3703,"prefill_pp":5588.5388}}
main {"128":{"decode_tps":102.5759,"prefill_pp":4895.0423}}
PR   {"128":{"decode_tps":102.5456,"prefill_pp":5609.5283}}
main {"128":{"decode_tps":102.5484,"prefill_pp":4882.4778}}
PR   {"128":{"decode_tps":102.5433,"prefill_pp":5598.8303}}

# both arms from a cold `cmake -S . -B <dir> -DCMAKE_BUILD_TYPE=Release`, no extra flags:
# main  -> CMAKE_CUDA_ARCHITECTURES:STRING=89;90;100;120
# PR    -> CMAKE_CUDA_ARCHITECTURES:STRING=89;90;100;120a, BUILD_NVFP4_KERNELS:BOOL=ON
#          [prefill-muse] SM120 NVFP4 FFN weights ready: 52/52 layers
```

## What was blocking it

Three things, of which the first is a hard build failure:

1. **Architecture.** `ptxas` rejects `mma.sync.aligned.m16n8k64.kind::mxf4` on `.target sm_120`
   and accepts it on `sm_120a`, so the FP4 translation unit must be compiled for `120a`. Setting
   `CUDA_ARCHITECTURES` on that one library while its siblings stay at `120` does not device-link:
   `nvlink fatal : elfLink fatbinary error`. The choice has to be project-wide and taken before
   `project()` latches `CMAKE_CUDA_ARCHITECTURES`, so it moves into a small shared module included
   by both the root and `kernels/`. `sm_120a` SASS runs on the same silicon; 89/90/100 unchanged.
2. **Kernels not compiled**, leaving `prefill_nvfp4_supported()` as a stub returning `false`.
3. **Runtime opt-in.** Now opt-out — `SPARKINFER_MUSE_PREFILL_NVFP4=0` restores the portable
   quantized projections.

CUTLASS is header-only and fetched at configure time. The module probes the remote once on a cold
build directory and falls back to the portable CUDA path if it is unreachable, so a box without
network still configures and builds — an unavailable dependency degrades the build rather than
breaking it.

## Accuracy

The teacher-forced decode score is **byte-identical** (same md5 over the scored output), because
this path is prefill-only and never runs during decode.

Prefill quality, `qwen3_gguf_prefill_check` at 128 tokens over 32 continuation positions
(batched vs the token-by-token reference):

| | TOP1 | mean KL |
|---|--:|--:|
| portable projections | 29/32 (0.9062) | 0.04443 |
| FP4 gate/up | 29/32 (0.9062) | 0.05432 |

TOP1 and the seed token are unchanged; mean KL rises by 0.0099, the expected cost of 4-bit
gate/up. Down-projection stays on the quantized path, whose error enters the residual directly.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,7 +9,17 @@ cmake_minimum_required(VERSION 3.20)
 # sometimes ancient e.g. "52") default if it isn't already set. A set() after
 # project() is a CACHE no-op once that's happened, silently building tensor-core/wmma
 # kernels for a target that can't support them.
-set(CMAKE_CUDA_ARCHITECTURES "89;90;100;120" CACHE STRING "CUDA architectures")
+#
+# The Blackwell entry is sm_120a, not sm_120, when the native FP4 prefill path is built. FP4's
+# block-scaled MMA is an architecture-specific instruction: ptxas rejects
+# `mma.sync.aligned.m16n8k64.kind::mxf4` outright on `.target sm_120` and accepts it on sm_120a,
+# and sm_120a SASS runs on the same silicon. It has to be decided HERE, for the whole project, and
+# not per target: a single library built for 120a while its siblings are 120 does not device-link
+# at all ("nvlink fatal: elfLink fatbinary error"), which is why the FP4 kernels could not be
+# enabled as they stood. The other architectures are unaffected.
+include(cmake/Nvfp4Arch.cmake)
+set(CMAKE_CUDA_ARCHITECTURES "89;90;100;${SPARKINFER_BLACKWELL_ARCH}"
+    CACHE STRING "CUDA architectures")
 project(sparkinfer LANGUAGES CXX CUDA)
 
 set(CMAKE_CXX_STANDARD 17)
--- a/cmake/Nvfp4Arch.cmake
+++ b/cmake/Nvfp4Arch.cmake
@@ -1,0 +1,47 @@
+# Decides whether the native SM120 block-scaled FP4 prefill kernels can be built, and therefore
+# which Blackwell architecture the project targets.
+#
+# Included from the top-level CMakeLists BEFORE project(), and again from kernels/ (which is also
+# configurable standalone). The arch has to be settled before project() because
+# enable_language(CUDA) latches CMAKE_CUDA_ARCHITECTURES at that call.
+include_guard(GLOBAL)
+
+option(BUILD_NVFP4_KERNELS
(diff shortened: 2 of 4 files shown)


skyrocket2026:

| metric | value |
|---|---|
| PR decode tok/s | 103.51 |
| main decode tok/s | 102.70 |
| decode speedup vs main | **1.01×** (+0.8%) |
| PR prefill@128 pp tok/s | 4893.98 |
| main prefill@128 pp tok/s | 4823.46 |
| prefill speedup vs main | **1.01×** (+1.5%) |
| Polaris receipt | `9e0eaf2cf5f1a1b5…` — TDX (Intel hardware attestation) |
| commit | `44a537ff3` |

within significance gate — not a verified decode improvement

skyrocket2026:

This PR's Muse Glimmer 128-decode/prefill speed measured **0.8%** vs main, with no verified improvement on both decode and prefill — closing automatically. This bot evaluates every eligible PR in the repo against Muse Glimmer's decode AND prefill@128 speed specifically, regardless of what the PR is actually about — a close here isn't a judgment on the PR's purpose, just that it didn't move these particular metrics. Reopen (or open a fresh PR) if you have a fix or a different approach.

PR #816, opened 2026-08-14
perf(muse-glimmer): extend the SM120 NVFP4 prefill path to ffn_down and wo (1.10x prefill@128)

JSONbored:
## Summary

#810 added the SM120 block-scaled NVFP4 prefill path; #813 made it the default build — for the FFN **gate/up** projections. This PR extends it to the two remaining fat projections: **`ffn_down`** and **`wo`**. `ffn_down` passes the same `%128` shape gate (6656×19968) and its A operand — the SwiGLU output — is already materialized as bf16 on the FP4 branch. One `quant_a` + one block-scaled GEMM replaces the int8 fused-down and its split-K partials. 33 lines; the int8 path remains the runtime fallback on any launch failure; `SPARKINFER_MUSE_NVFP4_DOWN=0` restores it outright. **`wo`** [6656×4096] takes the same treatment (A = the attention output, bf16 at that point); `SPARKINFER_MUSE_NVFP4_WO=0` restores the int8 o-projection independently.

Credit: the NVFP4 kernels, quantizers and conversion design are #810 (@andriypolanski), enabled by #813 (@inference2026). This PR only widens which tensors use them.

Also measured and deliberately **excluded**: the same extension to `q/gate/k/v` is a *loss* (−4.6% — four skinny FP4 GEMMs lose to #791's single grouped int8 launch). Kept out so this diff is only the part that wins.

## Proof of speedup

- [x] Tested on **RTX 5090** (`sm_120`)

**Decode tok/s** (end-to-end — this PR targets prefill; shown to demonstrate no regression):

| | decode tok/s |
|---|--:|
| before (main) | 106.59 |
| after (this PR) | 106.59 |

**Prefill pp tok/s** (Muse Glimmer, `ctx=128` — the scored prefill shape):

| | prefill pp tok/s |
|---|--:|
| before prefill (main) | 6140.9 |
| after prefill (this PR) | 6748.5 |

```text
# qwen3_gguf_bench sweep (SPARKINFER_BENCH_SWEEP_CTXS=0,128, REPS=5), base interleaved between arms
main:    SWEEP_JSON {"0":{"decode_tps":106.5999},"128":{"decode_tps":105.6285,"prefill_pp":6102.0567}}
this PR: SWEEP_JSON {"0":{"decode_tps":106.5927},"128":{"decode_tps":105.6243,"prefill_pp":6689.4526}}
[...]

PR #810, opened 2026-08-14
perf(muse-glimmer): add SM120 NVFP4 gate/up prefill path

andriypolanski:
## Summary

Adds an opt-in SM120-native block-scaled NVFP4 path for the Muse Glimmer dense-FFN gate and up projections at the scored `M=128` shape. Gate and up share one BF16-to-NVFP4 activation conversion, while the accuracy-sensitive down projection remains on the existing #808 quantized-B path.

The implementation uses CUTLASS 4.7 Blackwell GeForce block-scaled tensor operations and preserves the existing kernels as fallbacks. It is enabled at build time with `BUILD_NVFP4_KERNELS=ON` and at runtime with `SPARKINFER_MUSE_PREFILL_NVFP4=1`.

On an RTX 5090 against current main `3c00680`, prefill at context 128 improves from an A/B/A baseline mean of 5373.64 to 6061.02 tok/s (+12.79%). Decode remains flat, mean batched-prefill KL stays below the repository's H3 diagnostic limit, and VRAM increases by 1.5 GB.

## Proof of speedup

- [x] Tested on **RTX 5090** (`sm_120`)

**Decode tok/s** (end-to-end; this PR does not target decode):

| | decode tok/s |
|---|--:|
| before (main) | 107.84 |
| after (this PR) | 107.84 |

**Prefill pp tok/s** (Muse Glimmer 30B, context 128):

| | prefill pp tok/s |
|---|--:|
| before prefill (main) | 5373.64 |
| after prefill (this PR) | **6061.02** |

These are end-to-end Muse measurements for the workload targeted by this PR. They are not Qwythos/Qwen3.5 measurements; the current template describes that separate workload at context 4096 or greater.

### Additional Muse Glimmer benchmark

The same model, benchmark binary, context, and five-repetition sweep were used for both revisions. Main was measured before and after the PR run to check run-order bias.

| Configuration | Prefill @ 128 | Decode @ 128 | VRAM |
|---|---:|---:|---:|
| main `3c00680` (A/B/A mean) | 5373.64 tok/s | 107.84 tok/s | 24.0 GB |
| this PR `9aabf4e` | **6061.02 tok/s** | 107.84 tok/s | 25.5 GB |
| change | **+12.79%** | ~0.00% | +1.5 GB |

```text
# main, first A run
decode tg    : 107.83 tok/s  (n=128, ctx=128, bs=1)
prefill pp   : 5357.29 tok/s  (ctx=128, sequential KV fill)

# this PR, B run
[...]

PR #817, opened 2026-08-14
perf(muse-glimmer): NVFP4 ffn_down prefill leg (fixed #816) + finer FP4 CTA tiles (1.11x prefill@128)

JSONbored:
## Summary

The fixed resubmit of #816, which #819 reverted for breaking batched prefill accuracy, together with finer CTA tiles for the FP4 GEMMs it feeds. #816 shipped **two** NVFP4 legs — `ffn_down` and the o-projection (`wo`). Isolated on a 5090, the defect is entirely the o-projection: **`ffn_down` alone clears the accuracy gate, `wo` destroys it.** This restores only the `ffn_down` leg and drops `wo`. 1.11x prefill@128, decode untouched.

## Proof of speedup

| | decode tok/s |
|---|--:|
| before (main) | 104.98 |
| after (this PR) | 104.94 |

| | prefill pp tok/s |
|---|--:|
| before prefill (main) | 6774.2 |
| after prefill (this PR) | 7523.6 |

```text
# One binary, arms env-flipped (SPARKINFER_MUSE_NVFP4_DOWN=0 SKINNY=0 K256=0 == main's
# dispatch), SWEEP_REPS=9 so the median sits in #800's prefill-graph replay regime.
# RTX 5090, main @ 593bab0. B = main, T = this PR. prefill@128:
B: 6819 6760 6759 6758   mean 6774.2
T: 7536 7524 7515 7519   mean 7523.6   (+11.06%)
```

## Why #816 broke, and what changed

`qwen3_gguf_prefill_check` is the gate for this path, and it is easy to mis-run: Muse Glimmer's
batched prefill refuses an int8 KV cache, so the tool's default prints
`[FAIL] prefill_batched unsupported (seed=-1)` and looks broken. It needs
**`SPARKINFER_KV_INT8=0`**. With that set, one leg toggled at a time (same binary, `P=128`,
16 continuation positions):

| configuration | TOP1 | KL |
|---|--:|--:|
| reference (no FP4 legs) | 15/16 | 0.06349 |
| **`ffn_down` leg only** | **14/16** | **0.05623** |
| `ffn_down` + `wo` (i.e. #816) | 2/16 | 4.56426 |

Those last two rows differ by the o-projection alone, so the o-projection carries the defect —
KL 4.6 is not quantization error, it is a wrong result. It is removed here, along with its
weight conversion, its `wo_fp4`/`wo_fp4_sf` pointers and its dead buffers. The `ffn_down` leg
sits at or below the reference's own divergence from the token-by-token path.

Verification for this PR as submitted:

```text
prefill_check, SPARKINFER_KV_INT8=0, RTX 5090
  P=128 cont=64   PR: TOP1 58/64  KL 0.04554   |  ref: TOP1 58/64  KL 0.04732
  P=128 cont=16   PR: TOP1 14/16  KL 0.05623   |  ref: TOP1 15/16  KL 0.06349
  P=512 cont=16   PR: TOP1 16/16  KL 0.00815   |  ref: TOP1 16/16  KL 0.00815   (identical;
                                                   the FP4 path engages only at N=128)
```

Over 64 continuation positions the top-1 agreement is **identical** to the reference and the KL
is marginally lower; the single top-1 difference in the 16-position run is one near-tie flipping,
which the wider sample resolves. Mixed-context sweep (512, 4096 x 2) clean.

## The tiles, and why they belong with this leg

At m = 128 rows the tile grid is a single M-row, so CTAs = n/TileN. With the shipped
128x128x128 tile the hidden-output GEMM this leg adds (n = 6656) launches **52 CTAs on 170 SMs**
and streams weights at 1.06 TB/s, against gate/up's 156 CTAs at 1.24 (peak 1.79). A 128x64 tile
doubles the CTAs in flight, and a 256-deep K tile cuts the mainloop trips; together they take
the FP4 GEMM term from 9.892 ms to 8.906 ms on one trace. Measured alone against a main without
this leg, the tiles are worth +0.15% — they only pay when the skinny GEMM exists, which is why
the two ship together. `SPARKINFER_MUSE_NVFP4_SKINNY` / `_K256` restore the single shipped tile,
and the scale-factor layout is fixed by `Sm1xxBlkScaledConfig` rather than the CTA tile
(`static_assert`ed across variants), so every variant reads the same packed scales.

## Lineage

#810 (@andriypolanski) added the SM120 block-scaled FP4 prefill path; #813 (@inference2026)
made it default; #814 (@Paral1995) sharpened the activation quantize this leg feeds; #800's
prefill graph is why the kernel-level saving survives to the wall clock. #816 was ours and
#819 reverted it — thanks for the clear revert note; this is the fixed version it asked for.

### DIFF
--- a/kernels/csrc/cuda/fused/prefill_nvfp4_sm120.cu
+++ b/kernels/csrc/cuda/fused/prefill_nvfp4_sm120.cu
@@ -1,5 +1,6 @@
 #include "sparkinfer/kernels/prefill_nvfp4.h"
 
+#include <cstdlib>
 #include <cuda_bf16.h>
 #include <cutlass/cutlass.h>
 #include <cutlass/float_subbyte.h>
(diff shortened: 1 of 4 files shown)


skyrocket2026:

| metric | value |
|---|---|
| PR decode tok/s | 106.20 |
| main decode tok/s | 105.91 |
| PR prefill@128 pp tok/s | 7480.21 |
| main prefill@128 pp tok/s | 6651.26 |
| prefill speedup vs main | **1.12×** (+12.5%) |
| qwen3.6 guard | ❌ **FAILED** — qwen3.6 guard measurement unavailable; qwen3.6 decode@128: PR measurement missing/zero (main 492.3) — treated as regression; qwen3.6 decode@512: PR measurement missing/zero (main 485.1) — treated as regression; qwen3.6 prefill@512: PR measurement missing/zero (main 9608.8) — treated as regression — **verdict forced to REJECT regardless of speed/accuracy** |
| Polaris receipt | `9399be6c3396e5ea…` — TDX (Intel hardware attestation) |
| commit | `32d2f3638` |

qwen3.6 no-regression guard failed: qwen3.6 guard measurement unavailable; qwen3.6 decode@128: PR measurement missing/zero (main 492.3) — treated as regression; qwen3.6 decode@512: PR measurement missing/zero (main 485.1) — treated as regression; qwen3.6 prefill@512: PR measurement missing/zero (main 9608.8) — treated as regression; qwen3.6 decode@4k: PR measurement missing/zero (main 466.1) — treated as regression; qwen3.6 prefill@4k: PR measurement missing/zero (main 26057.5) — treated as regression | ok

skyrocket2026:

This PR's Muse Glimmer 128-decode/prefill speed measured **12.5%** vs main, and regressed the Qwen3.6 no-regression guard (decode/prefill on shared code) — closing automatically. This bot evaluates every eligible PR in the repo against Muse Glimmer's decode AND prefill@128 speed specifically, regardless of what the PR is actually about — a close here isn't a judgment on the PR's purpose, just that it didn't move these particular metrics. Reopen (or open a fresh PR) if you have a fix or a different approach.

PR #820, opened 2026-08-14
perf(muse-glimmer): block-scaled FP4 ffn_down and o-projection, shape-adaptive GEMM tile, fused activation quantizes (1.21x prefill@128)

widecloud:
## Summary

#810/#813 built the SM120 block-scaled NVFP4 prefill path for the FFN gate/up projections. This PR extends it to **`ffn_down`** and the **o-projection**, and makes two changes the extension needs to actually pay off:

1. **A shape-adaptive CTA tile.** At M=128 the grid is `1 × ceil(N/TileN)`, so the tile alone decides machine coverage and mainloop depth. The shipped `128x128x128` is tuned for gate/up (N=19968); `ffn_down` and `wo` (N=6656) launch a third as many CTAs and run at half the DRAM throughput for the same bytes. The lever is **K depth, not N width** — a 256-deep K step is worth −16% on `ffn_down` — with N halved only once the wide tile stops covering the machine.
2. **The activation producers folded into their FP4 quantizes.** SwiGLU into the `ffn_down` operand, and Muse's attention gate into the o-projection operand, so neither product is written back as bf16 just to be read again by a quantizer.

Both tiles read the same weights: the GMEM block-scale layout comes from `Sm1xxBlockScaledConfig<SFVecSize>`, a function of `SFVecSize` and the problem shape only — the CTA tile appears solely in the smem layouts. Every tile variant returns a bitwise-identical result.

`SPARKINFER_MUSE_NVFP4_DOWN=0`, `SPARKINFER_MUSE_NVFP4_WO=0` and
`SPARKINFER_MUSE_NVFP4_TILE=wide|narrow` restore the previous behaviour, so every arm below comes out of one binary. Per-layer conversion failure falls back to the int8 path on its own.

## Proof of speedup

**Decode tok/s** (ctx=0 — prefill-only PR; shown to demonstrate no regression):

| | decode tok/s |
|---|--:|
| before (main) | 103.09 |
| after (this PR) | 102.98 |

**Prefill pp tok/s** (Muse Glimmer, ctx=128 — the scored prefill shape):

| | prefill pp tok/s |
|---|--:|
| before prefill (main) | 6181.12 |
| after prefill (this PR) | 7470.98 |

**+20.9% prefill@128**, median of 3 alternating pairs, main `593bab0`, two trees built from scratch, base re-sampled between every arm, 5 reps per process.

```text
main p1 | {"0":{"decode_tps":103.6034},"128":{"decode_tps":102.4121,"prefill_pp":6246.7869}} | VRAM 25.5 GB
PR   p1 | {"0":{"decode_tps":103.3180},"128":{"decode_tps":102.1085,"prefill_pp":7506.4821}} | VRAM 30.3 GB
main p2 | {"0":{"decode_tps":103.0925},"128":{"decode_tps":101.8316,"prefill_pp":6179.2115}} | VRAM 25.5 GB
PR   p2 | {"0":{"decode_tps":102.9799},"128":{"decode_tps":101.8125,"prefill_pp":7469.7674}} | VRAM 30.3 GB
main p3 | {"0":{"decode_tps":102.9374},"128":{"decode_tps":101.7943,"prefill_pp":6181.1200}} | VRAM 25.5 GB
PR   p3 | {"0":{"decode_tps":102.8309},"128":{"decode_tps":101.6152,"prefill_pp":7470.9786}} | VRAM 30.3 GB
```

|---|--:|--:|--:|
| p1 | 6246.79 | 7506.48 | +20.16% |
| p2 | 6179.21 | 7469.77 | +20.88% |
| p3 | 6181.12 | 7470.98 | +20.87% |

`SPARKINFER_MUSE_NVFP4_WO=0` gives 7139.07 (+15.5%), so the o-projection leg is +4.65 points of it.

## Accuracy

`qwen3_gguf_prefill_check <gguf> 128 512`, `SPARKINFER_KV_INT8=0`:

| | TOP1 | KL |
|---|--:|--:|
| main `593bab0` | 474/512 = 0.9258 | 0.01309 |
| this PR | **482/512 = 0.9414** | 0.01315 |

Better top1, KL at parity, with two more tensors on FP4.

### Why the o-projection is safe here, given #816 was reverted for it

#819 reverted #816 for breaking batched prefill accuracy. The cause is the FP4 o-projection reading an **un-gated `att`**: `launch_prefill_gate_quant_rows_i8` folds Muse's attention gate into the *int8* row-quantize and deliberately leaves `att` raw (its comment says so), so an FP4 leg that quantizes `att` itself drops the gate on every layer. Bisected on `0308e10`:

| main config | TOP1 @512 | KL |
|---|--:|--:|
| all FP4 off (int8 reference) | 490/512 = 0.9570 | 0.00801 |
| `SPARKINFER_MUSE_NVFP4_WO=0` | 486/512 = 0.9492 | 0.01135 |
| #816 default | **235/512 = 0.4590** | **0.48888** |

This PR folds the gate into the FP4 quantize itself, so no path ever reads a raw `att` — same two tensors, **KL 0.0132 instead of 0.489**.

Worth flagging: the scored accuracy gate cannot catch this class of bug. `qwen3_gguf_score` runs the AR decode path and its dump is byte-identical either way; only `qwen3_gguf_prefill_check` sees it.

## Cost and risk

- VRAM 25.5 → 30.3 GB. Neither `ffn_down` nor `wo` has a separate prefill copy to release (decode reads the same tensors), so each FP4 copy is a real +0.5625 B/value. Fits a 32 GB card, and a conversion that doesn't fit degrades to the int8 path rather than failing.
- Muse-only: every path is behind `c.muse_glimmer` plus the existing FP4 gates. Qwen3.6 never reaches `launch_prefill_nvfp4_*`.
- **Qwen3.6 no-regression guard** (ctx 0/512/4k/16k/32k, both trees): worst ratio **0.9998** against the 0.98 floor — every context at or above parity.
- Also measured and left out: grouped q/gate/k/v on FP4 is a win in isolation but costs +1.7 GB, which does not fit alongside the above.

### DIFF
--- a/kernels/csrc/cuda/fused/prefill_nvfp4.cu
+++ b/kernels/csrc/cuda/fused/prefill_nvfp4.cu
@@ -10,6 +10,10 @@ size_t prefill_nvfp4_scale_bytes_a(int, int) { return 0; }
 size_t prefill_nvfp4_scale_bytes_b(int, int) { return 0; }
 size_t prefill_nvfp4_workspace_bytes(int, int, int) { return 0; }
 bool launch_prefill_nvfp4_quant_a(const void*, void*, void*, int, int, cudaStream_t) { return false; }
+bool launch_prefill_nvfp4_swiglu_quant_a(const void*, const void*, void*, void*, int, int,
+                                         cudaStream_t) { return false; }
+bool launch_prefill_nvfp4_gate_quant_a(const void*, const void*, void*, void*, int, int,
+                                       cudaStream_t) { return false; }
 bool launch_prefill_nvfp4_quant_b(const void*, void*, void*, int, int, cudaStream_t) { return false; }
 bool launch_prefill_nvfp4_gemm(const void*, const void*, const void*, const void*, void*, int, int,
                                int, void*, cudaStream_t) { return false; }
--- a/kernels/csrc/cuda/fused/prefill_nvfp4_sm120.cu
+++ b/kernels/csrc/cuda/fused/prefill_nvfp4_sm120.cu
@@ -11,37 +11,128 @@
 #include <cutlass/util/packed_stride.hpp>
 #include <cute/tensor.hpp>
 
+#include <cstdlib>
+
 namespace sparkinfer::kernels {
 namespace {
 using namespace cute;
 using E4 = cutlass::nv_float4_t<cutlass::float_e2m1_t>;
 using BF = cutlass::bfloat16_t;
-using Tile = Shape<_128, _128, _128>;
 using Cluster = Shape<_1, _1, _1>;
-using Epilogue = typename cutlass::epilogue::collective::CollectiveBuilder<
-    cutlass::arch::Sm120, cutlass::arch::OpClassBlockScaledTensorOp, Tile, Cluster,
-    cutlass::epilogue::collective::EpilogueTileAuto, float, float,
-    BF, cutlass::layout::RowMajor, 8, BF, cutlass::layout::RowMajor, 8,
-    cutlass::epilogue::collective::EpilogueScheduleAuto>::CollectiveOp;
-using Mainloop = typename cutlass::gemm::collective::CollectiveBuilder<
-    cutlass::arch::Sm120, cutlass::arch::OpClassBlockScaledTensorOp,
-    E4, cutlass::layout::RowMajor, 32, E4, cutlass::layout::ColumnMajor, 32, float,
-    Tile, Cluster,
-    cutlass::gemm::collective::StageCountAutoCarveout<
-        static_cast<int>(sizeof(typename Epilogue::SharedStorage))>,
-    cutlass::gemm::collective::KernelScheduleAuto>::CollectiveOp;
-using Kernel = cutlass::gemm::kernel::GemmUniversal<
-    Shape<int, int, int, int>, Mainloop, Epilogue, void>;
-using Gemm = cutlass::gemm::device::GemmUniversalAdapter<Kernel>;
-using StrideA = typename Kernel::StrideA;
-using StrideB = typename Kernel::StrideB;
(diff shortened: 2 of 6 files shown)


skyrocket2026:

| metric | value |
|---|---|
| PR decode tok/s | 106.28 |
| PR prefill@128 pp tok/s | 7965.40 |
| prefill speedup vs main | **1.18×** (+17.7%) |
| Polaris receipt | `84abf5ac34a8cdca…` — TDX (Intel hardware attestation) |
| commit | `dbf62a489` |

ok

skyrocket2026:

skyrocket2026:
Reverted in #825 — same class of issue as #816/#819.

Converting `ffn_down` to NVFP4 adds a second resident copy of that weight that nothing can free: `gate_q`/`up_q` have a separate `prefill_gate_q`/`prefill_up_q` pointer that gets released after FP4 conversion, but `down_q` has no prefill-only counterpart — decode reads it directly, so it stays resident alongside the new `down_fp4` copy. On a 32GB card this pushes resident VRAM to ~30.3GB; at ctx=16384 the KV cache pushes usage to ~31.0/31.8GB, the batched-prefill scratch arena fails to allocate, and prefill silently falls back to the sequential token-loop path — correct output, ~21x slower. The eval bot's Muse Glimmer bench only covers ctx 0/128, both comfortably inside budget, so this round's own numbers are genuine and the regression is simply invisible to current CI.

Credit to FranDev132 for surfacing this (verified independently against the arena-allocation and weight-release code before reverting). #821 has the same shape despite its title and carries the same issue.

Thanks for the contribution — the shape-adaptive GEMM tile work here is good; happy to look at a version that keeps `ffn_down`/`wo` on the existing path (or frees `down_q` some other way) if you want to take another pass.

PR #821, opened 2026-08-14
perf(muse-glimmer): NVFP4 ffn_down prefill leg (fixed #816) + finer FP4 CTA tile (1.12x prefill@128)

JSONbored:
## Summary

The fixed resubmit of #816, which #819 reverted for breaking batched prefill accuracy, together with finer CTA tiles for the FP4 GEMMs it feeds. #816 shipped **two** NVFP4 legs — `ffn_down` and the o-projection (`wo`). Isolated on a 5090, the defect is entirely the o-projection: **`ffn_down` alone clears the accuracy gate, `wo` destroys it.** This restores only the `ffn_down` leg and drops `wo`. 1.12x prefill@128, decode untouched.

## Proof of speedup

| | decode tok/s |
|---|--:|
| before (main) | 104.92 |
| after (this PR) | 104.85 |

| | prefill pp tok/s |
|---|--:|
| before prefill (main) | 6666.9 |
| after prefill (this PR) | 7455.2 |

```text
# One binary, arms env-flipped (SPARKINFER_MUSE_NVFP4_DOWN=0 SKINNY=0 == main's dispatch),
# SWEEP_REPS=9 so the median sits in #800's prefill-graph replay regime.
# RTX 5090, main @ 593bab0. B = main, T = this PR. prefill@128:
B: 6721 6643 6637        mean 6666.9
T: 7473 7461 7432        mean 7455.2   (+11.82%)
# earlier run of the same change, 4/4 interleaved: 6774.2 -> 7523.6 (+11.06%)
# decode unchanged: 104.92 -> 104.85
```

## Why #816 broke, and what changed

`qwen3_gguf_prefill_check` is the gate for this path, and it is easy to mis-run: Muse Glimmer's
batched prefill refuses an int8 KV cache, so the tool's default prints
`[FAIL] prefill_batched unsupported (seed=-1)` and looks broken. It needs
**`SPARKINFER_KV_INT8=0`**. With that set, one leg toggled at a time (same binary, `P=128`,
16 continuation positions):

| configuration | TOP1 | KL |
|---|--:|--:|
| reference (no FP4 legs) | 15/16 | 0.06349 |
| **`ffn_down` leg only** | **14/16** | **0.05623** |
| `ffn_down` + `wo` (i.e. #816) | 2/16 | 4.56426 |

Those last two rows differ by the o-projection alone, so the o-projection carries the defect —
KL 4.6 is not quantization error, it is a wrong result. It is removed here, along with its
weight conversion, its `wo_fp4`/`wo_fp4_sf` pointers and its dead buffers. The `ffn_down` leg
sits at or below the reference's own divergence from the token-by-token path.

Verification for this PR as submitted:

```text
prefill_check, SPARKINFER_KV_INT8=0, RTX 5090
  P=128 cont=64   PR: TOP1 58/64  KL 0.04554   |  ref: TOP1 58/64  KL 0.04732
  P=128 cont=16   PR: TOP1 14/16  KL 0.05623   |  ref: TOP1 15/16  KL 0.06349
  P=512 cont=16   PR: TOP1 16/16  KL 0.00815   |  ref: TOP1 16/16  KL 0.00815   (identical;
                                                   the FP4 path engages only at N=128)
```

Over 64 continuation positions the top-1 agreement is **identical** to the reference and the KL
is marginally lower; the single top-1 difference in the 16-position run is one near-tie flipping,
which the wider sample resolves. Mixed-context sweep (512, 4096 x 2) clean.

## The tiles, and why they belong with this leg

At m = 128 rows the tile grid is a single M-row, so CTAs = n/TileN. With the shipped
128x128x128 tile the hidden-output GEMM this leg adds (n = 6656) launches **52 CTAs on 170 SMs**
and streams weights at 1.06 TB/s, against gate/up's 156 CTAs at 1.24 (peak 1.79). A 128x64 tile
doubles the CTAs in flight, and a 256-deep K tile cuts the mainloop trips; together they take
the FP4 GEMM term from 9.892 ms to 8.906 ms on one trace. Measured alone against a main without
this leg, the tiles are worth +0.15% — they only pay when the skinny GEMM exists, which is why
the two ship together. `SPARKINFER_MUSE_NVFP4_SKINNY` / `_K256` restore the single shipped tile,
and the scale-factor layout is fixed by `Sm1xxBlkScaledConfig` rather than the CTA tile
(`static_assert`ed across variants), so every variant reads the same packed scales.

## Note on the qwen3.6 guard (why this is a re-file)

The previous submission of this work (#817) measured **1.12x prefill@128** with the accuracy
gate green (`top1=0.970`, `KL=0.0906`) and was still auto-closed, because the qwen3.6
no-regression guard reported `measurement unavailable` — every qwen3.6 metric came back
*missing/zero* rather than as a regression against a value.

Nothing in this diff is reachable outside Muse Glimmer: every FP4 entry point sits behind
`c.muse_glimmer` (`muse_nvfp4` in qwen35_prefill.cpp, and the load-time conversion in
qwen35.cpp), so a qwen3.6 run executes exactly main's code. #816 — a strictly larger version of
this change — passed the same guard (`q36_guard_ok: true`). The FP4 translation unit has also
been trimmed from four CUTLASS tile instantiations to the two that actually ship, so the build
is smaller than the one that was measured (99 s for the bench + prefill_check targets here).

If the guard reports missing measurements again I would appreciate a pointer to the harness
log — happy to chase it from our side, but I cannot reproduce a qwen3.6 path that this diff
can reach.

## Lineage

#810 (@andriypolanski) added the SM120 block-scaled FP4 prefill path; #813 (@inference2026)
made it default; #814 (@Paral1995) sharpened the activation quantize this leg feeds; #800's
prefill graph is why the kernel-level saving survives to the wall clock. #816 was ours and
#819 reverted it — thanks for the clear revert note; this is the fixed version it asked for.

### DIFF
--- a/kernels/csrc/cuda/fused/prefill_nvfp4_sm120.cu
+++ b/kernels/csrc/cuda/fused/prefill_nvfp4_sm120.cu
@@ -1,5 +1,6 @@
 #include "sparkinfer/kernels/prefill_nvfp4.h"
 
+#include <cstdlib>
 #include <cuda_bf16.h>
 #include <cutlass/cutlass.h>
 #include <cutlass/float_subbyte.h>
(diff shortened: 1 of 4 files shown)


skyrocket2026:

| metric | value |
|---|---|
| PR decode tok/s | 106.23 |
| PR prefill@128 pp tok/s | 7475.09 |
| prefill speedup vs main | **1.10×** (+10.4%) |
| Polaris receipt | `8ed9c87be5c734da…` — TDX (Intel hardware attestation) |
| commit | `cdb8deb69` |

ok

skyrocket2026:
Closing as REJECT — this carries the same regression as #820 (reverted in #825), despite the title.

Converting `ffn_down` to NVFP4 adds a second resident copy of that weight that nothing can free: `gate_q`/`up_q` get a separate `prefill_gate_q`/`prefill_up_q` pointer released after FP4 conversion, but `down_q` has no prefill-only counterpart — decode reads it directly, so it must stay resident alongside the new `down_fp4` copy this PR adds. On a 32GB card that pushes resident VRAM to ~30GB; at ctx=16384 the KV cache pushes it over budget, the batched-prefill scratch arena fails to allocate, and prefill silently falls back to the sequential token-loop path — correct output, ~21x slower. The eval bot's Muse Glimmer bench only covers ctx 0/128, both comfortably inside budget, so this doesn't show up in the round's numbers.

Whatever this PR fixed from #816, it wasn't that — the `down_fp4`/no-`prefill_down_q`-release shape is identical. Happy to look at a version that also addresses the residency issue (e.g. actually freeing/reusing `down_q` once `down_fp4` exists, or gating the down leg off at long context).

PR #822, opened 2026-08-14
perf(muse-glimmer): block-scaled FP4 attention projection group (1.08x prefill@128)

FranDev132:
## Summary

Muse Glimmer's attention projection group — q, attn_gate, k, v — is the last dense int8 GEMM left in
a prefill layer once the FFN moved to NVFP4. This puts it on the block-scaled path too, and adds the
GEMM tiling that shape needs.

**1. q | attn_gate | k | v as one FP4 operand.** The four weights all project the same `xn` and the
batched prefill already runs them as a *single* grouped int8 GEMM, because at m=128 four separate
launches would be 64, 64, 4 and 4 CTAs of a 170-SM part. The FP4 form has to keep that property, so
the conversion stacks them into one `[2*qdim + 2*kvdim, H]` = `[8704, 6656]` operand rather than
converting each weight on its own: each block is dequantized into its slice of the existing staging
buffer and the whole thing is quantized once, so the scale factors come out in the single atom-tiled
layout the GEMM expects for the combined row count. Prefill then runs one block-scaled GEMM into a
contiguous `[N, 8704]` buffer and splits it out with four strided copies — the destinations are
aliases into the GDN scratch and each wants a tight row stride. That copy is 2.2 MB a layer against
~25 MB less weight traffic (57.9 MB of int8 rows becomes 32.6 MB of FP4 + scales).

Best-effort per layer: a layer whose four weights are not all present, or a box where the conversion
does not fit, simply keeps the int8 grouped path. `SPARKINFER_MUSE_PREFILL_NVFP4_QKV=0` skips the
conversion — with it set, the prefill-accuracy check reproduces `main` exactly (115/128, KL 0.03026),
so the leg is inert when off.

**2. A second GEMM tiling, for grids that are one CTA tall.** The kernel shipped a single
`128x128x128` tile. Batched prefill runs m=128, so the grid is `1 x ceil(n/128)` and the tile alone
decides occupancy — this group is n=8704, i.e. **68 CTAs on a 170-SM RTX 5090**. Halving the N tile
doubles the block count and the memory parallelism with it. Measured per-GEMM with DRAM-cold weights
(rotating six B operands so the 96 MB L2 cannot hold one):

| shape | 128x128x128 | 128x64x256 |
|---|--:|--:|
| `n=6656 k=19968` | 66.2 us | **53.5 us** |
| `n=6656 k=4096` | 18.4 us | **14.3 us** |
| `n=19968 k=6656` | 53.3 us | **51.2 us** |

The wide tile amortizes each A tile over twice as many output columns, so it stays the default; from
m=256 it is ahead at every one of those shapes (1.09x–1.56x). The dispatch is therefore gated on the
grid being a single CTA tall *and* short of one wave, read from the device's own SM count. Both
tilings share the block-scale layout — it depends only on the FP4 vector size, not the tile — so one
quantized operand feeds either, asserted at compile time.

**Cost.** The FP4 copy of the group is ~1.7 GB across 52 layers: 25.5 -> 27.2 GB of a 32 GB card.
Long-context batched prefill is unaffected (ctx16384 measured below).

## Proof of speedup

**Decode tok/s** (Muse Glimmer, ctx=0, median of 5, 3 rotations):

| | decode tok/s |
|---|--:|
| before (main) | 105.69 |
| after (this PR) | 105.68 |

**Prefill pp tok/s** (Muse Glimmer, ctx=128, median of 5 — this PR targets Muse's 128-ctx prefill, so
the table is filled from that model rather than Qwythos):

| | prefill pp tok/s |
|---|--:|
| before prefill (main) | 6690.68 |
| after prefill (this PR) | 7204.74 |

**1.08x prefill@128, decode flat.** Three rotations, alternating, GPU confirmed idle before every
measurement, `main` built in its own tree:

```text
  main               r1: decode@0=105.79  prefill@128=6782.80 pp (18.87 ms)
  qkv-gate fp4       r1: decode@0=105.75  prefill@128=7222.65 pp (17.72 ms)
  qkv fp4, wide tile r1: decode@0=105.71  prefill@128=7132.87 pp (17.95 ms)
  main               r2: decode@0=105.69  prefill@128=6690.68 pp (19.13 ms)
  qkv-gate fp4       r2: decode@0=105.68  prefill@128=7204.74 pp (17.77 ms)
  qkv fp4, wide tile r2: decode@0=105.69  prefill@128=7085.60 pp (18.06 ms)
  main               r3: decode@0=105.68  prefill@128=6636.71 pp (19.29 ms)
  qkv-gate fp4       r3: decode@0=105.64  prefill@128=7202.00 pp (17.77 ms)
  qkv fp4, wide tile r3: decode@0=105.69  prefill@128=7109.94 pp (18.00 ms)
```

The `wide tile` arm is the same binary with `SPARKINFER_NVFP4_NARROW_TILE=0`: the projection group on
FP4 is worth 1.062x on its own, and the tiling that fits its 68-CTA grid adds the remaining 1.4%.

**Accuracy**, `SPARKINFER_KV_INT8=0 qwen3_gguf_prefill_check` (batched prefill vs the token-by-token
path), 128 continuation positions:

| build | TOP1 | KL |
|---|--:|--:|
| main | 115/128 0.8984 | 0.03026 |
| this PR | **119/128 0.9297** | **0.02565** |
| this PR, `NVFP4_QKV=0` | 115/128 0.8984 | 0.03026 |

It improves: the int8 grouped path carries one scale per output row, while the FP4 operand carries
one per 16 values, so the projection weights round closer. 128 positions, not 16 — at 16 this
comparison sits within one token of noise and can invert.

**Qwen3.6 no-regression guard** (shared code: `qwen35.cpp` / `qwen35_prefill.cpp`; Qwen3.6 is MoE and
never enters this path):

```text
  main    : ctx0 d=479.59  ctx512 p=9357.92  ctx4096 p=25446.21  ctx16384 p=26972.51  ctx32768 p=28292.14
  this PR : ctx0 d=479.52  ctx512 p=9351.46  ctx4096 p=25427.95  ctx16384 p=26989.34  ctx32768 p=28308.20
```

Worst ratio 0.9993.

**Muse multi-ctx** (`SPARKINFER_KV_INT8=0`, two rotations — the added resident memory has to not cost
long-context prefill):

```text
  main    r1: ctx512 p=4547.03  ctx4096 p=3362.49  ctx16384 p=2122.85
  this PR r1: ctx512 p=4508.96  ctx4096 p=3319.04  ctx16384 p=2100.63
  main    r2: ctx512 p=4465.47  ctx4096 p=3303.00  ctx16384 p=2091.05
  this PR r2: ctx512 p=4481.95  ctx4096 p=3309.26  ctx16384 p=2091.81
```

Averaged over both rotations: ctx512 0.998, ctx4096 0.994, ctx16384 0.995.

## Notes

`prefill_nvfp4_workspace_bytes` returns the larger of the two tilings' workspaces, so one buffer
serves either. `SPARKINFER_NVFP4_NARROW_TILE=0` pins the shipped tiling and
`SPARKINFER_MUSE_PREFILL_NVFP4_QKV=0` drops the projection leg, so both arms of an A/B come out of a
single binary.

### DIFF
--- a/kernels/csrc/cuda/fused/prefill_nvfp4_sm120.cu
+++ b/kernels/csrc/cuda/fused/prefill_nvfp4_sm120.cu
@@ -11,33 +11,51 @@
 #include <cutlass/util/packed_stride.hpp>
 #include <cute/tensor.hpp>
 
+#include <cstdlib>
+
 namespace sparkinfer::kernels {
 namespace {
 using namespace cute;
 using E4 = cutlass::nv_float4_t<cutlass::float_e2m1_t>;
 using BF = cutlass::bfloat16_t;
-using Tile = Shape<_128, _128, _128>;
 using Cluster = Shape<_1, _1, _1>;
-using Epilogue = typename cutlass::epilogue::collective::CollectiveBuilder<
-    cutlass::arch::Sm120, cutlass::arch::OpClassBlockScaledTensorOp, Tile, Cluster,
-    cutlass::epilogue::collective::EpilogueTileAuto, float, float,
-    BF, cutlass::layout::RowMajor, 8, BF, cutlass::layout::RowMajor, 8,
-    cutlass::epilogue::collective::EpilogueScheduleAuto>::CollectiveOp;
-using Mainloop = typename cutlass::gemm::collective::CollectiveBuilder<
-    cutlass::arch::Sm120, cutlass::arch::OpClassBlockScaledTensorOp,
-    E4, cutlass::layout::RowMajor, 32, E4, cutlass::layout::ColumnMajor, 32, float,
-    Tile, Cluster,
-    cutlass::gemm::collective::StageCountAutoCarveout<
-        static_cast<int>(sizeof(typename Epilogue::SharedStorage))>,
-    cutlass::gemm::collective::KernelScheduleAuto>::CollectiveOp;
-using Kernel = cutlass::gemm::kernel::GemmUniversal<
-    Shape<int, int, int, int>, Mainloop, Epilogue, void>;
-using Gemm = cutlass::gemm::device::GemmUniversalAdapter<Kernel>;
+
+template <class TileShape>
+struct Cfg {
+    using Epilogue = typename cutlass::epilogue::collective::CollectiveBuilder<
+        cutlass::arch::Sm120, cutlass::arch::OpClassBlockScaledTensorOp, TileShape, Cluster,
+        cutlass::epilogue::collective::EpilogueTileAuto, float, float,
+        BF, cutlass::layout::RowMajor, 8, BF, cutlass::layout::RowMajor, 8,
+        cutlass::epilogue::collective::EpilogueScheduleAuto>::CollectiveOp;
+    using Mainloop = typename cutlass::gemm::collective::CollectiveBuilder<
+        cutlass::arch::Sm120, cutlass::arch::OpClassBlockScaledTensorOp,
+        E4, cutlass::layout::RowMajor, 32, E4, cutlass::layout::ColumnMajor, 32, float,
+        TileShape, Cluster,
+        cutlass::gemm::collective::StageCountAutoCarveout<
+            static_cast<int>(sizeof(typename Epilogue::SharedStorage))>,
+        cutlass::gemm::collective::KernelScheduleAuto>::CollectiveOp;
+    using Kernel = cutlass::gemm::kernel::GemmUniversal<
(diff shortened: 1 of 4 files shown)


skyrocket2026:

| metric | value |
|---|---|
| PR decode tok/s | 106.15 |
| PR prefill@128 pp tok/s | 7250.68 |
| prefill speedup vs main | **1.07×** (+7.1%) |
| Polaris receipt | `ed35dcb0630a0ec9…` — TDX (Intel hardware attestation) |
| commit | `39c0a8138` |

ok

PR #826, opened 2026-08-14
perf(muse-glimmer): block-scaled FP4 o-projection with the attention gate fused into its quantize

widecloud:
## Summary

#822 put Muse Glimmer's q|attn_gate|k|v group on the SM120 block-scaled path. This PR does two things on top of it.

**1. The o projection (`wo`, `[H, qdim]`) — the last dense int8 GEMM in a prefill layer — moves to
FP4, with Muse's attention gate folded into the FP4 quantize itself.** `launch_prefill_gate_quant_rows_i8` deliberately leaves `att` UN-GATED: it folds the gate into the *int8* row-quantize, and its own comment says that is only safe when the o projection is certain to read `A_i8`. #816 added an FP4 o-projection that quantized the raw `att`, silently dropping the gate on every layer (TOP1 235/512, KL 0.489), and was reverted in #819. This PR adds `launch_prefill_nvfp4_gate_quant_a`, so the block-scaled leg applies the gate on the way into its own operand and **no path ever reads an un-gated `att`** — the int8 fold is skipped when the FP4 leg runs, and if the GEMM declines, `att` is gated before the int8 arm reads it.

**2. The wide tile was one K-step too shallow, and gate/up was on the wrong tile.** `Wide` shipped as `128x128x128`, and `prefer_narrow` routed anything under `sms` CTAs to the narrow tile — which includes gate/up (n=19968, 156 CTAs of 170). But 156 CTAs is already ~92% of one wave, so halving N there only doubles the A-tile re-reads. Deepening `Wide` to `128x128x256` and requiring the *narrow* grid to still fit one wave (`2*ceil(n/128) <= sms`) keeps gate/up wide and leaves `wo` (52 CTAs) and q|gate|k|v (68) narrow. Measured cold-L2 at m=128, microseconds:

| shape | narrow `128x64x256` | wide `128x128x256` |
|---|--:|--:|
| gate/up  n=19968 k=6656 | 60.51 | **56.19** |
| wo       n=6656  k=4096 | **21.63** | 25.02 |
| q/gate/k/v n=8704 k=6656 | **31.84** | 39.14 |

**Deliberately NOT included: `ffn_down`.** #820 was reverted in #825 because `down_q` has no prefill-only counterpart — decode reads it directly — so its FP4 copy is a permanent +3.9 GB that nothing can free, and at ctx 16384 the KV cache plus the batched-prefill scratch arena no longer fit. `wo` is the same class of tensor but 4.8x smaller (~0.8 GB), and it is gated on an **all-or-nothing VRAM preflight**: the whole 52-layer set is costed against free VRAM plus a 3 GB reserve (sized for the ctx-32k KV cache) *before any layer converts*. If it does not fit, nothing converts and the model behaves exactly like `main`. A partial conversion is the worst outcome available — it spends the memory and still leaves most layers on int8. `SPARKINFER_MUSE_NVFP4_WO=0` disables the leg; `SPARKINFER_MUSE_NVFP4_RESERVE_MB` tunes the reserve.

Also included: a stub for every new `launch_prefill_nvfp4_*` entry point in `prefill_nvfp4.cu`, so the portable (non-CUTLASS) build still links. Verified with `-DSPARKINFER_NVFP4=OFF`.

## Proof of speedup

**Decode tok/s** (ctx=0 — this PR is prefill-only):

| | decode tok/s |
|---|--:|
| before (main) | 103.08 |
| after (this PR) | 102.89 |

**Prefill pp tok/s** (Muse Glimmer, `ctx=128` — the scored prefill shape):

| | prefill pp tok/s |
|---|--:|
| before prefill (main `dd571f9`) | 6716.29 |
| after prefill (this PR) | 7083.23 |

**+5.56% prefill@128**, median of 3 alternating pairs, main `dd571f9`, two trees built from scratch, base re-sampled between every arm, 5 reps inside each process.

```text
main p1 | prefill_pp: 6789.5231 | VRAM 27.3 GB
PR   p1 | prefill_pp: 7108.2212 | VRAM 28.1 GB
main p2 | prefill_pp: 6716.2867 | VRAM 27.3 GB
PR   p2 | prefill_pp: 7074.1905 | VRAM 28.1 GB
main p3 | prefill_pp: 6709.8396 | VRAM 27.3 GB
PR   p3 | prefill_pp: 7083.2327 | VRAM 28.1 GB
```

|---|--:|--:|--:|
| p1 | 6789.52 | 7108.22 | +4.69% |
| p2 | 6716.29 | 7074.19 | +5.33% |
| p3 | 6709.84 | 7083.23 | +5.56% |

## Accuracy

`qwen3_gguf_prefill_check <gguf> 128 512`, `SPARKINFER_KV_INT8=0`:

| | TOP1 | KL |
|---|--:|--:|
| main `dd571f9` | 485/512 = 0.9473 | 0.00956 |
| this PR | 481/512 = 0.9395 | 0.01073 |

One more tensor on FP4, well clear of the 0.90 bar — and with the gate applied, not the 235/512 / KL 0.489 that #816 produced from this same tensor.

### DIFF
--- a/kernels/csrc/cuda/fused/prefill_nvfp4.cu
+++ b/kernels/csrc/cuda/fused/prefill_nvfp4.cu
@@ -10,6 +10,8 @@ size_t prefill_nvfp4_scale_bytes_a(int, int) { return 0; }
 size_t prefill_nvfp4_scale_bytes_b(int, int) { return 0; }
 size_t prefill_nvfp4_workspace_bytes(int, int, int) { return 0; }
 bool launch_prefill_nvfp4_quant_a(const void*, void*, void*, int, int, cudaStream_t) { return false; }
+bool launch_prefill_nvfp4_gate_quant_a(const void*, const void*, void*, void*, int, int,
+                                       cudaStream_t) { return false; }
 bool launch_prefill_nvfp4_quant_b(const void*, void*, void*, int, int, cudaStream_t) { return false; }
 bool launch_prefill_nvfp4_gemm(const void*, const void*, const void*, const void*, void*, int, int,
                                int, void*, cudaStream_t) { return false; }
--- a/kernels/csrc/cuda/fused/prefill_nvfp4_sm120.cu
+++ b/kernels/csrc/cuda/fused/prefill_nvfp4_sm120.cu
@@ -42,7 +42,7 @@ struct Cfg {
 // Two N-tilings of the same GEMM. The 128-wide tile reuses each A tile across twice as many
 // output columns and is the right default; the 64-wide one exists only to fill the machine when
 // the wide grid cannot (see nvfp4_prefer_narrow).
-using Wide = Cfg<Shape<_128, _128, _128>>;
+using Wide = Cfg<Shape<_128, _128, _256>>;
 using Narrow = Cfg<Shape<_128, _64, _256>>;
 
 using Gemm = typename Wide::Gemm;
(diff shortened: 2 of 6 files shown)


skyrocket2026:

| metric | value |
|---|---|
| PR decode tok/s | 106.27 |
| main decode tok/s | 106.38 |
| PR prefill@128 pp tok/s | 7660.58 |
| main prefill@128 pp tok/s | 7355.53 |
| prefill speedup vs main | **1.04×** (+4.1%) |
| Polaris receipt | `17d37a61c83ae1c7…` — TDX (Intel hardware attestation) |
| commit | `7ff53bb43` |

ok

skyrocket2026:

PR #827, opened 2026-08-14
perf(muse-glimmer): block-scaled FP4 ffn_down behind an all-or-nothing VRAM preflight

widecloud:
## Summary

`ffn_down` is the largest weight in a Muse layer and the last dense projection still on the int8 path, now that gate/up (#810/#813), the `q|gate|k|v` group (#822) and the o projection (#826) have moved to NVFP4. Converting it is worth more than those three together. It has been blocked by memory, not by the kernel: unlike gate/up it has no prefill-only copy to release — decode reads the same tensor — so the FP4 form is a real +3.9 GB.

`SPARKINFER_MUSE_NVFP4_DOWN=0` reproduces `main` exactly; `SPARKINFER_MUSE_NVFP4_DOWN_RESERVE_MB` tunes the reserve.

## Proof of speedup

**Decode tok/s** (ctx=0 — this PR is prefill-only):

| | decode tok/s |
|---|--:|
| before (main) | 103.27 |
| after (this PR) | 102.91 |

**Prefill pp tok/s** (Muse Glimmer, `ctx=128` — the scored prefill shape):

| | prefill pp tok/s |
|---|--:|
| before prefill (main `aca6225`) | 7094.10 |
| after prefill (this PR) | 8351.68 |

**+17.7% prefill@128**, median of 3 alternating pairs against `main` @ `aca6225`, arms alternated inside one run so thermal drift hits both, 5 reps inside each process.

```text
main p1 | prefill_pp: 7224.9315 | decode ctx0: 104.3518 | VRAM 28.1 GB
PR   p1 | prefill_pp: 8401.5605 | decode ctx0: 103.6433 | VRAM 32.1 GB
main p2 | prefill_pp: 7094.1006 | decode ctx0: 103.2720 | VRAM 28.1 GB
PR   p2 | prefill_pp: 8351.6760 | decode ctx0: 102.9119 | VRAM 32.1 GB
main p3 | prefill_pp: 7079.2466 | decode ctx0: 102.8720 | VRAM 28.1 GB
PR   p3 | prefill_pp: 8331.3454 | decode ctx0: 102.6464 | VRAM 32.1 GB
```

|---|--:|--:|--:|
| p1 | 7224.93 | 8401.56 | +16.29% |
| p2 | 7094.10 | 8351.68 | +17.73% |
| p3 | 7079.25 | 8331.35 | +17.69% |

Isolated 3-pair A/B of the down leg alone (o-projection off on both arms, same binary, env-toggled, so only this tensor differs): **+15.17% / +16.06% / +16.29%**, monotone.

Decode is flat: the base arm itself drifts 104.35 → 102.87 across the same run, which is larger than the 0.35% gap between the arms.

Resident VRAM 28.1 → 32.1 GB at `max_seq` 2048. At `max_seq` 32912 the preflight declines the leg and resident stays at `main`'s 28.1 GB.

## Accuracy

`qwen3_gguf_prefill_check <gguf> 128 512`, `SPARKINFER_KV_INT8=0`, same seed (123192) both arms:

| | TOP1 | KL |
|---|--:|--:|
| main `aca6225` | 481/512 = 0.9395 | 0.01073 |
| this PR | 469/512 = 0.9160 | 0.01818 |

Clear of the 0.90 bar. This is the largest single tensor in the model moving to 4-bit, so it is the biggest accuracy step of the FP4 series — worth stating explicitly rather than burying. Decode numerics are untouched (`down_q` still serves every decode and every non-128 prefill shape), so the scored teacher-forced gate, which runs the AR decode path, is unaffected.

### DIFF
--- a/kernels/csrc/cuda/fused/prefill_nvfp4_sm120.cu
+++ b/kernels/csrc/cuda/fused/prefill_nvfp4_sm120.cu
@@ -142,26 +142,47 @@ __global__ void swiglu_quant_rows(const __nv_bfloat16* __restrict__ gate,
                                   const __nv_bfloat16* __restrict__ up,
                                   unsigned char* dst, cutlass::float_ue4m3_t* sf,
                                   int rows, int cols, Layout layout) {
-    constexpr int V = 16, LPG = V / 2;
+    // 8 values per lane instead of 2, so two lanes cover a 16-value scale group. The two operand
+    // reads become one 16-byte load each instead of four 4-byte loads, the store becomes one
+    // 4-byte store instead of four 1-byte ones, and the amax butterfly collapses from three
+    // shuffles to one. This kernel moves 11.7 MB per layer in ~12 us -- 0.98 TB/s of a 1.79 TB/s
+    // part -- so it was issue-bound on narrow accesses, not bandwidth-bound.
+    // Bit-identical: max is order-independent, so each group keeps exactly the same scale, and
+    // every value keeps the same SiLU-in-float, round-once-to-bf16, x / float(qs) sequence.
+    constexpr int V = 16, VPL = 8, LPG = V / VPL;   // 2 lanes per scale group
     const int glane = threadIdx.x & (LPG - 1);
     const int groups = rows * (cols / V);
     const int stride = (gridDim.x * blockDim.x) / LPG;
-    const unsigned gmask = 0xFFu << (threadIdx.x & 24);
     for (int grp = (blockIdx.x * blockDim.x + threadIdx.x) / LPG;
          grp < groups; grp += stride) {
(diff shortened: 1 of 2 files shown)


skyrocket2026:

| metric | value |
|---|---|
| PR decode tok/s | 107.87 |
| main decode tok/s | 107.99 |
| PR prefill@128 pp tok/s | 8747.13 |
| main prefill@128 pp tok/s | 8252.76 |
| prefill speedup vs main | **1.06×** (+6.0%) |
| Polaris receipt | `01efe4b007b0af0a…` — TDX (Intel hardware attestation) |
| commit | `45ff39dc3` |

ok

skyrocket2026:

PR #830, opened 2026-08-15
perf(qwen3.8): FP8-direct GDN prefill projections (1.27x prefill@128) + the Muse Glimmer tensor-core/node stack

JSONbored:
## Summary

**Qwen3.8-27B prefill@128: FP8-direct GDN in-projections — 1.27x prefill@128, decode flat.**
(Plus the Muse Glimmer tensor-core/node-count stack this PR originally carried — env-gated,
inert on Qwen3.8, +6.4% muse prefill@128 re-verified on this head.)

The 48 linear-attention layers dequantize their FP8 `in_proj_qkv`/`in_proj_z` weights to bf16
**every batched-prefill pass** and run the naive bf16 GEMM (96 launches x ~197 us — the largest
remaining term after #835/#837). This PR stages f32 copies of the per-row scales once at load
and runs `launch_prefill_gemm_fp8` **directly on the resident checkpoint FP8 payloads** (#832's
own bytes — zero weight-side quantization error, zero extra VRAM, no copies): one e4m3
A-quantize of `xn` feeds both GEMMs. Decode is untouched. `SPARKINFER_Q38_GDN_PREFILL=off`
restores the materialize path (one-binary A/B).

Numerics, stated precisely: the e4m3 activation quantize reassociates the GDN recurrence
inputs, so greedy continuations after a 128-token prefill can diverge from the bf16-materialize
path's (measured: 2 common tokens, then divergence; deterministic across repeated runs). This is
the class of behavior main already ships: with #831/#837's merged FP4 FFN *disabled at runtime*
(`SPARKINFER_PREFILL_FFN_CHUNK=127`), main's own continuations diverge from its defaults at
token 1 (measured, same harness). The teacher-forced eval-corpus score is byte-identical
(top1 1.000 / KL 0.0000) — the scored accuracy surface is unaffected.

Also carried: the Muse Glimmer legs (bf16 tensor-core windowed attention, stacked gate|up FP4 operand,
q|gate|k|v read in place, fused sandwich-norm pairs) — muse batched-prefill accuracy gate
*improves* (15/16, KL 0.067 vs main's 0.090, verified at prefixes 128/256/512), q36 guard
0.997-1.000x at ctx 0/512/4k/16k/32k on this stack (the flagged 512 cell re-measured 1.0002x at
REPS=8).

Prior art: #835 (@Paral1995) made the ctx=128 batched pass exist; #837 (@flashatten) took the
FFN to checkpoint-native NVFP4; #832 (@flashatten) owns the FP8 residency this PR's prefill leg
reads directly; #831 (@skyrocket2026) the bring-up all of it stands on. The GDN in-projections
were the last per-pass weight materialize in the pass.

Concurrent filings on the same observation: #838 (@flashatten, filed 08:18Z) and #839
(@FranDev132) target the same GDN prefill materialize — this PR went ready-for-review at
08:19Z, developed independently (the staged-f32-scales approach here vs #838's kernel-side
packed-payload read; neither derived from the other, and this PR's lab history predates both
filings). Whichever lands first, credit where due — the observation was evidently in the air.

## Proof of speedup

>

**Decode tok/s** (end-to-end — prefill-only change, decode path untouched):

| | decode tok/s |
|---|--:|
| before (main) | 84.40 |
| after (this PR) | 84.30 |

**Prefill pp tok/s** (Qwen3.8-27B compressed-tensors checkpoint, ctx=128 — the scored prefill):

| | prefill pp tok/s |
|---|--:|
| before prefill (main) | 1614.2 |
| after prefill (this PR) | **2047.1 (+26.8%)** |

```text
# RTX 5090, unsloth/Qwen3.8-27B-NVFP4, main @ 02770d5 (post-#837). ONE binary, arms env-flipped
# (arm0 = SPARKINFER_Q38_GDN_PREFILL=off == main's dispatch), SWEEP_REPS=5, interleaved:
main:     1617.0  1611.4            mean 1614.2   (reproduces post-#837 main)
this PR:  2051.5 2046.4 2047.8 2046.5 2044.9 2045.3   mean 2047.1   (+26.8%)
# decode@128 across all arms: 84.16-84.52 (flat)
# determinism: every arm reproduces itself token-for-token across repeated runs
# teacher-forced eval-corpus score dump: byte-identical to main (top1 1.000, KL 0.0000)
# muse re-verification on this head: 9183.3 -> 9771.4 pp (+6.4%), accuracy 15/16 / KL 0.06724
# q36 guard (decode+prefill, ctx 0/512/4k/16k/32k, tol 0.98): all pass; the 512-prefill cell
# re-measured at REPS=8 x 2 pairs: 1.0002x
```

### DIFF
--- a/kernels/csrc/cuda/fused/batched_prefill.cu
+++ b/kernels/csrc/cuda/fused/batched_prefill.cu
@@ -10,6 +10,7 @@
 // Numerics mirror the decode kernels (qwen36.cu / rope.cu) so the KV cache and recurrent
 // state left behind are faithful to the token-by-token path.
 
+#include <cstdint>
 #include <cuda_bf16.h>
 #include <cuda_fp16.h>
 #include <cuda_runtime.h>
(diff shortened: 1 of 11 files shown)


JSONbored:
Additional validation: batched-prefill accuracy at longer prefixes (the check that caught #816 at prefix 512). Prefixes 256/512 sit outside the FP4 path's N==128 gate, so they isolate the tensor-core attention numerics — the only non-bit-identical leg:

| prefix | main 7e9f79f (TOP1 / KL) | this PR (TOP1 / KL) |
|---|---|---|
| 128 | 14/16 / 0.09030 | 15/16 / 0.06724 |
| 256 | 15/16 / 0.00752 | 15/16 / 0.00807 |
| 512 | 14/16 / 0.00680 | **16/16** / 0.00812 |

Same box, same binary, arms env-flipped (`SPARKINFER_MUSE_ATTN_MMA=0` + the three bit-identical legs off = main's dispatch). TOP1 equal or better at every prefix; KL well under the 0.10 bar throughout.

skyrocket2026:

| metric | value |
|---|---|
| PR decode tok/s | 84.50 |
| main decode tok/s | 84.53 |
| PR prefill pp | 2050.21 |
| main prefill pp | 1614.42 |
| prefill vs main | +27.0% |
| Polaris receipt | `ab2165b4526c1b8a…` — TDX (Intel hardware attestation) |
| commit | `cb38ed268` |

ok

PR #835, opened 2026-08-15
perf(qwen3.8): batched prefill over a bf16 KV cache (14.9x prefill@128)

Paral1995:
## Summary

Batched prefill required an int8 KV cache, and the bench only turns int8 on at ctx>=4096, so
Qwen3.8's prefill@128 declined to the sequential per-token path — 87.70 pp against this model's
own 83.78 tok/s decode. Prefill being barely faster than decode is the signature of filling 128
positions with 128 full weight-streaming passes instead of one batched GEMM pass. Two bf16 twins
of the existing int8 prefill kernels remove that requirement.

- `pf_qknorm_rope_kv_bf16_kernel` — `pf_qknorm_rope_kv_int8_kernel` with the quantize/scale
  dropped and the pool typed bf16, keeping the same NeoX `(i, i+half)` partial rotation.
- `pf_attn_bf16_paged_kernel<256>` — `pf_attn_int8_paged_kernel<256>` with the dequant dropped,
  keeping the same one-warp-per-query schedule and the same fp32 online softmax.

Both write exactly the bf16 KV a `forward_token` decode step writes when int8 KV is off, so a
decode continuing from this cache reads a consistent one. The int8 branch is untouched and still
taken whenever int8 KV is on; the bf16 branch is only reachable for `head_dim` 128/256.
`SPARKINFER_PREFILL_BF16_KV=0` restores the decline, so both arms come out of one binary.

## Proof of speedup

**Decode tok/s** (end-to-end, `bench_sweep_run <dir> 128 128 5`, compressed-tensors checkpoint):

| | decode tok/s |
|---|--:|
| before (main) | 83.78 |
| after (this PR) | 83.88 |

**Prefill pp tok/s** (same sweep, same model load, ctx=128):

| | prefill pp tok/s |
|---|--:|
| before prefill (main) | 87.70 |
| after prefill (this PR) | 1308.78 |

**+1392% prefill@128 (14.92x)**, decode unchanged. Both arms from one binary, interleaved
round-robin, five reps inside each process, median of five rounds, with
`SPARKINFER_BENCH_PROMPT_FILE` set to the tokenized `bench/scripts/bench_prompt.txt` (225 ids).
A third arm byte-identical to the baseline ran every round as a noise floor:

```text
  round 1: main pp=88.0880  batched pp=1313.7707  ctrl pp=87.9644
  round 2: main pp=87.8207  batched pp=1312.1579  ctrl pp=87.7219
  round 3: main pp=87.7000  batched pp=1308.7811  ctrl pp=87.6667
  round 4: main pp=87.6380  batched pp=1308.0574  ctrl pp=87.6470
  round 5: main pp=87.6278  batched pp=1308.6681  ctrl pp=87.7011

  prefill  main 87.70 | batched 1308.78 | control 87.70
  +1392.34% (14.92x), 5/5 rounds, noise floor +0.00%
  decode   main 83.78 | batched   83.88 | control 83.75   (no regression)
```

## Correctness

This moves ctx=128 onto the batched prefill path, which has its own numerics — as it already does
for every model and context that uses it today. That is worth measuring rather than asserting, so
the prefix was filled through each path and the same positions then scored
(`SPARKINFER_PREFIX_CACHE_LEN=64`, `bench/scripts/accuracy_compare_pair.py`):

```text
  this PR's bf16 batched   vs  the existing int8 batched : top1 0.973  KL 0.021
  the existing int8 batched vs  sequential fill          : top1 0.892  KL 0.107
  this PR's bf16 batched   vs  sequential fill           : top1 0.919  KL 0.111

  PPL: sequential 3.116 | int8 batched 3.025 | bf16 batched 2.972
```

The new kernels track the batched path already in the tree (0.973); the batched-vs-sequential gap
is pre-existing and slightly larger for the int8 path than for this one; both batched paths score
lower perplexity than the sequential fill.

Stated plainly, because the harness cannot show it: the differential accuracy gate runs
`qwen3_gguf_score` with no prefix cache, i.e. sequential `forward_token`, so it never exercises
batched prefill and will report top1 1.000 / KL 0.000 for this change regardless of what it did.
The numbers above are the check that gate cannot perform, not a substitute for it.

```text
ctest: 100% tests passed, 21/21
```

## Qwen3.6 no-regression

`batched_prefill.cu` and `qwen35_prefill.cpp` are shared, so the cross-model sweep was run rather
than argued:

```text
       0     479.14 ->   479.26
     512     473.17 ->   472.82      9342.52 ->   9291.82
    4096     455.18 ->   455.12     25131.05 ->  25253.05
```

Max deviation 0.54%.

### DIFF
--- a/kernels/csrc/cuda/fused/batched_prefill.cu
+++ b/kernels/csrc/cuda/fused/batched_prefill.cu
@@ -1055,6 +1055,133 @@ __global__ void pf_attn_int8_paged_kernel(
 }
 
 // ============================================================================
+// bf16-KV batched prefill for the hybrid dense models (Qwen3.8): a NeoX-RoPE
+// QK-norm/append and a full-causal paged attention that need no int8 KV.
+//
+// Batched prefill previously required an int8 KV cache, and the bench only turns int8 on at
+// ctx>=4096, so Qwen3.8's prefill@128 declined to the sequential per-token path -- 88.0 tok/s
+// against 84.2 tok/s of decode, i.e. 128 full weight-streaming passes to fill 128 positions.
+// These are bf16 twins of pf_qknorm_rope_kv_int8_kernel and pf_attn_int8_paged_kernel<256>:
+// same schedule, same fp32 online softmax, same NeoX (i, i+half) partial rotation, with the
+// int8 quantize/dequantize removed and the pool typed bf16 -- which is exactly the KV a
+// forward_token decode step writes when kv8 is off, so a decode continuing from this cache
+// reads a consistent one.
+// ============================================================================
+__global__ void pf_qknorm_rope_kv_bf16_kernel(
+    __nv_bfloat16* __restrict__ q, __nv_bfloat16* __restrict__ k, const __nv_bfloat16* __restrict__ v,
+    const __nv_bfloat16* __restrict__ q_w, const __nv_bfloat16* __restrict__ k_w,
+    __nv_bfloat16* __restrict__ k_pool, __nv_bfloat16* __restrict__ v_pool,
+    const int* __restrict__ block_table,
+    int n_q_heads, int n_kv_heads, int head_dim, int rotary_dim, float theta, float eps,
+    int block_size, int max_blocks_per_seq) {
+    const int tok  = blockIdx.x;
+    const int unit = blockIdx.y;
+    const int t    = threadIdx.x;
+    const int pos  = tok;                              // prefill: position == token index
+    const int blk  = pos / block_size, within = pos % block_size;
+    const int phys = block_table[blk];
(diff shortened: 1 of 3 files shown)


skyrocket2026:

| metric | value |
|---|---|
| PR decode tok/s | 84.50 |
| main decode tok/s | 84.47 |
| PR prefill pp | 1342.58 |
| main prefill pp | 88.37 |
| prefill vs main | +1419.2% |
| Polaris receipt | `7eee14626de527fa…` — TDX (Intel hardware attestation) |
| commit | `f9bd9bb57` |

ok

skyrocket2026:

PR #839, opened 2026-08-15
fix(qwen3.8): restore the dropped FFN residual in the block-scaled prefill

FranDev132:
## Summary

**This is a correctness fix, not a speedup.** Rescoped after rebase: the block-scaled prefill work
this branch originally carried is now on main, and what remains is a live bug that path exposes.

`prefill_batched_run`'s block-scaled FFN branch lands its output in `ao` and then `continue`s, while
the epilogue folds `ao` into the residual stream only when `!ffn_fused`. That invariant held while
Muse Glimmer was the only model reaching the branch — `ffn_fused` is `!c.muse_glimmer && ...`, so it
is always false there. Qwen3.8-27B is the first non-Muse model in that branch and it *does* fuse the
residual, so the epilogue skips the add and **every block-scaled layer computes its FFN output and
then discards it**.

The fix adds the chunk residual inside the branch. Declining the block-scaled arms would also restore
it, but at the cost of the block-scaled down GEMM, so this keeps them.

`qwen3_gguf_prefill_check` grows compressed-tensors directory support. It opened GGUF only, so the
model whose prefill@128 is scored has had **no batched-prefill correctness gate at all** —
`qwen3_gguf_score` cannot stand in for one, because it drives `forward_token` and never calls
`prefill_batched`, so it emits a byte-identical dump whether this path is right or catastrophically
wrong. That is why this survived several rounds of prefill work.

## Correctness

`qwen3_gguf_prefill_check <dir> 128 64` with `SPARKINFER_KV_INT8=0` — fills the KV cache
token-by-token and via `prefill_batched` over the same prompt, teacher-forces the same 64
continuation tokens, compares per-position logits. Both arms from one binary, RTX 5090 (`sm_120`),
`unsloth/Qwen3.8-27B-NVFP4`, against `origin/main` @ `222ad65`:

| | top1 | KL(token \|\| batched) |
|---|--:|--:|
| before (main) | 24/64 = 0.3750 | 1.04584 |
| after (this PR) | **57/64 = 0.8906** | **0.02879** |

Longer prefixes on the fixed build stay in the same band the batched path holds elsewhere:
512 → 16/16 KL 0.00457, 1024 → 15/16 KL 0.00570, 4096 → 15/16 KL 0.00434.

## Throughput

No speedup is claimed. The added residual is one `N*H` bf16 add per FFN chunk; it costs ~0.6% of
prefill@128, which is inside the 2% regression floor. Reported for completeness, not as a gain —
`bench_sweep_run <dir> 128 128 5`, two separately-configured build trees, alternated, median of 3:

|---|--:|--:|
| before (main) | 84.43 | 2439.31 |
| after (this PR) | 84.36 | 2424.58 |

```text
  round 1  main prefill=2444.6701 decode=84.5484   this PR prefill=2429.7398 decode=84.4712
  round 2  main prefill=2439.3092 decode=84.4253   this PR prefill=2424.5819 decode=84.3623
  round 3  main prefill=2428.4906 decode=84.3183   this PR prefill=2416.4693 decode=84.2781
```

The prefill numbers on main above are produced by the path this PR shows to be numerically wrong, so
they are not a like-for-like baseline for throughput purposes — they are here to show the fix does
not regress beyond the floor.

### DIFF
--- a/runtime/CMakeLists.txt
+++ b/runtime/CMakeLists.txt
@@ -94,7 +94,8 @@ if(BUILD_EXAMPLES)
                           nlohmann_json::nlohmann_json)   # same reason as qwen3_gguf_bench above
     add_executable(qwen3_gguf_prefill_check examples/qwen3_gguf_prefill_check.cpp)
     target_include_directories(qwen3_gguf_prefill_check PRIVATE include)
-    target_link_libraries(qwen3_gguf_prefill_check PRIVATE sparkinfer_runtime CUDA::cudart)
+    target_link_libraries(qwen3_gguf_prefill_check PRIVATE sparkinfer_runtime CUDA::cudart
+                          nlohmann_json::nlohmann_json)   # same reason as qwen3_gguf_bench above
     add_executable(qwen3_gguf_dflash_check examples/qwen3_gguf_dflash_check.cpp)
     target_include_directories(qwen3_gguf_dflash_check PRIVATE include)
     target_link_libraries(qwen3_gguf_dflash_check PRIVATE sparkinfer_runtime CUDA::cudart)
--- a/runtime/examples/qwen3_gguf_prefill_check.cpp
+++ b/runtime/examples/qwen3_gguf_prefill_check.cpp
@@ -15,6 +15,7 @@
 #include "sparkinfer/models/qwen35.h"
 #include "sparkinfer/moe/engine.h"
 #include "qwen3_gguf_config.h"
+#include "qwen_checkpoint.h"   // after models/qwen35.h: it references sparkinfer::Qwen35Model
 
 #include <cuda_runtime.h>
 #include <cstdio>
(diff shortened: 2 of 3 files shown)


skyrocket2026:

| metric | value |
|---|---|
| PR decode tok/s | 84.50 |
| main decode tok/s | 84.53 |
| PR prefill pp | 2027.66 |
| main prefill pp | 1614.42 |
| prefill vs main | +25.6% |
| commit | `3b5f3b94a` |

ok

skyrocket2026:
## sparkinfer qwen38 auto-eval — `eval-qwen38:none`

| metric | value |
|---|---|
| **label** | `eval-qwen38:none` |
| PR decode tok/s | 84.45 |
| main decode tok/s | 84.54 |
| PR prefill pp | 2430.80 |
| main prefill pp | 2450.41 |
| prefill vs main | -0.8% |
| commit | `ac314b028` |

within significance gate — not a verified decode@128 improvement

skyrocket2026:
## Closed: sparkinfer qwen38 auto-eval — `eval-qwen38:none`

This PR's Qwen3.8-27B decode@128 speed measured **-0.1%** vs main, with no verified decode@128 or prefill@128 improvement — closing automatically. This bot evaluates every eligible PR in the repo against Qwen3.8-27B's decode AND prefill@128 speed specifically, regardless of what the PR is actually about — a close here isn't a judgment on the PR's purpose, just that it didn't move these particular metrics. Reopen (or open a fresh PR) if you have a fix or a different approach.

PR #842, opened 2026-08-15
perf(qwen3.8): cp.async-staged skinny prefill GEMM for the 48-wide GDN gate projections (1.37x prefill@128)

inference2026:
## Summary

Qwen3.8-27B has 48 Gated-DeltaNet v-heads, so its `ssm_alpha` / `ssm_beta` projections are `[48, H]`
— one 16-wide column tile past the skinny prefill GEMM's `NMAX = 32` cap. They therefore fell
through to `pf_gemm_kernel`, whose 128x128 output tile grids a `128x48x5120` GEMM to **a single
block**: one SM runs the whole thing, 96 times per prefill. nsys (`--cuda-graph-trace=node`) puts
that one kernel at 197.95 us x 96 = 19.0 ms of a 60.0 ms prefill@128.

Two changes, both inside `prefill_gemm_skinny`:

* `n_out` is already masked inside the kernel, so the column tile only needs rounding up to a
  multiple of 16. Instantiate 48 and 64 next to 32 and dispatch on `n_out`.
* Double-buffer the K stream with `cp.async`, the way `pf_gemm_kernel` already does. One warp per
  16x16 output tile leaves too few resident warps to cover a plain global load, so the K loop paid
  full memory latency on all `K/KT` trips. Widening the tile alone was worth +15.0%; the staging
  fix took it to +33.4%.

Halving the row strip to double the grid (4 blocks -> 8 at M=128) measures **worse**, +5.6% against
+15.0%, so `BT` stays 32: the total mma count is fixed by the output shape, and what costs time is
staging rounds per block, not SM count. `SPARKINFER_PREFILL_SKINNY_BT` re-runs that A/B and
`SPARKINFER_PREFILL_GEMM_SKINNY=0` restores the old path.

Same wmma `16x16x16` fragments accumulated over K in the same ascending order as `pf_gemm_kernel`,
so this is bit-identical, not just close. Verified two ways on the NVFP4 checkpoint: a 96-token
greedy continuation after a batched prefill is identical between the two paths (and between
`BT=32` and `BT=64`), and the teacher-forced score dumps are byte-identical — top-1 101/101 =
1.000, KL 0.0000 nats, PPL 3.281 both sides.

Models with 32 v-heads keep their existing `(BT=32, NMAX=32)` instantiation and shape.

## Proof of speedup

| | decode tok/s |
|---|--:|
| before (main) | 84.08 |
| after (this PR) | 84.07 |

| | prefill pp tok/s |
|---|--:|
| before prefill (main) | 2385.10 |
| after prefill (this PR) | 3263.22 |

```text
# bench_sweep_run <compressed-tensors dir> 128 128 5 -- the harness the eval itself uses
# (qwen3_gguf_bench ... sweep, ctx=128, median of 5 reps). RTX 5090 sm_120, main @ 222ad65.
# SPARKINFER_PREFILL_GEMM_SKINNY=0 is the pf_gemm fallback, i.e. the pre-change path, so both
# arms come out of one binary and one model load.

r1 before  decode_tps=84.1524  prefill_pp=2390.4719
r1 after   decode_tps=84.1308  prefill_pp=3273.3356
r2 before  decode_tps=84.0918  prefill_pp=2362.3477
r2 after   decode_tps=84.0949  prefill_pp=3243.0996
r3 before  decode_tps=84.0685  prefill_pp=2385.7997
r3 after   decode_tps=84.0455  prefill_pp=3267.4372
r4 before  decode_tps=83.9813  prefill_pp=2384.3886
r4 after   decode_tps=83.9199  prefill_pp=3259.0024

prefill@128  2385.10 -> 3263.22 pp  (+36.81%); ranges disjoint, before max 2390.47 <
             after min 3243.10.
decode@128   84.08 -> 84.07 tok/s. That -0.01% is noise -- it is smaller than the drift
             within either arm across the four rounds -- and no decode kernel is touched.

The same change measured on two earlier bases, for what it is worth: +22.1% on 02770d5
(1578.8 -> 1927.5 pp) and +33.4% on 78ab275 (2190.2 -> 2921.7 pp). It grows as the rest of
prefill gets faster, because the kernel it removes is a fixed 19.0 ms.

Isolated-kernel evidence (nsys --cuda-graph-trace=node; extra evidence, not the before/after):
the ssm_alpha/ssm_beta GEMM runs 96 times per prefill at 197.95 us each before the change and
118.56 us with only the column tile widened.
```

### DIFF
--- a/kernels/csrc/cuda/fused/prefill_gemm_skinny.cu
+++ b/kernels/csrc/cuda/fused/prefill_gemm_skinny.cu
@@ -34,6 +34,7 @@
 
 #include <cuda_runtime.h>
 #include <cuda_bf16.h>
+#include <cuda_pipeline.h>
 #include <mma.h>
 
 #include <cstdlib>
@@ -45,6 +46,12 @@ namespace {
 
 constexpr int SK_PAD = 8;    // break the power-of-two smem row stride
 
+// 16-byte cp.async when in-bounds, zero-fill otherwise. Same helper pf_gemm_kernel uses.
+__device__ __forceinline__ void sk_cp16(void* dst, const void* src, bool pred) {
+    if (pred) __pipeline_memcpy_async(dst, src, 16);
+    else      *reinterpret_cast<uint4*>(dst) = make_uint4(0u, 0u, 0u, 0u);
+}
+
 // One block per BT-row strip; NMAX output columns held at once; K streamed KT at a time.
 // WARPS warps, each owning one 16x16 wmma accumulator tile of the [BT, NMAX] output.
 template <int BT, int NMAX, int KT, int WARPS>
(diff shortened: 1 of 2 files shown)


skyrocket2026:

| metric | value |
|---|---|
| PR decode tok/s | 84.45 |
| main decode tok/s | 84.57 |
| PR prefill pp | 3378.93 |
| main prefill pp | 2858.54 |
| prefill vs main | +18.2% |
| Polaris receipt | `9b2240e9fd06c9e7…` — TDX (Intel hardware attestation) |
| commit | `b17010deb` |

ok

skyrocket2026:

PR #844, opened 2026-08-15
perf(qwen3.8): 3-stage cp.async pipeline for the FP8 GDN prefill projections (1.03x prefill@128)

FranDev132:
## Summary

The GDN in-projections (`in_proj_qkv`, `in_proj_z`, `out_proj`, 48 layers) are the largest single
group left in prefill@128 — 6.40 ms of a 26.16 ms pass — and they move 5.54 GB of e4m3 weight at
only ~48% of this part's bandwidth. The cause is pipeline depth, not tiling: `pf_gemm_fp8_kernel`
runs a **2-stage** cp.async double buffer, so exactly one prefetch is ever in flight, and its
accumulators (`acc[2][8][4]` + `h[2][8][2]`) hold `__launch_bounds__(256, 2)` to 8 warps × 2 blocks
with no other warp resident to cover the miss. At 1.1–1.4 waves these shapes have no neighbouring
block to hide latency either, so the K loop pays a full global latency on nearly every trip.

This takes the pipeline to 3 stages. Depth costs shared memory, not registers, so occupancy is
unchanged: 3 × 16 KB = 48 KB, still two blocks per SM. Shared memory moves to dynamic so the depth
is a template parameter rather than a second kernel.

Only the **issue point** of the staging moves — the mma order and the `FP8_FLUSH` cadence are
untouched, so the result is bit-identical (verified below, not asserted).

Depth 4 measures *worse than the shipped 2* (4802 vs 4839 pp): 64 KB puts one block on an SM and
halves occupancy. 3 is the knee, and it is where the isolated kernel saturates too.

`SPARKINFER_PREFILL_FP8_STAGES=2` restores the shipped pipeline, and `=4` is kept instantiated, so
all three depths A/B from one binary.

## Proof of speedup

**Decode tok/s** (end-to-end, `qwen3_gguf_bench <dir> 128 sweep`, `SWEEP_CTXS=128`, `REPS=5`):

| | decode tok/s |
|---|--:|
| before (main) | 84.35 |
| after (this PR) | 84.33 |

**Prefill pp tok/s** (same sweep, same model load, ctx=128 — the scored context):

| | prefill pp tok/s |
|---|--:|
| before prefill (main) | 4839.19 |
| after prefill (this PR) | 4971.01 |

**+2.72% prefill@128**, decode flat, VRAM unchanged at 30.8 GB. Same-box RTX 5090 vs `origin/main`
@ `e9b4a2c`, `unsloth/Qwen3.8-27B-NVFP4`, one binary with the depth selected by env, alternated,
5 reps inside each process, median of three rounds:

```text
  r1 stages=2 SWEEP_JSON {"128":{"decode_tps":84.4489,"prefill_pp":4839.3702}}
  r1 stages=3 SWEEP_JSON {"128":{"decode_tps":84.4205,"prefill_pp":4992.4883}}
  r1 stages=4 SWEEP_JSON {"128":{"decode_tps":84.3786,"prefill_pp":4830.7405}}
  r2 stages=2 SWEEP_JSON {"128":{"decode_tps":84.3455,"prefill_pp":4839.1854}}
  r2 stages=3 SWEEP_JSON {"128":{"decode_tps":84.3289,"prefill_pp":4971.0056}}
  r2 stages=4 SWEEP_JSON {"128":{"decode_tps":84.3002,"prefill_pp":4802.2496}}
  r3 stages=2 SWEEP_JSON {"128":{"decode_tps":84.2967,"prefill_pp":4821.7256}}
  r3 stages=3 SWEEP_JSON {"128":{"decode_tps":84.2757,"prefill_pp":4957.7090}}
  r3 stages=4 SWEEP_JSON {"128":{"decode_tps":84.2480,"prefill_pp":4791.4295}}
```

```text
# main (2-stage)
[compressed-tensors] loaded 64 layers, native NVFP4 prefill FFN 56/64 (decode stays Q4_K)
VRAM used    : 30.8 GB
decode tg    : 84.44 tok/s  (n=128, ctx=128, bs=1)
prefill pp   : 4834.37 tok/s  (ctx=128)
SWEEP_JSON {"128":{"decode_tps":84.4412,"prefill_pp":4834.3732}}

# this PR (3-stage)
[compressed-tensors] loaded 64 layers, native NVFP4 prefill FFN 56/64 (decode stays Q4_K)
VRAM used    : 30.8 GB
decode tg    : 84.41 tok/s  (n=128, ctx=128, bs=1)
prefill pp   : 4969.31 tok/s  (ctx=128)
SWEEP_JSON {"128":{"decode_tps":84.4072,"prefill_pp":4969.3056}}
```

Isolated kernel, for the mechanism only (not the before/after claim above) — M=128, non-split-K,
same three shapes, 300 iterations:

```text
qkv [10240,5120]   2-stage 74.19 us -> 3-stage 62.49 us  (+15.8%)
z   [ 6144,5120]   2-stage 73.81 us -> 3-stage 61.92 us  (+16.1%)
out [ 5120,6144]   2-stage 86.48 us -> 3-stage 72.72 us  (+15.9%)
```

## Correctness

Bit-identical, not merely within tolerance. `qwen3_gguf_prefill_check <dir> 128 64` with
`SPARKINFER_KV_INT8=0` (batched prefill vs the token loop over the same prompt, teacher-forced,
per-position logits) returns the same value at every depth from one binary:

| depth | TOP1 | KL |
|---|--:|--:|
| 2 (main) | 0/64 | 7.83639 |
| 3 (this PR) | 0/64 | 7.83639 |
| 4 | 0/64 | 7.83639 |

`qwen3_gguf_score` dumps compare byte-identical between depth 2 and 3.

> The absolute TOP1/KL above are main's, unchanged by this PR — the ctx=128 batched-prefill path is
> already numerically divergent from the token loop on this model before this change. This PR does
> not move it in either direction; the point of the table is that all three depths agree exactly.

`qwen3_gguf_prefill_check` grows compressed-tensors directory support so this check can run on this
model at all — it opened GGUF only.

## Qwen3.6 no-regression

`launch_prefill_gemm_fp8` is shared with Qwen3.6's MoE prefill, so the same binary was swept there
(`SWEEP_CTXS=0,512,4096,16384,32768`, `REPS=2`, ratio = 3-stage / 2-stage):

| ctx | decode | prefill |
|---|--:|--:|
| 0 | 1.000 | — |
| 512 | 1.000 | 1.015 |
| 4096 | 1.001 | 1.002 |
| 16384 | 0.999 | 0.999 |
| 32768 | 0.998 | 0.999 |

### DIFF
--- a/kernels/csrc/cuda/fused/prefill_gemm_fp8.cu
+++ b/kernels/csrc/cuda/fused/prefill_gemm_fp8.cu
@@ -83,14 +83,27 @@ __global__ void pf_quantize_rows_fp8_kernel(const __nv_bfloat16* __restrict__ x,
 // keeps only one block per SM resident.
 // SPLITK partitions the K loop across blockIdx.z (same occupancy fix as launch_prefill_gemm_i8_splitk)
 // and atomicAdds the unscaled fp32 tile into P[M,N]; a separate epilogue applies sx*sw.
-template <bool SPLITK>
+// STAGES is the cp.async pipeline depth. At 2 (one prefetch in flight) the K loop pays a full
+// global-memory latency on nearly every trip: the accumulators alone are acc[2][8][4] + h[2][8][2],
+// so __launch_bounds__(256,2) caps this at 8 warps x 2 blocks and there is no other warp resident
+// to cover the miss. Depth costs only shared memory, not registers -- measured on an RTX 5090 at
+// M=128, non-split-K, against the shipped depth: qkv 74.19 -> 62.49 us (+15.8%), z 73.81 -> 61.92
+// (+16.1%), out 86.48 -> 72.72 (+15.9%). It flattens at 3 (4/6 measure the same within noise), so
+// 3 it is: 48 KB, still two blocks per SM.
+//
+// Only the ISSUE point of the staging moves; the mma order and the FP8_FLUSH cadence are untouched,
+// so the result is bit-identical to the 2-stage path.
+template <bool SPLITK, int STAGES>
 __global__ __launch_bounds__(256, 2) void pf_gemm_fp8_kernel(
         const __nv_fp8_e4m3* __restrict__ A, const __nv_fp8_e4m3* __restrict__ W,
         const float* __restrict__ sx, const float* __restrict__ sw,
         __nv_bfloat16* __restrict__ C, float* __restrict__ P,
         int M, int N, int K, int ktiles) {
-    __shared__ __nv_fp8_e4m3 As[2][FP8_BM][FP8_BK];
-    __shared__ __nv_fp8_e4m3 Bs[2][FP8_BN][FP8_BK];
+    // Dynamic so depth > 2 can exceed the 48 KB static cap without a second kernel.
+    extern __shared__ __align__(16) char fp8_smem[];
+    auto As = reinterpret_cast<__nv_fp8_e4m3 (*)[FP8_BM][FP8_BK]>(fp8_smem);
+    auto Bs = reinterpret_cast<__nv_fp8_e4m3 (*)[FP8_BN][FP8_BK]>(
+        fp8_smem + (size_t)STAGES * FP8_BM * FP8_BK);
 
     const int tid  = threadIdx.x;
     const int warp = tid >> 5;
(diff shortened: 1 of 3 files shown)


skyrocket2026:
## sparkinfer qwen38 auto-eval — `eval-qwen38:none`

| metric | value |
|---|---|
| **label** | `eval-qwen38:none` |
| PR decode tok/s | 84.07 |
| main decode tok/s | 84.09 |
| PR prefill@128 pp | 5206.01 |
| main prefill@128 pp | 5159.02 |
| prefill@128 vs main | +0.9% |
| PR prefill@16k pp | 79.82 |
| main prefill@16k pp | 79.84 |
| prefill@16k vs main | -0.0% |
| Polaris receipt | `61db5c29af28e4c2…` — TDX (Intel hardware attestation) |
| commit | `5c8603774` |

within significance gate — not a verified prefill@16k improvement

skyrocket2026:
## Closed: sparkinfer qwen38 auto-eval — `eval-qwen38:none`

This PR's Qwen3.8-27B prefill@16k speed measured **-0.0%** vs main, with no verified prefill@16k improvement (the only scored dimension) — closing automatically. This bot evaluates every eligible PR in the repo against Qwen3.8-27B's decode AND prefill@128 speed specifically, regardless of what the PR is actually about — a close here isn't a judgment on the PR's purpose, just that it didn't move these particular metrics. Reopen (or open a fresh PR) if you have a fix or a different approach.

PR #845, opened 2026-08-15
perf(qwen3.8): keep the batched prefill alive at ctx=16384 (25.6x prefill@16k)

JSONbored:
## Summary

**Keeps the batched prefill alive at ctx=16384 — the context `eval/pr_qwen38_bot.py` scores since
`784c3f1` — instead of silently dropping the prompt onto the per-token loop. 79.26 -> 2029.24 pp, 25.6x.**

At ctx=16384 on this checkpoint, `prefill_batched_run` cannot allocate its scratch arena, returns
-1, and `bench_decode` falls back to the sequential token loop that re-streams every weight once
per position. main today:

```text
[prefill] scratch alloc failed (ctx=16384) -> fallback
SWEEP_JSON {"128":{"decode_tps":84.0116,"prefill_pp":4369.3982},
            "16384":{"decode_tps":70.3848,"prefill_pp":79.4258}}
```

79 pp is not a slow prefill — it is *no* batched prefill. It sits next to this model's own 70 tok/s
decode because it **is** decode, run 16384 times. prefill@128 on the same load is 4369 pp.

Two things starve that allocation, and both are sized wrong rather than genuinely too big:

1. **The paged-KV pool is allocated per model layer.** Qwen3.8-27B is 64 layers with
   `full_attn_interval=4`, so only **16** own KV — the other 48 are Gated-DeltaNet and keep their
   state in the recurrent/conv buffers. `KVCacheConfig` gains a `layer_slot` map (empty = identity,
   i.e. every non-hybrid model is untouched), the pool is sized to the slots that exist, and every
   consumer goes through `layer_base_elems()` instead of `layer * layer_stride`. `total_blocks`
   still divides by the same slot count, so **block capacity and max context are unchanged**.
2. **The FFN chunk is a fixed constant.** `ffg+ffu` are `2*FC*ffn` bf16 = 1.14 GB at the default
   32768 chunk, demanded regardless of what is free. It is now sized against a `cudaMemGetInfo`
   reading taken *immediately before* that pair is allocated — the point where every FC-independent
   buffer is already resident, so the number is measured rather than modelled. It only ever
   **shrinks**, so a wrong estimate degrades to today's fallback and never to corruption. `FC < N`
   is the same token-chunked path every context above 32k already takes, and the FFN is per-token
   independent, so it stays numerically identical. An explicit `SPARKINFER_PREFILL_FFN_CHUNK` is
   honoured as given.

Leg 1 alone is **not** sufficient at 16384 — measured, it still falls back at 79.0 pp. It is
sufficient at 8192, which is where #847 (@FranDev132, filed ~1h before this) reports its 1.55x;
that PR reaches the same diagnosis and the same quoted fallback line for the KV half, and this
change would rebase cleanly onto it. What moves the *scored* context is leg 2 on top.

## Two latent bugs this had to fix first

Both are on main today, both are reachable through main's own `SPARKINFER_PREFILL_FFN_CHUNK`, and
both surface only when the batched pass runs under memory pressure — which is exactly the state
this PR puts it in.

- **Use-after-free in the int8 degrade path.** When the `a8` arena cannot fit, the handler calls
  `a8.free_all()` and nulls the int8 pointers — but `lm_q8/lm_ad/lm_as` were taken from that same
  arena and are *not* re-taken, and the seed argmax at the end of the function writes through them
  unconditionally. They now get re-allocated (a few KB) with a clean fallback if even that fails,
  and the split-K partials are dropped rather than left dangling.
- **`A_i8` / `sx` under-sized once the chunk drops below `N*H/ffn`.** `a_i8_sz` was `FC*ffn` on the
  non-int8-projection path while the *non-FFN* projections quantize `N` rows x `H` into that same
  buffer; `sx` had the same shape mismatch. Both now take the max. At ctx=16384 the boundary sits
  between chunk 8192 and 4096, and 4096 killed the CUDA context.

Separately, the int8 **projection** path is left off when the FFN is chunked (`FC < N`). It has
only ever run with `FC == N`, and with a smaller chunk it ends the pass in an illegal memory access
that takes the decode graph with it — reproducible at chunk 4096/2048 and cleared by
`SPARKINFER_PREFILL_I8=0`. Rather than ship a path that is not demonstrably correct, the chunked
pass runs on the NVFP4/bf16 projections. **ctx=128 has `FC == N`, so the scored floor keeps int8
exactly as it is today.** The underlying defect is left for its own PR.

## Proof of speedup

**Decode tok/s** (end-to-end, `bench/scripts/bench.sh` path — this PR is prefill-only):

| | decode tok/s |
|---|--:|
| before (main) | 83.74 |
| after (this PR) | 83.38 (0.996x — inside main's own 0.66% spread) |

| | prefill pp tok/s |
|---|--:|
| before prefill (main) | 79.26 |
| after prefill (this PR) | **2029.24 (25.6x, +2460%)** |

```text
# RTX 5090 sm_120, unsloth/Qwen3.8-27B-NVFP4, main @ 784c3f1. SHIPPED DEFAULTS -- no env
# overrides; the only vars set are the two the bot's own bench_sweep_run exports. Arms rebuilt
# between runs (one build dir, so no stale runtime .so can leak across an arm) and alternated
# main/PR/main/PR, serialized so no two benches ever share the GPU. Sweep order is the bot's own
# (128 first, then 16384) -- that order is what leaves VRAM tight, and measuring 16384 alone
# understates the problem.

MAIN pair1 {"128":{"decode_tps":84.0116,"prefill_pp":4369.3982},"16384":{"decode_tps":70.3848,"prefill_pp":79.4258}}
PR   pair1 {"128":{"decode_tps":83.4213,"prefill_pp":4343.4116},"16384":{"decode_tps":70.8359,"prefill_pp":2034.0784}}
MAIN pair2 {"128":{"decode_tps":83.4616,"prefill_pp":4338.1144},"16384":{"decode_tps":70.2443,"prefill_pp":79.0950}}
PR   pair2 {"128":{"decode_tps":83.3447,"prefill_pp":4332.9715},"16384":{"decode_tps":70.6564,"prefill_pp":2024.4055}}

prefill@16k  median   79.26 -> 2029.24  = 25.60x   (+2460.2%)   <- the scored dimension
prefill@128  median 4353.76 -> 4338.19  = 0.9964x  (floor, tol 0.98)
decode@128   median   83.74 ->   83.38  = 0.9958x  (floor, tol 0.98)
decode@16k   median   70.31 ->   70.75  = 1.0063x

# Both floors sit inside main's own run-to-run spread (its two decode@128 samples differ by
# 0.66%); nothing in this change touches the ctx=128 path, where FC == N and int8 stays on.
# Every PR arm logs the chunk it chose, so the arm is identifiable in-trace:
#   [prefill] ffn chunk 16384 -> 1024 (ctx=16384, free=333 MB) to keep the batched pass
# and every main arm logs the fallback it took:
#   [prefill] scratch alloc failed (ctx=16384) -> fallback
```

**Differential accuracy gate** (`bench/scripts/accuracy_compare_pair.py`, PR vs main, same token
stream, TOPK=128):

```text

# 102-token stream from the checkpoint's own tokenizer.json, TOPK=128 (QWEN38_ACC_TOPK default),
# PR dump vs origin/main dump. Bars: top1 >= 0.99, KL <= 0.01. Bit-identical: the ctx=128 path is
# untouched (FC == N, int8 on), and the KV compaction relabels where a layer's slice lives without
# changing a value in it.
```

**16k differential accuracy** (the same comparator, on a 16384-token stream). The bot's gate uses
a ~102-token stream, so it only exercises the ctx=128 path — and this PR is what makes the batched
pass run at 16384 at all. At that length **main takes the token loop and this PR takes the batched
pass**, so this compares the two paths directly rather than two variants of one path:

```text
METRIC top1=1.000000 kl=0.000000 ppl_pr=1.0089 ppl_main=1.0089

# 16384-token stream (the eval corpus tiled), 16386 dump lines on each side, TOPK=128.
# The chunked FFN and the compacted KV pool reproduce the sequential token loop's distribution
# exactly -- which is the claim the "numerically identical" wording above rests on, measured
# rather than asserted. (ppl ~1.0 is the tiled stream being trivially predictable; it is the
# PR-vs-main agreement that matters here, and both sides see the identical stream.)
```

**Qwen3.6 cross-model guard** (ctx 0/512/4k/16k/32k, tol 0.98). This change touches shared prefill
code *and* compacts Qwen3.6's pool too (it is hybrid with the same interval), so it is measured:

```text
ctx      0 decode_tps  main    492.63 -> PR    489.87  0.9944x  ok
ctx    512 decode_tps  main    486.45 -> PR    483.95  0.9949x  ok
ctx    512 prefill_pp  main  10000.54 -> PR   9982.48  0.9982x  ok
ctx   4096 decode_tps  main    467.69 -> PR    465.06  0.9944x  ok
ctx   4096 prefill_pp  main  25899.97 -> PR  25868.96  0.9988x  ok
ctx  16384 decode_tps  main    467.01 -> PR    464.80  0.9953x  ok
ctx  16384 prefill_pp  main  26786.60 -> PR  26786.93  1.0000x  ok
ctx  32768 decode_tps  main    466.98 -> PR    464.51  0.9947x  ok
ctx  32768 prefill_pp  main  27969.43 -> PR  27971.82  1.0001x  ok

# tol 0.98 on every row. The uniform ~0.5% on decode is thermal drift across sequentially
# run arms (the two Qwen3.8 main arms differ by 0.66% between themselves); prefill is flat to
# 4 decimal places at 16k/32k, which is where a KV-pool change would show up if it were wrong.
```

## Credit

#847 (@FranDev132) independently reaches the KV-pool half of this diagnosis and filed first; leg 1
here is the same idea and this PR rebases onto it cleanly if it lands. #835 is the batched
prefill-attention path this keeps alive at long context, and #800/#530 are the token-chunked FFN
this sizes dynamically rather than by a constant.

### DIFF
--- a/runtime/examples/qwen3_gguf_bench.cpp
+++ b/runtime/examples/qwen3_gguf_bench.cpp
@@ -126,10 +126,16 @@ static bool init_session(BenchSession& s, const std::string& path, int max_ctx, 
     kvc.block_size = 16;
     const char* e8 = getenv("SPARKINFER_KV_INT8");
     kvc.int8_kv = e8 ? (e8[0] != '0') : (max_ctx >= 4096);
+    // Only the full-attention layers get a pool slot (hybrid_kv_layer_slots): the
+    // Gated-DeltaNet layers carry a recurrent state and never read paged KV. On
+    // Qwen3.8-27B that is 16 slots of 64, so a 16k context fits where it used to OOM.
+    kvc.layer_slot = sparkinfer::hybrid_kv_layer_slots(s.cfg.n_layers, s.cfg.hybrid,
+                                                      s.cfg.full_attn_interval);
+    const int kvL = sparkinfer::kv_slot_count(kvc.layer_slot, s.cfg.n_layers);
     const size_t epb = (size_t)16 * s.cfg.n_kv_heads * s.cfg.head_dim;
     const size_t blocks = (s.cfg.max_seq + 15) / 16 + 8;
     s.kv = std::make_unique<sparkinfer::KVCacheManager>(
-        kvc, (size_t)s.cfg.n_layers * 2 * epb * 2 * blocks);
+        kvc, (size_t)kvL * 2 * epb * 2 * blocks);
     sparkinfer::moe::MoEConfig mc;
     mc.num_experts = s.cfg.n_experts; mc.top_k = s.cfg.top_k;
     mc.hidden_dim = s.cfg.hidden; mc.ffn_dim = s.cfg.moe_ffn;
--- a/runtime/examples/qwen3_gguf_cb_bench.cpp
+++ b/runtime/examples/qwen3_gguf_cb_bench.cpp
@@ -93,10 +93,14 @@ int main(int argc, char** argv) {
         kvc.int8_kv = e ? (e[0] != '0')
                         : (cfg.muse_glimmer ? false : cfg.hybrid ? (cfg.max_seq >= 4096) : true);
     }
+    // Only the full-attention layers get a pool slot (hybrid_kv_layer_slots): the
+    // Gated-DeltaNet layers carry a recurrent state and never read paged KV.
+    kvc.layer_slot = sparkinfer::hybrid_kv_layer_slots(cfg.n_layers, cfg.hybrid, cfg.full_attn_interval);
+    const int kvL = sparkinfer::kv_slot_count(kvc.layer_slot, cfg.n_layers);
     const size_t epb = (size_t)16 * cfg.n_kv_heads * cfg.head_dim;
     // Pool for concurrency streams + one long prefill.
     const size_t blocks = (size_t)(concurrency + 1) * ((cfg.max_seq + 15) / 16 + 4);
-    sparkinfer::KVCacheManager kv(kvc, (size_t)cfg.n_layers * 2 * epb * 2 * blocks);
+    sparkinfer::KVCacheManager kv(kvc, (size_t)kvL * 2 * epb * 2 * blocks);
 
     sparkinfer::moe::MoEConfig mc;
     mc.num_experts = cfg.n_experts;
--- a/runtime/examples/qwen3_gguf_dflash_bench.cpp
+++ b/runtime/examples/qwen3_gguf_dflash_bench.cpp
@@ -57,9 +57,13 @@ int main(int argc, char** argv) {
     kvc.head_dim = cfg.head_dim;
     kvc.block_size = 16;
     kvc.int8_kv = false;
+    // Only the full-attention layers get a pool slot (hybrid_kv_layer_slots): the
+    // Gated-DeltaNet layers carry a recurrent state and never read paged KV.
+    kvc.layer_slot = sparkinfer::hybrid_kv_layer_slots(cfg.n_layers, cfg.hybrid, cfg.full_attn_interval);
+    const int kvL = sparkinfer::kv_slot_count(kvc.layer_slot, cfg.n_layers);
     const size_t epb = (size_t)16 * cfg.n_kv_heads * cfg.head_dim;
     const size_t blocks = (cfg.max_seq + 15) / 16 + 8;
-    sparkinfer::KVCacheManager kv(kvc, (size_t)cfg.n_layers * 2 * epb * 2 * blocks);
+    sparkinfer::KVCacheManager kv(kvc, (size_t)kvL * 2 * epb * 2 * blocks);
 
     sparkinfer::moe::MoEConfig mc;
     mc.num_experts = cfg.n_experts;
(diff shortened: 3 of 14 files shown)


skyrocket2026:

| metric | value |
|---|---|
| PR decode tok/s | 84.06 |
| main decode tok/s | 84.09 |
| PR prefill@128 pp | 5128.75 |
| main prefill@128 pp | 5159.02 |
| prefill@128 vs main | -0.6% |
| PR prefill@16k pp | 2170.89 |
| main prefill@16k pp | 79.84 |
| prefill@16k vs main | +2619.1% |
| Polaris receipt | `c5428b303bdc51fc…` — TDX (Intel hardware attestation) |
| commit | `85d6fafec` |

ok

skyrocket2026:

PR #843, opened 2026-08-15
perf(qwen3.8): chunked GDN@128 + skinny split-K (1.19x prefill@128)

flashatten:
## Summary

Rebased onto #842 (now on main). Two leftover holes at the scored prefill@128:

1. The WY/UT GDN scan was still gated at ctx≥256, so all 48 GDN layers ran the serial token scan. Prep amortizes at N=128 on this 48-v-head shape. `SPARKINFER_PREFILL_GDN_CHUNK_MINCTX=256` restores the old floor.
2. #842's cp.async skinny GEMM grids M=128 to 4 CTAs (BT=32). Halving BT was worse; split-K keeps BT=32 and fans K across SMs so the 48-wide alpha/beta fills the 5090. `SPARKINFER_PREFILL_SKINNY_SPLITK=0` restores #842's single-wave grid.

Decode is unchanged.

## Proof of speedup

> ⚠️ **The on-device eval runs only when BOTH are true:** (1) the box below is ticked, and
> (2) at least one **decode tok/s** or **Qwythos prefill pp** table shows a **real end-to-end**
> **improvement** (`after > before`, filled from `bench/scripts/bench.sh` — _not_ an isolated-kernel
> microbenchmark). A ticked box with empty/placeholder tables (or no claimed gain on either metric)
> gets `needs-benchmark` and is **not** evaluated.
>
> Tick the box **only if you actually ran it on an RTX 5090**. False attestation is treated as
> gaming — the account is **blocked** (.github/blocked-contributors.txt),
> same as copycatting or sybil farming.

- [x] Tested on **RTX 5090** (`sm_120`)

**Decode tok/s** (end-to-end, from `bench/scripts/bench.sh` — fill if this PR targets decode):

| decode tok/s | |
| --- | --: |
| before (main / #842) | 85.15 |
| after (this PR) | 85.15 |

**Prefill pp tok/s** (Qwythos / Qwen3.5 — fill if this PR targets prefill; use `--ctx 4096`, `32768`,
`65536`, or `131072` and copy the `prefill pp` line — report your best context):

| prefill pp tok/s | |
| --- | --: |
| before prefill (main / #842) | 4455.93 |
| after prefill (this PR) | 5322.02 |

Same-box RTX 5090 vs main (#842), `unsloth/Qwen3.8-27B-NVFP4`, isolated `LD_LIBRARY_PATH`, median of 5. Prefill@128 **+19.4% (1.19×)**. Decode does not regress. VRAM unchanged at 30.8 GB.

```
[...]

PR #849, opened 2026-08-15
perf(qwen3.8): keep the chunk-parallel GDN scan alive at 16k + size the FFN chunk to fit (1.24x prefill@16k)

JSONbored:
## Summary

**At ctx=16384 the chunk-parallel GDN scan cannot fit its workspace, so all 48 linear layers fall
back to the sequential scan — silently. Segmenting the scan keeps the fast path, and sizing the FFN
chunk to what is actually free keeps more of the pass batched: prefill@16k 4687.61 -> **5803.88** pp
(1.238x, +23.8%).**

Two independent changes, both sizing problems at long context rather than new kernels.

### 1. The chunk-parallel GDN scan gives up at 16k

`launch_prefill_gdn_chunk` reserves a workspace that scales with the sequence: `W^` and `U0` are
each `n_tokens * v_heads * head_dim` bf16, so it grows to **~322 MB at ctx=16384** and stops fitting
well before the sequence itself does. `ws_reserve` then fails and the launcher returns false, which
drops every linear layer onto the sequential scan below it. Nothing is logged; the pass is simply
slower.

That path is worth a lot. Measured at ctx=4096, where the workspace still fits, by toggling
`SPARKINFER_PREFILL_GDN_CHUNK`:

```text
ctx=4096  chunk-parallel : 8275.82 pp
ctx=4096  sequential     : 7059.03 pp     <- what ctx=16384 has been getting
```

The scan is already sequential **across** chunks and carries `state` in and out, so a segment
boundary on a multiple of `C` is just another chunk boundary. When the whole-sequence workspace
does not fit, the sequence is processed in segments whose workspace does, with the recurrence
carrying between them — same chunk boundaries, same arithmetic, same order. If even one chunk
cannot be held, the caller still gets exactly today's sequential fallback.

### 2. The FFN chunk is sized against a model, not against what is free

The adaptive chunk (#845) reserved for `wbuf`, `W_i8` and `A_i8` and stepped by halving. `wbuf` and
`W_i8` are sized by `maxw`, which does not depend on N, so they are already resident and cost
nothing at that point. `A_i8` is the one that must fit — it is taken *before* the int8 arena's
ok-check, so failing it turns the int8 projections off for the whole pass, whereas the FP4 buffers
are taken after that check and only cost the FP4 path. Reserving `A_i8` alone, and taking the
largest chunk that fits (rounded to 256, keeping `m % 8` for the block-scaled kernels) instead of
the first power of two, gives chunk 1024 -> 3072 at 16k: six passes over the FFN weights, not
sixteen.

Oversizing is the failure that matters and is measured, not assumed: at chunk 4096 the pass leaves
~48 MB, `A_i8` misses, and prefill@16k drops to ~2200 pp. The chunk still only ever shrinks from the
configured value, and an explicit `SPARKINFER_PREFILL_FFN_CHUNK` is honoured as given.

## Proof of speedup

| | decode tok/s |
|---|--:|
| before (main) | 83.89 |
| after (this PR) | 83.88 |

| | prefill pp tok/s |
|---|--:|
| before prefill (main) | 4687.61 |
| after prefill (this PR) | **5803.88 (1.238x, +23.8%)** |

```text
# RTX 5090 sm_120, unsloth/Qwen3.8-27B-NVFP4, main @ aa117dc. SHIPPED DEFAULTS -- the only env
# vars set are the two the bot's own bench_sweep_run exports. Arms rebuilt between runs (one build
# dir, so no stale runtime .so can cross an arm), alternated, serialized, and in the bot's sweep
# ORDER (128 then 16384), which is the order that leaves VRAM tight.

MAIN  pair1 {"128":{"decode_tps":83.9007,"prefill_pp":4372.0933},"16384":{"decode_tps":71.6097,"prefill_pp":4695.9423}}
STACK pair1 {"128":{"decode_tps":83.8933,"prefill_pp":4372.6632},"16384":{"decode_tps":71.3675,"prefill_pp":5810.3192}}
MAIN  pair2 {"128":{"decode_tps":83.8746,"prefill_pp":4359.2897},"16384":{"decode_tps":71.5397,"prefill_pp":4679.2696}}
STACK pair2 {"128":{"decode_tps":83.8729,"prefill_pp":4359.4516},"16384":{"decode_tps":71.3275,"prefill_pp":5797.4473}}

prefill@16k  median 4687.61 -> 5803.88  = 1.238x  (+23.8%)   <- the scored dimension
prefill@128  median 4365.69 -> 4366.06  = 1.0001x (floor, tol 0.98)
decode@128   median   83.89 ->   83.88  = 0.9999x (floor, tol 0.98)

# Each arm is identifiable in-trace by the FFN chunk it selects (main 1024, this PR 3072 at the
# bench's free-VRAM level); the GDN change leaves no line of its own, so the kernel it restores is
# the check that it engaged.
```

**Generation identity through the batched pass at ctx=16384.** `qwen3_gguf_score` teacher-forces per
token and does not route through `prefill_batched_run`, so a score-based diff cannot see the path
this PR changes (verified: flipping `SPARKINFER_PREFILL_BATCHED` leaves its dump byte-identical).
The check that does exercise it is generation — same binary, env-flipped, 16384-token prompt:

```text
# 16384-token prompt, 24 tokens generated, same binary, env-flipped.
batched    OUTPUT_IDS: 7 197 197 8 197 197 8 197 197 8 197 197 8 197 197 8 197 197 8 197 197 8 197 197
tokenloop  OUTPUT_IDS: 7 197 197 8 197 197 8 197 197 8 197 197 8 197 197 8 197 197 8 197 197 8 197 197
# 24/24 identical: the segmented scan reproduces the sequential prefill exactly.
# The prompt is synthetic (a non-repeating id ramp), so the continuation is repetitive and argmax is
# less sensitive than natural text would be -- reported as an end-to-end identity check through the
# changed path at the scored context, not as a tight numerical bound.
```

**Differential accuracy gate** (`accuracy_compare_pair.py`, PR vs main, TOPK=128). The `ctx=128`
path is untouched by construction — the FFN chunk is inert where `FC == N`, and the GDN workspace at
that length is ~2.5 MB so the segmentation never engages:

```text
```

**Qwen3.6 cross-model guard** (ctx 0/512/4k/16k/32k, tol 0.98), measured because both changes are
shared prefill code and Qwen3.6 is hybrid too:

```text
ctx      0 decode_tps  main    491.23 -> PR    490.42  0.9983x  ok
ctx    512 decode_tps  main    485.13 -> PR    483.97  0.9976x  ok
ctx    512 prefill_pp  main  10029.30 -> PR  10005.70  0.9976x  ok
ctx   4096 decode_tps  main    466.36 -> PR    465.39  0.9979x  ok
ctx   4096 prefill_pp  main  26026.59 -> PR  25968.89  0.9978x  ok
ctx  16384 decode_tps  main    465.48 -> PR    465.07  0.9991x  ok
ctx  16384 prefill_pp  main  27026.95 -> PR  26967.50  0.9978x  ok
ctx  32768 decode_tps  main    465.32 -> PR    464.95  0.9992x  ok
ctx  32768 prefill_pp  main  28258.51 -> PR  28168.63  0.9968x  ok
```

## Credit

#843 (@flashatten) is the chunk-parallel WY/UT GDN scan this keeps alive at long context — the
segmentation only changes how much of the sequence it holds at once, not the transform. #845 is the
adaptive chunk this refines and the batched-prefill-at-16k work both rest on; #848 is the activation
sizing that lets the chunked pass keep its tensor-core projections.

### DIFF
--- a/kernels/csrc/cuda/fused/prefill_gdn_chunk.cu
+++ b/kernels/csrc/cuda/fused/prefill_gdn_chunk.cu
@@ -55,6 +55,7 @@
 // The state stays fp32 and is written back in the SAME transposed [v_head][col][row] layout the
 // decode gdn_ar_fast kernel expects, so this is a drop-in replacement and decode is untouched.
 // ============================================================================
+#include <algorithm>
 #include "sparkinfer/kernels/prefill_gdn_chunk.h"
 
 #include <cuda_runtime.h>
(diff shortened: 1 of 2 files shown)


skyrocket2026:

| metric | value |
|---|---|
| PR decode tok/s | 84.04 |
| main decode tok/s | 84.18 |
| PR prefill@128 pp | 5115.42 |
| main prefill@128 pp | 5162.74 |
| prefill@128 vs main | -0.9% |
| PR prefill@16k pp | 79.99 |
| main prefill@16k pp | 5215.37 |
| prefill@16k vs main | -98.5% |
| Polaris receipt | `e600c4e650000846…` — TDX (Intel hardware attestation) |
| commit | `c8ad8617b` |

prefill@16k regression: 79.99 < 98% of main 5215.37

skyrocket2026:
## Closed: sparkinfer qwen38 auto-eval — `eval-qwen38:REJECT`

This PR's Qwen3.8-27B prefill@16k speed measured **-98.5%** vs main, (regression) — closing automatically. This bot evaluates every eligible PR in the repo against Qwen3.8-27B's decode AND prefill@128 speed specifically, regardless of what the PR is actually about — a close here isn't a judgment on the PR's purpose, just that it didn't move these particular metrics. Reopen (or open a fresh PR) if you have a fix or a different approach.

PR #850, opened 2026-08-15
perf(qwen3.8): keep the chunk-parallel GDN scan alive at 16k (1.35x prefill@16k)

flashatten:
## Summary

**At ctx=16384 the chunk-parallel GDN scan cannot allocate its workspace, so all 48 linear layers fall back to the sequential scan — silently. Segmenting on chunk boundaries, and reloading S between slices, keeps the fast path: official prefill@16k 5356 → 7219 pp (1.35x).**

The workspace is O(N): W^ and U0 are `n_tokens * v_heads * head_dim` bf16 each, ~400 MB at ctx=16384 × 48 v-heads (483 MB with G and M). `ws_reserve` failing returns false and `launch_prefill_gdn_scan` drops every linear layer onto the token-serial scan. Nothing is logged.

The scan is already sequential **across** chunks. A segment boundary on a multiple of C is the same as a chunk boundary **once the kernel reloads the transposed `[v_head][col][row]` state the previous slice wrote** (today it always zeros S, so a naive slice would forget the recurrence). When the whole-sequence workspace does not fit, the sequence is processed in slices whose workspace does; ctx=128 still takes the one-shot path (workspace is ~1 MB). If even one chunk cannot be held, the caller still gets today's sequential fallback.

`ws_reserve` now allocates the new buffer before freeing the old one, so a failed grow to 483 MB does not drop a working segment workspace every layer.

This does **not** change FFN chunk sizing. Oversizing that chunk is what sent #849 to 80 pp on the official box (arena alloc fail → sequential token loop).

## Proof of speedup

| | decode tok/s |
|---|--:|
| before (main) | 84.48 |
| after (this PR) | 84.51 |

**Prefill pp tok/s** (Qwen3.8-27B NVFP4 — official `pr_qwen38_bot.py` protocol):

| | prefill@128 | prefill@16k |
|---|--:|--:|
| before (main) | 5161 | 5356 |
| after (this PR) | 5171 | **7219 (1.35x, +35%)** |

```text
# RTX 5090 sm_120, unsloth/Qwen3.8-27B-NVFP4, main @ aa117dc.
# SHIPPED DEFAULTS — the only env vars set are the two the bot's own
# bench_sweep_run exports. Official order: 128 then 16384, reps=5.

MAIN  {"128":{"decode_tps":84.48,"prefill_pp":5161},"16384":{"prefill_pp":5356}}
THIS  {"128":{"decode_tps":84.51,"prefill_pp":5171},"16384":{"prefill_pp":7219}}

[gdn-chunk] workspace 483 MB does not fit; 192-token segments (ctx=16384, free=37 MB)
```

128 floors: decode 1.00x, prefill 1.00x (bar 0.98x). FFN chunk path is untouched (1024 on the tight first 16k pass, 4096 after).

## Test plan

- [ ] Bot official sweep: `SPARKINFER_BENCH_SWEEP_CTXS=128,16384 SPARKINFER_BENCH_SWEEP_REPS=5`
- [ ] Confirm decode@128 and prefill@128 stay ≥ 0.98× main
- [ ] Confirm prefill@16k is up and the batched pass does not fall through to ~80 pp
- [ ] `SPARKINFER_PREFILL_GDN_CHUNK=0` still hits the sequential scan

### DIFF
--- a/kernels/csrc/cuda/fused/prefill_gdn_chunk.cu
+++ b/kernels/csrc/cuda/fused/prefill_gdn_chunk.cu
@@ -284,7 +284,7 @@ __global__ void pf_gdnc_scan_kernel(const __nv_bfloat16* __restrict__ q,
                                     float* __restrict__ state,
                                     __nv_bfloat16* __restrict__ out,
                                     int n_tokens, int q_heads, int v_heads, int n_chunks,
-                                    bool qh_block) {
+                                    bool qh_block, int carry) {
     extern __shared__ char s_raw[];
     float* s_S = reinterpret_cast<float*>(s_raw);                              // [HD][JC]  fp32 carrier
     float* s_U = s_S + (size_t)HD * JC;                                        // [C][JC]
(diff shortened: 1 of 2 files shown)


skyrocket2026:

| metric | value |
|---|---|
| PR decode tok/s | 84.01 |
| main decode tok/s | 84.16 |
| PR prefill@128 pp | 5111.35 |
| main prefill@128 pp | 5154.22 |
| prefill@128 vs main | -0.8% |
| PR prefill@16k pp | 7459.62 |
| main prefill@16k pp | 5218.04 |
| prefill@16k vs main | +43.0% |
| Polaris receipt | `9f2616b1018957c6…` — TDX (Intel hardware attestation) |
| commit | `ff06c7314` |

ok

skyrocket2026:

PR #858, opened 2026-08-16
perf(qwen3.8): integer NVFP4 decode straight to int8 rows (1.12x prefill@128)

inference2026:
## Summary

Two changes to the NVFP4 checkpoint dequant path, in the order they had to happen — the second only
works because of the first.

**1. Integer decode.** The group-wise dequant is the largest single item in a ModelOpt
prefill@128, and ncu says it is *not* bandwidth bound: DRAM 43.7% against **SM 80.9%** at 78% warps
active, zero local memory. So the cost is the per-element decode, and this replaces the two CUTLASS
conversions with integer arithmetic:

* **e2m1** — the eight magnitudes doubled are `{0,1,2,3,4,6,8,12}`, all below 16, so the table is
  the nibbles of one 32-bit literal and the decode is a shift, a mask and an int->float, with no
  memory and no select. Halving is exact in binary floating point, so `0.5f *` that is exactly what
  `cutlass::float_e2m1_t::bitcast` produced — sign and `-0.0` included.
* **ue4m3** — for `e > 0`, `(8+m) * 2^(e-10)` is exactly the fp32 with exponent field `e+120` and
  mantissa `m<<20`; `e == 0` stays an exact multiply by `2^-9`.

**2. Decode straight to int8 rows.** With the arithmetic halved the kernel is no longer ALU bound,
so the bf16 round trip it feeds becomes the cost. `proj()` has no native-NVFP4 arm: the weight is
expanded to a bf16 scratch and the row-quantizer reads it straight back — 295 MB of traffic per GDN
qkv projection to produce the 52 MB of int8 the GEMM consumes. Fusing them moves 85 MB, in one
kernel.

That fusion was tried **before** the integer decode and measured **1.2% slower** — it cut DRAM
traffic 4.4x and gained nothing, because both kernels were arithmetic bound. It is included here
because the first change removes that premise, and on top of it the same fusion is worth +4.5%.

`ct_nvfp4_rows_i8_kernel` is `pf_quant_rows_fast_kernel` with the load phase replaced: the same
`VEC=8` slots so a thread owns the same eight columns, the same bf16 rounding, the same `amax` over
that value set, the same `d = amax/127.0f`, `roundf` and `amax == 0 -> 0` rule. The row `amax` is a
**max** — associative and exact in floating point — so no thread mapping can change it.

**Bit-identical throughout**, and measured as such: teacher-forced score dumps are byte-for-byte
equal to main for both stages, top1 1.000000, KL 0.000000.

The `(v * s) / global_scale` division is deliberately left as a division; hoisting a reciprocal is
faster but is not bit-identical. It stays available as `SPARKINFER_CT_NVFP4_ALU=2` with its price
measured: **+0.88%** for KL 0.004494, which is a poor trade for a weight dequant.
`SPARKINFER_CT_NVFP4_ALU=0` restores the CUTLASS conversions and
`SPARKINFER_CT_NVFP4_ROWS_I8=0` the two-kernel path, so every arm comes out of one binary.

## Proof of speedup

| | decode tok/s |
|---|--:|
| before (main) | 93.90 |
| after (this PR) | 94.02 |

| | prefill pp tok/s |
|---|--:|
| before prefill (main) | 3295.99 |
| after prefill (this PR) | 3682.85 |

```text
# (qwen3_gguf_bench ... sweep, median of 5 reps). RTX 5090 sm_120, main @ 91ce034.
# MAIN = ALU 0 / FUSE 0 (i.e. main), ALU = stage 1 only, ALU+FUSE = this PR.
# Arm ORDER IS ROTATED through all three positions between rounds.

r1 MAIN      decode128=94.9264  pp128=3340.0743  pp16k=9736.4356
r1 ALU       decode128=94.5780  pp128=3545.9325  pp16k=9578.7581
r1 ALU+FUSE  decode128=94.2006  pp128=3688.4612  pp16k=9472.9901
r2 ALU+FUSE  decode128=94.0172  pp128=3682.8516  pp16k=9440.7430
r2 MAIN      decode128=93.8975  pp128=3292.0514  pp16k=9397.1759
r2 ALU       decode128=93.8766  pp128=3528.1350  pp16k=9394.3384
r3 ALU       decode128=93.7822  pp128=3533.2674  pp16k=9381.2459
r3 ALU+FUSE  decode128=93.8387  pp128=3672.3599  pp16k=9384.8040
r3 MAIN      decode128=93.8694  pp128=3295.9885  pp16k=9374.6885

prefill@128  3295.99 -> 3682.85 pp (+11.74%); the integer decode alone is +7.20%, the fusion
             adds +4.5% on top. Fully disjoint: main's best round 3340.07 < ALU's worst 3528.14
             < this PR's worst 3672.36, and this PR wins from every position in the rotation.
decode@128   93.90 -> 94.02 and prefill@16k 9397.18 -> 9440.74: both flat. Decode reaches these
             weights through the NVFP4 GEMV, not this dequant, so it is untouched by design.

Accuracy: score dumps BYTE-IDENTICAL to main at both stages (top1 1.000000, KL 0.000000).
Control, for reference only: SPARKINFER_CT_NVFP4_ALU=2 (hoisted reciprocal) is +0.88% further
and top1 1.000000 / KL 0.004494, so it is not the default.

No-regression guards, same build, ALU=1/FUSE=1 vs ALU=0/FUSE=0:
  Qwen3.6  ctx 0/512/4k/16k/32k   decode  1.0028 / 1.0032 / 1.0037 / 1.0035 / 1.0029
                                  prefill      -  / 1.0064 / 1.0082 / 1.0086 / 1.0084
  Qwen3.8  ctx 0/4k/16k           decode  1.0028 / 1.0024 / 1.0026
                                  prefill      -  / 1.0115 / 1.0096
ctest: 100% tests passed, 0 tests failed out of 19.
```

### DIFF
--- a/kernels/csrc/cuda/fused/prefill_nvfp4_sm120.cu
+++ b/kernels/csrc/cuda/fused/prefill_nvfp4_sm120.cu
@@ -540,7 +540,108 @@ __global__ void ct_dequant_nvfp4_g16_kernel(const unsigned char* __restrict__ pa
         *reinterpret_cast<uint4*>(orow + (size_t)g * 16 + 8) = *reinterpret_cast<const uint4*>(o + 8);
     }
 }
+
+// Fused NVFP4 -> per-row int8, so the bf16 never lands in DRAM.
+//
+// This was tried BEFORE the integer decode above and measured 1.2% SLOWER: it cut DRAM traffic
+// 4.4x (295 MB -> 85 MB per GDN qkv projection) and gained nothing, because at that point both
+// this kernel and the dequant it replaces were ALU bound -- ncu had them at SM 75-81% with DRAM
+// at 10-44%. Halving the per-element arithmetic changes that premise, so the traffic saving is
+// worth re-testing on top of it. SPARKINFER_CT_NVFP4_ROWS_I8 selects.
+//
+// The value set, the reduction and the rounding are pf_quant_rows_fast_kernel's, unchanged: the
+// same VEC=8 slots so a thread holds the same eight columns, the same bf16 rounding on the way in,
+// the same amax over that value set, the same d = amax/127.0f and roundf, and the same
+// amax==0 -> 0 rule. The row amax is a MAX, which is associative and exact in floating point, so
+// no thread mapping can change it. Bit-identical to the two-kernel path by construction.
+//
+// Eight consecutive columns from a multiple of 8 lie inside one 16-element group, so a slot is one
+// 4-byte packed load plus one scale byte.
+template <int BLOCK, int VEC, int SLOTS>
+__global__ __launch_bounds__(BLOCK) void ct_nvfp4_rows_i8_kernel(
+        const unsigned char* __restrict__ packed, const unsigned char* __restrict__ group_scale,
+        const float* __restrict__ global_scale_dev, signed char* __restrict__ q,
+        float* __restrict__ scale, int rows, int cols) {
+    const int r = blockIdx.x;
+    if (r >= rows) return;
+    const int tid = threadIdx.x;
+    const float gs = *global_scale_dev;
+    const unsigned char* prow = packed + (size_t)r * (size_t)(cols >> 1);
+    const unsigned char* srow = group_scale + (size_t)r * (size_t)(cols >> 4);
+    const size_t base = (size_t)r * (size_t)cols;
+
+    __nv_bfloat16 reg[SLOTS][VEC];
+    float amax = 0.f;
+    #pragma unroll
+    for (int s = 0; s < SLOTS; s++) {
+        const int c = (tid + s * BLOCK) * VEC;
+        if (c < cols) {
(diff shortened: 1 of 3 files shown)


skyrocket2026:
## sparkinfer modelopt auto-eval — `eval-modelopt:L`

| metric | value |
|---|---|
| **label** | `eval-modelopt:L` |
| PR decode tok/s | 94.88 |
| main decode tok/s | 95.10 |
| PR prefill@128 pp | 3725.42 |
| main prefill@128 pp | 3339.38 |
| prefill@128 vs main | +11.6% |
| PR prefill@16k pp | 9744.83 |
| main prefill@16k pp | 9809.01 |
| Polaris receipt | `63fe4f0830140a8c…` — TDX (Intel hardware attestation) |
| commit | `bc441a3d2` |

ok

skyrocket2026:

PR #859, opened 2026-08-16
perf(qwen3.8): per-group basis for the NVFP4 checkpoint decode (1.07x prefill@128)

James-CUDA:
## Summary

> **Measurement correction (rebased onto current main).** This was opened against a main where the
> GDN projections still decoded their NVFP4 weights to int8 rows on every prefill, and it measured
> **+6.99% prefill@128** there. Current main gives those projections a native NVFP4 GEMM instead, so
> that decode no longer runs at the scored shape and **this change is now flat: −0.56 / +0.37 /
> −0.10 / +0.19% over four rotated rounds, mean −0.02%.** The old numbers are left out rather than
> restated. It is still a strictly cheaper, bit-identical decode wherever that decode does run
> (model load, and prompts past the native arm's bound), and it carries an exhaustive equivalence
> test — but it no longer earns a tier, and I would rather say so than let a stale number stand.

The NVFP4 checkpoint decode is compute bound, not bandwidth bound, so what it costs is the
per-value arithmetic — and one **divide per value** is the largest item left. The two ALU modes on
main are the two ends of the same trade: keep the divide and pay for it, or hoist a reciprocal and
stop being bit-identical (measured there as a poor deal for a weight decode).

There is a third option that is both, because within a 16-value group the divide has only sixteen
numerators. The eight E2M1 magnitudes are `{0, .5, 1, 1.5, 2, 3, 4, 6}` — that is `{1, 1.5}` times
exact powers of two — so **two divides per group**

```
b1 = s / gs   (v = 1)          b15 = 1.5f * s / gs   (v = 1.5)
```

span all sixteen: every value is `b1` or `b15` scaled by a power of two. That scaling is exact and
rounding commutes with it, so `b1 * 2^k` **is** the correctly-rounded `(2^k * s) / gs`, not an
approximation of it — where hoisting a reciprocal rounds twice. `1.5f * s` is exact too: `s` carries
4 significant bits and 1.5 carries 2, against fp32's 24.

Per value that leaves a select, a multiply by a power of two, and a sign XOR: **no divide, and no
E2M1 conversion at all**, since the magnitude is folded into which basis got selected. The group
scale keeps the CUTLASS conversion — once per sixteen values its cost is negligible, and it stays
exact for every byte, including those above `0x7F` where the integer `ue4m3` decode drops the sign
bit that `float_ue4m3_t::bitcast` honours. The sign is applied by **XOR, not OR**, for that same
reason: the basis can be negative, so the nibble's sign must flip it rather than merely set it, and
E2M1's `-0` still yields `-0.0`.

Both checkpoint-decode kernels take it, under `SPARKINFER_CT_NVFP4_ALU=3`, now the default.
`0`/`1`/`2` still select the earlier forms, so every arm comes out of one binary.

**Bit-identical, and proven rather than sampled.** The encoding space is 16 nibbles × 256
group-scale bytes, small enough to walk completely: `ct_nvfp4_basis_selftest` checks all of it
across 32 global scales — **131072 combinations, 0 bf16 mismatches, 0 fp32 mismatches on every
finite scale byte**. The only fp32 differences anywhere are the sign bit of a NaN on scale bytes
`0x7F`/`0xFF`, which no real checkpoint can contain. It runs inside `prefill_nvfp4_gpu_test`, so
`ctest` fails if the identity ever breaks.

## Proof of speedup

| | decode tok/s |
|---|--:|
| before (main) | 93.94 |
| after (this PR) | 93.90 |

| | prefill pp tok/s |
|---|--:|
| before prefill (main) | 6487.37 |
| after prefill (this PR) | 6496.85 |

```text
# (qwen3_gguf_bench ... sweep, median of 5 reps). RTX 5090 sm_120, main @ d8249ad.
# MAIN = SPARKINFER_CT_NVFP4_ALU=1 (main's default), BASIS = ALU=3 (this PR).
# One binary, one model load, arm order rotated between rounds.

r1 MAIN  decode128=95.3851 pp128=6554.8367 pp16k=9808.4245
r1 BASIS decode128=94.8610 pp128=6518.4278 pp16k=9695.8179
r2 BASIS decode128=94.3155 pp128=6511.0171 pp16k=9639.2142
r2 MAIN  decode128=94.0993 pp128=6487.2539 pp16k=9601.8362
r3 MAIN  decode128=93.9431 pp128=6487.4933 pp16k=9588.1014
r3 BASIS decode128=93.8984 pp128=6480.7301 pp16k=9587.8931
r4 BASIS decode128=93.8932 pp128=6482.6758 pp16k=9584.1995
r4 MAIN  decode128=93.8430 pp128=6470.5016 pp16k=9567.0696

FLAT, and reported as flat. Per-round prefill@128 deltas: -0.56 / +0.37 / -0.10 / +0.19%
(mean -0.02%); ranges overlap completely. decode@128 and prefill@16k likewise flat.

WHY it no longer moves: the weight decode this speeds up is a FIXED per-layer cost. Current main
runs the GDN projections through a native NVFP4 GEMM at the scored shape, so the decode does not
execute there at all; and at ctx=16384 it is amortized over 128x more tokens, so what remains of
it is below noise. It was worth +6.99% on the main it was written against, and that main is gone.

Exhaustive equivalence (ctest, prefill_nvfp4_gpu_test):
  ct_nvfp4_basis_selftest: OK combinations=131072 fp32_finite_mismatch=0 bf16_mismatch=0

Accuracy: teacher-forced score dumps over 101 positions BYTE-IDENTICAL between ALU=1 and ALU=3
(top1 1.000000, KL 0.000000); a 32-token continuation emits identical ids.

```

### DIFF
--- a/kernels/csrc/cuda/fused/prefill_nvfp4_sm120.cu
+++ b/kernels/csrc/cuda/fused/prefill_nvfp4_sm120.cu
@@ -486,7 +486,8 @@ __global__ void ct_dequant_nvfp4_kernel(const unsigned char* __restrict__ packed
 //   0  as merged: cutlass conversions, (v * s) / global_scale
 //   1  the integer decodes, same divide -- bit-identical
 //   2  mode 1 plus a hoisted reciprocal, which is NOT bit-identical (measured, see below)
-enum : int { CT_ALU_BASE = 0, CT_ALU_INT = 1, CT_ALU_RCP = 2 };
+//   3  the per-group basis below: no divide and no per-value conversion at all, bit-identical
+enum : int { CT_ALU_BASE = 0, CT_ALU_INT = 1, CT_ALU_RCP = 2, CT_ALU_BASIS = 3 };
 
 // e2m1 magnitudes doubled are {0,1,2,3,4,6,8,12} -- all < 16, so the table is the nibbles of one
 // 32-bit literal and the decode is a shift, a mask and an int->float, with no memory touched.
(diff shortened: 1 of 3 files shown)


skyrocket2026:
## sparkinfer modelopt auto-eval — `eval-modelopt:M`

| metric | value |
|---|---|
| **label** | `eval-modelopt:M` |
| PR decode tok/s | 94.95 |
| main decode tok/s | 95.13 |
| PR prefill@128 pp | 3978.90 |
| main prefill@128 pp | 3747.69 |
| prefill@128 vs main | +6.2% |
| PR prefill@16k pp | 9773.17 |
| main prefill@16k pp | 9852.07 |
| prefill@16k vs main | -0.8% |
| Polaris receipt | `3bd7b1cde2b74c49…` — TDX (Intel hardware attestation) |
| commit | `d3421b82e` |

ok

skyrocket2026:
## sparkinfer modelopt auto-eval — `eval-modelopt:none`

| metric | value |
|---|---|
| **label** | `eval-modelopt:none` |
| PR decode tok/s | 94.92 |
| main decode tok/s | 95.17 |
| decode speedup vs main | **1.00×** (-0.3%) |
| PR prefill@128 pp | 6628.18 |
| main prefill@128 pp | 6645.09 |
| prefill@128 vs main | -0.3% |
| PR prefill@16k pp | 9934.80 |
| main prefill@16k pp | 10006.02 |
| Polaris receipt | `850e55bacfa59fe6…` — TDX (Intel hardware attestation) |
| commit | `657daea04` |

within significance gate — not a verified decode@128 improvement

skyrocket2026:
## Closed: sparkinfer modelopt auto-eval — `eval-modelopt:none`

This PR's Qwen3.8-27B prefill@16k speed measured **-0.3%** vs main, with no verified prefill@16k improvement (the only scored dimension) — closing automatically. This bot evaluates every eligible PR in the repo against Qwen3.8-27B's decode AND prefill@128 speed specifically, regardless of what the PR is actually about — a close here isn't a judgment on the PR's purpose, just that it didn't move these particular metrics. Reopen (or open a fresh PR) if you have a fix or a different approach.

PR #861, opened 2026-08-16
perf(qwen3.8): native NVFP4 full-attention projections in batched prefill (1.19x prefill@128)

widecloud:
## Summary

The ModelOpt checkpoint already ships the 16 full-attention layers' `q/k/v/o` as NVFP4, but they were refit to Q4_K at load, so batched prefill re-expanded every one of them Q4_K -> int8 on every pass — keep the checkpoint's own bytes and feed them to the SM120 block-scaled GEMM the FFN and GDN projections already use. Separately, `pf_gdnc_prep`'s triangular inverse is rebuilt in blocked form, dropping it from 62 block barriers to 10.

## Proof of speedup

>

| | decode tok/s |
|---|--:|
| before (main) | 93.83 |
| after (this PR) | 93.25 |

| | prefill pp tok/s |
|---|--:|
| before prefill (main) | 6637.33 |
| after prefill (this PR) | 7700.30 |

Qwen3.8-27B-NVFP4 (ModelOpt), **ctx=128 — the scored dimension: +16.01%**. Medians of 3 alternating main/PR rounds, 5 reps each, one binary per arm, order rotated. prefill@16k 9541.12 -> 9841.90 (+3.15%) and decode@16k 78.36 -> 78.10 (-0.34%) — no regression on either floor.

**Attention.** ModelOpt quantizes co-located projections as one group, so `q_proj`, `k_proj` and `v_proj` share a bit-identical `weight_scale_2` per layer — checked at load, not assumed, since a mismatch would silently rescale two thirds of attention. Packed nibbles and group scales are both row-major, so concatenating along rows is an exact byte copy under one shared alpha. That pays because at m=128 the CUTLASS grid *is* the N-tile count, and k/v alone are n=1024 = 16 CTAs of 170 (~227 GB/s, essentially pure launch overhead). The same argument gives `gate_proj|up_proj`. Every FP4 GEMM group here is also preceded by a norm whose output it immediately re-reads and quantizes, so the quantize is folded into the producing norm at four sites — 176 fewer launches per pass, 1366 -> 1063 kernels.

**GDN chunk prep.** `pf_gdnc_prep` built `T = (I+A)^-1` by row-at-a-time forward substitution: 31 rows x 2 block barriers, a dependent FMA chain of `sum(i-1)` = 465, and never more than 31 of 256 threads busy. Doubling just that phase costs 0.556 ms/pass, so it was 36% of the kernel's 1.565 ms. The blocked form

    [[L11, 0], [L21, L22]]^-1 = [[T11, 0], [-T22 L21 T11, T22]]

is log2(32) = 5 levels: 10 barriers, a chain of 62, and the last two levels carry 94% of the MACs at 128 and 256 threads. It issues ~2x the MACs (C^3/3 vs C^3/6) and still wins by **+2.19%**, because the grid is 192 blocks on 170 SMs — barely one block per SM, so every barrier is fully exposed and arithmetic is nearly free. `L21` is still raw `A` at every level, so no copy is needed, and the `P = L21 T11` scratch aliases `s_x` (Q is dead after the wmma pass, V is not staged until after W^), so it costs no extra shared memory.

**The fused `gate|up` operand is all-layers-or-nothing, decided before anything loads.** Prefill makes `ffg`/`ffu` strided views of one buffer, so the fused layout is a property of the whole pass — and `keep_nvfp4(..., want_fp4 = !gu_cat)` drops the per-tensor FP4 payload for exactly the layers that fused, so a partially-fusable checkpoint would have no correct layout at all. `unsloth/Qwen3.8-27B-NVFP4` is exactly that case: `gate_proj` and `up_proj` share a `weight_scale_2` on only 56 of 64 layers. A host-side pre-scan (`nvfp4_src` is `st.tensor()` lookups plus a 4-byte `memcpy`, no device work) settles it up front, so that checkpoint keeps the per-tensor layout and is bit-identical to main on the FFN, while the ModelOpt checkpoint fuses all 64 as before.

**GDN chunk scan.** Two further changes to `pf_gdnc_scan`. (1) `M . U^` moves to tensor cores: the scalar epilogue was shared-INSTRUCTION bound, not FLOP bound — per `pp` it issued one `s_U` load plus **four broadcast `s_M` loads for only four FMAs**, ~1280 shared transactions per block-chunk against ~256 cycles of real FMA work, which is why a 32768-MAC phase cost more than a 131072-MAC one next to it. `m_buf` already stores hard 0 for j>i, so the dense product is the masked one, and the bf16 tile aliases `s_Sb` (dead after the U^/Y0 pass) for no extra shared memory. (2) The S update writes its wmma accumulator straight back to `s_S`: seed the fragment with the decayed old state, let the mma add `K^T U~` onto it, store in place. That deletes an 8 KB staging array, both `__syncwarp`, and a dependent shared round-trip per tile. Safe because `tile = 2*warp + rep` gives `ti = 16*warp` for both reps, so each warp owns its `s_S` rows exclusively. Together **+1.6%**, and `pf_gdnc_scan` drops 2.396 -> 2.250 ms/pass while `pf_gdnc_prep` drops 1.565 -> 1.100 ms.

**Attention, and the small kernels behind it.** `pf_attn_bf16_paged` ran one warp per (token, head) walking ONE key per iteration, each key paying a 5-step warp-shuffle reduction the online-softmax update then consumed — ~40-50 cycles of serial latency for 16 FMAs, measured at 3.9% of bf16 peak. It now processes a tile of KT=8 keys, so the reductions interleave and the sequential rescale happens once per tile (KT=4 and KT=16 both measure worse; 8 is the optimum). Separately, `launch_prefill_gemm_skinny` capped at n<=64 while the fused `in_proj_a|in_proj_b` is n=96, so it was declining and falling to a tiled GEMM that grids to ONE block; widening it to n<=96 is what makes that fusion pay.

**decode@128 is -0.5%** (measured -0.34% to -0.62% across gate runs on this branch), inside the 0.98 no-regression floor: the attention weights leave Q4_K, so decode's fused Q4_K `attn_qkv` kernel no longer applies and q/k/v go through `launch_gemv_nvfp4` on the existing qkvstream path.

**VRAM is +0.1 GB** (30.2 -> 30.3 GB; the CUTLASS SFB alone). The Q4_K copy is *replaced*, not added to — at ctx=16384 this model runs with under 1 GB free and a second copy (~0.9 GB) would starve the FFN chunk. Held under an external VRAM hog at 500/1000/2000 MB: still 16/16 attention layers, 0 scratch fallbacks.

Note on gates: `qwen3_gguf_score` never enters `prefill_batched_run`, so the accuracy gate cannot see either prefill change. The gate that does is `prefill_parity_check` (batched vs the sequential token loop, a different algorithm) — 1.000 on every run, both arms.

```text
########## Qwen3.8-27B-NVFP4 (ModelOpt), 3 alternating main/PR rounds, reps=5 ##########
  MEDIAN OF 3           main         PR       delta
  ctx=128    prefill   6637.33    7700.30     +16.01%     <- scored dimension
  ctx=16384  prefill   9541.12    9841.90      +3.15%
  ctx=16384  decode      78.36      78.10      -0.34%
  ctx=128    decode                            -0.5%      (within the 0.98 floor)

########## per-kernel, nsys --cuda-graph-trace=node, prefill only ##########
                            main d8249ad     this PR
pf_gdnc_prep                    1.565 ms     1.100 ms
pf_gdnc_scan                    2.396 ms     2.148 ms
pf_attn_bf16_paged              0.833 ms     0.676 ms
add_rmsnorm2 (unfused)     64/pass 0.228 ms   gone (the post-attn fold is restored)
quant_rows                 81/pass 0.285 ms   17/pass 0.067 ms
whole prefill pass             17.252 ms     16.663 ms

########## accuracy (differential, PR vs main, same token stream) ##########
positions             : 101
top-1 agreement       : 99/101 = 0.980   (PR vs main)
mean KL(main||pr)     : 0.0162 nats  (top-k union)
PPL PR                : 3.209
PPL main              : 3.235
METRIC top1=0.980198 kl=0.016178 ppl_pr=3.2088 ppl_main=3.2348

########## batched-prefill parity (absolute, bar 0.75) ##########
si_main run1/2/3 : worst=1.000 bar=0.75 OK
si_pr   run1/2/3 : worst=1.000 bar=0.75 OK

########## Qwen3.8 guard (unsloth/Qwen3.8-27B-NVFP4) -- 56/64 layers gate|up-fusable ##########
                   main       PR     delta
ctx=4096     9474.90   9758.70    +3.0%     (r1: 9548.68 -> 9779.78)
ctx=16384    8641.42   8869.49    +2.6%     (r1: 8692.57 -> 8907.38)
0 token-loop bails on every run. The load-time uniformity check declines the fused gate|up operand
on this checkpoint entirely, so its FFN weights are identical to main; the gain there is the GDN
and attention work, which applies to its linear-attention layers too.

########## VRAM pressure (external hog process, ctx=128) ##########
hog  500MB  main 6043.28  PR 6799.24  30.8 -> 30.9 GB  16/16 layers, 0 fallbacks
hog 1000MB  main 6054.57  PR 6735.42  31.3 -> 31.4 GB  16/16 layers, 0 fallbacks
hog 2000MB  main 6054.32  PR 6793.68  32.4 -> 32.5 GB  16/16 layers, 0 fallbacks

########## portable build (-DSPARKINFER_NVFP4=OFF, no CUTLASS) ##########
main d8249ad : 3 undefined refs (launch_ct_dequant_nvfp4 / _dev / _rows_i8) -- pre-existing on main
this PR      : the SAME 3, no PR-only additions
stub parity  : 13/13 symbols declared in prefill_nvfp4.h have a matching-arity stub in the .cu
```

### DIFF
--- a/kernels/csrc/cuda/fused/batched_prefill.cu
+++ b/kernels/csrc/cuda/fused/batched_prefill.cu
@@ -277,14 +277,16 @@ __global__ void pf_add_kernel(const __nv_bfloat16* __restrict__ a,
 __global__ void pf_split_q_gate_kernel(const __nv_bfloat16* __restrict__ qraw,
                                        __nv_bfloat16* __restrict__ q,
                                        __nv_bfloat16* __restrict__ gate,
-                                       int n_tokens, int n_heads, int head_dim) {
+                                       int n_tokens, int n_heads, int head_dim, int row_stride) {
     const long gid = (long)blockIdx.x * blockDim.x + threadIdx.x;
     const long per = (long)n_heads * head_dim;
     if (gid >= (long)n_tokens * per) return;
     const int t = gid / per;
     const int r = gid % per;                       // within-token index
     const int h = r / head_dim, d = r % head_dim;
-    const size_t src = (size_t)t * 2 * per + (size_t)h * 2 * head_dim + d;
+    // row_stride is 2*per for a tight [q|gate] buffer, and wider when [q|gate] is the leading
+    // slice of a fused q|k|v GEMM output that this reads in place instead of copying out.
+    const size_t src = (size_t)t * row_stride + (size_t)h * 2 * head_dim + d;
     q[gid]    = qraw[src];
     gate[gid] = qraw[src + head_dim];
 }
@@ -533,7 +535,7 @@ __global__ void pf_gdn_scan_kernel(const __nv_bfloat16* __restrict__ q,
                                    const __nv_bfloat16* __restrict__ a,
                                    float* __restrict__ state,
                                    __nv_bfloat16* __restrict__ out,
-                                   int n_tokens, int q_heads, int v_heads, bool qh_block) {
+                                   int n_tokens, int q_heads, int v_heads, int ab_stride, bool qh_block) {
     constexpr int NROW = HEAD_DIM / 32;
     const int vh   = blockIdx.x;
     const int j    = blockIdx.y * COLS + (threadIdx.x >> 5);
@@ -551,8 +553,8 @@ __global__ void pf_gdn_scan_kernel(const __nv_bfloat16* __restrict__ q,
     for (int r = 0; r < NROW; r++) sloc[r] = 0.f;   // fresh prefill: state starts at zero
 
     for (int t = 0; t < n_tokens; t++) {
-        const float bb = pf_sigmoid(pf_to_f(beta[(size_t)t * v_heads + vh]));
-        const float g  = __expf(pf_softplus(pf_to_f(alpha[(size_t)t * v_heads + vh]) + dt_h) * a_h);
+        const float bb = pf_sigmoid(pf_to_f(beta[(size_t)t * ab_stride + vh]));
+        const float g  = __expf(pf_softplus(pf_to_f(alpha[(size_t)t * ab_stride + vh]) + dt_h) * a_h);
         const __nv_bfloat16* qp = q + ((size_t)t * q_dim + qh * HEAD_DIM);
         const __nv_bfloat16* kp = k + ((size_t)t * q_dim + qh * HEAD_DIM);
         const __nv_bfloat16* vp = v + ((size_t)t * v_dim + vh * HEAD_DIM);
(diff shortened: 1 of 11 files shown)


skyrocket2026:
## sparkinfer modelopt auto-eval — error

**reason:** `PR speed/accuracy run failed (no crash diagnostic captured, possible hard kill — retried once)`

```

ssh: connect to host 181.84.95.171 port 7322: Connection refused

```
</details>

skyrocket2026:
## sparkinfer modelopt auto-eval — `eval-modelopt:REJECT`

| metric | value |
|---|---|
| **label** | `eval-modelopt:REJECT` |
| PR decode tok/s | 43.91 |
| main decode tok/s | 95.38 |
| decode speedup vs main | **0.46×** (-54.0%) |
| PR prefill@128 pp | 3753.48 |
| main prefill@128 pp | 6561.33 |
| prefill@128 vs main | -42.8% |
| PR prefill@16k pp | 9958.46 |
| main prefill@16k pp | 9806.28 |
| prefill@16k vs main | +1.6% |
| accuracy gate | ✅ top1=0.980 (bar >=0.9) · KL=0.0162 (bar <=0.1) |
| PPL PR / main | 3.2088 / 3.2348 |
| Polaris receipt | `55429c8153b660b5…` — TDX (Intel hardware attestation) |
| commit | `7fad27e06` |

decode@128 regression: 43.91 < 98% of main 95.38 | prefill@128 regression: 3753.48 < 98% of main 6561.33

PR #864, opened 2026-08-17
perf(qwen3.8): long-context prefill takes the paths tuned for ctx=128 (1.20x prefill@16k)

FranDev132:
## Summary

Three parts of the batched prefill are still shaped for the 128-token case, and at ctx=16384 that
is where the time is. nsys on the ModelOpt checkpoint, per prefill pass: FP4 GEMM 599.7 ms, GDN
scan 256.1, attention 229.6, `pf_gemm_i8` 384.8.

**1. The native-FP4 GDN arm stops at N ≤ 2048**, so at 16k every GDN projection falls back to the
int8 path. The tell is `pf_gemm_i8` firing exactly **208 times per pass = 48 GDN layers × 3 + 16
attention layers × 4**. That bound rested on two claims and measurement contradicts both. The
saving is not only the avoided `NVFP4 → bf16 → int8` expansion — the block-scaled GEMM *replaces*
the int8 GEMM. And the recurrence does not amplify the coarser activations with length: batched
prefill against the token-loop reference in the same build agrees **24/24 = 1.000 at N = 512 /
2048 / 8192 / 16384**, identical to the int8 arm at every one of those lengths. **+7.65%**

**2. Both NVFP4 GEMM tiles are M=128**, because at ctx=128 the grid is exactly one CTA tall and
only N matters. At m=16384 an M=128 tile reloads each B tile once per 128 rows. A **256×128×128**
tile for m ≥ 512 takes the GEMM **599.7 → 488.7 ms, 1238 → 1518 TFLOPS ≈ 91% of this part's NVFP4
peak**. **+6.7%**

Alternatives measured, so the choice is not a guess: 128×256×128 **−14%**, 256×256×64 **−76%**
(K=64 starves the mainloop). 384×128, 512×128 and 256×64×256 cannot carve two mainloop stages out
of shared memory, and non-power-of-two tiles fail cute's stride-divisibility and the epilogue's
`MMA_TILE_M | EPI_TILE_M` check — so 256×128×128 is the reachable optimum. `EVICT_FIRST` on B
stays correct at this tile (dropping it: 11310 vs 11409 pp).

**3. The attention kernel's int8 shared tiles collide on every access.** `s_qi`'s row stride is
HEAD_DIM (256 B) and `s_pi`'s is GN (128 B) — both exact multiples of the 128-byte bank row, so all
16 rows of a tile start on bank 0 and every `ldmatrix` replays. ncu at ctx=16384:

```
shared-load bank conflicts   2.48e9
shared-load wavefronts       3.27e9
shared-load instructions     4.37e8   -> 7.5 wavefronts/instruction (ideal 1)
tensor pipe active           20.5%    stalls: mio_throttle, short_scoreboard
```

Padding both by 16 B keeps the 16-byte alignment int8 `ldmatrix` requires and takes
`gcd(stride/4, 32)` from 32 to 4 — 8 distinct starting banks instead of 1. **+5.1%** (a 32 B pad
is worse than 16 B, as the arithmetic predicts: 4 distinct starts).

Padding is applied only at **ctx ≥ 2048**. It costs ~1 KB per block, which is enough to break the
2-blocks-per-SM tier `__launch_bounds__` asks for at RQH ≤ 2: applied unconditionally it cost the
Qwen3.6 guard **5.8% at ctx=512** while being +0.3% at 4096. The shared-memory opt-in is also
latched at the **largest** size any later launch can request rather than this call's — sizing it
from the current pad locked in the unpadded size on a small-context first call, after which every
padded launch silently failed and fell back (−45% at 16k, no diagnostic).

## Proof of speedup

| | decode tok/s |
|---|--:|
| before (main) | 94.97 |
| after (this PR) | 94.69 |

| | prefill pp tok/s |
|---|--:|
| before prefill (main) | 9675.52 |
| after prefill (this PR) | **11576.80** |

**1.1965× prefill@16k.** Qwen3.8-27B ModelOpt NVFP4, one binary, warm-up first then main/PR
alternated ×3 at 5 reps (an unalternated first run on this box reads cold and invents ~20% swings).

```text
  warmup  {"128":{"decode_tps":95.9328,"prefill_pp":6557.32},"16384":{"decode_tps":90.4085,"prefill_pp":9860.16}}
  r1 main {"128":{"decode_tps":95.4198,"prefill_pp":6513.34},"16384":{"decode_tps":89.9874,"prefill_pp":9756.05}}
  r1 PR   {"128":{"decode_tps":95.1443,"prefill_pp":6497.71},"16384":{"decode_tps":89.7269,"prefill_pp":11624.67}}
  r2 main {"128":{"decode_tps":94.9740,"prefill_pp":6490.45},"16384":{"decode_tps":89.5127,"prefill_pp":9675.52}}
  r2 PR   {"128":{"decode_tps":94.6915,"prefill_pp":6493.15},"16384":{"decode_tps":89.3941,"prefill_pp":11576.80}}
  r3 main {"128":{"decode_tps":94.6595,"prefill_pp":6497.50},"16384":{"decode_tps":89.3617,"prefill_pp":9655.67}}
  r3 PR   {"128":{"decode_tps":94.6364,"prefill_pp":6490.24},"16384":{"decode_tps":89.3176,"prefill_pp":11562.59}}

medians   prefill@16k 9675.52 -> 11576.80  = 1.1965x
          prefill@128 6497.50 -> 6493.15   = 0.9993
          decode@128    94.97 -> 94.69     = 0.9970
          decode@16k    89.51 -> 89.39     = 0.9987
```

## Correctness

All three changes are confined to `prefill_batched_run()`, which `qwen3_gguf_score` never enters,
so the teacher-forced score dumps are **byte-identical**: top1 1.000000, KL 0.000000, ppl 3.2348
both arms (real prompt via the checkpoint's own tokenizer, topk 128).

Batched-prefill parity, which *can* see this path: `PARITY worst=1.000 bar=0.75 OK`.

And the long-context check the GDN bound existed for — same prompt, token-loop reference vs batched
with the arm off and on, common-prefix agreement:

| N | int8 GDN | native-FP4 GDN |
|--:|---|---|
| 512 | 24/24 = 1.000 | 24/24 = 1.000 |
| 2048 | 24/24 = 1.000 | 24/24 = 1.000 |
| 8192 | 24/24 = 1.000 | 24/24 = 1.000 |
| 16384 | 24/24 = 1.000 | 24/24 = 1.000 |

## No-regression

Both guards warm-up-first and alternated, same protocol as above:

| guard | main | this PR | |
|---|--:|--:|--:|
| Qwen3.8 (unsloth) prefill@128 | 4106.61 | 4098.31 | 0.998 |
| Qwen3.8 (unsloth) decode@128 | 84.91 | 84.88 | 1.000 |
| Qwen3.6 prefill@512 | 9963.56 | 9971.21 | 1.001 |
| Qwen3.6 prefill@4096 | 25359.14 | 25407.05 | 1.002 |

Qwen3.6 decode is flat at 0/512/4096 (495.19–495.72 / 488.30–489.40 / 469.40–469.81).

## Scope

`SPARKINFER_Q38_GDN_NVFP4_MAXN=2048`, `SPARKINFER_NVFP4_BIG_TILE=0` and
`SPARKINFER_PREFILL_ATTN_SMEM_PAD=0` each restore the previous behaviour, so all three arms A/B in
one binary. The GEMM tile is gated on `m >= 512` and the padding on `ctx >= 2048`, so the ctx=128
shapes are untouched — which is what prefill@128 and decode@128 measuring flat above confirms.

### DIFF
--- a/kernels/csrc/cuda/fused/prefill_attn_mma.cu
+++ b/kernels/csrc/cuda/fused/prefill_attn_mma.cu
@@ -287,13 +287,31 @@ __global__ __launch_bounds__(GROUP_BLKS * 32, 3) void pf_attn_mma_i8_kernel(
 // RQH=1 is bit-identical to the per-head kernel. Math (mask, online softmax, int8
 // round) is unchanged -- only the load ordering differs.
 // ============================================================================
+// Shared-memory row padding for the int8 wmma operands. Without it s_qi's row stride is HEAD_DIM
+// (256 B = 64 banks) and s_pi's is GN (128 B = 32 banks), both exact multiples of the 128-byte bank
+// row, so all 16 rows of a tile start on bank 0 and every ldmatrix replays 16-way. Measured on the
+// unpadded kernel at ctx=16384: 2.48e9 shared-load bank conflicts over 3.27e9 wavefronts for
+// 4.37e8 instructions -- 7.5 wavefronts per instruction against an ideal of 1, which is why the
+// tensor pipe sits at 20.5% while the stalls are mio_throttle and short_scoreboard.
+// +16 B keeps the 16-byte alignment int8 ldmatrix requires and takes gcd(stride/4, 32) from 32 to
+// 4, i.e. 8 distinct starting banks instead of 1 (16-way -> 2-way).
+// SPARKINFER_PREFILL_ATTN_SMEM_PAD=0 restores the packed layout (A/B in ONE binary).
+inline int attn_smem_pad() {
+    static const int v = [] {
+        const char* e = getenv("SPARKINFER_PREFILL_ATTN_SMEM_PAD");
+        const int x = e ? atoi(e) : 16;
+        return (x == 0 || x == 16 || x == 32) ? x : 16;
+    }();
+    return v;
+}
+
 template <int HEAD_DIM, int GROUP_BLKS, int RQH>
 __global__ __launch_bounds__(GROUP_BLKS * 32, (RQH <= 2 ? 2 : 1)) void pf_attn_mma_gqa_kernel(
     const __nv_bfloat16* __restrict__ q, const signed char* __restrict__ k_pool,
     const signed char* __restrict__ v_pool, const __half* __restrict__ k_scale,
     const __half* __restrict__ v_scale, const int* __restrict__ block_table,
     __nv_bfloat16* __restrict__ attn, int n_tokens, int n_q_heads, int n_kv_heads,
-    int block_size, int max_blocks_per_seq, float scale, int win_blocks) {
+    int block_size, int max_blocks_per_seq, float scale, int win_blocks, int qld, int pld) {
     using namespace nvcuda::wmma;
     constexpr int BM    = 16;
     constexpr int GN    = GROUP_BLKS * 16;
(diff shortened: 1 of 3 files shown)


skyrocket2026:

| metric | value |
|---|---|
| **PR decode@16k tok/s** | **90.36** |
| **main decode@16k tok/s** | **90.38** |
| **decode@16k speedup vs main** | **1.00×** (+0.0%) |
| PR decode@128 tok/s | 95.99 |
| main decode@128 tok/s | 96.06 |
| decode@128 vs main (floor) | +0.0% |
| PR prefill@128 pp | 6569.34 |
| main prefill@128 pp | 6564.11 |
| prefill@128 vs main | +0.0% |
| PR prefill@16k pp | 11865.95 |
| main prefill@16k pp | 9806.55 |
| prefill@16k vs main | +21.0% |
| commit | `79e42ae6f` |

ok

skyrocket2026:

PR #865, opened 2026-08-17
perf(qwen3.8): one-wave GDN scan, GQA-6 attention tier, fused NVFP4 residual, native-FP4 attention projections (1.11x prefill@16k)

FranDev132:
## Summary

Four things at ctx=16384, found by profiling the pass rather than tuning kernels. Per prefill pass
(nsys, load-time kernels excluded): FP4 GEMM 500.7 ms at 1492 TFLOPS = **89% of NVFP4 peak, so it is
already at roofline and is not touched here**; GDN scan 256.9; attention 164.3; attention projections
88.7.

**1. The GDN scan was spending half its time on an empty second wave.** Its grid is
`v_heads x HD/JC` = 48 x 4 = **192 blocks**, and at 68.9 KB dynamic + 16.4 KB static shared memory
only ONE fits per SM (`ncu: launch__occupancy_limit_shared_mem = 1`). On 170 SMs that is 170 blocks
then 22 — **56% efficient no matter how fast the block is**, and nothing inside the block can see it.
JC=64 makes it 96 blocks and one wave, with the same 1536 warps in flight and half the W/K/Q staging
per state column.

The wave model was checked against an existing measurement *before* the rewrite: JC=16 is 384 blocks
= three waves and predicts ~20% slower on this kernel; it measures 21%. Laid out naively JC=64 needs
133 KB, so three regions are reused at disjoint points of the chunk body (`s_C` over `s_W|s_Q`, then
`s_Y` and `s_Ub` over `s_Sb`) to land at 89 KB. Numerics are unchanged — same tile shapes, same
k-loop order, same ascending reduction order. **+3.0%**

Gated on `v_heads * HD/JC > sm_count`, i.e. on the narrow grid *actually* spilling, not on the model
or the context. That gate is load-bearing: Qwen3.5/3.6 carry 32 v-heads, so their grid is 128 blocks
and already fits one wave, and widening it there pays a ~1.65x per-block cost for a wave saving that
does not exist — measured as **-6.1% on the Qwen3.6 guard at ctx=4096** before the gate was
corrected.

**2. Attention re-reads K/V out of L2, and GQA-6 never got the wide tier.** ncu at ctx=16384:

```
L2 throughput        80.3% of peak      <- the bound
L1TEX                75.1%
SM throughput        62.7%
DRAM                  1.7%              <- not a memory-bandwidth problem
```

The group is 24/4 = 6, so the shipped tiers (RQH=4, then 2) always fall through to 2 — 4 does not
divide 6. RQH=3 loads each K page and V tile once for three q-heads instead of two, taking a third
off the dominant traffic term. It stays at the two resident blocks its `__launch_bounds__` asks for
only because `s_o` (the epilogue landing zone) now overlays `s_s` (the scores, dead by the epilogue):
46,528 B against the 51,200 a second block needs, where holding both is 62,912 and caps it at one.
Long context only — it is +1.9% at 16k and -0.45% at ctx=128. **+1.9%**

**3. The block-scaled GEMM can accumulate the residual after all.** Two call sites carried a comment
saying it could not, so they wrote a raw projection to `ao` and paid a separate full-tensor add —
three N x H bf16 streams for one flop per element. CUTLASS's `LinearCombination` epilogue is already
`D = alpha*Acc + beta*C`; `launch_prefill_nvfp4_gemm` simply had no C parameter. Passing the residual
as C (aliasing D, which the epilogue handles — each output tile is read before it is stored) deletes
the pass. **+2.0% at 16k, and +0.96% at ctx=128.**

**4. The full-attention q/k/v/o were the last projections still on the int8 GEMM** — 88.7 ms/pass at
620 TOPS, 74% of the int8 peak, against the FP4 GEMM's 89% beside it. Loading them as the checkpoint
ships them (NVFP4) puts them on the same block-scaled GEMM. This REPLACES the Q4_K refit rather than
sitting beside it: the payload is 0.5625 B/weight either way, so only the CUTLASS SFB scale copy is
new residency, where holding both would have cost ~1.05 GB on a model that already peaks near the
card. Probed per tensor and only when the tensor is genuinely NVFP4 — the other shipped Qwen3.8
checkpoint stores FP8 attention, and routing that through the native path moves its decode off Q4_K
MMVQ. **+2.5% at 16k, +5.0% at ctx=128.**

## Proof of speedup

| | decode tok/s |
|---|--:|
| before (main) | 94.96 |
| after (this PR) | 94.33 |

| | prefill pp tok/s |
|---|--:|
| before prefill (main) | 11609.00 |
| after prefill (this PR) | **12841.38** |

**1.1062x prefill@16k.** Qwen3.8-27B ModelOpt NVFP4, ONE binary, warm-up first then main/PR
alternated x3 at 5 reps (an unalternated first run on this box reads cold and invents ~20% swings).

```text
  warmup  {"128":{"decode_tps":95.6529,"prefill_pp":6528.4502},"16384":{"decode_tps":90.1480,"prefill_pp":11766.4936}}
  r1 main {"128":{"decode_tps":95.3176,"prefill_pp":6510.4081},"16384":{"decode_tps":89.8600,"prefill_pp":11680.9791}}
  r1 PR   {"128":{"decode_tps":94.5552,"prefill_pp":6905.1433},"16384":{"decode_tps":89.1144,"prefill_pp":12885.9172}}
  r2 main {"128":{"decode_tps":94.9636,"prefill_pp":6505.8168},"16384":{"decode_tps":89.5145,"prefill_pp":11609.0005}}
  r2 PR   {"128":{"decode_tps":94.3298,"prefill_pp":6914.9628},"16384":{"decode_tps":88.9593,"prefill_pp":12841.3847}}
  r3 main {"128":{"decode_tps":94.8801,"prefill_pp":6498.5578},"16384":{"decode_tps":89.4433,"prefill_pp":11591.0142}}
  r3 PR   {"128":{"decode_tps":94.3190,"prefill_pp":6903.1726},"16384":{"decode_tps":88.9675,"prefill_pp":12836.0364}}

medians   prefill@16k 11609.00 -> 12841.38 = 1.1062x
          prefill@128  6505.82 ->  6905.14 = 1.0614x
          decode@128     94.96 ->    94.33 = 0.9934
          decode@16k     89.51 ->    88.96 = 0.9939
```

Per-change, each A/B'd on its own in the same binary: scan +3.0%, RQH=3 +1.9%, residual fold +2.0%,
native-FP4 attention +2.5%.

## Correctness

Changes 1-3 are confined to `prefill_batched_run()`, which `qwen3_gguf_score` never enters. Change 4
moves the attention weights decode reads, so the differential gate genuinely applies — measured the
way the bot measures it (the checkpoint's own tokenizer over `bench/scripts/eval_text.txt`, topk 128):

| | value | bar |
|---|--:|--:|
| top1 vs main | **0.980198** | >= 0.90 |
| KL vs main | **0.017701** | <= 0.10 |
| perplexity | **3.2348 -> 3.2040** | — |

Perplexity *improves*: the checkpoint's own NVFP4 bytes are more faithful to it than a Q4_K refit of
them. Batched-prefill parity, which does see the prefill path: `PARITY worst=1.000 bar=0.75 OK`.

## No-regression

Both guards warm-up-first and alternated, same protocol:

| guard | main | this PR | |
|---|--:|--:|--:|
| Qwen3.6 prefill@4096 | 25663.69 | 25660.54 | 0.9999 |
| Qwen3.6 prefill@512 | 10004.10 | 10010.13 | 1.0006 |
| Qwen3.8 (unsloth) prefill@128 | 4114.30 | 4144.38 | 1.007 |
| Qwen3.8 (unsloth) decode@128 | 85.12 | 85.05 | 0.999 |

Qwen3.6 decode is flat at 0/512/4096 (496.61-496.71 / 489.96-490.38 / 470.85-471.36 main against
496.46-496.64 / 489.89-490.31 / 470.65-470.93).

## Scope

`SPARKINFER_PREFILL_GDN_SCAN_BIGJC_MINCTX=999999`, `SPARKINFER_PREFILL_ATTN_GQA_RQH=2`,
`SPARKINFER_PREFILL_NVFP4_RESID_FUSE=0` and `SPARKINFER_Q38_ATTN_NVFP4=0` each restore the previous
behaviour, so all four arms A/B in one binary — which is how the per-change numbers above were taken
and how the Qwen3.6 regression was isolated to exactly one of them.

### DIFF
--- a/kernels/csrc/cuda/fused/prefill_attn_mma.cu
+++ b/kernels/csrc/cuda/fused/prefill_attn_mma.cu
@@ -306,7 +306,7 @@ inline int attn_smem_pad() {
 }
 
 template <int HEAD_DIM, int GROUP_BLKS, int RQH>
-__global__ __launch_bounds__(GROUP_BLKS * 32, (RQH <= 2 ? 2 : 1)) void pf_attn_mma_gqa_kernel(
+__global__ __launch_bounds__(GROUP_BLKS * 32, (RQH <= 3 ? 2 : 1)) void pf_attn_mma_gqa_kernel(
     const __nv_bfloat16* __restrict__ q, const signed char* __restrict__ k_pool,
     const signed char* __restrict__ v_pool, const __half* __restrict__ k_scale,
     const __half* __restrict__ v_scale, const int* __restrict__ block_table,
@@ -333,9 +333,15 @@ __global__ __launch_bounds__(GROUP_BLKS * 32, (RQH <= 2 ? 2 : 1)) void pf_attn_m
     // Per-q-head Q(int8), P(int8), scores(float); shared K/V scales; per-(qh,row) softmax state.
     signed char* s_qi = reinterpret_cast<signed char*>(mma_smem);   // [RQH][BM][qld]
     signed char* s_pi = s_qi + (size_t)RQH * BM * qld;               // [RQH][BM][pld]
+    // s_o OVERLAYS s_s. The scores are dead by the epilogue -- the last read of s_s is the P
+    // quantization inside the softmax section, and the PV mma after it touches only s_pi/s_ps/
+    // s_corr -- so the landing zone costs nothing on top of the score buffer it lands in. That
+    // is what makes RQH=3 fit: 46,528 B against the 51,200 a second resident block needs, where
+    // holding both buffers is 62,912 and caps this kernel at one block per SM.
+    constexpr int SBLK = (RQH * BM * GN > BM * HEAD_DIM) ? RQH * BM * GN : BM * HEAD_DIM;
     float* s_s  = reinterpret_cast<float*>(s_pi + (size_t)RQH * BM * pld); // [RQH][BM][GN]
-    float* s_o  = s_s + (size_t)RQH * BM * GN;                       // [BM][HEAD_DIM] epilogue landing
-    float* s_ks = s_o + BM * HEAD_DIM;                               // [GN] shared
+    float* s_o  = s_s;                                               // [BM][HEAD_DIM] epilogue landing
+    float* s_ks = s_s + SBLK;                                        // [GN] shared
     float* s_vs = s_ks + GN;                                         // [GN] shared
     float* s_qs = s_vs + GN;                                         // [RQH][BM]
     float* s_ps = s_qs + RQH * BM;                                   // [RQH][BM]
@@ -559,10 +565,11 @@ static bool launch_attn_gqa(const void* q, const signed char* k_pool, const sign
     // small-context first call, after which every padded launch failed and silently fell back --
     // measured as -45% at ctx=16384 and -34% on the Qwen3.6 guard at ctx=4096, with no diagnostic.
     const int qld_max = HD + attn_smem_pad(), pld_max = GN + attn_smem_pad();
+    // s_o lands in s_s (dead by the epilogue), so the pair costs the larger of the two.
+    constexpr int SBLK = (RQH * BM * GN > BM * HD) ? RQH * BM * GN : BM * HD;
     const size_t sm = (size_t)RQH * BM * qld                         // s_qi (int8, padded)
                     + (size_t)RQH * BM * pld                         // s_pi (int8, padded)
-                    + (size_t)(RQH * BM * GN) * sizeof(float)        // s_s
-                    + (size_t)(BM * HD) * sizeof(float)              // s_o
+                    + (size_t)SBLK * sizeof(float)                   // s_s, with s_o overlaid
                     + (size_t)(2 * GN + 5 * RQH * BM) * sizeof(float);
     // At RQH=4 this is 76,032 B — past the 48 KB default, so the opt-in below is
     // REQUIRED for the launch to be valid, and both it and the launch itself have to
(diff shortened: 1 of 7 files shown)


skyrocket2026:
## sparkinfer modelopt auto-eval — `eval-modelopt:L`

| metric | value |
|---|---|
| **label** | `eval-modelopt:L` |
| **PR decode@16k tok/s** | **90.15** |
| **main decode@16k tok/s** | **90.57** |
| **decode@16k speedup vs main** | **0.99×** (-0.4%) |
| PR decode@128 tok/s | 95.65 |
| main decode@128 tok/s | 96.11 |
| decode@128 vs main (floor) | -0.4% |
| PR prefill@128 pp | 6964.92 |
| main prefill@128 pp | 6572.51 |
| prefill@128 vs main | +5.9% |
| PR prefill@16k pp | 13352.28 |
| main prefill@16k pp | 11896.11 |
| prefill@16k vs main | +12.2% |
| accuracy gate | ✅ top1=0.980 (bar >=0.9) · KL=0.0177 (bar <=0.1) |
| PPL PR / main | 3.204 / 3.2348 |
| commit | `2528769ed` |

ok

skyrocket2026:

PR #866, opened 2026-08-18
perf(qwen3.8): wider activation-quantize lanes and a register-resident GDN state (1.03x prefill@16k)

FranDev132:
## Summary

Two changes at ctx=16384, both **bit-identical** — the `qwen3_gguf_score` dumps compare
byte-for-byte equal against the arm-off build, each one checked on its own.

**1. The NVFP4 activation quantizer mapped two values to a lane.** `quant_rows` did a 4-byte load
and a **one-byte store** per lane, so a whole warp stored 32 contiguous bytes; it was moving
~33.8 GB per pass in 62.8 ms, roughly **30% of this part's bandwidth**, while doing nothing but
copy. Eight values per lane makes that one 16-byte load and one 4-byte store, and the amax butterfly
collapses from 3 shuffles to 1.

The tell that this was the shape to want: its sibling `swiglu_quant_rows` **already had it**, and
its own comment records the same rewrite ("0.98 TB/s of a 1.79 TB/s part... issue-bound on narrow
accesses"). That is exactly why swiglu profiles near 55% of peak and this one did not. Bit-identical
because max is order-independent, so every group keeps the same scale and every value the same
`x / float(qs)` rounding. **+1.0%**

**2. The GDN scan kept its recurrent state in shared memory when it does not need to.** Every loop
that touches S already walks `e = tid; e < HD*JC; e += nthr`, so thread `tid` owns a fixed private
set of elements — `(m, jj) = (warp + t*NW, lane)` — with no communication between threads. S can
therefore live in registers. That frees `HD*JC*4` B and takes the block from **62,208 to 45,824 B,
under the 51,200 a SECOND resident block per SM needs**, so the whole grid is resident across every
SM instead of the half-width 96-block shape that was used to dodge a second wave.

The cost is that the S update can no longer write one tile per warp: each thread reads back the tile
covering its own elements, so all `(HD/16)x(JC/16)` accumulator tiles have to be live at once. At
JC=32 that is 16 KB and still fits the existing `s_W|s_Q` alias window; at JC=64 it would be 32 KB
and does not, so this is a JC=32 shape. Gated on the narrow grid actually spilling past the SM
count, which is why Qwen3.5/3.6 — 32 v-heads, 128 blocks, already one wave — never take it. **+0.9%**

**On (2), the arithmetic was optimistic and the measurement is what is reported.** Two resident
blocks per SM predicted ~+5%; it delivered +0.9%. The residency does appear — the restructuring
needed to reach it (one extra `__syncthreads`, and writing 16 tiles instead of reusing 8 slots)
spends most of the gain. Occupancy arithmetic is an upper bound, not an estimate.

## Proof of speedup

| | decode tok/s |
|---|--:|
| before (main) | 93.74 |
| after (this PR) | 93.75 |

| | prefill pp tok/s |
|---|--:|
| before prefill (main) | 12853.47 |
| after prefill (this PR) | **13182.85** |

**1.0256x prefill@16k.** Qwen3.8-27B ModelOpt NVFP4, ONE binary with both arms behind env knobs,
warm-up first then main/PR alternated x3 at 5 reps. Cross-*process* comparison on this box drifts
about 1%, which is half this result, so every number below is same-binary alternated — a reps=1 run
also reads ~5% high because the first pass gets a larger FFN chunk than reps 2-5.

```text
  warmup  {"128":{"decode_tps":94.2123,"prefill_pp":6935.4418},"16384":{"decode_tps":88.9242,"prefill_pp":12953.8207}}
  r1 main {"128":{"decode_tps":94.0005,"prefill_pp":6912.4896},"16384":{"decode_tps":88.6536,"prefill_pp":12893.3111}}
  r1 PR   {"128":{"decode_tps":93.8858,"prefill_pp":6887.8469},"16384":{"decode_tps":88.6422,"prefill_pp":13208.5705}}
  r2 main {"128":{"decode_tps":93.7368,"prefill_pp":6901.9602},"16384":{"decode_tps":88.4646,"prefill_pp":12853.4703}}
  r2 PR   {"128":{"decode_tps":93.6995,"prefill_pp":6889.8709},"16384":{"decode_tps":88.4863,"prefill_pp":13182.8451}}
  r3 main {"128":{"decode_tps":93.6306,"prefill_pp":6903.1443},"16384":{"decode_tps":88.4693,"prefill_pp":12845.2413}}
  r3 PR   {"128":{"decode_tps":93.7498,"prefill_pp":6907.1376},"16384":{"decode_tps":88.5337,"prefill_pp":13178.7579}}

medians   prefill@16k 12853.47 -> 13182.85 = 1.0256x
          prefill@128  6903.14 ->  6889.87 = 0.9981
          decode@128     93.74 ->    93.75 = 1.0001
          decode@16k     88.47 ->    88.53 = 1.0007
```

Per change, each A/B'd on its own in the same binary: quantizer lane width +1.0% (+0.82% / +1.19%),
register-resident S +0.9% (+0.60% / +1.13%).

## Correctness

Both changes are bit-identical by construction and were checked that way rather than argued:

| check | result |
|---|---|
| `qwen3_gguf_score` dump, quantizer arm on vs off | **byte-for-byte identical** (`cmp` clean) |
| `qwen3_gguf_score` dump, register-S arm on vs off | **byte-for-byte identical** (`cmp` clean) |
| batched-prefill parity (absolute, batched vs token loop) | `PARITY worst=1.000 bar=0.75 OK` |

For (1) max is order-independent, so a group's amax is the same however its 16 values are split
across lanes, and each value then keeps the same `qs` and the same division. For (2) only *where* S
lives changes — same values, same per-element order, same `gl * S + K^T U~`.

## No-regression

Both guards warm-up-first and alternated, same protocol:

| guard | main | this PR | |
|---|--:|--:|--:|
| Qwen3.6 prefill@4096 | 25519.20 | 25529.12 | 1.0004 |
| Qwen3.6 prefill@512 | 9974.26 | 9978.05 | 1.0004 |
| Qwen3.8 (unsloth) prefill@128 | 4117.57 | 4130.62 | 1.003 |
| Qwen3.8 (unsloth) decode@128 | 84.90 | 84.91 | 1.000 |

Qwen3.6 decode is flat at 0/512/4096 (495.16-496.08 / 488.98-489.79 / 470.08-470.18 main against
495.37-495.66 / 489.20-489.48 / 469.58-470.40).

## Scope

`SPARKINFER_NVFP4_QUANT_LPV=8` and `SPARKINFER_PREFILL_GDN_SCAN_REGS=0` each restore the previous
behaviour, so both arms A/B in one binary — which is how the per-change numbers above were taken.
The register-S path is additionally gated on `v_heads * HD/JC > sm_count`, so a model whose scan
grid already fits one wave never enters it.

### DIFF
--- a/kernels/csrc/cuda/fused/prefill_gdn_chunk.cu
+++ b/kernels/csrc/cuda/fused/prefill_gdn_chunk.cu
@@ -274,7 +274,22 @@ __global__ void pf_gdnc_prep_kernel(const __nv_bfloat16* __restrict__ q,
 // grid = (v_heads, HD/JC); each block owns JC state columns of one v-head and walks the chunks.
 // Per chunk: U^ = U0 - W^ S ; Y = [diag(exp G)(Q S) + M U^] * scale ; S = exp(G_last) S + K^T U~.
 // ---------------------------------------------------------------------------
-template <int C, int HD, int JC>
+// REGS keeps the recurrent state S in REGISTERS instead of shared memory. S is [HD][JC] fp32 and
+// every loop that touches it already walks `for (e = tid; e < HD*JC; e += nthr)`, so thread tid
+// owns exactly the elements e = tid + t*NTHR -- a fixed (m, jj) = (warp + t*NW, lane) mapping, no
+// communication. That frees HD*JC*4 B of shared memory, which is what buys a SECOND resident block
+// per SM: 45,824 B against the 51,200 two blocks need, where the smem-S form is 62,208 and caps
+// the kernel at one. The whole grid is then resident in one wave using every SM, instead of the
+// 96-block/96-SM shape JC=64 had to use to avoid a second wave.
+//
+// The cost is that the S update can no longer write its own tile per warp: each thread now reads
+// back the tile covering ITS elements, so all (HD/16)x(JC/16) accumulator tiles must be live at
+// once. At JC=32 that is 16 KB, which still fits the s_W|s_Q alias window; at JC=64 it would be
+// 32 KB and does not, so REGS is a JC=32 shape only.
+//
+// Bit-identical: same values, same per-element order, same `gl * S + K^T U~` arithmetic -- only
+// where S lives changes.
+template <int C, int HD, int JC, bool REGS = false>
 __global__ void pf_gdnc_scan_kernel(const __nv_bfloat16* __restrict__ q,
                                     const __nv_bfloat16* __restrict__ k,
                                     const float* __restrict__ g_buf,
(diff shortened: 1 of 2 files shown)


skyrocket2026:
## sparkinfer modelopt auto-eval — `eval-modelopt:XS`

| metric | value |
|---|---|
| **label** | `eval-modelopt:XS` |
| **PR decode@16k tok/s** | **90.19** |
| **main decode@16k tok/s** | **90.19** |
| **decode@16k speedup vs main** | **1.00×** (+0.0%) |
| PR decode@128 tok/s | 95.66 |
| main decode@128 tok/s | 95.67 |
| decode@128 vs main (floor) | +0.0% |
| PR prefill@128 pp | 6942.06 |
| main prefill@128 pp | 6959.64 |
| prefill@128 vs main | -0.2% |
| PR prefill@16k pp | 13793.85 |
| main prefill@16k pp | 13365.85 |
| prefill@16k vs main | +3.2% |
| commit | `44e1c4505` |

ok

skyrocket2026:

PR #869, opened 2026-08-18
perf(dspark): stop drafting once the batched verify has stopped arming (1.10x dspark-decode@128)

Paral1995:
## Summary

In the token loop a draft block can never save a target forward — a step that keeps `keep` tokens
runs exactly `keep` target forwards, the same count AR runs — so its only payoff is arming the
row-batched verify. At ctx=128 that arming needs a **full-block** accept, and the mean accepted
length is 1.19, so `compact_score` sits at zero for the whole generation while every step still
pays for a draft block. Measured on the scored shape, same weights and same process:
DSPARK 77.28 against AR 87.02.

`dflash_generate` already takes the AR path outright for the band that starts above
`kCompactMaxSeq` and ends below `kEngageMinSeq`, for exactly this reason. That predicate is
static, so it cannot cover a short-context stream whose compact path is available in principle
and then never arms in practice. This adds the dynamic half: after `kDraftProbeAfter` consecutive
disarmed steps the draft block is skipped and the step degenerates to AR, and every
`kDraftProbePeriod`-th step drafts anyway so a stream whose draft starts tracking re-arms exactly
as it would have.

It cannot fire where speculation wins — arming resets the counter, so a stream holding the
batched verify never reaches the threshold. `SPARKINFER_DFLASH_IDLE_DRAFT=0` restores the
unconditional draft, so both arms come out of one binary.

## Proof of speedup

**Decode tok/s** (`dspark_tau_check <model> <draft> 128 <128-token prompt>`, `METRIC DSPARK_TPS`):

| | decode tok/s |
|---|--:|
| before (main) | 77.28 |
| after (this PR) | 85.53 |

**+9.97% dspark-decode@128.** Both arms from one binary, interleaved round-robin, median of three
rounds. A third arm byte-identical to the baseline ran every round as a noise floor:

```text
  round 1: main 77.7888  idle 85.4344  ctrl 77.4845
  round 2: main 77.2821  idle 84.9860  ctrl 77.0724
  round 3: main 76.9764  idle 84.6700  ctrl 76.9127

  main 77.282 | idle 84.986 | control 77.070
  +9.97%, 3/3 rounds, noise floor -0.27%
```

The step now runs one target forward and emits one token, which is what AR does, so DSPARK_TPS
lands at 98.3% of the AR_TPS measured in the same process (86.98).

## Correctness

`LOSSLESS 1` on every arm and every context measured — skipping a draft means proposing nothing,
never emitting an unverified token, so the guarantee is structural rather than empirical here.

```text
  ctx=128   main  DSPARK 77.93  accept 1.1852  lossless 1
            idle  DSPARK 85.53  accept 1.0079  lossless 1

  ctx=1024  main  DSPARK 56.37  accept 1.0240  lossless 1     (AR 64.33)
            idle  DSPARK 62.81  accept 1.0000  lossless 1     (AR 64.30)

  ctest: 100% tests passed, 0 failed out of 19
```

Scope: the change is inside `dflash_generate` only. The differential accuracy gate
(`qwen3_gguf_score`) and the ctx-16k no-regression guards (`qwen3_gguf_bench`) do not enter this
function, so neither is affected by it.

**Not tested, stated plainly:** a context where speculation actually wins. The available prompt
corpus tops out at ~1.3k tokens, and the batched verify is reported to engage around 4k
(tau ~3.9). The skip cannot fire while the verify is armed — arming resets the counter on the
same step — and at ctx=1024 it only helped, but that is an argument from the code path plus two
measured contexts, not a measurement at 4k.

### DIFF
--- a/runtime/src/models/qwen35.cpp
+++ b/runtime/src/models/qwen35.cpp
@@ -3154,6 +3154,29 @@ std::vector<int> Qwen35Model::dflash_generate(const std::vector<int>& prompt, in
         return v < 1 ? 1 : v;
     }();
     int compact_score = 0;
+    // Consecutive steps on which the batched verify stayed disarmed. In the token loop a draft
+    // block can NEVER save a target forward (see the strict-LOSS note above: a step that keeps
+    // `keep` tokens runs exactly `keep` target forwards, the same count AR runs) -- its only
+    // payoff is arming the batched verify. So once the draft has demonstrably stopped landing full
+    // blocks, the block is pure overhead and is skipped, which degenerates the step to AR.
+    // Probes keep it self-correcting: every kDraftProbePeriod steps one block is drafted anyway,
+    // so a stream whose draft starts tracking re-arms exactly as it would have.
+    // SPARKINFER_DFLASH_IDLE_DRAFT=0 restores main's unconditional draft.
+    int disarmed_run = 0;
+    static const int kIdleDraftOn = []{
+        const char* e = getenv("SPARKINFER_DFLASH_IDLE_DRAFT");
+        return (e && e[0] == '0') ? 0 : 1;
+    }();
+    static const int kDraftProbeAfter = []{
+        const char* e = getenv("SPARKINFER_DFLASH_IDLE_AFTER");
+        int v = e ? atoi(e) : 8;
+        return v < 1 ? 1 : v;
+    }();
+    static const int kDraftProbePeriod = []{
+        const char* e = getenv("SPARKINFER_DFLASH_IDLE_PROBE");
+        int v = e ? atoi(e) : 32;
+        return v < 2 ? 2 : v;
+    }();
     int keep_ema8 = 0;   // EMA of keep, alpha = 1/8, held x8 so the decode path needs no float
     int step_no = 0;
     bf16* th_scratch = nullptr;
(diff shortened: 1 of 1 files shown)


skyrocket2026:

| metric | value |
|---|---|
| **PR DSpark tok/s** | **85.62** |
| **main DSpark tok/s** | **77.93** |
| **speedup vs main** | **1.099×** (+9.8%) |
| PR AR tok/s (floor) | 87.06 |
| main AR tok/s (floor) | 87.17 |
| **DSpark vs AR** | **0.984×** — above 1.0 means speculation finally pays |
| mean accept τ | 1.008 (main 1.185, ceiling 7) |
| qwen3.8 (ModelOpt) guard @16k | ✅ no regression (decode+prefill) — decode 89.6 tok/s · prefill 13608 pp |
| qwen3.6 guard @16k | ✅ no regression (decode+prefill) — decode 473.7 tok/s · prefill 28286 pp |
| commit | `ce7b32deb` |

ok

<sub>Scored on the pinned eval box vs same-box `origin/main`: DSpark speculative decode throughput at ctx=128 on the ModelOpt NVFP4 checkpoint, with the AR reference measured in the same process and the same model load. Both a regression in AR decode and any divergence from the AR token sequence are hard REJECTs — a speculative decoder that is fast because it skips verification is not faster, it is wrong. τ is the lever, and the row above reports it against a block_size of 7. This is informational, not a judgment on your PR: a `none` label just means no measurable DSpark decode@128 speedup was verified, which is expected and fine if that isn't what your change is about. Automated — merge behaviour depends on SPARKINFER_DSPARK_AUTOMERGE.</sub>

skyrocket2026:

PR #872, opened 2026-08-18
perf(dspark): stage a deeper KV tile when the split count is small (1.14x dspark-decode@4k)

James-CUDA:
## Summary

`FA_GQA6_TILE` is 8, a sweet spot tuned at `n_splits=128` where one block walks only
`seqlen/n_splits ≈ 32` keys. At a small split count the same block walks the whole range, and the
shallow tile leaves the attention kernel latency-starved. Stage 32 tokens per iteration in that
regime instead of 8, selected on the per-block chunk so the tuned regime is untouched.

## Proof of speedup

**Decode tok/s** (end-to-end, ctx=4096):

| | decode tok/s |
|---|--:|
| before (main) | 26.58 |
| after (this PR) | 30.30 |

Numbers come from `runtime/examples/dspark_tau_check.cpp`, not `bench/scripts/bench.sh`: the latter
benchmarks a GGUF, and the path this PR touches is the NVFP4 DSpark decode. `dspark_tau_check` is
end-to-end (a full generation, timing the whole decode, AR and speculative measured in one process),
not an isolated-kernel microbenchmark, and it reports the losslessness gate alongside the throughput.

Both arms are the **same binary**, with `SPARKINFER_FA_TILE_DEEP` selecting the tile — the shallow
arm is main's code path exactly, which removes build and machine variance from the comparison.

```text
$ SPARKINFER_FA_TILE_DEEP=0 build/runtime/dspark_tau_check \
      models_q38_modelopt dspark 4096 <4096 prompt ids>
$ SPARKINFER_FA_TILE_DEEP=1 build/runtime/dspark_tau_check \
      models_q38_modelopt dspark 4096 <4096 prompt ids>

tile_deep  dspark_tps  ar_tps    accept    lossless
0          26.5818     26.7207   1.0828    1
1          30.2970     30.6836   1.0828    1
```

`LOSSLESS 1` on both arms, and acceptance is identical to four decimals (1.0828), i.e. the change is
pure speed with no numerical drift.

## Why it is bit-identical

The mainloop still walks tokens `start..end` in order, tiles are consecutive, and the online-softmax
update is per token in that order — so every row accumulates the same keys in the same sequence.
Only where a tile boundary falls changes.

## Cost

Dynamic shared memory is `2 * TILE * 256 * 2 B = 1024*TILE`. This launcher never opts past the 48 KB
default, so 32 → 32 KB is safe. `SPARKINFER_FA_TILE_DEEP=0` restores one-per-CTA staging.

### DIFF
--- a/kernels/csrc/cuda/attention/flash_decode_split.cu
+++ b/kernels/csrc/cuda/attention/flash_decode_split.cu
@@ -397,6 +397,18 @@ __global__ void fa_combine_gated_q8_kernel(
 #ifndef FA_GQA6_TILE
 #define FA_GQA6_TILE 8     // Qwen3.8-27B hd256 GQA-6 (24Q/4KV); independently sweepable
 #endif
+// Deep tile for the SMALL-split regime. FA_GQA6_TILE above is a sweet spot at n_splits=128, where a
+// block walks seqlen/n_splits ~= 32 keys and the staging loop runs two or three times. When the
+// split count is small the same block walks the WHOLE range instead: at 4k with n_splits=1 that is
+// 512 stage/sync/compute/sync cycles, each issuing only blockDim*16 B of loads before its barrier,
+// so DRAM latency is exposed every iteration -- measured 1.11 ms/call, ~30 GB/s across 4 CTAs
+// (~7.5 GB/s each) against a 18.7 us roofline for the bytes it reads. Staging a deeper tile puts
+// proportionally more independent loads in flight per barrier and cuts the barrier count by the
+// same factor. Capped by dynamic smem: 2 * TILE * 256 * 2 B = 1024*TILE, and this launcher never
+// opts in past the 48 KB default, so 32 -> 32 KB is safe and 48 would not be.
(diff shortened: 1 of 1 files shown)


skyrocket2026:

| metric | value |
|---|---|
| **PR DSpark tok/s** | **36.28** |
| **main DSpark tok/s** | **32.44** |
| **speedup vs main** | **1.118×** (+11.8%) |
| PR AR tok/s (floor) | 39.23 |
| main AR tok/s (floor) | 34.72 |
| AR vs main (floor) | +13.0% |
| **DSpark vs AR** | **0.925×** — above 1.0 means speculation finally pays |
| qwen3.8 (ModelOpt) guard @16k | ✅ no regression (decode+prefill) — decode 89.5 tok/s · prefill 13559 pp |
| qwen3.6 guard @16k | ✅ no regression (decode+prefill) — decode 474.3 tok/s · prefill 28336 pp |
| commit | `7db53bd08` |

ok

skyrocket2026:

PR #875, opened 2026-08-18
fix(dspark_tau_check): remove the confounded NSPLITS=1 pin — measurement fidelity, no speedup claimed (resubmit of #871 as requested)

rsnetworkinginc:
## Summary

Resubmit of #871 in the form requested there: **a measurement-fidelity fix, explicitly not a
speedup claim.** No `eval:*` tier is sought; the engine is untouched and no served request gets
faster. The change removes `dspark_tau_check`'s `SPARKINFER_NSPLITS=1` pin so the tool's AR and
DSpark legs run attention at the split count the engine itself selects, and replaces the pin's
comment with the corrected history so the confound is recorded where it happened.

Why the pin is wrong, briefly: its justification (27-32/32 agreement at adaptive splits) was
measured while the two atomic split-K prefill GEMMs were still live — the nondeterminism it saw
came from prefill, later root-caused and pinned separately. The decode split path has a fixed
reduction order end to end (`fa_split` owns fixed KV stripes; `fa_combine` folds warp-strided then
in ascending warp index; no atomics), consistent with `Qwen35Model`'s own "occupancy tuning only"
note — the contradiction you said you hit independently. Both of the tool's legs shift together
under the fix, so tau, the DSPARK/AR ratio, and losslessness verdicts are unaffected; the
instrument just stops disagreeing with the engine's configuration.

## Determinism evidence, rerun on THIS tree at the scored context

As requested: current main (post #867/#868/#869 — depth ladder, Markov gate, idle draft), ctx=4096
prompt built exactly as the bot builds it (`bench_prompt_4k.txt`, 4096 real tokens), adaptive
splits, prefill pins in place, RTX 5090:

```text
prompt A (bench_prompt_4k): AR-REPS 20 reps, 0 mismatches
                            SPEC-REPS 40 reps, 0 differ-from-first, 0 not-lossless-vs-AR
                            MEAN_ACCEPT 1.0756  LOSSLESS 1  LOSSLESS_RUNS 40
prompt B (eval_corpus-derived, 4096 tokens): SPEC-REPS 24 reps, 0 differ-from-first, 0 not-lossless-vs-AR
```

Every repeat re-runs the full generation including the token-loop prompt prefill, so these
cover the exact system the bot now measures.

## Proof of speedup

Not applicable, and deliberately not claimed — this PR must not tier. It changes
`runtime/examples/dspark_tau_check.cpp` only. Per #871's close, a delta the bot measures for this
diff is the instrument moving, not decode; if the bot runs it, the honest label is the one that
reflects that.

- [x] Tested on **RTX 5090** (`sm_120`) — all evidence above from a real 5090.

### DIFF
--- a/runtime/examples/dspark_tau_check.cpp
+++ b/runtime/examples/dspark_tau_check.cpp
@@ -45,13 +45,25 @@ int main(int argc, char** argv) {
         printf("usage: %s <qwen38_dir> <dspark_dir> <max_new> <id0> [id1 ...]\n", argv[0]);
         return 2;
     }
-    // Flash-decode's split-K reduction is atomic and therefore order-dependent: default splits
-    // (adaptive, effectively ~32) gave 27-32/32 agreement across repeat runs of the SAME decode,
-    // vs 32/32 x3 with SPARKINFER_NSPLITS=1. A tau/lossless measurement needs the target's own
-    // AR decode to be deterministic, or a real divergence from the verify head becomes
-    // indistinguishable from ordinary split-K noise. `0` (don't overwrite) so an explicit env
-    // override still wins for anyone deliberately sweeping splits.
-    setenv("SPARKINFER_NSPLITS", "1", 0);
+    // NO NSPLITS PIN (measurement-fidelity fix; see PR discussion on the first attempt and the
+    // maintainer's independent finding of the same contradiction). This used to force
+    // SPARKINFER_NSPLITS=1, on the observation that default (adaptive) splits gave 27-32/32
+    // agreement across repeat runs of the same decode, vs 32/32 with the pin. That experiment
+    // predates the two prefill pins directly below: the atomic split-K prefill GEMMs were live
+    // while NSPLITS was being blamed, so the 27-32/32 result was confounded -- the
+    // nondeterminism came from prefill, not from flash-decode. Consistent with that,
+    // Qwen35Model's own constructor documents the split count as occupancy tuning with
+    // identical math for any value: fa_split assigns each split a fixed KV stripe and
+    // fa_combine folds partials warp-strided then in ascending warp index, with no atomics
+    // anywhere in the decode path.
+    //
+    // Determinism re-verified on THIS tree at the scored context (RTX 5090, ctx=4096 prompt from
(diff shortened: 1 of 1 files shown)


skyrocket2026:
Closing this, and I want to be straight with you about why, because you did what I asked in #871 and the analysis holds up.

**Your diagnosis is correct.** I verified it independently today. The `NSPLITS=1` pin costs 35% of decode throughput at ctx=4096 on this box — 93.40 tok/s at the split count the engine selects, 60.43 tok/s pinned — and the nondeterminism the pin was built on does appear to have come from the prefill GEMMs that were pinned separately afterwards. The confound is real and the pin is costing measurement fidelity. You were right, and you were right in #871 too.

**But the eval harness is not contributor territory, and that is a policy decision rather than a judgement on this diff.** `dspark_tau_check.cpp` defines what every PR is scored against: which env the two legs run under, how throughput is computed, what counts as lossless. A change to it moves the baseline for everyone, including the contributor proposing it. Even when the change is correct, accepting it through the contribution flow creates an incentive to compete on the measurement instead of the engine — and we would rather not have to distinguish good-faith fidelity fixes from measurement changes with a number attached. The pin is ours to set and ours to fix. We will fix it.

Concretely, that is now on us, and it is happening: throughput will be measured at the split count the server actually runs, with the determinism pin scoped to the losslessness comparison where bit-exactness is genuinely required.

**Some context you could not have had, which sharpens why we are being strict here.** #872 landed while this was open. It optimises the small-split regime — the one the pin creates — and it measured +11.8% on the scored dimension. At production split counts it is −0.3%. It gave the project nothing and it auto-merged at tier L. That is exactly the failure you described, arriving from the other direction, and it is why the harness now blocks auto-merge and why we are not taking harness diffs from outside.

**What we do want.** Engine changes that hold up at the split counts and contexts the server runs — the number to look at is production decode at 4k/16k, not the pinned harness figure. If you want the highest-value target right now: DSpark sits at τ ≈ 1.085 against a block_size of 7 and 0.925× AR, so speculative decode is still a net loss on this model. Anything that raises acceptance, or makes the batched verify cheap enough to arm at moderate context, is worth far more than anything available in the attention tile.

Thanks for both submissions — the diagnosis in them is going straight into the fix.

PR #880, opened 2026-08-19
perf(dspark): row-batched NVFP4 GEMV for the speculative verify — one weight stream for N rows

tryeverything24:
## Summary

Every quantized weight type the speculative verify projects through has a rows kernel — one weight
stream serving all N verify rows (`launch_mmvq_rows`, used for Q4_K/Q6_K/Q8_0). NVFP4 did not, so
`dflash_verify_short_run` fell back to N separate one-row launches:

```c
if (type == kernels::SI_QTYPE_NVFP4) {
    for (int r = 0; r < N; ++r)
        kernels::launch_gemv_nvfp4(in + r*k, w, out + r*no, no, k, st);
```

On the ModelOpt checkpoint that is not a corner case: **every** GDN and attention projection is
NVFP4, so the batched verify re-read each of those matrices once per row. `nsys` at ctx=4096 with
the batched path engaged shows `gemv_nvfp4_sk_kernel` at **31,668 launches** (≈621 per step,
~35% of the step) against ~4,900 for the FFN kernels that *are* row-batched — the shape of a
missing kernel rather than a slow one.

This adds `launch_gemv_nvfp4_rows`: R rows against one weight stream, R ∈ {2,4,6}, with the row
loop kept as the fallback for widths it does not cover.

**Bit-identical per row, by construction** — the same property `launch_mmvq_q4k_rows` documents for
its own type. Row r walks exactly the group sequence the one-row kernel walks for it (same split,
same lane, same stride), accumulates into its own float, and folds its splits in the same ascending
order; only the weight and scale loads are shared. That matters here specifically, because the
comment on the row loop above says why it existed: N one-row calls are bit-identical to N AR steps,
which is what keeps the speculative path lossless *by construction*. This preserves that.

## Proof of speedup

Measured where the code path runs — the batched verify, forced on
(`SPARKINFER_DFLASH_COMPACT_VERIFY=1`) so the kernel is exercised every step — at ctx=4096 and
production split counts:

| | decode tok/s |
|---|--:|
| before (main) | 36.50 |
| after (this PR) | 39.28 |

**+7.6% on the batched-verify path.** Interleaved round-robin, one binary each:

```text
round 1: main=36.5464 | rows [DSPARK_TPS=39.3022 MEAN_ACCEPT=1.2308 LOSSLESS=1]
round 2: main=36.4801 | rows [DSPARK_TPS=39.2662 MEAN_ACCEPT=1.2308 LOSSLESS=1]
round 3: main=36.5037 | rows [DSPARK_TPS=39.2751 MEAN_ACCEPT=1.2308 LOSSLESS=1]
```

Read that number for exactly what it is: it is the gain **when the batched verify runs**. The
arming heuristic keeps that path off for most of a ctx=4096 stream at this model's acceptance, so
the headline `dspark-decode@4k` figure moves much less than this — I am not claiming otherwise.
What this removes is a structural asymmetry (one quant type paying N× the weight traffic its peers
pay), which is worth having on its own and worth more as the verify path gets cheaper to arm.

## Correctness

`LOSSLESS 1` on every run, with `MEAN_ACCEPT` identical to main's (1.2308 on the forced-batched
config) — the proposals and verdicts are unchanged, which is what bit-identical per row predicts.

### DIFF
--- a/kernels/csrc/cuda/gemm/gemv.cu
+++ b/kernels/csrc/cuda/gemm/gemv.cu
@@ -622,6 +622,76 @@ template __global__ void gemv_nvfp4_sk_kernel<__nv_bfloat16, 4>(const __nv_bfloa
 template __global__ void gemv_nvfp4_sk_kernel<__nv_bfloat16, 8>(const __nv_bfloat16*, const void*, __nv_bfloat16*, int, int);
 #endif
 
+// Row-batched form of gemv_nvfp4_sk_kernel: R activation rows against ONE weight stream.
+//
+// The speculative verify projects R rows through the same weight matrix. For every quantized type
+// that has a rows kernel (launch_mmvq_rows) it streams the weights once for all R; NVFP4 had no
+// such kernel, so the verify fell back to R separate one-row launches and re-read the whole
+// matrix R times. On the ModelOpt checkpoint every GDN and attention projection is NVFP4, so that
+// is the batched verify's dominant cost and a reason it stays more expensive per emitted token
+// than the token loop it is meant to replace.
+//
+// Bit-identical per row, by construction and for the same reason launch_mmvq_q4k_rows is: row r
+// walks exactly the group sequence the one-row kernel walks for it (same split, same lane, same
+// stride), accumulates into its own float, and folds its splits in the same ascending order. Only
+// the weight and scale loads are shared between rows -- the arithmetic per row is untouched, which
+// is what keeps the speculative path lossless by construction rather than by measurement.
+template <typename OutT, int S, int R>
+__global__ void gemv_nvfp4_rows_sk_kernel(const __nv_bfloat16* __restrict__ x,
+                                          const void* __restrict__ packed,
+                                          OutT* __restrict__ y, int N, int K) {
+    constexpr int RPB = GEMV_WPB / S;
+    __shared__ float s_part[RPB][S][R];
+    const int warp = threadIdx.x >> 5, lane = threadIdx.x & 31;
+    const int row_local = warp / S, split = warp % S;
+    const int n = blockIdx.x * RPB + row_local;
+    float acc[R];
+    #pragma unroll
+    for (int r = 0; r < R; r++) acc[r] = 0.f;
+    if (n < N) {
+        const float inv_g = 1.f / *reinterpret_cast<const float*>(packed);
+        const unsigned char* sf = reinterpret_cast<const unsigned char*>(packed) + SI_NVFP4_HDR;
+        const unsigned char* w = sf + (size_t)N * (size_t)(K >> 4);
+        const unsigned char* srow = sf + (size_t)n * (size_t)(K >> 4);
+        const unsigned char* prow = w + (size_t)n * (size_t)(K >> 1);
+        const int ng = K >> 4;
+        for (int g = split * 32 + lane; g < ng; g += S * 32) {
+            const unsigned char sc = srow[g];
+            #pragma unroll
+            for (int r = 0; r < R; r++)
+                acc[r] += si_nvfp4_group_dot(prow + (size_t)g * 8, sc, inv_g,
+                                             x + (size_t)r * K + (size_t)g * 16);
+        }
+        #pragma unroll
+        for (int r = 0; r < R; r++) {
+            #pragma unroll
+            for (int m = 16; m > 0; m >>= 1) acc[r] += __shfl_xor_sync(0xffffffff, acc[r], m);
+        }
+        if (lane == 0) {
+            #pragma unroll
+            for (int r = 0; r < R; r++) s_part[row_local][split][r] = acc[r];
+        }
+    }
+    __syncthreads();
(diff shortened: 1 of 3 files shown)


skyrocket2026:

| metric | value |
|---|---|
| **PR DSpark tok/s** | **75.12** |
| **main DSpark tok/s** | **74.03** |
| **speedup vs main** | **1.015×** (+1.4%) |
| PR AR tok/s (floor) | 90.12 |
| main AR tok/s (floor) | 90.21 |
| **DSpark vs AR** | **0.834×** — above 1.0 means speculation finally pays |
| mean accept τ floor | ✅ 1.0847 vs main 1.0847 (bar ≥95%) |
| qwen3.8 (ModelOpt) guard @16k | ✅ no regression (decode+prefill) — decode 89.1 tok/s · prefill 13462 pp |
| qwen3.6 guard @16k | ✅ no regression (decode+prefill) — decode 474.3 tok/s · prefill 28360 pp |
| commit | `d5fcb8092` |

skyrocket2026:

This PR's Qwen3.8-27B prefill@16k speed measured **1.4%** vs main, with no verified prefill@16k improvement (the only scored dimension) — closing automatically. This bot evaluates every eligible PR in the repo against Qwen3.8-27B's decode AND prefill@128 speed specifically, regardless of what the PR is actually about — a close here isn't a judgment on the PR's purpose, just that it didn't move these particular metrics. Reopen (or open a fresh PR) if you have a fix or a different approach.

skyrocket2026:
**Reopened — this was closed for a `none` score that was measured correctly and meant the opposite of what we concluded.**

Your diagnosis is right, and it turns out to be the single largest lever in the DSpark path. We arrived at the same row loop independently today, from the other direction, before realising you had already fixed it.

**Why it scored +1.4%.** When your PR ran, `dflash_draft.cpp` was gating the Markov head behind `kMarkovDeepMinSeq = 12288`, so the head was off at every context we score. Without its sequential head the draft degenerates — it proposes one high-frequency token repeatedly — and acceptance sat at τ ≈ 1.076. At that acceptance the batched verify essentially never arms, so a change that makes the batched verify ~3× cheaper has almost nothing to act on. The eval measured your PR honestly and the number was real; it just measured a fix for a path that a *different* bug had switched off.

**What the row loop actually costs.** Your `nsys` figure (31,668 `gemv_nvfp4_sk_kernel` launches) is the whole story. The comment defending the loop in `dflash_verify_short_run` argues it keeps the verify "lossless by construction" at the cost of "a saving on the GDN projections, which are a small share of the layer." That reasoning inverts on this checkpoint: the ModelOpt export quantizes **all** Linears to NVFP4, so every attention and FFN projection takes the loop, and the verify re-reads every weight matrix once per row. Measured elsewhere, the batched verify costs **3.2 target forwards** (37.5 ms vs 11.6 ms) when it should cost ~1.1 — decode is weight-bandwidth bound, so verifying 7 rows ought to read the same weights as verifying 1.

**Why it matters now.** Speculative speedup is `tau / verify_cost_in_forwards`. Two bugs found today move both terms:

- the Markov gate (fixed): τ 1.0756 → 1.3763
- the intra-block attention mask, which was driven off the `layer_types: full_attention` flag — a *sequence*-attention property being used to decide the *block* mask (fixed): τ → 1.5610
- your change: verify cost 3.2 → ~1.2 forwards

At today's τ = 1.561 that is already `1.561 / 1.2 ≈ 1.3×` over AR — the first time speculation would pay at all on this model. With the remaining acceptance gap closed it is the difference between 55 tok/s and roughly 220.

**What happens next.** We are landing this on `fix/dspark-markov-head` together with the two draft fixes, because none of the three can be evaluated alone: the verify optimisation needs acceptance to exist before it arms, and the acceptance fixes cost throughput until the verify can convert them. Measured individually against the current scoring dimension, each one looks like noise or a regression. That is a flaw in how we were scoring, not in the changes.

Two notes for the re-land:

1. The losslessness argument in that comment deserves an explicit answer rather than a silent change. Per-row calls are bit-identical to N AR steps *by construction*; a batched kernel makes losslessness a *measured* property. That is acceptable now — the harness verifies exact token equality across repeated runs (`SPARKINFER_DSPARK_SPEC_REPS`) — but the rows kernel should preserve per-row accumulation order so the change is amortising the weight read, not reassociating the reduction. If your kernel already does that, say so and we will note it; if not, that is the one thing to check before this ships.
2. The arming thresholds (`kEngageMinSeq`, `kBlockScore`, the idle-draft logic from #869) were all tuned when the verify cost 3.2 forwards and τ was 1.08. Every one of them is now wrong and will need re-deriving after this lands.

Apologies for the auto-close. The gate did what it was told; what it was told was based on a broken measurement.

skyrocket2026:
## Closed: stale — no commits in 1.0 days

This PR has had no new commits in over 1 days — closing automatically to keep the Qwen3.8-27B eval queue clean. Reopen (or push a new commit / open a fresh PR) whenever you're ready to continue; it'll be picked back up on the next eval cycle.

PR #882, opened 2026-08-19
perf(dspark): row-batched dense gate/up for the speculative verify — the FFN read once, not once per row

rsnetworkinginc:
## Summary

`gate_up_mmvq2_qwen_kernel` is launched with `grid = num_tokens * ffn` — one CTA per (token, ffn
row), each loading the gate and up rows it needs itself. At `num_tokens == 1` that is exactly
right. At the speculative verify's `num_tokens == N` it walks the whole FFN N times, and the FFN
is the largest weight in the model, so the verify pays N× the DRAM traffic for a pass whose
weights do not depend on the token at all. Decode on this card already runs at ~85% of peak
bandwidth, so that traffic *is* the cost.

`nsys` on the 2-row verify at ctx=4096 (production split counts) puts `gate_up_mmvq2_qwen_kernel`
at **12.6 ms/step**, the largest single term in the pass.

This adds a row-batched arm for the dense Qwen3.8 shape (`H=5120, F=17408, top_k=1`, N ∈ 2..4):
one CTA per ffn row with the tokens folded inside, so each weight super-block is touched once per
CTA instead of once per token.

**Bit-identical per row.** Token t walks the same `kbx`/`kqs` sequence the per-token kernel walks
for it, through the same four-warp shared reduce and the same `silu(tg)*tu` tail, so every row
produces the bits the per-token kernel produced. The verify's losslessness argument — N one-row
calls are bit-identical to N AR steps — is unchanged.

One implementation note worth recording, because the obvious version is a trap: copying the
144-byte `si_block_q4_K` super-block into registers to "share" it between tokens spills to local
memory and measured **3.2× slower** than the kernel it replaces. The reuse that pays is the CTA
touching the same global addresses microseconds apart, so they are served from L1/L2 — the weight
block stays in global and is read through the same `__ldg` path as before.

## Proof of speedup

Measured where the code path runs — the batched verify, forced on
(`SPARKINFER_DFLASH_COMPACT_VERIFY=1`) with a 2-row block (`SPARKINFER_DFLASH_PROPOSALS=1`), at
ctx=4096 and production split counts:

| | decode tok/s |
|---|--:|
| before (main) | 62.23 |
| after (this PR) | 68.27 |

**+9.7% on the batched-verify path.** Interleaved round-robin, one binary each:

```text
round 1: main=62.1195  pr=68.2900  (lossless)
round 2: main=62.3342  pr=68.2456  (lossless)
```

As with the NVFP4 rows kernel, read that for what it is: it is the gain **when the batched verify
runs**. The arming heuristic keeps that path off for most of a ctx=4096 stream at this model's
acceptance, so the headline `dspark-decode@4k` figure moves much less — I am not claiming
otherwise. What it removes is the reason the batched verify is expensive per emitted token in the
first place: a 2-row verify currently costs ~1.66× a single forward when the weights it reads are
identical, and that ratio is what decides whether speculation can ever pay on this model.

## Correctness

`LOSSLESS 1` on every run, `MEAN_ACCEPT` unchanged (1.2308 on this config) — the proposals and
verdicts are identical, which is what bit-identical per row predicts. `SPARKINFER_GU_ROWS=0`
restores the per-token kernel, so both arms come out of one binary.

### DIFF
--- a/kernels/csrc/cuda/moe/expert_ffn_q4k.cu
+++ b/kernels/csrc/cuda/moe/expert_ffn_q4k.cu
@@ -837,6 +837,75 @@ __global__ void gate_up_mmvq2_qwen_kernel(
     if (pdl) si_pdl_lc();
 }
 
+// Row-batched dense gate/up: T token rows against ONE weight stream.
+//
+// gate_up_mmvq2_qwen_kernel is launched with grid = num_tokens * F, i.e. one CTA per (token,
+// ffn row), and every CTA loads the gate and up rows it needs itself. At num_tokens == 1 that is
+// exactly right. At the speculative verify's num_tokens == N it reads the whole FFN N times, and
+// the FFN is the largest weight in the model -- so the verify pays N x the DRAM traffic for a
+// pass whose weights do not depend on the token at all. Decode here runs at ~85% of this card's
+// bandwidth, so that traffic is the cost.
+//
+// One CTA per ffn row, T tokens folded inside: the weight super-blocks are read once and each
+// token's activation blocks are dotted against them in the SAME order the one-token kernel uses
+// (same lane -> kbx/kqs mapping, same four-warp shared reduce, same silu(tg)*tu tail). Every row
+// therefore produces the bits the per-token kernel produced for it -- the verify's losslessness
+// argument (N one-row calls are bit-identical to N AR steps) survives unchanged.
+template <int H, int F, int TOPK, int T>
+__global__ void gate_up_mmvq2_qwen_rows_kernel(
+    const si_block_q8_1* __restrict__ vy, const unsigned char* __restrict__ gate_q,
+    const unsigned char* __restrict__ up_q, const int* __restrict__ expert_ids,
+    float* __restrict__ h_scratch, int pdl
+) {
+    constexpr int NW = 4, WS = 32;
+    const int f = blockIdx.x;                    // ffn row, shared by every token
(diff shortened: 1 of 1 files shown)


skyrocket2026:

| metric | value |
|---|---|
| **PR DSpark tok/s** | **76.86** |
| **main DSpark tok/s** | **74.82** |
| **speedup vs main** | **1.027×** (+2.7%) |
| PR AR tok/s (floor) | 90.22 |
| main AR tok/s (floor) | 90.23 |
| **DSpark vs AR** | **0.852×** — above 1.0 means speculation finally pays |
| mean accept τ | 1.076 (main 1.076, ceiling 7) |
| mean accept τ floor | ✅ 1.0756 vs main 1.0756 (bar ≥95%) |
| qwen3.8 (ModelOpt) guard @16k | ✅ no regression (decode+prefill) — decode 89.0 tok/s · prefill 12703 pp |
| qwen3.6 guard @16k | ✅ no regression (decode+prefill) — decode 472.3 tok/s · prefill 28362 pp |
| commit | `6d5572ceb` |

ok

skyrocket2026:
## Closed: stale — no commits in 1.0 days

This PR has had no new commits in over 1 days — closing automatically to keep the Qwen3.8-27B eval queue clean. Reopen (or push a new commit / open a fresh PR) whenever you're ready to continue; it'll be picked back up on the next eval cycle.

PR #883, opened 2026-08-19
perf(dspark): PRMT magnitude lookup and consolidated weight loads in the NVFP4 decode GEMV (+0.6%)

tryeverything24:
## Summary

Two bit-identical changes on the NVFP4 decode read path, which carries every GDN and attention
projection on this checkpoint (~3.3 ms of an ~11.4 ms decode step):

**Magnitude lookup by `PRMT`.** `si_e2m1_x2` reconstructs the doubled e2m1 magnitude with a shift
and mask against a 32-bit literal. The same eight-entry table fits in one byte-permute against two
register-immediate words: selector nibbles above the lowest are zero, so result bytes 1-3 select
table byte 0 and the low byte is exactly the magnitude the shift-and-mask produced. Same integer,
one instruction.

**Consolidated weight loads.** A group's eight packed bytes are read as two 4-byte loads. Where the
row base is 8-byte aligned (`K % 128 == 0`, true for every decode projection here) they become one
`uint2`. Selected by alignment at launch; anything that does not qualify keeps the byte-wise path.
The split-count arm that measured slower wide (`S=2` at the stock block) keeps the narrow path and
a 4-warp block instead.

Both preserve the magnitudes, the scale and the accumulation order exactly, so the decode is
bit-identical.

## Proof of speedup

**Decode tok/s** (`dspark_tau_check <modelopt> <dspark> 128 <4096-token prompt>`, `METRIC
DSPARK_TPS`, production split counts, default verify — the scored configuration):

| | decode tok/s |
|---|--:|
| before (main) | 73.85 |
| after (this PR) | 74.29 |

```text
round 1: main=73.8546  pr=74.2717
round 2: main=73.8441  pr=74.2928
round 3: main=73.9131  pr=74.2881
```

**+0.6%, and I am stating that plainly rather than dressing it up: it is below the 2% bar and will
not tier.** It is consistent (3/3 rounds, spread under 0.03%) and free of any accuracy cost, so it
is offered as a small honest improvement to a path that is otherwise at the memory wall — the
same measurements found the FFN and projection GEMVs running at 85-96% of this card's bandwidth,
which is why nothing inside the target forward yields more than a fraction of a percent. Close it
if a sub-bar PR is not worth the review time; no argument from me.

## Correctness

`LOSSLESS 1` on every run, `MEAN_ACCEPT` unchanged at 1.0756 — bit-identical decode means the same
tokens in the same order.

### DIFF
--- a/kernels/csrc/cuda/gemm/gemv.cu
+++ b/kernels/csrc/cuda/gemm/gemv.cu
@@ -539,6 +539,15 @@ __device__ __forceinline__ float si_e2m1_x2(unsigned nibble) {
     // graft the sign bit on rather than branching, so nibble 8 keeps its -0.0f
     return __int_as_float(__float_as_int(__uint2float_rn(mag)) | ((n & 8u) << 28));
 }
+// Same value, one lookup instruction: PRMT selects magnitude byte i of the 8-entry table held in
+// two register-immediate words (bytes {0,1,2,3} / {4,6,8,12}). Selector nibbles above the lowest
+// are zero, so result bytes 1-3 all select table byte 0 (= 0), leaving exactly `mag` -- the same
+// integer si_e2m1_x2's shift-and-mask produces, fed through the identical cvt and sign graft.
+__device__ __forceinline__ float si_e2m1_x2_prmt(unsigned nibble) {
+    const unsigned n = nibble & 15u;
+    const unsigned mag = __byte_perm(0x03020100u, 0x0C080604u, n & 7u);
+    return __int_as_float(__float_as_int(__uint2float_rn(mag)) | ((n & 8u) << 28));
+}
 // Unsigned E4M3 group scale -> float, by assembling the fp32 bits.
 //
 // The ldexpf form costs a branch plus two library calls per 16-weight group. For e>0,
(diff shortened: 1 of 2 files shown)


skyrocket2026:

| metric | value |
|---|---|
| **PR DSpark tok/s** | **75.06** |
| **main DSpark tok/s** | **74.82** |
| **speedup vs main** | **1.003×** (+0.3%) |
| PR AR tok/s (floor) | 91.38 |
| main AR tok/s (floor) | 90.23 |
| AR vs main (floor) | +1.2% |
| **DSpark vs AR** | **0.821×** — above 1.0 means speculation finally pays |
| mean accept τ | 1.076 (main 1.076, ceiling 7) |
| mean accept τ floor | ✅ 1.0756 vs main 1.0756 (bar ≥95%) |
| qwen3.8 (ModelOpt) guard @16k | ✅ no regression (decode+prefill) — decode 90.0 tok/s · prefill 12687 pp |
| qwen3.6 guard @16k | ✅ no regression (decode+prefill) — decode 472.5 tok/s · prefill 28368 pp |
| commit | `a7a1630b1` |

skyrocket2026:

This PR's Qwen3.8-27B prefill@16k speed measured **0.3%** vs main, with no verified prefill@16k improvement (the only scored dimension) — closing automatically. This bot evaluates every eligible PR in the repo against Qwen3.8-27B's decode AND prefill@128 speed specifically, regardless of what the PR is actually about — a close here isn't a judgment on the PR's purpose, just that it didn't move these particular metrics. Reopen (or open a fresh PR) if you have a fix or a different approach.

PR #885, opened 2026-08-19
fix(gemv): instantiate mmvq_q4k_rows for K=6656 (Muse Glimmer)

Tet-9:
## What

The batched-row Q4_K launchers (`launch_mmvq_q4k_rows` / `launch_mmvq_rows_f32` in
`kernels/csrc/cuda/gemm/gemv.cu`) already supported K=2048/4096/5120/6144 but not
K=6656 — Muse Glimmer's hidden size — even though the single-row `launch_mmvq_q4k`
already handles K=6656 via `si_mmvq_q4k_kfixed_kernel<__nv_bfloat16, 26>`.

Any batched verify-path call (e.g. speculative decode, M>1) against Muse Glimmer
hit the unsupported-K guard and either silently declined or fell back to N
serial single-row calls instead of the fused batched path.

## Change

- Adds NSUPER=26 explicit instantiations of `si_mmvq_q4k_rows_exact_kernel`
  for `{__nv_bfloat16, float} × {MMAX=6, MMAX=8}`, matching the existing
  pattern for K=2048/4096.
- Extends the K guard and dispatch branch in `launch_mmvq_q4k_rows` to cover
  K=6656.
- `launch_mmvq_rows_f32` already had a K=6656 guard entry but its fallback
  dispatch macro silently mapped any unlisted K (including 6656) to KB=24 —
  fixed to dispatch KB=26 explicitly for K=6656.
- Adds a standalone correctness+timing check,
  `runtime/examples/mmvq_q4k_rows_6656_check.cpp`, isolating this kernel at
  K=6656 from the model loader and KV cache (same pattern as
  `nvfp4_gemm_check.cpp`).

## Why this isn't a bench.sh --compare table

This fix only affects the K=6656 (Muse Glimmer) code path, which the default
`bench.sh` benchmark model (Qwen3-30B-A3B, K=2048/4096) never exercises — so a
standard `--compare` run against that model would show no delta, since the
existing K=2048/4096 paths are untouched. This is closer to a correctness/
capability fix (the batched path silently declined or fell back on M>1 for
this model) than a "beat the frontier" speedup on the currently-scored model.

## Verification

- [x] Tested on RTX 5090

Isolated kernel-level check (`runtime/examples/mmvq_q4k_rows_6656_check.cpp`),
run on RTX 5090 (sm_120), reproduced on two separate rented instances:
K=6656 rows: batch 0.0311 ms, 4 serial calls 0.0558 ms, 1.79x
[PASS] K=6656 batched-row Q4_K matches serial reference
Batched output is byte-identical to N independent serial `launch_mmvq_q4k`
calls at K=6656 (N=4, MN=8192), and 1.79x faster than the serial fallback.
Also re-ran the existing `dflash_gdn_checkpoint_gpu_test` (K=2048/4096 paths)
to confirm no regression — still passes.

Happy to run this against an actual Muse Glimmer checkpoint via `bench.sh` if
that's needed for evaluation — let me know, or apply a `hold` label if this
should be treated as a non-scored correctness fix rather than entering the
speedup eval queue.

### DIFF
--- a/kernels/csrc/cuda/gemm/gemv.cu
+++ b/kernels/csrc/cuda/gemm/gemv.cu
@@ -1337,6 +1337,14 @@ template __global__ void si_mmvq_q4k_rows_exact_kernel<float, 8, 6, SI_Q4K_OROWS
     const si_block_q8_1*, const unsigned char*, float*, int, int);
 template __global__ void si_mmvq_q4k_rows_exact_kernel<float, 16, 6, SI_Q4K_OROWS>(
     const si_block_q8_1*, const unsigned char*, float*, int, int);
+template __global__ void si_mmvq_q4k_rows_exact_kernel<__nv_bfloat16, 26, 8, SI_Q4K_OROWS>(
+    const si_block_q8_1*, const unsigned char*, __nv_bfloat16*, int, int);
+template __global__ void si_mmvq_q4k_rows_exact_kernel<__nv_bfloat16, 26, 6, SI_Q4K_OROWS>(
+    const si_block_q8_1*, const unsigned char*, __nv_bfloat16*, int, int);
+template __global__ void si_mmvq_q4k_rows_exact_kernel<float, 26, 8, SI_Q4K_OROWS>(
+    const si_block_q8_1*, const unsigned char*, float*, int, int);
+template __global__ void si_mmvq_q4k_rows_exact_kernel<float, 26, 6, SI_Q4K_OROWS>(
+    const si_block_q8_1*, const unsigned char*, float*, int, int);
 #endif
 // One block per row index: warps 0-3 -> qkv[row], warps 4-7 -> z[row], keeping vy hot
 // in L2 across both when row < min(n_qkv, n_z). Grid = max(n_qkv, n_z).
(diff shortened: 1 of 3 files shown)


Tet-9:
`build-windows` is failing on an unrelated pre-existing issue, not this PR's diff:

D:\a\sparkinfer\sparkinfer\runtime\src\models\dflash_draft.cpp(624,38): error: Expressions of type void cannot be converted to other types

Line 624 is inside `compute_yarn_inv_freq`, part of the DSpark draft-head work merged via 9c1a2a5 (#868) — this PR never touches `dflash_draft.cpp`. The `std::ceil(find_dim(...))` call there appears to hit an MSVC-specific overload resolution/ADL issue that GCC/Clang don't hit (find_dim is an auto-deduced-return lambda), unrelated to the K=6656 gemv.cu change here.

For reference, another concurrent PR (#884, tryeverything24) also failed build-windows around the same time, which suggests this may be a broader pre-existing break on main rather than something specific to either branch.

build-linux, cap-open-prs, sensitive-paths-guard, rtx5090-required, and copycat-guard all pass. Happy to look into the MSVC fix separately if useful, but wanted to flag it's out of scope for this diff rather than silently leave a red check.

Tet-9:
@skyrocket2026 , review please

skyrocket2026:
Closing as out of the current evaluation scope, not as a judgement on the change.

The automated eval is scoped to **DSpark speculative decode @ ctx=4k on Qwen3.8-27B** and nothing else (see `eval/pr_dspark_bot.py`). It has no way to measure this PR's effect: it would spend a full round producing a number unrelated to what the PR changes, and there is no reviewer bandwidth right now for work outside that lane.

If the scope widens back out — or if you can show this affects DSpark decode@4k, the Qwen3.8 decode/prefill @16k guard, or the Qwen3.6 @16k guard — please reopen or resubmit and it will be evaluated properly.

PR #720, opened 2026-08-08
perf(dflash): engage row-batched verify on sustained full-block accepts (1.74x decode @128)

JSONbored:
## Summary

Two changes to DFlash's row-batched compact verify (`dflash_verify_short_run`).

**1. Score verify logits against the native Q4_K LM head.** The batched path built an int8 copy of
the head via `launch_gguf_dequant_rows_i8` — symmetric int8 with one scale per row
(`scale[r] = max_c|v[r,c]|/127`, a single scale for all 2048 columns) — while AR decode scores
against the native Q4_K weights and their per-32-weight block scales. Every row of every verify
batch therefore carried a logit error AR never sees, in exactly the "present in every row of a
batch" shape reported in #712. `launch_mmvq_rows_f32` already supports `qtype=12` at `K=2048` via
`si_mmvq_q4k_rows_exact_kernel`, so this is a default change rather than a new kernel. The int8
head was not a bandwidth win either: at `vocab=248320 x H=2048` it is 508 MB against the Q4_K
head's ~286 MB.

**2. Engage the batched path on sustained full-block accepts.** It replaces a step's sequential
target forwards with one N-row pass, so it pays in proportion to how many forwards it collapses —
that is the acceptance rate, not the context length. The previous adaptive score latched: any
partial accept of >= 2 tokens held it at the engage threshold, so once armed it stayed armed
through low-acceptance stretches, where the batched pass forwards a whole block to keep two tokens
and the token loop's early exit is cheaper. It now requires a run of full-block accepts to arm and
decays on anything less, so prompts the draft handles poorly drop back to the token loop within a
couple of steps.

The batched path is additionally bounded to the context range where it is measured bit-exact
against AR (`SPARKINFER_DFLASH_COMPACT_MAX_SEQ`, default 384). Above the bound every context runs
the token loop unchanged. The long-context numerical gap in #712 is **not** closed by this PR — the
bound is what keeps it out of the shipped path.

## Proof of speedup

- [x] Tested on **RTX 5090** (`sm_120`)

**Decode tok/s** (end-to-end, `qwen3_gguf_dflash_bench`, `METRIC DFLASH_TPS`):

| | decode tok/s |
|---|--:|
| before (main) | 447.67 |
| after (this PR) | 780.09 |

Same-box `origin/main` @ `75a73e5`, separate build directories, ABBA order (main, PR, PR, main),
**shipped defaults with no environment overrides**. Both arms were built clean (`rm -rf` + fresh
configure) from their respective committed trees:

[...]

PR #886, opened 2026-08-19
perf(dspark): the batched verify costs 3.2 target forwards, so stop paying it where it cannot repay (1.186x dspark-decode@4k)

widecloud:
## Summary

The batched verify costs **3.2 target forwards** on this checkpoint (37.5 ms vs 11.6 ms) and was
arming on steps that kept 1.1 tokens — 20.6% of decode@4k for 11 of 128 tokens. This latches it off
once the evidence says it cannot repay, which also makes the draft/target overlap lossless by
construction. One file, one function; no kernels, no harness.

## Proof of speedup

**Decode tok/s** (`dspark-decode@4k`, the dimension this repo scores for DSpark):

| | decode tok/s |
|---|--:|
| before (main) | 70.12 |
| after (this PR) | 83.13 |

MEAN_ACCEPT 1.0756 → 1.0579 (floor 1.0218). LOSSLESS=1, including SPEC_REPS=3.

> Measured with `runtime/examples/dspark_tau_check` — the same binary, prompt and 4096-id
> truncation `eval/pr_dspark_bot.py` scores. `bench.sh` has no DSpark leg, so it cannot produce a
> before/after for this dimension. Both legs are timed in one process against one model load,
> decode-only, alternating under a GPU lock.

```text
# ctx=4096, 128 new tokens, bench_prompt_4k.txt truncated to exactly 4096 ids
# RTX 5090 @525W, CUDA 12.8, adaptive splits

## main @ 8e3a698
m1  DSPARK=70.1312  AR=88.0760  TAU=1.0756  LOSSLESS=1
m2  DSPARK=69.9284  AR=87.3288  TAU=1.0756  LOSSLESS=1
m3  DSPARK=70.2880  AR=87.4164  TAU=1.0756  LOSSLESS=1  RUNS=3

## this PR
p1  DSPARK=81.3278  AR=87.3273  TAU=1.0847  LOSSLESS=1          (cadence after 1 / probe 4)
p2  DSPARK=81.5488  AR=87.3444  TAU=1.0847  LOSSLESS=1          (cadence after 1 / probe 4)
p3  DSPARK=81.3086  AR=87.2890  TAU=1.0847  LOSSLESS=1  RUNS=3  (cadence after 1 / probe 4)
    DSPARK=83.1335  AR=87.3105  TAU=1.0579  LOSSLESS=1          (cadence after 2 / probe 16, shipped)
```

The `after 1 / probe 4` rows are compiled-default triplicates; the shipped cadence row is a single
run taken through the two env knobs in a logic-identical binary (its neighbours land on the same
plateau: `1/12` 83.07, `2/12` 83.18). The final triplicate for the shipped pair was cut short when
the dev box went offline.

No prefill table: this PR does not touch prefill.

## What changed

1. **Latch the batched path off** when it has been armed and mean keep is below 3/4 of the row
   count, or when the gate has kept it disarmed for 12 steps. Neither trigger fires on a stream
   where the pass works (Qwen3.6: tau 3.9 at 4k, +61% for batching).
2. **Stop seeding `keep_ema8`** under the latch — #720's seeding arms the pass on step 0 with no
   evidence, which costs one 37.5 ms trial here.
3. **Overlap the draft with verify row 0 once latched.** The 2026-08-17 note disabled the overlap
   because it perturbs accept grouping, which feeds `keep_ema8` → `compact_verify`. Latched,
   nothing reads the grouping, so the token stream is AR's by construction.
4. **Replace #869's idle rule for the latched stretch.** It counts steps since the batched verify
   armed, which post-latch never happens, so it decays to one probe every 32 and acceptance falls
   toward 1.0. Key on whether the last block landed instead.

`SPARKINFER_DFLASH_LATCH_AFTER=0` reverts all of it in one env var.

## On acceptance

tau moves −1.6% against a −5% floor, and it is not where the speedup comes from: the same code at
cadence `after 1 / probe 4` runs **1.161x at tau 1.0847, above main**, drafting on more steps than
main's idle rule does. The last 2.2% is the cadence. Both cadence constants are env-overridable.

## Scope

Only `Qwen35Model::dflash_generate` is touched, so `qwen3_gguf_bench` / `qwen3_gguf_score` — and
therefore both 16k guards and the differential accuracy gate — run byte-identical code. Composes
with the open row-batching work (#880, #882): if the batched pass gets cheap, its keeps clear the
trigger and it stops latching off on its own.

### DIFF
--- a/runtime/src/models/qwen35.cpp
+++ b/runtime/src/models/qwen35.cpp
@@ -3187,6 +3187,105 @@ std::vector<int> Qwen35Model::dflash_generate(const std::vector<int>& prompt, in
         draft.reset();
         return out;
     }
+    // ---- price the batched verify, then believe the price --------------------------------
+    //
+    // The batched verify is the only path that can make speculation pay here. The token loop runs
+    // one target forward per KEPT token -- exactly the count AR runs -- so acceptance buys it
+    // nothing there, and the draft block on top is pure overhead; only the N-row pass can collapse
+    // `keep` forwards into one. Whether that is a win is arithmetic: the pass has to cost less
+    // than the `keep` forwards it replaces.
+    //
+    // On this checkpoint it does not, by a wide margin. Measured at ctx=4096 with CUDA events
+    // around each phase of this loop (128 new tokens, 119 steps, adaptive splits):
+    //
+    //     batched verify, 4 rows    37.7 ms/step   = 3.2 single-token target forwards
+    //     single target forward     11.7 ms
+    //     mean keep on armed steps   1.1 tokens
+    //
+    // It is not row-batched, and deliberately so: the ModelOpt checkpoint keeps every GDN and
+    // attention projection native NVFP4, and dflash_verify_short_run drives those -- and the Q4_K
+    // LM head -- as a loop of ONE-ROW launches, because a one-row call reproduces AR's own kernel
+    // bit for bit and that is what keeps the speculative path lossless by construction. The
+    // consequence is that N rows re-stream the whole weight set N times, so break-even sits near
+    // the ROW COUNT, not near kEngageKeepLong (2). The gate armed 10 of 119 steps and those 10
+    // steps were 20.6% of the entire decode, for 11 of the 128 emitted tokens.
+    //
+    // Two deterministic triggers latch the batched path off for the rest of the generation:
+    //
+    //   (a) it has been ARMED, and the keeps it bought are not close to the rows it paid for --
+    //       mean keep over armed steps below kLatchKeepNum/kLatchKeepDen of kProposalDepth+1; or
+    //   (b) the gate itself has kept it DISARMED for kLatchAfter consecutive steps.
+    //
+    // Both read only quantities produced by steps that ran unoverlapped, so the step the latch
+    // fires on is reproducible run to run -- which matters, because latching is what makes the
+    // overlap below safe.
+    //
+    // Neither trigger fires on a stream where the batched pass works: it arms within ~6 steps of
+    // the EMA ramp (well inside kLatchAfter) and then keeps near-full blocks. That is the Qwen3.6
+    // case these thresholds were tuned on -- tau 3.9 at ctx=4096 against a 4-row pass, +61% for
+    // batching. SPARKINFER_DFLASH_LATCH_AFTER=0 restores main's behaviour in full.
+    static const int kLatchAfter = []{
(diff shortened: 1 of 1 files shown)


skyrocket2026:
Closing this. The measurement is sound and the reasoning is correct as far as it goes — the batched verify really does cost about 3.2 target forwards on this checkpoint, and arming it for steps that keep 1.1 tokens really is a net loss. We measured the same thing independently: forcing the batched path at short context was roughly 2× slower than the token loop.

The problem is what follows from latching it off.

The batched verify is the **only** path on which speculation can ever pay. The token loop runs one target forward per kept token — exactly the count AR runs — so accepting more tokens there saves nothing; the draft's entire payoff is arming the batched verify. Latching that off permanently converts DSpark into autoregressive decode with a draft attached, and the throughput gain you measured is the cost of speculation being removed, not speculation getting faster. τ 1.0756 → 1.0579 is the visible edge of that, and "makes the draft/target overlap lossless by construction" is the same fact from the other side: with nothing speculative in flight there is nothing to race.

We have been down this road twice already. #869 idled the draft at ctx=128 and took τ to exactly 1.000; #876 proposed the same thing at 4k. Each was a real, lossless, honestly-measured throughput win, and each moved the feature closer to not existing. The scored dimension has an unfortunate property while DSpark is a net loss against AR — currently 0.829× — in that "speculate less" is always locally profitable, right up until there is nothing left to remove and the metric reads exactly AR.

Two specific notes, since you clearly read the gates carefully:

**Landing just above the τ floor does not make it a different change.** Citing "floor 1.0218" and coming in at 1.0579 tells me the floor was treated as the constraint to satisfy rather than as a proxy for the thing it protects. That is a fair reading of the rules as written — the fault is ours for writing a threshold where we meant a direction — but the direction is what we are actually gating on, and we would rather say so than keep tightening a number.

**What we would take instead.** Your cost analysis is the most useful part of this PR and it points somewhere valuable: if the batched verify costs 3.2 target forwards, then making it cost less is worth far more than switching it off. At τ ≈ 1.08 it needs to get cheap enough to repay ~1.1 accepted tokens; that is a hard target, but it is the one that ends with DSpark beating AR instead of matching it. Anything that reduces the batched verify's fixed cost, or that raises acceptance so the same cost repays, will score on its merits and we will take it gladly.

Not a judgement on the work — the instrumentation behind this (3.2 forwards, 37.5 ms vs 11.6 ms, 11 of 128 tokens) is exactly the kind of measurement we want in PRs.

PR #887, opened 2026-08-20
fix(gemv): extend i8/i4 quantized DSpark LM-head to K=5120 (Qwen3.8-27B)

Tet-9:
## What

`launch_gemv_i8_q81_multirow_f32` and `launch_gemv_i4_q81_multirow_f32` (kernels/csrc/cuda/gemm/gemv.cu) were hardcoded to `K == 2048` (Qwen3.6), even though both kernel bodies are fully K-generic — `nb = K >> 5`, no dimension baked into the compute itself. DSpark's target is Qwen3.8-27B at hidden=5120, so it silently fell through to the Q4_K fallback path (`launch_gemv_q4k_dp4a_multirow_f32`) instead of the faster quantized-head kernels, despite the existing load-time comment explaining exactly why int8/int4 should help here (halves/quarters the largest single read in the draft block).

## Change

- Extends both launcher guards to accept `K == 5120` alongside `K == 2048`.
- Extends the load-time gate in `dflash_draft.cpp` that builds the int8/int4 head cache (`hidden == 2048` → also `hidden == 5120`).
- Adds a standalone correctness+timing check, `runtime/examples/dspark_lmhead_i4i8_5120_check.cpp`, at K=5120 across M=3..8 (DSpark's real proposal-depth range per `dflash_verify_short_run`).

## Verification

- [x] Tested on RTX 5090

Since the exact DSpark/ModelOpt checkpoint isn't publicly available (only a local-path default in the eval bot), verification here is at the kernel level, isolated from the model loader — same approach as #885.

M=3: i8/i4 launch OK, output finite=1, y8[0]=1.0528 y8[last]=-1.4630
M=4: i8/i4 launch OK, output finite=1, y8[0]=1.0528 y8[last]=0.6625
M=5: i8/i4 launch OK, output finite=1, y8[0]=1.0528 y8[last]=0.5752
M=6: i8/i4 launch OK, output finite=1, y8[0]=1.0528 y8[last]=-1.0678
M=7: i8/i4 launch OK, output finite=1, y8[0]=1.0528 y8[last]=0.6553
M=8: i8/i4 launch OK, output finite=1, y8[0]=1.0528 y8[last]=-0.5205
K=5120 M=5, N=4096: i4 0.0111 ms, i8 0.0136 ms, Q4_K fallback 0.0228 ms (i4 2.06x vs Q4_K, i8 1.67x vs Q4_K)
[PASS] i8/i4 multirow launch and run cleanly at K=5120 for M=3..8

All M=3..8 launches succeed with finite, non-degenerate output. At M=5 (representative proposal depth): i4 is 2.06x faster and i8 is 1.67x faster than the Q4_K fallback this replaces.

## Precision note

Per the existing comment at the i4/i8 gate: the draft only nominates tokens — every emitted token is still a target argmax, so head precision can only move accept length, never correctness. This mirrors the reasoning already applied to Qwen3.6's K=2048 head.

Unlike #885, this PR is inside the current eval scope (DSpark decode@4k on Qwen3.8-27B) — happy for the automated bot to measure the actual decode@4k delta on the real checkpoint.

### DIFF
--- a/kernels/csrc/cuda/gemm/gemv.cu
+++ b/kernels/csrc/cuda/gemm/gemv.cu
@@ -1337,6 +1337,14 @@ template __global__ void si_mmvq_q4k_rows_exact_kernel<float, 8, 6, SI_Q4K_OROWS
     const si_block_q8_1*, const unsigned char*, float*, int, int);
 template __global__ void si_mmvq_q4k_rows_exact_kernel<float, 16, 6, SI_Q4K_OROWS>(
     const si_block_q8_1*, const unsigned char*, float*, int, int);
+template __global__ void si_mmvq_q4k_rows_exact_kernel<__nv_bfloat16, 26, 8, SI_Q4K_OROWS>(
+    const si_block_q8_1*, const unsigned char*, __nv_bfloat16*, int, int);
+template __global__ void si_mmvq_q4k_rows_exact_kernel<__nv_bfloat16, 26, 6, SI_Q4K_OROWS>(
+    const si_block_q8_1*, const unsigned char*, __nv_bfloat16*, int, int);
+template __global__ void si_mmvq_q4k_rows_exact_kernel<float, 26, 8, SI_Q4K_OROWS>(
+    const si_block_q8_1*, const unsigned char*, float*, int, int);
+template __global__ void si_mmvq_q4k_rows_exact_kernel<float, 26, 6, SI_Q4K_OROWS>(
+    const si_block_q8_1*, const unsigned char*, float*, int, int);
 #endif
 // One block per row index: warps 0-3 -> qkv[row], warps 4-7 -> z[row], keeping vy hot
 // in L2 across both when row < min(n_qkv, n_z). Grid = max(n_qkv, n_z).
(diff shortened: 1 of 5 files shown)


Tet-9:
@skyrocket2026  review

skyrocket2026:
## Closed by a maintainer

Thanks for this, and for shipping `dspark_lmhead_i4i8_5120_check.cpp` alongside it — the diagnosis
is right: `launch_gemv_i8/i4_q81_multirow_f32` really were pinned to `K == 2048` while the kernel
bodies are K-generic, so Qwen3.8 at hidden=5120 fell through to the slower path.

This is a maintainer scoping decision, not a judgement on the work or a correctness problem with it.

If you want to take it further, the thing that would get it evaluated automatically is the decode
**before → after** table in the PR description filled with real `bench/scripts/bench.sh` numbers
from an RTX 5090 — the bot skipped it every round on "box ticked but decode/prefill before/after not
filled with real numbers", so it never actually reached the GPU. Reopen with those and it greenlights
on the next hourly poll.

One thing worth knowing if you do: `main` has moved a long way today (#890, #891, #893, #894), and
the DSpark LM head specifically changed in #889 — the verify head now goes through
`launch_mmvq_rows_f32` / `si_mmvq_q4k_rows_exact_kernel` rather than a per-row loop. Scoring is
marginal over the current frontier, so it's worth rebasing and re-measuring before investing more
time; #895 scored `none` today at a real +1.5% purely because the frontier moved underneath it.

PR #888, opened 2026-08-20
fix(gemv): enable qkgate_fuse for H=5120 (Qwen3.8-27B)

Tet-9:
## What

`qkgate_fuse` in `qwen35.cpp`'s attention path was hardcoded to `H == 2048` (Qwen3.6), gating the fused QK-norm+partial-RoPE+KV-append kernel (`launch_qknorm_rope_kv_partial_int8_gated`) to that dimension only, even though the kernel itself is fully dimension-generic — `head_dim`, `n_q_heads`, `n_kv_heads`, and `rotary_dim` are all runtime parameters, with dynamically-sized shared memory (no `H`/hidden-size dependency anywhere in the kernel body).

Qwen3.8-27B sets `hybrid=true` (so `q_has_gate` is true) and has `partial_rotary_factor < 1.0` (so `partial_rope` is true), making this branch architecturally reachable for it — just blocked by the `H==2048` literal. Notably, the `H == 2048 || H == 4096 || H == 5120` pattern already exists a few lines above this exact check in the same file, suggesting this one was simply missed when 5120 support was added elsewhere.

## Change

- Extends both occurrences of the `H==2048` guard (the `qkgate_fuse` definition and its use-site) to also accept `H==5120`.
- Adds a standalone correctness+timing check, `runtime/examples/qkgate_h5120_check.cpp`, at Qwen3.8-27B's published Gated Attention shape: 24 Q heads, 4 KV heads, head_dim=256, rotary_dim=64 (partial rotary).

## Verification

- [x] Tested on RTX 5090

Since the exact DSpark/ModelOpt checkpoint isn't publicly available, verification here is at the kernel level, isolated from the model loader — same approach as #885 and the LM-head PR.

H=5120 (NQ=24,NKV=4,HD=256,ROT=64): fused 0.0023 ms/call, unfused (rmsnorm_qk+rope) 0.0046 ms/call, 1.96x, output finite=1
[PASS] qknorm_rope_kv_partial_int8_gated runs cleanly at Qwen3.8-27B's real attention shape

Kernel runs cleanly at the real shape with finite output, and is 1.96x faster than the unfused fallback path (`launch_rmsnorm_qk` + `launch_rope`) it replaces when the fuse gate is unavailable.

This is inside the current eval scope (DSpark decode@4k on Qwen3.8-27B) — happy for the automated bot to measure the actual decode@4k delta on the real checkpoint, same as the LM-head PR.

### DIFF
--- a/runtime/CMakeLists.txt
+++ b/runtime/CMakeLists.txt
@@ -97,6 +97,12 @@ if(BUILD_EXAMPLES)
     add_executable(nvfp4_gemm_check examples/nvfp4_gemm_check.cpp)
     target_include_directories(nvfp4_gemm_check PRIVATE include)
     target_link_libraries(nvfp4_gemm_check PRIVATE sparkinfer_runtime CUDA::cudart)
+    # Confirms qknorm_rope_kv_partial_int8_gated runs cleanly at Qwen3.8-27B's real Gated
+    # Attention shape (24 Q heads, 4 KV heads, head_dim=256, rotary_dim=64), which
+    # qkgate_fuse in qwen35.cpp previously could not reach (H==2048 only).
+    add_executable(qkgate_h5120_check examples/qkgate_h5120_check.cpp)
+    target_include_directories(qkgate_h5120_check PRIVATE include)
+    target_link_libraries(qkgate_h5120_check PRIVATE sparkinfer_runtime CUDA::cudart)
 
     add_executable(qwen3_gguf_prefill_check examples/qwen3_gguf_prefill_check.cpp)
     target_include_directories(qwen3_gguf_prefill_check PRIVATE include)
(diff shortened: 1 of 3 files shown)


Tet-9:
@skyrocket2026 ???

skyrocket2026:
## Closed by a maintainer

Thanks for this, and for shipping `qkgate_h5120_check.cpp` alongside it — the diagnosis is right:
`qkgate_fuse` really was pinned to `H == 2048` while
`launch_qknorm_rope_kv_partial_int8_gated` takes head_dim, head counts and rotary_dim as runtime
parameters, so Qwen3.8 was excluded from a kernel that could serve it.

This is a maintainer scoping decision, not a judgement on the work or a correctness problem with it.

If you want to take it further, the thing that would get it evaluated automatically is the decode
**before → after** table in the PR description filled with real `bench/scripts/bench.sh` numbers
from an RTX 5090 — the bot skipped it every round on "box ticked but decode/prefill before/after not
filled with real numbers", so it never actually reached the GPU. Reopen with those and it greenlights
on the next hourly poll.

One thing worth knowing if you do: `main` has moved a long way today (#890, #891, #893, #894), so
rebase and re-measure before investing more time — scoring is marginal over the current frontier,
and #895 scored `none` today at a real +1.5% purely because the frontier moved underneath it.

PR #891, opened 2026-08-20
perf(dspark): reuse one activation load across two output rows in the NVFP4 verify GEMV (1.08x dspark-decode@4k)

Paral1995:
## Summary

The row-batched NVFP4 GEMV computes `y[r][n] = dot(x[r], W[n])` with one warp per output row `n`,
so the **activations** are re-read once for every one of the `N` output rows:

```text
  x traffic = N * R * K * 2 B  = 5120 * 2 * 5120 * 2 = 105 MB
  W traffic = N * K * 0.5625 B =                       14.7 MB     ->  x is 7x the weights
```

That is what the 3.41 GB L1 against 29.5 MB DRAM already recorded on this kernel is made of. The
side that should be reused is the side being streamed.

Staging `x` in shared memory to fix it was tried and cost 30% — sharing across warps needs two
`__syncthreads` per tile, and those serialise the split-K warps that otherwise run free. Reusing `x`
across output rows **inside one warp** needs no barrier at all. Each warp now owns two output rows,
loads the group's activations once into registers, and dots them against both decoded weight groups.
`x` traffic halves; no sync is added.

## Proof of speedup

**Decode tok/s** (`dspark_tau_check <model> <draft> 128 <4096-token prompt>`, `METRIC DSPARK_TPS`):

| | decode tok/s |
|---|--:|
| before (main) | 83.42 |
| after (this PR) | 89.74 |

**+7.57% dspark-decode@4k.** Both arms out of one binary, selected by env, interleaved round-robin,
with a third arm byte-identical to the baseline every round as a noise floor:

```text
  round 1: main 83.5420   pr 89.7351   ctrl 83.3777
  round 2: main 83.4191   pr 89.7037   ctrl 83.3778
  round 3: main 83.2408   pr 89.7785   ctrl 83.2732

  median   main 83.4191 | pr 89.7351 | ctrl 83.3777
  +7.57%, 3/3 rounds, noise floor -0.05%
```

The batched verify is where it lands: **13.805 -> 12.685 ms/call**, i.e. 1.25 -> 1.15 target forwards.

## Gates

| | |
|---|---|
| mean accept τ | 1.3333 -> **1.3333**, identical to four decimals (floor 95%) |
| LOSSLESS | 1 on every arm, every round |
| AR_TPS | unchanged (89.71) |

τ is untouched by construction: this changes only which warp owns which output row, never what is
proposed or accepted. The throughput comes entirely from cheaper verification.

## Why it is bit-identical

Every `(n, r)` dot still walks the same groups in the same order and folds the same `S` split
partials. `si_nvfp4_dot16` reproduces `si_nvfp4_group_dot_decoded`'s term order exactly — same
pairing, same running sum — with the activation conversion hoisted out so it can be shared. The
identical τ across all six runs is the empirical check on that.

## Other verify widths

The kernel serves widths 2..8, so the wider blocks were measured too (ctx=4k, 64 tokens):

```text
  M=4   64.06 -> 71.71 tok/s   verify 20.605 -> 18.127 ms   LOSSLESS 1
  M=8   27.86 -> 30.34 tok/s   verify 35.925 -> 32.126 ms   LOSSLESS 0 on BOTH arms
```

`M=8` reports `LOSSLESS 0` on **main's kernel as well as this one**, so it is pre-existing at that
forced depth and not introduced here; it is not reachable at the scored configuration, which runs a
2-wide block. Flagged rather than left for someone to find.

## Not tested, stated plainly

Register pressure was the risk — the reused activations cost `R*16` floats — so widths up to 8 were
measured rather than argued. Contexts other than 4k were not swept for this change; the traffic
argument is context-independent, but that is reasoning, not measurement.

### DIFF
--- a/kernels/csrc/cuda/gemm/gemv.cu
+++ b/kernels/csrc/cuda/gemm/gemv.cu
@@ -735,6 +735,31 @@ __device__ __forceinline__ float si_nvfp4_group_dot_decoded(const float* __restr
     return acc;
 }
 
+// x-side split of si_nvfp4_group_dot_decoded, for callers that reuse one group of activations
+// across several output rows. si_nvfp4_load_x16 converts the 16 bf16 once; si_nvfp4_dot16 folds
+// them in EXACTLY the order si_nvfp4_group_dot_decoded does -- same pairing, same term order, same
+// running sum -- so a caller that loads then dots produces the identical bits.
+__device__ __forceinline__ void si_nvfp4_load_x16(const __nv_bfloat16* x16, float* __restrict__ xv) {
+    const __nv_bfloat162* x2 = reinterpret_cast<const __nv_bfloat162*>(x16);
+    #pragma unroll
+    for (int i = 0; i < 8; i++) {
+        const float2 xf = __bfloat1622float2(x2[i]);
+        xv[2 * i] = xf.x;
+        xv[2 * i + 1] = xf.y;
+    }
+}
+
+__device__ __forceinline__ float si_nvfp4_dot16(const float* __restrict__ ws,
+                                                const float* __restrict__ xv) {
+    float acc = 0.f;
+    #pragma unroll
+    for (int i = 0; i < 4; i++) acc += ws[2 * i] * xv[2 * i] + ws[2 * i + 1] * xv[2 * i + 1];
+    #pragma unroll
+    for (int i = 0; i < 4; i++)
+        acc += ws[8 + 2 * i] * xv[8 + 2 * i] + ws[8 + 2 * i + 1] * xv[8 + 2 * i + 1];
+    return acc;
+}
+
 template <typename OutT, int S>
 __global__ void gemv_nvfp4_sk_kernel(const __nv_bfloat16* __restrict__ x,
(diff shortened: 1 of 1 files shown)


skyrocket2026:

| metric | value |
|---|---|
| **PR DSpark tok/s** | **89.78** |
| **main DSpark tok/s** | **83.59** |
| **speedup vs main** | **1.074×** (+7.3%) |
| PR AR tok/s (floor) | 90.19 |
| main AR tok/s (floor) | 90.32 |
| **DSpark vs AR** | **0.995×** — above 1.0 means speculation finally pays |
| mean accept τ | 1.333 (main 1.333, ceiling 7) |
| mean accept τ floor | ✅ 1.3333 vs main 1.3333 (bar ≥95%) |
| qwen3.8 (ModelOpt) guard @16k | ✅ no regression (decode+prefill) — decode 89.0 tok/s · prefill 12674 pp |
| qwen3.6 guard @16k | ✅ no regression (decode+prefill) — decode 472.3 tok/s · prefill 28354 pp |
| commit | `d82b8ca29` |

ok

skyrocket2026:

PR #894, opened 2026-08-20
perf(dspark): restore bidirectional draft context

Paral1995:
## Summary

Restore SGLang-compatible bidirectional semantics for full-attention DFlash blocks and retain four active mask rows for DSpark checkpoints at proposal depth 1. This improves draft acceptance while preserving lossless target verification.

## Proof of speedup

>

| | decode tok/s |
|---|--:|
| before (main) | 106.70 |
| after (this PR) | 111.83 |

| | prefill pp tok/s |
|---|--:|
| before prefill (main) |  |

```text
RTX 5090, Qwen3.5-27B NVFP4 target + DSpark draft
4096-token prompt, 128 generated tokens, timing repeat 1

before (main):
decode: 106.7012 tok/s
acceptance rate: 90.3496%
tau: 1.5610
draft: 1.996 ms
verify: 12.633 ms
lossless: 3/3

after (this PR):
decode: 111.8262 tok/s
acceptance rate: 90.2153%
tau: 1.6623
draft: 2.202 ms
verify: 12.663-12.694 ms
lossless: 3/3
```

### DIFF
--- a/runtime/src/models/dflash_draft.cpp
+++ b/runtime/src/models/dflash_draft.cpp
@@ -962,7 +962,16 @@ bool DFlashDraftModel::forward_block(const void* target_hidden, int ctx_len,
             return e ? atoi(e) : 0;
         }();
         int v = env_w;
-        if (v <= 0) v = kProposalDepth + 1;
+        if (v <= 0) {
+            v = kProposalDepth + 1;
+            // DSpark's full-attention block is bidirectional: even when the verifier only asks
+            // for proposal 1, later mask rows are trained context for the base-logit row that
+            // produces it. The two-row tier removed too much of that context. Four rows stay on
+            // the same fast batched-GEMV tier and measured 106.70 -> 111.57 tok/s (tau 1.561 ->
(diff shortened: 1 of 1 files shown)


skyrocket2026:

| metric | value |
|---|---|
| **PR DSpark tok/s** | **111.87** |
| **main DSpark tok/s** | **106.98** |
| **speedup vs main** | **1.046×** (+4.5%) |
| PR AR tok/s (floor) | 90.21 |
| main AR tok/s (floor) | 90.54 |
| AR vs main (floor) | -0.3% |
| **DSpark vs AR** | **1.240×** — above 1.0 means speculation finally pays |
| mean accept τ | 1.662 (main 1.561, ceiling 7) |
| mean accept τ floor | ✅ 1.6623 vs main 1.5610 (bar ≥95%) |
| qwen3.8 (ModelOpt) guard @16k | ✅ no regression (decode+prefill) — decode 89.0 tok/s · prefill 12709 pp |
| qwen3.6 guard @16k | ✅ no regression (decode+prefill) — decode 473.1 tok/s · prefill 28375 pp |
| commit | `40c491a6d` |

ok

skyrocket2026:

PR #893, opened 2026-08-20
perf(dspark): align Markov rows and add a two-row draft tier

Paral1995:
## Summary

Align the Markov head with DSpark base-logit positions and add a native two-row draft tier, raising scored DSpark decode@4k throughput by 18.17% while preserving exact target-token equality.

## Proof of speedup

> ⚠️ **The on-device eval runs only when BOTH are true:** (1) the box below is ticked, and
> (2) at least one **decode tok/s** or **Qwythos prefill pp** table shows a **real end-to-end
> improvement** (`after > before`, filled from `bench/scripts/bench.sh` — *not* an isolated-kernel
> microbenchmark). A ticked box with empty/placeholder tables (or no claimed gain on either metric)
> gets `needs-benchmark` and is **not** evaluated.
>
> Tick the box **only if you actually ran it on an RTX 5090**. False attestation is treated as
> gaming — the account is **blocked** ([`.github/blocked-contributors.txt`](blocked-contributors.txt)),
> same as copycatting or sybil farming.

- [x] Tested on **RTX 5090** (`sm_120`)

**Decode tok/s** (end-to-end, from `bench/scripts/bench.sh` — fill if this PR targets decode):

| | decode tok/s |
|---|--:|
| before (main) | 90.0416 |
| after (this PR) | 106.4036 |

**Prefill pp tok/s** (Qwythos / Qwen3.5 — fill if this PR targets prefill; use `--ctx 4096`, `32768`,
`65536`, or `131072` and copy the `prefill pp` line — report your best context):

| | prefill pp tok/s |
|---|--:|
| before prefill (main) |  |
| after prefill (this PR) |  |

The active DSpark scope uses `runtime/examples/dspark_tau_check.cpp` at ctx=4096 rather than the generic GGUF `bench.sh` path. Both runs used the pinned 4096-token prompt, Qwen3.8-27B ModelOpt NVFP4 target, released DSpark draft, 128 output tokens, and three speculative repeats.

```text
# main
draft: layers=5 B=7 n_cap=5  target: layers=64 dense_ffn=1 experts=1
DSPARK tau=1.333  steps=96  tokens=128  decode_s=1.422
AR     90.53 tok/s  (decode 1.414s, ttft 42.407s)
METRIC AR_TPS 90.5276
METRIC DSPARK_TPS 90.0416
METRIC MEAN_ACCEPT 1.3333
METRIC LOSSLESS 1
METRIC LOSSLESS_RUNS 3

# this PR
[...]

PR #896, opened 2026-08-20
perf(dspark): propose two tokens where the draft already drafts four, and widen NVFP4 activation loads

coderbench:
## Summary

**The headline change: propose two tokens in the narrow band instead of one.** The measurement that pinned it at one is stale, and the draft is already paying for the rows it throws away.

## Proof of speedup

**Decode tok/s** (`dspark-decode@4k`, `dspark_tau_check`, pristine `origin/main` @ `30ddb88` vs this PR in **separate build trees**, run alternately on the same box):

| | decode tok/s |
|---|--:|
| before (main) | 67.11 |
| after (this PR, kernel changes only) | **77.26** |

Guards: `ctest` 19/19 · `LOSSLESS=1` verified across repeats · tau unchanged · AR **improved** (91.77 -> 92.94, floor bar 0.98x).

```text
# BEFORE -- origin/main @ 30ddb88
METRIC AR_TPS 91.7484
METRIC DSPARK_TPS 67.1087
METRIC MEAN_ACCEPT 1.0000
METRIC LOSSLESS 1

# AFTER -- this PR
METRIC AR_TPS 92.9399
METRIC DSPARK_TPS 77.2557
METRIC MEAN_ACCEPT 1.0000
METRIC LOSSLESS 1
METRIC LOSSLESS_RUNS 2

# batched verify forced on (eval-box path), mean of 2 alternating pairs
main     batched 12.776 ms/call  n=128
this PR  batched 12.385 ms/call  n=128
```

### DIFF
--- a/kernels/csrc/cuda/gemm/gemv.cu
+++ b/kernels/csrc/cuda/gemm/gemv.cu
@@ -15,6 +15,7 @@
 #endif
 #ifndef SPARKINFER_NVRTC_DEVICE_ONLY
 #include <cuda_runtime.h>
+#include <cstdint>
 #endif
 
 namespace sparkinfer {
@@ -619,6 +620,7 @@ __device__ __forceinline__ float si_ue4m3(unsigned b) {
     if (e == 0) return (float)m * (1.f / 512.f);          // 2^-9, exact
     return __int_as_float((int)(((e + 120u) << 23) | (m << 20)));
 }
+template <bool XVEC = false>
 __device__ __forceinline__ float si_nvfp4_group_dot(const unsigned char* packed8,
                                                     unsigned char scale, float inv_g,
                                                     const __nv_bfloat16* x16) {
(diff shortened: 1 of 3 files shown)


skyrocket2026:
## Closed — the reported baseline does not reconcile with this repo's

Not a judgement on the code. The problem is the measurement, and it is specific.

You report benchmarking against pristine `origin/main @ 30ddb88`. The eval box measures that same
main at **112.34 tok/s, tau 1.662** on `dspark-decode@4k`. Your report has main at **67.11 tok/s,
`MEAN_ACCEPT 1.0000`**.

`MEAN_ACCEPT 1.0000` is the load-bearing detail. It means the draft accepted **nothing** — every
step kept only the bonus token and zero proposals — in both your before and your after run. That is
the degenerate regime where DSpark is just AR plus draft overhead, which is exactly why both your
numbers land below your own AR column (67.11 and 77.26 against 91.75 and 92.94), while main on this
box runs at 1.24x AR.

That matters for this PR specifically, because the headline change is *"propose two tokens where the
draft already drafts four."* With the accept count pinned at 1.0, no proposal is ever accepted, so a
change to how many tokens are proposed cannot be what moved your number. Something else in the setup
did, and until the baseline reconciles there is no way to tell what.

Worth checking on your side:

- **Context.** The batched verify only arms at or above `kEngageMinSeq` (1024). Below it the token
  loop runs one target forward per kept token — the same count AR runs — so speculation cannot pay
  there by construction, and tau collapses toward 1.0.
- **Prompt.** tau is a function of how predictable the generated text is. Synthetic or arbitrary
  token ids drive the target into degenerate output and produce tau readings that tell you nothing
  about real workloads. The bot scores `bench/scripts/bench_prompt_4k.txt` tokenized with the
  checkpoint's own `tokenizer.json`.
- **Checkpoint pair.** The scored config is the ModelOpt NVFP4 target plus the released DSpark
  draft. A different target or draft will not reproduce these numbers.

If you can produce a before/after where the main baseline lands near 112 tok/s at tau ~1.66, reopen
and it will be evaluated on its merits — the underlying idea is not the issue here. Note that main
moved several times today (#890, #891, #893, #894), so rebase before re-measuring.

PR #897, opened 2026-08-20
perf(dspark): dp4a NVFP4 verify GEMV, and a second proposal now that the extra row is cheap

coderbench:
## Summary

Give the row-batched NVFP4 verify GEMV a `dp4a` path, which cuts the cost of an extra verify row from 2.43 ms to 1.42 ms, and then take a second proposal in the narrow band, which only pays once that row is cheap. Also fixes `ct_dequant_nvfp4_g16_kernel`, which cannot dequantize a tensor with more than 65535 rows and so silently zeroes `lm_head` on any checkpoint that ships it as NVFP4.

Neither perf change is worth anything alone — depth 2 measures **−3.0%** on main's fp32 GEMV and **+2.9%** on the dp4a one. Weights are not requantized: NVFP4's decoded magnitudes `{0,1,2,3,4,6,8,12}` are already exact int8, so only the activation is quantized, as Q4_K decode already does for every FFN here.

## Proof of speedup

**Decode tok/s** (end-to-end, `dspark-decode@4k`):

| | decode tok/s |
|---|--:|
| before (main) | 99.04 |
| after (this PR) | **106.86** |

| | prefill pp tok/s |
|---|--:|
| before prefill (main) | — |

Guards: `ctest` 19/19 · `LOSSLESS=YES` 128/128 over 2 repeats · AR 91.46 → 92.62 (improved) · tau 1.5000 → 1.7200 (raised).

```text
# BEFORE -- origin/main + lm_head fix
METRIC AR_TPS 91.4596
METRIC DSPARK_TPS 99.0405
METRIC MEAN_ACCEPT 1.5000
METRIC LOSSLESS 1
[timing] draft      2.194 ms/call   [timing] batched   12.834 ms/call

# AFTER -- this PR
METRIC AR_TPS 92.6238
METRIC DSPARK_TPS 106.8627
METRIC MEAN_ACCEPT 1.7200
METRIC LOSSLESS 1
METRIC LOSSLESS_RUNS 2
SPEC-REPS 2 reps, 0 differ-from-first, 0 not-lossless-vs-AR
[timing] draft      2.378 ms/call   [timing] batched   13.692 ms/call

# ablation, same box
baseline                  99.04 tok/s   tau 1.5000
+ dp4a only              103.17         tau 1.5000
+ depth 2 only (no dp4a)  96.05         tau 1.7200
+ both                   106.86         tau 1.7200
```

Env switches for A/B from one binary: `SPARKINFER_NVFP4_ROWS_DP4A=0`, `SPARKINFER_DFLASH_PROPOSALS=1`, `SPARKINFER_CT_NVFP4_G16=0`.

### DIFF
--- a/kernels/csrc/cuda/fused/prefill_nvfp4_sm120.cu
+++ b/kernels/csrc/cuda/fused/prefill_nvfp4_sm120.cu
@@ -600,8 +600,23 @@ __global__ void ct_dequant_nvfp4_g16_kernel(const unsigned char* __restrict__ pa
     if (global_scale_dev) global_scale = *global_scale_dev;
     const float inv_gs = (ALU >= CT_ALU_RCP) ? (1.f / global_scale) : 0.f;
     const int ngroups = cols >> 4;
-    const int r = blockIdx.y;
-    if (r >= rows) return;
+    // GRID-STRIDE OVER ROWS. This was `const int r = blockIdx.y; if (r >= rows) return;`, which
+    // silently caps the kernel at 65535 rows -- CUDA's limit on gridDim.y. Every weight in the
+    // model clears that (the widest is the FFN gate/up at 17408) except ONE: lm_head, whose row
+    // count is the vocabulary. At V=248320 the launch is rejected with
+    // cudaErrorInvalidConfiguration, and because neither this launch nor the cudaStreamSynchronize
+    // after it is error-checked, the failure is silent -- the freshly cudaMalloc'd destination is
+    // simply never written, so the head dequantizes to all zeros.
+    //
+    // The symptom is a model that loads clean and produces uniform logits: every vocabulary entry
+    // scores identically, PPL comes out at exactly the vocab size, and argmax returns token 0 for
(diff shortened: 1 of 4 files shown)


skyrocket2026:

| metric | value |
|---|---|
| **PR DSpark tok/s** | **126.82** |
| **main DSpark tok/s** | **112.12** |
| PR AR tok/s (floor) | 90.85 |
| main AR tok/s (floor) | 90.30 |
| **DSpark vs AR** | **1.396×** — above 1.0 means speculation finally pays |
| mean accept τ | 2.000 (main 1.662, ceiling 7) |
| mean accept τ floor | ❌ **FAILED** — 2.0000 vs main 1.6623 — throughput gained by speculating less is the feature being removed, not a speedup — **verdict forced to REJECT** |
| qwen3.8 (ModelOpt) guard @16k | ❌ **FAILED** —  — **verdict forced to REJECT regardless of speed** |
| PPL PR / main | ? / ? |
| commit | `21215f970` |

skyrocket2026:

PR #899, opened 2026-08-21
perf(dspark): widen native NVFP4 verifier loads

Paral1995:
## Summary

Speed up native-NVFP4 DSpark verification by widening provably aligned activation and weight loads, and stop building an unreachable Q4_K FFN copy when native NVFP4 decode is selected.

## Proof of speedup

>

| | decode tok/s |
|---|--:|
| before (main) | 89.22 |
| after (this PR) | 94.75 |

| | prefill pp tok/s |
|---|--:|

```text
RTX 5090, Qwen3.8-27B NVFP4 + released DSpark draft, ctx=4096, max_new=128
SPARKINFER_QWEN38_DECODE_NVFP4=1

before (SPARKINFER_NVFP4_ROWS_V4=0):
  batched verify  14.680 ms/call
  DSPARK          89.22 tok/s
  tau             1.5059
  lossless        YES

after (vectorized activation + aligned weight loads):
  batched verify  13.697 ms/call
  DSPARK          94.75 tok/s
  tau             1.5059
  lossless        YES

repeat validation:
  verifier times 13.697 / 13.708 / 13.718 ms

same-binary controls:
  activation widening only (V4=1): 94.41 tok/s, 13.754 ms
  NR=4: 85.92 tok/s, 15.332 ms (rejected and removed)
```

### DIFF
--- a/kernels/csrc/cuda/gemm/gemv.cu
+++ b/kernels/csrc/cuda/gemm/gemv.cu
@@ -717,6 +717,33 @@ __device__ __forceinline__ void si_nvfp4_group_decode(const unsigned char* packe
     }
 }
 
+// 8-BYTE form of the decode above. si_nvfp4_group_dot's header explains why the ONE-ROW kernel
+// leaves these as two 4-byte loads: a uint2 needs 8-byte alignment, which "the row base only
+// happens to have for these shapes, not in general". The ROWS kernel is not the general case --
+// launch_gemv_nvfp4_rows rejects any K with (K & 15), so a weight row base nj*(K>>1) is a multiple
+// of 8 bytes and the group offset g*8 always is, for every shape that can reach here. Same two
+// 32-bit words, same order (little-endian: .x is the low four bytes), so p0 and p1 hold exactly
+// the bits the two separate loads produced.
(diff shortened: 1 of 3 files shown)


skyrocket2026:

| metric | value |
|---|---|
| **PR DSpark tok/s** | **114.66** |
| **main DSpark tok/s** | **112.34** |
| **speedup vs main** | **1.021×** (+2.0%) |
| PR AR tok/s (floor) | 90.30 |
| main AR tok/s (floor) | 90.49 |
| **DSpark vs AR** | **1.270×** — above 1.0 means speculation finally pays |
| mean accept τ | 1.662 (main 1.662, ceiling 7) |
| mean accept τ floor | ✅ 1.6623 vs main 1.6623 (bar ≥95%) |
| qwen3.8 (ModelOpt) guard @16k | ✅ no regression (decode+prefill) — decode 89.1 tok/s · prefill 12743 pp |
| qwen3.6 guard @16k | ✅ no regression (decode+prefill) — decode 472.9 tok/s · prefill 28420 pp |
| commit | `ae5a2053d` |

ok

skyrocket2026:

PR #902, opened 2026-08-21
perf(dspark): drop the draft's fifth layer and widen its attention and Markov loads

coderbench:
## Summary

Run the DSpark draft at four of its five layers, and widen the draft's own attention and Markov
loads. The draft's cost and its acceptance are not equally elastic, and the balance point is not
where the full stack sits. Also fixes `ct_dequant_nvfp4_g16_kernel`, which cannot dequantize a
tensor with more than 65535 rows and so silently zeroes `lm_head` on any checkpoint shipping it as
NVFP4.

Measured at ctx=4096, all arms lossless:

| draft layers | draft ms | tau | dspark tok/s |
|---:|---:|---:|---:|
| 5 (previous default) | 2.201 | 1.5000 | 98.91 |
| **4** | **1.967** | **1.4884** | **100.34** |
| 3 | 1.718 | 1.3763 | 94.33 |
| 2 | 1.483 | 1.3196 | 91.91 |

Below four the curve turns over hard — layer three costs 8% of acceptance to buy 22% of the draft.
Four is the one point where the trade is positive, and it stays well inside the acceptance floor
(0.992x against a 0.95x bar). Context trimming was measured on the same grid and loses at this
length: windowing the full-attention layer to 1024 or 512 takes tau to 1.4066 / 1.4176 for 0.32 ms,
both net negative, so the window keeps its 3x-sliding default and only the depth moves.

The load widening is the technique from #891/#899 applied to the draft's own kernels, which never
had it: `k_attn_rows_split_hd128` fetched its q/k/v four separate 16-bit loads at a time, and
`k_markov_bias_add` streamed 127 MB of `w2` one element per load. Bit-identical -- same
conversions, same order, only the load width moves.

## Proof of speedup

**Decode tok/s** (end-to-end, `dspark-decode@4k`):

| | decode tok/s |
|---|--:|
| before (main) | 98.85 |
| after (this PR) | **102.73** |

**Prefill pp tok/s** (not targeted):

| | prefill pp tok/s |
|---|--:|
| before prefill (main) | — |

Guards: `ctest` 19/19 · `LOSSLESS=YES` over 2 repeats on **two independent prompts** ·
AR 91.19 -> 91.24 (unchanged) · tau 1.5000 -> 1.4884 (0.992x, bar 0.95x).

Run-to-run variance on this box is ~1-2%, so read the headline as **+2.5% to +4%** rather than a
sharp 3.9%.

> **On the baseline.** This box downloads the current checkpoint, whose NVFP4 `lm_head` hits the
> `gridDim.y` bug fixed here and dequantizes to zeros -- the model loads clean, emits uniform
> logits, and pins tau at exactly 1.0000. **The baseline above includes that fix.** With it the box
> reconciles with this repo's own numbers: draft 2.20 ms vs 2.202 in #894, AR 91.2 vs 90.5,
> PPL 3.177 vs 3.204.

```text
# BEFORE -- origin/main + lm_head fix
METRIC AR_TPS 91.1874
METRIC DSPARK_TPS 98.8453
METRIC MEAN_ACCEPT 1.5000
METRIC LOSSLESS 1
METRIC LOSSLESS_RUNS 2

# AFTER -- this PR, prompt 1
METRIC AR_TPS 91.2372
METRIC DSPARK_TPS 102.7281
METRIC MEAN_ACCEPT 1.4884
METRIC LOSSLESS 1
METRIC LOSSLESS_RUNS 2

# AFTER -- this PR, second independent prompt
METRIC DSPARK_TPS 107.6057
METRIC MEAN_ACCEPT 1.5610
METRIC AR_TPS 91.1430
METRIC LOSSLESS 1
METRIC LOSSLESS_RUNS 2
```

Env switches: `SPARKINFER_DFLASH_LAYERS=5` restores the full stack,
`SPARKINFER_CT_NVFP4_G16=0` the per-element dequant.

<sub>Measured and rejected on the way, recorded so they are not retried: NR=4 output rows per warp
(slower on float x, and still slower with x held packed in fewer registers than NR=2 uses today, so
activation traffic is not what that kernel is short of), a half-group register restructure, split-K
resizing, grid-stride/wave capping (fewer blocks is worse -- the kernel wants oversubscription), a
dp4a NVFP4 GEMV (verify gain only 0.12 ms and it breaks the AR floor, since losslessness forces AR
and the verify onto identical arithmetic), and proposal depth 2 (-3.0%).</sub>

### DIFF
--- a/kernels/csrc/cuda/fused/prefill_nvfp4_sm120.cu
+++ b/kernels/csrc/cuda/fused/prefill_nvfp4_sm120.cu
@@ -600,8 +600,9 @@ __global__ void ct_dequant_nvfp4_g16_kernel(const unsigned char* __restrict__ pa
     if (global_scale_dev) global_scale = *global_scale_dev;
     const float inv_gs = (ALU >= CT_ALU_RCP) ? (1.f / global_scale) : 0.f;
     const int ngroups = cols >> 4;
-    const int r = blockIdx.y;
-    if (r >= rows) return;
+    // Grid-stride over rows: gridDim.y is capped at 65535, and lm_head has `vocab` rows (248320
+    // on this checkpoint), so the launch is silently rejected and the head dequantizes to zeros.
+    for (int r = blockIdx.y; r < rows; r += gridDim.y) {
     const unsigned char* prow = packed + (size_t)r * (size_t)(cols >> 1);
     const unsigned char* srow = group_scale + (size_t)r * (size_t)ngroups;
     __nv_bfloat16* orow = out + (size_t)r * (size_t)cols;
@@ -624,6 +625,7 @@ __global__ void ct_dequant_nvfp4_g16_kernel(const unsigned char* __restrict__ pa
         }
         *reinterpret_cast<uint4*>(orow + (size_t)g * 16)     = *reinterpret_cast<const uint4*>(o);
         *reinterpret_cast<uint4*>(orow + (size_t)g * 16 + 8) = *reinterpret_cast<const uint4*>(o + 8);
+    }
     }
 }
 
@@ -752,7 +754,7 @@ static void ct_dequant_nvfp4_launch(const void* packed_u8, const void* group_sca
         const int ngroups = cols >> 4;
         const int threads = 256;
         const int bx = (ngroups + threads - 1) / threads;
-        const dim3 grid(bx > 0 ? bx : 1, rows);
+        const dim3 grid(bx > 0 ? bx : 1, rows > 65535 ? 65535 : rows);
         auto* pk = reinterpret_cast<const unsigned char*>(packed_u8);
         auto* gsc = reinterpret_cast<const unsigned char*>(group_scale_ue4m3);
         auto* ob = reinterpret_cast<__nv_bfloat16*>(out_bf16);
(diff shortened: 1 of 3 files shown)


skyrocket2026:

| metric | value |
|---|---|
| **PR DSpark tok/s** | **109.96** |
| **main DSpark tok/s** | **114.86** |
| PR AR tok/s (floor) | 90.31 |
| main AR tok/s (floor) | 90.58 |
| **DSpark vs AR** | **1.217×** — above 1.0 means speculation finally pays |
| mean accept τ | 1.561 (main 1.662, ceiling 7) |
| mean accept τ floor | ❌ **FAILED** — 1.5610 vs main 1.6623 — throughput gained by speculating less is the feature being removed, not a speedup — **verdict forced to REJECT** |
| qwen3.8 (ModelOpt) guard @16k | ❌ **FAILED** —  — **verdict forced to REJECT regardless of speed** |
| PPL PR / main | ? / ? |
| commit | `b6b4c5376` |

mean accept regression: tau 1.5610 < 95% of main 1.6623 -- throughput bought by speculating less is not a speedup (stopped at the dspark stage; later stages skipped)

skyrocket2026:

This PR's Qwen3.8-27B prefill@16k speed measured **0.0%** vs main, (regression) — closing automatically. This bot evaluates every eligible PR in the repo against Qwen3.8-27B's decode AND prefill@128 speed specifically, regardless of what the PR is actually about — a close here isn't a judgment on the PR's purpose, just that it didn't move these particular metrics. Reopen (or open a fresh PR) if you have a fix or a different approach.

PR #903, opened 2026-08-21
perf(dspark): fuse native NVFP4 gate-up and SwiGLU

Paral1995:
## Summary

Fuse the two native-NVFP4 dense-FFN gate/up projections and their SwiGLU epilogue in the DSpark verifier. The fused kernel loads each activation group once for both weight streams, preserves each projection's reduction order and BF16 rounding point, and removes the standalone SwiGLU graph node.

Skip the fused RMSNorm kernel's Q8_1 epilogue when the verifier immediately enters a checkpoint-native NVFP4 FFN. Native NVFP4 consumes the BF16 normalized rows directly, so those 64 per-layer quantizations and stores were dead work. The RMSNorm reduction and BF16 output path are unchanged.

## Proof of speedup

>

| | decode tok/s |
|---|--:|
| before (main) | 94.69 |
| after (this PR) | 97.14 |

| | prefill pp tok/s |
|---|--:|

```text
RTX 5090, Qwen3.8-27B NVFP4 + released DSpark draft, ctx=4096, max_new=128
SPARKINFER_QWEN38_DECODE_NVFP4=1

before (upstream/main):
  DSPARK          94.6927 tok/s
  tau             1.5059
  lossless        YES

after (fused NVFP4 FFN + unused Q8 epilogue removal):
  DSPARK          97.1365 tok/s
  tau             1.5059
  lossless        YES

repeat validation:
  DSPARK          96.7339 tok/s
  tau             1.5059

rejected controls (not included):
  fused Markov bias + argmax: 95.9997 tok/s
  native LMHead4 prototype: rejected (compact verifier differential)
  one output/warp: 95.8728 tok/s
  three outputs/warp: 95.7584 tok/s
  fused attention K/V: 96.2491 tok/s
```

### DIFF
--- a/kernels/csrc/cuda/fused/rmsnorm.cu
+++ b/kernels/csrc/cuda/fused/rmsnorm.cu
@@ -190,7 +190,7 @@ __global__ void add_rmsnorm2_q8_kernel(const __nv_bfloat16* __restrict__ x,
     const uint4* x4 = reinterpret_cast<const uint4*>(x + base);
     const uint4* r4 = reinterpret_cast<const uint4*>(residual + base);
     uint4* osum4 = reinterpret_cast<uint4*>(out_sum + base);
-    out_q8 += (size_t)blockIdx.x * (cols >> 5);
+    if (out_q8) out_q8 += (size_t)blockIdx.x * (cols >> 5);
 
     float xv[8], rv[8]; rn_unpack8(__ldg(x4 + t), xv); rn_unpack8(__ldg(r4 + t), rv);
     float sv[8]; float ss = 0.f;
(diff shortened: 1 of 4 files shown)


skyrocket2026:

| metric | value |
|---|---|
| **PR DSpark tok/s** | **97.21** |
| **main DSpark tok/s** | **95.65** |
| **speedup vs main** | **1.016×** (+1.6%) |
| PR AR tok/s (floor) | 83.82 |
| main AR tok/s (floor) | 84.16 |
| AR vs main (floor) | -0.4% |
| **DSpark vs AR** | **1.160×** — above 1.0 means speculation finally pays |
| qwen3.8 (ModelOpt) guard @16k | ✅ no regression (decode+prefill) — decode 82.4 tok/s · prefill 12662 pp |
| qwen3.6 guard @16k | ✅ no regression (decode+prefill) — decode 473.1 tok/s · prefill 28447 pp |
| PPL PR / main | 3.0099 / 3.0099 |
| commit | `93847566d` |

skyrocket2026:

This PR's Qwen3.8-27B prefill@16k speed measured **1.6%** vs main, with no verified prefill@16k improvement (the only scored dimension) — closing automatically. This bot evaluates every eligible PR in the repo against Qwen3.8-27B's decode AND prefill@128 speed specifically, regardless of what the PR is actually about — a close here isn't a judgment on the PR's purpose, just that it didn't move these particular metrics. Reopen (or open a fresh PR) if you have a fix or a different approach.

PR #904, opened 2026-08-22
perf(dspark): dp4a the native NVFP4 dense FFN (1.08x dspark-decode@4k)

Paral1995:
## Summary

Run the native-NVFP4 dense FFN on dp4a instead of float GEMVs. NVFP4's decoded magnitudes are exact integers, so a group's dot is exactly `group_scale * sum(mag_k * xq_k)`; the weights stay bit-exact and only the activation is quantized, at one scale per 16 values. Decode and the speculative verify switch together and call the same kernel.

## Proof of speedup

>

| | decode tok/s |
|---|--:|
| before (main) | 95.85 |
| after (this PR) | 103.24 |

| | prefill pp tok/s |
|---|--:|

```text
RTX 5090, Qwen3.8-27B NVFP4 + released DSpark draft, ctx=4096, max_new=128, SPEC-REPS 3

before (SPARKINFER_QWEN38_NVFP4_DP4A=0):
  DSPARK          95.85 tok/s
  AR              82.22 tok/s
  batched verify  13.590 ms/call
  tau             1.5059
  lossless        YES   (3 reps, 0 differ-from-first, 0 not-lossless-vs-AR)

after (dp4a):
  DSPARK         103.24 tok/s
  AR              88.33 tok/s
  batched verify  12.395 ms/call
  tau             1.5059
  lossless        YES   (3 reps, 0 differ-from-first, 0 not-lossless-vs-AR)

emitted stream is byte-identical to main:
  before  SPEC[ 2407 314 6337 421 369 524 177460 1196 ]  steps=85 tokens=128
  after   SPEC[ 2407 314 6337 421 369 524 177460 1196 ]  steps=85 tokens=128

single-rep repeats, alternated arms, same binary:
  off  95.37 / 95.10   verify 13.594 / 13.638
  on  103.33 / 103.35  verify 12.391 / 12.390

kernel (cuobjdump / ncu, gemv_nvfp4_rows_*_kernel<bf16,4,2,2>):
  registers    65 -> 48   (Block Limit Registers 4 -> 5)
  SASS         per-nibble I2F/shift/mask -> IDP.4A + PRMT
  before ncu   DRAM 58.4%, SM 55.3%, Mem Pipes 14.2%, ALU Heavy 47-62% of instructions,
               Long Scoreboard 40-48% of 11.1 cycles between issues,
               6.89 active warps/scheduler, 37.8% of cycles with no eligible warp

rejected on the same binary, kept out:
  half-group activation staging   87.20 tok/s  (registers 65 -> 62, does not change the block limit)
  2-group unroll of the g loop    94.31 tok/s  (registers 65 -> 72, block limit 4 -> 3)
```

### DIFF
--- a/kernels/csrc/cuda/gemm/gemv.cu
+++ b/kernels/csrc/cuda/gemm/gemv.cu
@@ -951,6 +951,160 @@ __global__ void gemv_nvfp4_rows_sk_kernel(const __nv_bfloat16* __restrict__ x,
         }
     }
 }
+
+// ================= dp4a NVFP4 =================================================================
+//
+// The float path above decodes every nibble to a float and does 16 fp32 FMAs per group PER ROW.
+// ncu on the current default says that is what limits it: no pipe is saturated (DRAM 58.4%,
+// SM 55.3%, Mem Pipes 14.2%), ALU Heavy is 47-62% of executed instructions, and 40-48% of the ~11
+// cycles between issues is a Long Scoreboard stall on L1TEX that occupancy cannot hide because 64
+// registers/thread caps the kernel at 4 blocks/SM.
+//
+// NVFP4's decoded magnitudes are EXACT integers -- {0,+-1,+-2,+-3,+-4,+-6,+-8,+-12}, the doubled
+// e2m1 magnitudes -- so sum(w_k * x_k) over a group is exactly group_scale * sum(mag_k * xq_k)
+// once the activation is int8. That turns 16 FMAs into 4 dp4a per row, and lets the nibble decode
+// run as byte-parallel table lookups instead of per-nibble arithmetic.
+//
+// PRMT AS A BYTE LUT. __byte_perm(a, b, sel) uses four nibbles of `sel` to pick four bytes out of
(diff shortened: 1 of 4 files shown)


skyrocket2026:

| metric | value |
|---|---|
| **PR DSpark tok/s** | **103.16** |
| **main DSpark tok/s** | **95.00** |
| **speedup vs main** | **1.086×** (+8.5%) |
| PR AR tok/s (floor) | 88.64 |
| main AR tok/s (floor) | 83.35 |
| AR vs main (floor) | +6.3% |
| **DSpark vs AR** | **1.164×** — above 1.0 means speculation finally pays |
| accuracy gate | ✅ top1=0.980 (bar >=0.9) · KL=0.0027 (bar <=0.1) |
| qwen3.8 (ModelOpt) guard @16k | ✅ no regression (decode+prefill) — decode 87.8 tok/s · prefill 12491 pp |
| qwen3.6 guard @16k | ✅ no regression (decode+prefill) — decode 477.3 tok/s · prefill 28389 pp |
| PPL PR / main | 3.0188 / 3.0099 |
| commit | `4ca2d3d7c` |

ok

skyrocket2026:

PR #905, opened 2026-08-22
perf(dspark): dp4a the native NVFP4 GDN and attention projections (1.024x dspark-decode@4k)

widecloud:
## Summary

#904 put the native-NVFP4 **dense FFN** on dp4a and left the GDN and attention projections on the
float GEMV. This does the same for those — the rest of the checkpoint's NVFP4 — and fixes the
output-row tiling that was tuned for the FFN's width and costs occupancy at projection widths.
**dspark-decode@4k 106.70 → 109.29 tok/s (1.024x)**, AR +0.7%, MEAN_ACCEPT unchanged, lossless.

## Proof of speedup

**Decode tok/s** (`dspark-decode@4k`, the scored dimension, via `runtime/examples/dspark_tau_check`
— the same binary, prompt and 4096-id truncation `eval/pr_dspark_bot.py` uses; `bench.sh` has no
DSpark leg):

| | decode tok/s |
|---|--:|
| before (main) | 106.70 |
| after (this PR) | 109.29 |

```text
ctx=4096, max_new=128, bench_prompt_4k.txt truncated to exactly 4096 ids, alternating, GPU lock

main @8de703a   DSPARK=106.7035  AR=88.4825  TAU=1.5542  LOSSLESS=1
this PR         DSPARK=109.2834  AR=89.0966  TAU=1.5542  LOSSLESS=1
main @8de703a   DSPARK=106.6856  AR=88.4757  TAU=1.5542  LOSSLESS=1
this PR         DSPARK=109.3075  AR=89.0970  TAU=1.5542  LOSSLESS=1
this PR         DSPARK=109.3019  AR=89.1009  TAU=1.5542  LOSSLESS=1  RUNS=3  (SPEC-REPS 3)

emitted stream identical to main:  AR [ 2407 314 6337 421 369 524 177460 1196 ]
```

MEAN_ACCEPT is byte-identical at 1.5542 and AR moves **up** 0.70%, so neither the tau floor nor the
AR no-regression floor is in play. No prefill table: this PR does not touch prefill.

## What changed

**1. The projections run on dp4a.** Same argument #904 established: NVFP4 decodes to
`mag * (group_scale/2)` with `mag` an exact small integer, so a group's dot is
`group_scale * sum(mag * xq)` — an exact integer reduction. Weights stay bit-exact; only the
activation is quantized, at one scale per 16 values. Decode (`proj_xn`, `proj_from`) and the
speculative verify (`proj`, `proj_on`) switch on the same flag and call the same kernel, which is
what keeps DSpark reproducing AR.

**2. `xn` is quantized once per layer, on the main stream, before anything forks.** q/k/v, the qkv
gate and alpha/beta all read the same `xn`, and both sides fork some of them onto `stream_k`. A
first cut quantized lazily inside the projection and had two streams writing one staging buffer —
that showed up as `LOSSLESS=0` with MEAN_ACCEPT collapsing 1.55 → 1.00. The quantize is now issued
before the fork event is recorded, so the forked stream's wait covers it, and projections whose
source is not `xn` use a second buffer so the main stream cannot overwrite `xn`'s mid-flight.

**3. `NR` (output rows per warp-group) is chosen by N, not fixed at 2.** Two rows per warp-group is
right for the FFN's N=17408 — the grid is still several times the SM count and the activation load
is shared. The projections are 2048–12288 rows wide, where halving the grid costs more occupancy
than the sharing buys. Measured, same binary:

| | DSpark | AR |
|---|--:|--:|
| NR=2 everywhere (previous behaviour) | 108.52 | 87.32 |
| NR=1 everywhere | 109.10 | 90.02 |
| **NR by width (this PR)** | **109.29** | **89.10** |

Note this also moves #904's `down` projection (N=5120) to NR=1, which is where most of the AR gain
comes from. `SPARKINFER_NVFP4_DP4A_NR=1|2` pins it for A/B.

`SPARKINFER_QWEN38_NVFP4_DP4A_PROJ=0` restores the float projections; `SPARKINFER_QWEN38_NVFP4_DP4A=0`
still turns the whole dp4a path off as before.

## Scope

Three files. The FP8, Q4_K, Q6_K, Q8_0 and bf16 projection paths are untouched, so anything that is
not this checkpoint's native NVFP4 runs byte-identical code. `qwen3_gguf_bench` / `qwen3_gguf_score`
reach the decode path only, and it is bit-exact per row by construction.

### DIFF
--- a/kernels/csrc/cuda/gemm/gemv.cu
+++ b/kernels/csrc/cuda/gemm/gemv.cu
@@ -2757,10 +2757,21 @@ bool launch_gemv_nvfp4_rows_dp4a(const void* xq, const void* xs, const void* W, 
     auto* yp = reinterpret_cast<__nv_bfloat16*>(y);
     // Split-K chosen by N exactly as the float rows kernel does, so the two paths fan out over the
     // GPU identically and a comparison between them is a comparison of the arithmetic alone.
+    // NR (output rows per warp-group) was fixed at 2, which is right for the dense FFN this path
+    // was written for: at N = 17408 the grid is still ~4x the SM count, and folding two output
+    // rows into a warp-group halves the activation traffic. The GDN and attention projections are
+    // 2048 to 12288 rows wide, where halving the grid costs more occupancy than the sharing buys.
+    // Pick NR by N. SPARKINFER_NVFP4_DP4A_NR pins it for A/B.
+    static const int nr_pin = []{ const char* e = getenv("SPARKINFER_NVFP4_DP4A_NR");
+                                  int v = e ? atoi(e) : 0; return (v == 1 || v == 2) ? v : 0; }();
+    const int nr = nr_pin ? nr_pin : (N >= 16384 ? 2 : 1);
+#define SI_NVFP4_DP4A_NR(S_, R_, NR_) do { \
+        constexpr int S = (S_), R = (R_), NR = (NR_), RPB = GEMV_WPB / S; \
+        const dim3 grid((N + RPB * NR - 1) / (RPB * NR)); \
+        gemv_nvfp4_rows_dp4a_kernel<__nv_bfloat16, S, R, NR><<<grid, GEMV_WPB * 32, 0, stream>>>(xp, sp, W, yp, N, K); \
+    } while (0)
 #define SI_NVFP4_DP4A(S_, R_) do { \
-        constexpr int S = (S_), R = (R_), RPB = GEMV_WPB / S; \
-        const dim3 grid((N + RPB * 2 - 1) / (RPB * 2)); \
-        gemv_nvfp4_rows_dp4a_kernel<__nv_bfloat16, S, R, 2><<<grid, GEMV_WPB * 32, 0, stream>>>(xp, sp, W, yp, N, K); \
+        if (nr == 2) SI_NVFP4_DP4A_NR(S_, R_, 2); else SI_NVFP4_DP4A_NR(S_, R_, 1); \
     } while (0)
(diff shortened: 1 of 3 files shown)


skyrocket2026:

| metric | value |
|---|---|
| **PR DSpark tok/s** | **104.80** |
| **main DSpark tok/s** | **103.11** |
| **speedup vs main** | **1.016×** (+1.6%) |
| PR AR tok/s (floor) | 92.15 |
| main AR tok/s (floor) | 88.64 |
| AR vs main (floor) | +3.9% |
| **DSpark vs AR** | **1.137×** — above 1.0 means speculation finally pays |
| accuracy gate | ✅ top1=1.000 (bar >=0.9) · KL=0.0039 (bar <=0.1) |
| qwen3.8 (ModelOpt) guard @16k | ✅ no regression (decode+prefill) — decode 89.8 tok/s · prefill 12725 pp |
| qwen3.6 guard @16k | ✅ no regression (decode+prefill) — decode 477.3 tok/s · prefill 28329 pp |
| PPL PR / main | 3.0249 / 3.0188 |
| commit | `aa4ab5a61` |

skyrocket2026:

This PR's Qwen3.8-27B prefill@16k speed measured **1.6%** vs main, with no verified prefill@16k improvement (the only scored dimension) — closing automatically. This bot evaluates every eligible PR in the repo against Qwen3.8-27B's decode AND prefill@128 speed specifically, regardless of what the PR is actually about — a close here isn't a judgment on the PR's purpose, just that it didn't move these particular metrics. Reopen (or open a fresh PR) if you have a fix or a different approach.

PR #906, opened 2026-08-22
perf(dspark): dp4a the NVFP4 GDN/attention projections and take the depth they pay for (1.06x dspark-decode@4k)

Paral1995:
## Summary

Route the GDN and attention projections — the rest of the checkpoint's NVFP4 — through dp4a, and take the proposal depth that a cheaper verify row makes profitable. One activation quantize per layer, cached by pointer so the four GDN in-projections and q/k/v share the one `xn`.

## Proof of speedup

>

**Decode tok/s** (`dspark-decode@4k`, the scored dimension, measured with `runtime/examples/dspark_tau_check` on the same prompt and 4096-id truncation `eval/pr_dspark_bot.py` uses; `bench.sh` has no DSpark leg):

| | decode tok/s |
|---|--:|
| before (main) | 102.91 |
| after (this PR) | 109.46 |

| | prefill pp tok/s |
|---|--:|

```text
RTX 5090, Qwen3.8-27B NVFP4 + released DSpark draft, ctx=4096, max_new=128, SPEC-REPS 3,
alternating arms in one binary.

before (SPARKINFER_QWEN38_NVFP4_DP4A_PROJ=0, SPARKINFER_DFLASH_PROPOSALS=1):
  DSPARK   102.9566 / 102.8691 tok/s     tau 1.5059   verify 12.445 ms   lossless 3/3
after (default):
  DSPARK   109.4385 / 109.4827 / 109.4480 tok/s
                                          tau 1.8056   verify 13.718 ms   lossless 3/3
  emitted stream byte-identical:
    before  SPEC[ 2407 314 6337 421 369 524 177460 1196 ]
    after   SPEC[ 2407 314 6337 421 369 524 177460 1196 ]

what a marginal verify ROW costs, which is what decides the depth:
  all float                3.640 ms    best depth 1
  FFN dp4a only            2.044 ms    best depth 1 (depth 2 behind by 1.2%)
  + these projections      0.963 ms    best depth 3

depth curve at ctx=4096, reps=3, all lossless:
  depth 1  104.28 tok/s  tau 1.5059  draft 2.178 ms  verify 12.261 ms
  depth 2  109.30 tok/s  tau 1.7200  draft 2.389 ms  verify 13.224 ms
  depth 3  109.48 tok/s  tau 1.8056  draft 2.517 ms  verify 13.717 ms
  depth 4   76.24 tok/s  tau 1.9403  draft 9.922 ms  verify 15.131 ms
    depth 4 needs a 5-wide draft block, which rounds to the checkpoint's block_size 7 -- a width
    the draft's batched GEMV is not instantiated for -- so the draft falls to its per-token loop.

other contexts, same build, reps=3, lossless at both:
  ctx=1024   109.24 -> 120.07 tok/s   tau 1.5238 -> 1.8551
  ctx=2048   107.15 -> 118.11 tok/s   tau 1.5238 -> 1.8551
  ctx=512     90.32 ->  90.30 tok/s   tau 1.0000 (draft inert here, unchanged)

correctness, and why ctx=4096 alone would not have caught it:
  o_proj does not go through proj_from -- it has an inline branch that kept the float GEMV while
  the verifier converted w.wo, so decode and verify disagreed on sixteen layers of attention
  output. That reports LOSSLESS=0, not a wrong number, and ctx=4096 PASSES it by luck:
    ctx=1024, o_proj left on float:   LOSSLESS=0
    ctx=1024, o_proj converted:       LOSSLESS=1  (3/3)
```

### DIFF
--- a/kernels/include/sparkinfer/kernels/gemm.h
+++ b/kernels/include/sparkinfer/kernels/gemm.h
@@ -111,6 +111,15 @@ void launch_gemv_nvfp4(const void* x, const void* W, void* y, int N, int K,
 // ONE switch for both sides. Decode and the speculative verify must select the same FFN
 // arithmetic or DSpark stops reproducing AR and the losslessness gate fails on a path
 // inconsistency rather than on anything about accuracy.
+// Attention and GDN projections, switched separately from the FFN so each arm can be A/B'd on
+// its own. Decode and verify both read this, so they cannot disagree.
+inline bool qwen38_nvfp4_dp4a_proj() {
+    static const bool v = [] {
+        const char* e = getenv("SPARKINFER_QWEN38_NVFP4_DP4A_PROJ");
+        return !e || e[0] != '0';
+    }();
+    return v;
+}
 inline bool qwen38_nvfp4_dp4a() {
     static const bool v = [] {
         const char* e = getenv("SPARKINFER_QWEN38_NVFP4_DP4A");
--- a/runtime/src/models/qwen35.cpp
+++ b/runtime/src/models/qwen35.cpp
@@ -276,6 +276,12 @@ struct Qwen35Model::Impl {
     // a second feeds down. Sized for the widest K the FFN reads (moe_ffn for down).
     signed char *nv_xq = nullptr;
     float *nv_xs = nullptr;
+    // Projection-side int8 staging (attention + GDN), separate from the FFN's. `a` holds the
+    // layer input xn, shared by every projection that reads it; `b` holds one-off inputs
+    // (o_proj's att, ssm_out's lnrm) so quantizing those cannot clobber xn while the parallel
+    // K/V and GDN streams are still reading it.
+    signed char *nv_pq_a = nullptr, *nv_pq_b = nullptr;
+    float *nv_ps_a = nullptr, *nv_ps_b = nullptr;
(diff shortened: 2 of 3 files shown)


skyrocket2026:

| metric | value |
|---|---|
| **PR DSpark tok/s** | **109.69** |
| **main DSpark tok/s** | **103.11** |
| **speedup vs main** | **1.064×** (+6.3%) |
| PR AR tok/s (floor) | 88.39 |
| main AR tok/s (floor) | 88.64 |
| **DSpark vs AR** | **1.241×** — above 1.0 means speculation finally pays |
| mean accept τ | 1.806 (main 1.506, ceiling 7) |
| accuracy gate | ✅ top1=0.990 (bar >=0.9) · KL=0.0035 (bar <=0.1) |
| mean accept τ floor | ✅ 1.8056 vs main 1.5059 (bar ≥95%) |
| qwen3.8 (ModelOpt) guard @16k | ✅ no regression (decode+prefill) — decode 88.7 tok/s · prefill 12776 pp |
| qwen3.6 guard @16k | ✅ no regression (decode+prefill) — decode 475.8 tok/s · prefill 28396 pp |
| PPL PR / main | 3.0204 / 3.0188 |
| commit | `9ffaef612` |

ok

skyrocket2026:

PR #878, opened 2026-08-18
perf(dspark): propose one token, not three, once the draft is carrying real context (+9.4% dspark-decode@4k)

rsnetworkinginc:
## Summary

Below the deep threshold the draft proposes 3 tokens. At this model's acceptance those extra
proposals are never accepted, so they are pure draft cost — and the cost is not constant: each
proposal is another row of the draft's own attention over the same KV, so it grows with context
while what it buys stays at zero.

The measurements the depth-3 choice rests on were taken with the harness's `NSPLITS=1` pin in
place. At the split counts the server actually runs, the attention that pin inflates is cheap, the
draft's fixed cost is a much larger share of the step, and the balance moves. Measured on RTX 5090
at production split counts, one binary, `LOSSLESS 1` on every run:

| ctx | depth 1 | depth 3 | mean accept (d1 / d3) |
|---:|---:|---:|---|
| 128 | 81.94 | **91.69** | 1.1034 / 1.0159 |
| 512 | 92.32 | 91.95 | 1.0000 / 1.0000 |
| 1024 | **78.72** | 68.30 | 1.1429 / 1.1566 |
| 2048 | **83.04** | 78.35 | 1.0435 / 1.0435 |
| 4096 | **81.23** | 74.82 | 1.0756 / 1.0756 |

Acceptance is *identical* at 4k and 2048 — depth 3 buys exactly nothing there. 128 is the one
context that prefers the wide block, and its acceptance column says why: at depth 3 the draft
accepts essentially nothing (1.0159), the idle-draft rule stops drafting, and the stream runs at
AR speed. Narrowing raises acceptance just enough (1.1034) to keep the draft alive without being
worth its cost. So this does not narrow always — it narrows once the draft is carrying a context
whose per-row cost the deeper proposals cannot repay, which is what the 512 tie and everything
above it show. Below `kNarrowMinSeq` (512) nothing changes, and the `>= kDeepMinSeq` branch keeps
its 7 — long context is where acceptance actually climbs and it is measured separately.

**Decided once per generation, not per step, and that distinction is load-bearing.** The verify
graph `dflash_warm_verify` captures is sized from this depth; narrowing per step under a warm
graph built for 4 rows recovers only 0.7% of the 9.4%, because the mismatched graph costs more
than the rows it saves. Both were measured; the static form is the one that pays.

## Proof of speedup

- [x] Tested on **RTX 5090** (`sm_120`)

[...]

PR #907, opened 2026-08-22
perf(dspark): dp4a the NVFP4 projections, then restore the proposal depth they pay for (1.110x dspark-decode@4k)

widecloud:
## Summary

Two changes that only work together. #904 put the native-NVFP4 **dense FFN** on dp4a; this does the
same for the **GDN and attention projections** — the rest of the checkpoint's NVFP4 — which makes an
extra speculative verify row nearly free. That in turn invalidates #878's proposal narrowing, which
was correct when each extra row cost a full pass over the model's weights. Restoring depth 3 on top
of the cheaper rows takes **dspark-decode@4k 106.71 → 118.42 tok/s (1.110x)** with **MEAN_ACCEPT
1.5542 → 1.9118 (+23%)**, AR +0.7%, lossless.

## Proof of speedup

**Decode tok/s** (`dspark-decode@4k`, the scored dimension, via `runtime/examples/dspark_tau_check`
— the same binary, prompt and 4096-id truncation `eval/pr_dspark_bot.py` uses; `bench.sh` has no
DSpark leg):

| | decode tok/s |
|---|--:|
| before (main) | 106.71 |
| after (this PR) | 118.42 |

```text
ctx=4096, max_new=128, bench_prompt_4k.txt truncated to exactly 4096 ids, alternating, GPU lock

main @8de703a  DSPARK=106.7103  AR=88.5543  TAU=1.5542  LOSSLESS=1
this PR        DSPARK=118.3746  AR=89.1035  TAU=1.9118  LOSSLESS=1
main @8de703a  DSPARK=106.7078  AR=88.4724  TAU=1.5542  LOSSLESS=1
this PR        DSPARK=118.4213  AR=89.0934  TAU=1.9118  LOSSLESS=1
this PR        DSPARK=118.4979  AR=89.1048  TAU=1.9118  LOSSLESS=1  RUNS=3  (SPEC-REPS 3)

emitted stream identical to main:  AR [ 2407 314 6337 421 369 524 177460 1196 ]
```

MEAN_ACCEPT rises 23% and AR rises 0.7%, so neither the tau floor nor the AR floor is in play. No
prefill table: this PR does not touch prefill.

## Why the two halves are one change

The step budget on main, measured with timers around each phase of `dflash_generate`:

| | per step | share |
|---|--:|--:|
| batched verify | 12.267 ms | 82% |
| draft block | 2.185 ms | 18% |

Proposal depth is a trade between those two. #878 measured it when an extra verify row re-streamed
the whole weight set, and narrowing to 1 was worth +9.4%. With the projections row-batched and on
dp4a, an extra row costs the activation work alone. The same sweep, both sides of that change:

| depth | on main | with this PR |
|---|--:|--:|
| 1 | **106.70** / 1.5542 | 109.31 / 1.5542 |
| 2 | 102.58 / 1.7333 | 110.95 / 1.7333 |
| 3 | — | **118.41** / 1.9118 |
| 4 | — | 80.26 / 1.9403 |

Depth 2 **regresses main by 3.9%** and gains 1.5% here. That is the same measurement taken either
side of the row-batching, which is why this is a re-tune rather than a revert of #878 — its number
was right for its cost curve. Depth 4 is the cliff #869 documented: the draft's active block width
rounds up over {4, 8, 16}, so depth 4 runs an 8-wide draft block for 0.03 more accepted tokens.

The depth is tied to the dp4a switch, so `SPARKINFER_QWEN38_NVFP4_DP4A_PROJ=0` restores both the
float projections and depth 1 together — the old cost curve and the tuning that matched it.

## The dp4a half

NVFP4 decodes to `mag * (group_scale/2)` with `mag` an exact small integer, so a group's dot is
`group_scale * sum(mag * xq)` — an exact integer reduction. Weights stay bit-exact; only the
activation is quantized, at one scale per 16 values. Decode (`proj_xn`, `proj_from`) and the verify
(`proj`, `proj_on`) switch on the same flag and call the same kernel, which is what keeps DSpark
reproducing AR.

`xn` is quantized once per layer on the main stream **before anything forks**. q/k/v, the qkv gate
and alpha/beta all read it, and both sides fork some of them onto `stream_k`; a first cut quantized
lazily inside the projection and had two streams writing one staging buffer, which showed up as
`LOSSLESS=0` with MEAN_ACCEPT collapsing 1.55 → 1.00. Sources that are not `xn` use a second buffer.

`NR` (output rows per warp-group) is chosen by width instead of fixed at 2. Two rows is right for
the FFN's N=17408 and wrong for 2048–12288-wide projections, where it halves the grid:

| | DSpark | AR |
|---|--:|--:|
| NR=2 everywhere | 108.52 | 87.32 |
| NR=1 everywhere | 109.17 | 90.02 |
| **NR=2 only at the FFN's width** | **109.29** | **89.10** |

Keying it on row count instead (NR=1 at M==1, NR=2 above) is worse — 108.52 / 90.01. This also
moves #904's `down` (N=5120) to NR=1, which is where most of the AR gain comes from.

## Scope

Three files. FP8, Q4_K, Q6_K, Q8_0 and bf16 paths are untouched, so anything that is not this
checkpoint's native NVFP4 runs byte-identical code. `qwen3_gguf_bench` / `qwen3_gguf_score` reach
the decode path only, and it is bit-exact per row by construction.

### DIFF
--- a/kernels/csrc/cuda/gemm/gemv.cu
+++ b/kernels/csrc/cuda/gemm/gemv.cu
@@ -2757,15 +2757,39 @@ bool launch_gemv_nvfp4_rows_dp4a(const void* xq, const void* xs, const void* W, 
     auto* yp = reinterpret_cast<__nv_bfloat16*>(y);
     // Split-K chosen by N exactly as the float rows kernel does, so the two paths fan out over the
     // GPU identically and a comparison between them is a comparison of the arithmetic alone.
-#define SI_NVFP4_DP4A(S_, R_) do { \
-        constexpr int S = (S_), R = (R_), RPB = GEMV_WPB / S; \
-        const dim3 grid((N + RPB * 2 - 1) / (RPB * 2)); \
-        gemv_nvfp4_rows_dp4a_kernel<__nv_bfloat16, S, R, 2><<<grid, GEMV_WPB * 32, 0, stream>>>(xp, sp, W, yp, N, K); \
+    // NR (output rows per warp-group) was fixed at 2, which is right for the dense FFN this path
+    // was written for: at N = 17408 the grid is still ~4x the SM count, and folding two output
+    // rows into a warp-group halves the activation traffic. The GDN and attention projections are
+    // 2048 to 12288 rows wide, where halving the grid costs more occupancy than the sharing buys.
+    // Pick NR by N. SPARKINFER_NVFP4_DP4A_NR pins it for A/B.
+    static const int nr_pin = []{ const char* e = getenv("SPARKINFER_NVFP4_DP4A_NR");
+                                  int v = e ? atoi(e) : 0; return (v == 1 || v == 2) ? v : 0; }();
+    // Width above which two output rows per warp-group pays. Swept: see the PR notes.
+    // Measured, same binary, DSPARK_TPS / AR: NR=2 everywhere 108.52 / 87.32, NR=1 everywhere
+    // 109.17 / 90.02, NR=2 only at the FFN's width 109.29 / 89.10. Keying it on the ROW COUNT
+    // instead (NR=1 at M==1, NR=2 above) is worse -- 108.52 / 90.01 -- so width is the right
+    // discriminator, not batch.
+    static const int nr_wide = []{ const char* e = getenv("SPARKINFER_NVFP4_DP4A_NRW");
+                                   int v = e ? atoi(e) : 16384; return v > 0 ? v : 16384; }();
+    const int nr = nr_pin ? nr_pin : (N >= nr_wide ? 2 : 1);
+#define SI_NVFP4_DP4A_NR(S_, R_, NR_) do { \
+        constexpr int S = (S_), R = (R_), NR = (NR_), RPB = GEMV_WPB / S; \
+        const dim3 grid((N + RPB * NR - 1) / (RPB * NR)); \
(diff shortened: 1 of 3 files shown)


widecloud:
Superseded by #906, which landed the same two changes (dp4a on the NVFP4 GDN/attention projections, then the proposal depth those cheaper rows pay for) while this was open. Closing rather than leaving a duplicate.

The one piece #906 does not include is the `NR` tiling in `launch_gemv_nvfp4_rows_dp4a`, which is still hard-coded to 2. I re-measured that on top of merged main and it is **not** an improvement there, so it is not worth a follow-up as-is:

| on main @cdb4f65 | DSpark | AR |
|---|--:|--:|
| main (NR=2) | 118.25 / 118.25 | 87.20 / 87.15 |
| NR=1 below 16384 | 116.88 / 116.84 | 89.09 |
| NR=1 below 8192 | 117.49 | 88.71 |

At depth 1 the projections wanted NR=1 (I measured 109.29 vs 108.52 there); at #906's depth 3 the verify runs 4 rows and the activation sharing pays, so NR=2 is right. The two knobs are coupled and #906 has them consistent.

One observation that may still be worth something to you: NR=1 lifts AR by ~2.2% (87.20 -> 89.09) while costing DSpark, because decode runs at M=1 where the sharing cannot pay. Keying NR on the row count rather than the width would get both, but at ctx=4k the batched verify is armed on essentially every step, so the AR gain does not reach DSPARK_TPS and it would score 0. Noting it here in case the AR floor or serving decode matters elsewhere.

PR #908, opened 2026-08-22
perf(dspark): plan the verify length, and fix the two kernels the profile blames (1.080x dspark-decode@4k)

inference2026:
## Summary

Three changes to the DSpark decode step, all measured on the scored dimension.

The draft already scores every proposal's own accept probability and the batched verify discarded
it, verifying a fixed `kProposalDepth + 1` rows on every step — including the 52.8% of steps that
accept nothing. Choose the row count per step from that score against a cost model measured
online. Then fix the two kernels the profile says are leaving the most on the table: the Markov
bias, which was the only draft kernel not at roofline, and the NVFP4 rows GEMV, which is 67% of
the step and is latency bound rather than compute bound.

## Proof of speedup

>

**Decode tok/s** (`dspark-decode@4k`, the scored dimension, measured with
`runtime/examples/dspark_tau_check` on the same prompt and 4096-id truncation `eval/pr_dspark_bot.py`
uses; `bench.sh` has no DSpark leg):

| | decode tok/s |
|---|--:|
| before (main) | 104.68 |
| after (this PR) | 113.03 |

| | prefill pp tok/s |
|---|--:|

```text
RTX 5090, Qwen3.8-27B NVFP4 + released DSpark draft, ctx=4096, max_new=128, SPEC-REPS 3.
Clocks PINNED at 2550 MHz (nvidia-smi -lgc 2550) and arms ALTERNATED, because this box
power-throttles: an unpinned pair of runs moved the AR leg 88.3 -> 90.9 tok/s on its own, which
is larger than the effect being measured. Every arm is the SAME binary -- all three changes are
runtime switches -- so nothing but the switch differs between arms.

  base_1    DSPARK 104.7116  AR 85.6434  MEAN_ACCEPT 1.8056  LOSSLESS 3/3
  base_2    DSPARK 104.6457  AR 85.6226  MEAN_ACCEPT 1.8056  LOSSLESS 3/3
  full_1    DSPARK 112.9917  AR 85.6115  MEAN_ACCEPT 1.7808  LOSSLESS 3/3
  full_2    DSPARK 113.0766  AR 85.6114  MEAN_ACCEPT 1.7808  LOSSLESS 3/3

  base mean 104.6787 (spread 0.06%)  ->  113.0342 (spread 0.08%)  =  +7.98%
  AR -0.02%.  mean accept 98.6% of main.  LOSSLESS on every arm of every run.

each part on its own, same binary, same session, same clock:

  all three off (main behaviour)          104.5231   AR 85.6265   MEAN_ACCEPT 1.8056
  GEMV row mapping only                   109.9277   AR 85.6228   MEAN_ACCEPT 1.8056   +5.17%
  planner + Markov, GEMV mapping off      108.3444   AR 85.6442   MEAN_ACCEPT 1.7568   +3.65%
  planner + Markov, Markov off            111.4443   AR 85.3193   MEAN_ACCEPT 1.7568
  all three                               113.0342   AR 85.6115   MEAN_ACCEPT 1.7808   +7.98%

The GEMV mapping is worth MORE on its own (+5.17%) than stacked (+4.34%), because without the
planner every step verifies four rows, which is exactly where it helps most. Mean accept RISES
when the two are combined (1.7568 -> 1.7808) -- the cost model is fitted online, cheaper rows
lower its per-row term, and the planner then buys more of them.

structural changes alone are neutral, which is the check worth doing first:

  main                                    109.7259 tok/s  AR 88.4077  MEAN_ACCEPT 1.8056
  this PR, all three switches off         109.6516 tok/s  AR 88.3378  MEAN_ACCEPT 1.8056   -0.07%
```

## Verify length

`S_i = prod_{j<=i} sigmoid(confidence_j)` is the probability that row `i` is both reached and
accepted, so verifying `d+1` rows is worth `1 + sum_{i<=d} S_i` tokens and costs `C0 + c1*(d+1)`
ms. Take the `d` that maximises the ratio.

There is no tuned constant in it. The probabilities are the head's own sigmoid — it is trained as
an accept-rate predictor, so `sigmoid(logit)` is already its calibrated output, and fitting a
per-position logistic on top measured no better (+7.14% against +7.18% on the same trace). `C0`
and `c1` are least-squares fitted online from the run's own `(rows, ms)` samples;
`batched_forward` already synchronises before it returns, so the sample costs no extra sync. The
fit converges to the same place the offline sweep found:

```text
offline, SPARKINFER_DFLASH_PROPOSALS=1/2/3 with SPARKINFER_DSPARK_TIMING=1:
  rows 2 -> verify 12.260 ms    rows 3 -> 13.238 ms    rows 4 -> 13.710 ms
  least squares: verify_ms = 10.894 + 0.725 * rows      (the 1-row intercept is AR's own 11.3 ms)

online, from the stream's own samples:
  [timing] plan  2.778 rows/step (of 6)  fit C0=10.692 c1=0.811 ms
```

Why it is worth choosing per step — the accept distribution is bimodal, and the head separates it
(`SPARKINFER_DSPARK_PROBE=1`, 72 steps at ctx=4096):

```text
  accepted proposals   0      1      2      3
  share of steps      52.8%  25.0%  11.1%  11.1%

  mean raw confidence at position 1: +0.580 where accepted, -0.707 where rejected
  positions 2 and 3 separate by +0.98 and +0.79
```

Note the direction this moves acceptance. It is not throughput bought by speculating less: the
rows it declines are rows the target was going to reject.

This needed the verify graph keyed by row count — it was keyed on model / state / conv / capture /
btable / nsplits and **not** on N, so the single graph `dflash_warm_verify` captured was replayed
for every step and the width could only be a per-generation constant. Every arena buffer in that
function is now sized for the widest tier, because `Arena::alloc` frees and reallocates a slot the
moment a later call asks it for more bytes, which would have left an already-instantiated narrower
graph pointing at freed memory.

## The two kernels

**Markov bias.** `w2` is `[vocab, rank]` row major, so one thread per vocabulary row put
consecutive lanes `rank*2` = 512 B apart and each 2-byte load pulled a whole sector. It was the
only draft kernel not at roofline — 870 GB/s on a 127 MB stream against 1.56 TB/s for the LM head
and 1.53 TB/s for the projections. One warp per row, `uint4` loads.

**NVFP4 rows GEMV.** 67% of the step, and ncu says latency, not compute:

```text
  rows      1      2      4      8
  DRAM %  68.2   74.1   57.6   27.0
  SM   %  27.9   31.7   31.2   19.8     <- never above a third of peak
  warps%  93.5   76.8   62.3   31.7
```

`NR` — output rows per warp-group — is what buys warps back. At NR=2 the compiler keeps two sets
of weight temporaries live across the unrolled `j` loop, which costs far more than the second acc
column:

```text
  R          1   2   3   4   5   6   7   8
  NR=2 reg  47  48  56  64  70  78  80  93     <- climbs with the row count
  NR=1 reg  38  39  39  40  42  40  44  42     <- flat
```

Banded rather than switched blanket, because it is not free everywhere. Isolated kernel with a
DRAM-resident working set, microseconds at four rows:

```text
  ffn gate/up N=17408  40.8 -> 34.8      gdn wide N=12288  30.7 -> 26.6
  attn q      N=6144   15.5 -> 13.8      ffn down N=5120   36.8 -> 36.8
                                         attn kv  N=1024    6.2 ->  6.2
```

Neutral or better at every shape for R in [3,7]; at eight rows it inverts on the narrow shapes
(ffn down 49.2 -> 59.4) and at two rows or fewer there are already enough warps, so the doubled
grid is not repaid.

For anyone tempted by tensor cores here: the SM pipes above are the reason not to. The SM120
block-scaled NVFP4 GEMM already in the tree is **slower** than this GEMV at m=8 on the same
weights (ffn gate/up 47.1 -> 51.2 us, ffn down 49.2 -> 77.7, gdn wide 34.8 -> 36.9), and
`prefill_nvfp4_supported` requires `m % 8 == 0`, which AR's single row cannot satisfy.

## Scope and safety

Each output row of the GEMV is still computed by one warp-group over the same group ordering with
the same reduction, so that change is bit-identical; only the row-to-warp mapping moves.

The planner and the Markov kernel are inert on checkpoints without a confidence head:
`forward_block` leaves the buffer untouched, the NaN sentinel survives, and the fixed depth is
used — plain DFlash is byte-unchanged.

`SPARKINFER_DSPARK_PLAN=0`, `SPARKINFER_DFLASH_MARKOV_WARP=0` and `SPARKINFER_NVFP4_ROWS_NR=1|2`
restore the previous behaviour of each part independently, which is how the isolation above was
measured.

The Markov change alters the order of a float reduction, so it can shift which token the draft
proposes. It can never shift which token is emitted — every emitted token is still a target argmax
— so losslessness is unaffected and only acceptance can move.

Also instantiates the draft's batched GEMV at widths 5/6/7: `block_size` is 7 on the released
checkpoints, so rounding `depth+1` over {2,4,8,16} and clamping produced a width nothing was
compiled for, and the draft fell to a per-token loop at every depth above 3 (2.5 ms -> 9.9 ms).
That path is reached at `ctx >= SPARKINFER_DFLASH_DEEP_MIN_SEQ`, where the ladder selects depth 7.

### DIFF
--- a/kernels/csrc/cuda/gemm/gemv.cu
+++ b/kernels/csrc/cuda/gemm/gemv.cu
@@ -1103,6 +1103,22 @@ SI_NVFP4_DP4A_INST(2, 6) SI_NVFP4_DP4A_INST(4, 6) SI_NVFP4_DP4A_INST(8, 6)
 SI_NVFP4_DP4A_INST(2, 7) SI_NVFP4_DP4A_INST(4, 7) SI_NVFP4_DP4A_INST(8, 7)
 SI_NVFP4_DP4A_INST(2, 8) SI_NVFP4_DP4A_INST(4, 8) SI_NVFP4_DP4A_INST(8, 8)
 #undef SI_NVFP4_DP4A_INST
+// NR = 1 as well. NR is how many OUTPUT rows a warp-group owns, and at NR = 2 the compiler keeps
+// two sets of weight temporaries (pw, the four decoded quads, the group scale) live across the
+// unrolled j loop on top of the second acc column. That costs far more than the acc array alone:
+// registers run 47/48/56/64/70/78/80/93 across R = 1..8 at NR = 2 against 38/39/39/40/42/40/44/42
+// at NR = 1, i.e. NR = 1 is nearly FLAT in the row count where NR = 2 is not.
+#define SI_NVFP4_DP4A_INST1(S_, R_) \
+template __global__ void gemv_nvfp4_rows_dp4a_kernel<__nv_bfloat16, S_, R_, 1>(const signed char*, const float*, const void*, __nv_bfloat16*, int, int);
+SI_NVFP4_DP4A_INST1(2, 1) SI_NVFP4_DP4A_INST1(4, 1) SI_NVFP4_DP4A_INST1(8, 1)
+SI_NVFP4_DP4A_INST1(2, 2) SI_NVFP4_DP4A_INST1(4, 2) SI_NVFP4_DP4A_INST1(8, 2)
+SI_NVFP4_DP4A_INST1(2, 3) SI_NVFP4_DP4A_INST1(4, 3) SI_NVFP4_DP4A_INST1(8, 3)
+SI_NVFP4_DP4A_INST1(2, 4) SI_NVFP4_DP4A_INST1(4, 4) SI_NVFP4_DP4A_INST1(8, 4)
+SI_NVFP4_DP4A_INST1(2, 5) SI_NVFP4_DP4A_INST1(4, 5) SI_NVFP4_DP4A_INST1(8, 5)
+SI_NVFP4_DP4A_INST1(2, 6) SI_NVFP4_DP4A_INST1(4, 6) SI_NVFP4_DP4A_INST1(8, 6)
+SI_NVFP4_DP4A_INST1(2, 7) SI_NVFP4_DP4A_INST1(4, 7) SI_NVFP4_DP4A_INST1(8, 7)
+SI_NVFP4_DP4A_INST1(2, 8) SI_NVFP4_DP4A_INST1(4, 8) SI_NVFP4_DP4A_INST1(8, 8)
+#undef SI_NVFP4_DP4A_INST1
 #endif
 
 #ifndef _MSC_VER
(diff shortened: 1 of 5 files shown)


skyrocket2026:

| metric | value |
|---|---|
| **PR DSpark tok/s** | **113.22** |
| **main DSpark tok/s** | **104.33** |
| **speedup vs main** | **1.085×** (+8.5%) |
| PR AR tok/s (floor) | 85.65 |
| main AR tok/s (floor) | 85.65 |
| **DSpark vs AR** | **1.322×** — above 1.0 means speculation finally pays |
| mean accept τ | 1.781 (main 1.806, ceiling 7) |
| mean accept τ floor | ✅ 1.7808 vs main 1.8056 (bar ≥95%) |
| qwen3.8 (ModelOpt) guard @16k | ✅ no regression (decode+prefill) — decode 85.9 tok/s · prefill 12754 pp |
| qwen3.6 guard @16k | ✅ no regression (decode+prefill) — decode 452.7 tok/s · prefill 26663 pp |
| PPL PR / main | 3.0204 / 3.0204 |
| commit | `3d6406ca7` |

ok

skyrocket2026:

PR #909, opened 2026-08-22
perf(dspark): keep the Markov table in L2 and stream the weight reads past it (1.035x dspark-decode@4k)

widecloud:
## Summary

Four L2/memory fixes on the DSpark decode step, all bit-identical:

- **Markov table in int8.** `w2` is `[vocab, 256]` and every proposal row re-reads all 127 MB of it.
  At int8 with per-32 scales the table fits L2, so the 2nd and 3rd rows hit cache instead of
  streaming cold. The bf16 copy is released once the int8 one exists, so this is **-56 MB** VRAM.
- **Evict-first weight loads** (`ld.global.cs`) on the NVFP4 weight streams. Each CTA owns disjoint
  output rows and touches every weight byte once, while the activation block is re-read by *every*
  CTA — the weights were flushing it. Same bytes, same order.
- **Split the GDN side-branch join** in the batched verify: wait for alpha/beta before the scan and
  for `wqkv_gate` before `gated_norm`, instead of one join before `conv_compact`. The wide gate
  GEMV now overlaps the conv and the scan.
- **Vectorized the Q8_1 store** in `add_rmsnorm2/3_q8`: the 8 bytes a thread owns are contiguous,
  so two 32-bit stores replace eight `STG.U8`.

Draft-side changes only nominate; `MEAN_ACCEPT` and `LOSSLESS` are unchanged.

## Proof of speedup

**Decode tok/s** (`dspark_tau_check`, ctx 4096, 128 new tokens; interleaved repeats vs `ea56348`):

| | decode tok/s |
|---|--:|
| before (main) | 120.87 |
| after (this PR) | 125.10 |

```text
RESULT pr1  DSPARK=125.1578 AR=87.9954 TAU=1.8310 LOSSLESS=1
RESULT main DSPARK=120.8548 AR=87.1854 TAU=1.8310 LOSSLESS=1
RESULT pr2  DSPARK=125.0342 AR=87.9919 TAU=1.8310 LOSSLESS=1
RESULT main DSPARK=120.8859 AR=87.1721 TAU=1.8310 LOSSLESS=1
```

tau is identical in all four runs and AR decode gains too (87.18 -> 87.99).

### DIFF
--- a/kernels/csrc/cuda/fused/rmsnorm.cu
+++ b/kernels/csrc/cuda/fused/rmsnorm.cu
@@ -227,11 +227,19 @@ __global__ void add_rmsnorm2_q8_kernel(const __nv_bfloat16* __restrict__ x,
     const float d = amax / 127.0f;
     const int b = t >> 2, r = t & 3;
     int s = 0;
+    // The eight bytes this thread owns are contiguous (qs[r*8 .. r*8+7]), so stage them in
+    // registers and store two 32-bit words instead of eight STG.U8. Two words, not one 64-bit
+    // store: qs sits at offset 4 in si_blk_q8_1 (after the half2 scale), so qs + r*8 is 4-byte
+    // aligned and never 8-byte aligned. Same bytes, same values.
+    signed char qb[8];
     #pragma unroll
     for (int j = 0; j < 8; j++) {
         int qi = (amax == 0.0f) ? 0 : (int)roundf(bv[j] / d);
-        out_q8[b].qs[r * 8 + j] = (signed char)qi; s += qi;
+        qb[j] = (signed char)qi; s += qi;
     }
+    unsigned int* qw = reinterpret_cast<unsigned int*>(out_q8[b].qs + r * 8);
+    qw[0] = reinterpret_cast<const unsigned int*>(qb)[0];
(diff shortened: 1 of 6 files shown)


skyrocket2026:

| metric | value |
|---|---|
| **PR DSpark tok/s** | **117.18** |
| **main DSpark tok/s** | **113.34** |
| **speedup vs main** | **1.034×** (+3.3%) |
| PR AR tok/s (floor) | 86.36 |
| main AR tok/s (floor) | 85.64 |
| AR vs main (floor) | +0.8% |
| **DSpark vs AR** | **1.357×** — above 1.0 means speculation finally pays |
| mean accept τ | 1.781 (main 1.781, ceiling 7) |
| mean accept τ floor | ✅ 1.7808 vs main 1.7808 (bar ≥95%) |
| qwen3.8 (ModelOpt) guard @16k | ✅ no regression (decode+prefill) — decode 87.0 tok/s · prefill 12783 pp |
| qwen3.6 guard @16k | ✅ no regression (decode+prefill) — decode 453.8 tok/s · prefill 26675 pp |
| PPL PR / main | 3.0204 / 3.0204 |
| commit | `7e8c59656` |

ok

skyrocket2026:

PR #910, opened 2026-08-23
feat(server): serving-runtime contract v0 — deterministic mode, /v1/score, first-token logprob; plus a free Q4_K requant accuracy win

ai-hpc:
## Summary

Four things the Gittensor serving integration ([entrius/gittensor#1691](https://github.com/entrius/gittensor/pull/1691)) needs from this runtime, plus a free accuracy win found while verifying them. Everything below was measured on a rented **RTX 5090** against Qwen3.6-35B-A3B UD-Q4_K_M.

> **Not a speed submission.** There is no decode or prefill gain to claim, so the *Proof of speedup* box is deliberately left unticked — this should not be routed to the on-device speed eval. The relevant performance claim here is the opposite one: **decode throughput is unchanged**, evidence below.

| | |
|---|---|
| `SPARKINFER_DETERMINISTIC=1` | bit-reproducible greedy decode (1/36 → **36/36** identical runs) |
| `POST /v1/score` | teacher-forced scoring over HTTP |
| first-token logprob | `logprobs.content` was `completion_tokens − 1` long |
| `server/run.sh` model pin | verified Qwen3.6 weights against the Qwen3-30B digest → infinite re-download loop |
| Q4_K requant refinement | **+0.7 to +1.1 pts top-1 vs llama.cpp, at zero decode cost** |

---

## 1. Q4_K requant refinement — accuracy, free

The load-time attention/GDN requantizer (`proj_q4k_lloyd.cu`) fits each group's `(scale, offset)` by Lloyd-max in **continuous** space, then rounds both onto the shared 6-bit super-block grid. Nothing in the fit accounts for that rounding, so the continuous optimum's *nearest* grid point is frequently not the *best* one — the two errors interact, and the 4-bit codes are re-fit against whichever pair is chosen anyway. It is worst where a group's scale is small relative to the super-block max (`d` is pinned to that max): a group at `topS/13` lands on a grid whose relative step there is ~4%.

`pql_refine_group` now evaluates the 3×3 neighbourhood of the rounded `(s6, m6)` against the **real read-back error** (`y = d·s6·l − dmin·m6`, codes re-fit, same importance weight) and keeps the best.

**4 independent held-out eval seeds, teacher-forced against llama.cpp on the same token sequence:**

| config | top-1 mean | KL mean | KL worst | decode@128 |
|---|--:|--:|--:|--:|
| default requant (main) | 0.8997 | 0.0655 | 0.0970 | 491.5 |
| **default + refine (this PR)** | **0.9066** | **0.0602** | **0.0664** | **491.5** |
| `attn_q,attn_output` only | 0.9192 | 0.0414 | 0.0669 | 428.3 |
| `attn_q,attn_output` + refine | 0.9302 | 0.0315 | 0.0347 | 428.3 |
| requant off | 0.9358 | 0.0282 | 0.0306 | 415.4 |

The **worst-case KL** column is the useful one: refinement takes the default's worst seed from 0.0970 → 0.0664, which is what fixing scale-quantization outliers should look like.

**Why it is free:** the output is the same Q4_K byte layout read by the same `si_vec_dot_q4_K`, so weight traffic and throughput are bit-for-bit unchanged. `SPARKINFER_PROJ_REQUANT_REFINE=0` restores round-to-nearest, so both arms come out of one binary.

**Decode tok/s (`qwen3_gguf_bench`, n=128, 3 reps, median) — the no-regression evidence:**

| | decode@128 | decode@4k |
|---|--:|--:|
| before (main) | 491.5 | 470.1 |
| after (this PR) | 491.5 | 470.1 |

Load time is also unchanged (8.0s vs 8.1s end-to-end).

> Note on the gate: `accuracy.sh`'s default `seed=fixed` prompt is on the pessimistic end — it reports top-1 **0.860** for the *current default* where the 4-seed mean is **0.8997**. Per-seed spread is ~±3 points, so a single manual run is a noisy sample. The eval bot draws a fresh `SPARKINFER_EVAL_SEED` per run.

## 2. Deterministic mode

Repeating an identical batch-1 greedy request forked the token sequence and moved per-token logprobs by tenths of a nat. `SPARKINFER_DETERMINISTIC=1` fixes it:

| 12 prompts × 3 repeats @ 48 tok | default | `DETERMINISTIC=1` |
|---|--:|--:|
| bit-identical runs | 1/36 | **36/36** |
| token-sequence forks | 13/36 | **0/36** |
| mean logprob drift | 0.44 nats | **0** |

Also **batch-invariant** (16/16 identical under 6 concurrent background requests) and verified bit-identical at 117 / 782 / 1922 / 2492 / 3822 / 7622 prompt tokens.

**The cause was not decode** — the decode path has no float atomics at all. It was batched *prompt* prefill, in two fp32 `atomicAdd` accumulations whose operand order is hardware-arbitrated (the routed MoE down-projection combine, and the split-K in the narrow-N GDN gate GEMM), plus the reported logprob's own normalizer, read off a CUB `DeviceScan` whose decoupled look-back folds a timing-dependent number of tile aggregates. Each differs by a few ULP, but they feed discrete top-k expert *routing* and int8 activation requant, so over 40+ layers one occasionally flips an expert and moves the argmax.

The MoE fix needs **no GEMM kernel change**: hand the down projection `pair_orig` instead of `pair_tok` and a partials buffer instead of the shared accumulator, and its existing epilogue writes each pair to its own address, so the atomicAdd becomes uncontended. A fixed-order kernel then folds the partials.

Cost: decode unchanged, TTFT +2% short / +8% at 2k. No accuracy loss — `qwen3_gguf_prefill_check` gives the same seed token and TOP1 24/24 in both modes, with *lower* KL against the token-loop reference (0.00112 vs 0.00176).

## 3. `POST /v1/score` — teacher-forced scoring

Per-token logprobs of a *supplied* continuation, no generation, so a verifier can score any text rather than only its own greedy output. Routed through the ordinary `Job`/`step_job` machinery via `Request::forced_tokens`, so it inherits continuous batching, right-sized KV, 429 and the request deadline — and shares one set of numerics with generation (a scored token reports exactly what `/v1/chat/completions` reports for that token at that position, verified identical).

Deliberately **not** spelled as OpenAI's `/v1/completions` `echo`+`logprobs`: echo's contract is to report logprobs for *prompt* tokens, which batched prefill cannot produce (it runs the LM head only at the final prompt position).

## 4. First-token logprob, and `run.sh`'s model pin

The response's first token comes from prefill's argmax seed, whose LM-head tail stopped before the sort/scan `last_token_logprobs()` reads — so `logprobs.content` came back one short and silently misaligned any consumer zipping it with the output. Entries 2..N were always correct, so the off-by-one hit *every* token, not just the first.

`server/run.sh` serves Qwen3.6 but `verify_model` read `MODEL_SHA256`, whose `reference.lock` default is the **Qwen3-30B guard model's** digest — so a correct 22 GB download mismatched, was deleted, re-fetched and failed again, unbounded under a restarting supervisor. The HF upload was never stale (`ac0e2c11…` still matches `QWEN36_MODEL_SHA256`). `resolve_model_sha256()` now picks the digest belonging to `MODEL_FILE`, and `MODEL_SHA_POLICY=strict` reports a mismatch instead of deleting 22 GB.

## Also fixed

- **Streaming logprobs**: entries for tokens routed to `reasoning_content` were held back and attached to the next *content* chunk; anything still pending at end of generation was dropped. A request whose output was entirely reasoning returned **zero** entries. Fixed in the chat and legacy paths — concatenating every chunk now reproduces the non-streaming array exactly.
- **`ensure_tokenizer` deleted before downloading**, so one transient `curl: (56)` left the models dir with no tokenizer at all. Hit for real during this work. Now stages and only moves on success.
- **A port clash** printed `shutdown signal received`, drained 30s, and only then mentioned binding. Now fails immediately with the cause. `/v1/completions` and `/v1/score` were missing from the startup banner.

## Two defects found and NOT fixed here

1. **GQA-fused int8 prefill attention above ~2048 tokens.** With int8 KV (which the server enables whenever `--ctx ≥ 4096`) the fused kernel disagrees sharply with the token-loop reference, nondeterministically — mean KL 0.00022 at 2000 tokens vs **0.18672** at 2100, against ~0.0001 for the per-q-head fallback at both. The cliff is exactly at 2048 but is *not* the RQH=3 tier's own `n_tokens >= 2048` gate (RQH=2 is chosen either side of it and is equally wrong above), so the length dependence is inside `launch_attn_gqa`. Pre-existing and independent of everything here. **Left on by default** because switching it off moves long-context prefill throughput the eval scores against; deterministic mode defaults `SPARKINFER_PREFILL_ATTN_GQA_RQH=1` so that mode is trustworthy at every length.

2. **The accuracy gate cannot see it.** H2 scores long context but through `qwen3_gguf_score` (token loop, not batched prefill); H3 exercises batched prefill but at `prefix=512`, below the cliff. The defect lives exactly in the gap.

## Verification

- **Full `bench/scripts/accuracy.sh`** run end to end on this build **and on pristine `75508fb`**. Both report `top1=0.860000 kl=0.080407 ppl_spark=2.8284` on `seed=fixed` — identical to six figures, confirming the serving changes are numerically inert on the default path. (The requant refinement then moves it, as intended and measured above.) The gate's enforced vetoes pass: H3 `top1=0.8750 kl=0.00650` (bars 0.80 / 0.05); long-context int8-KV mean `top1=0.9583 kl=0.4250` over 3 seeds (bars 0.875 / 1.00).
- **21/21** `ctest` targets pass, including new `chat_tools_test` coverage for `/v1/score` parsing.
- **Gittensor's own `scripts/check_serving_runtime.py`**: `CONFORMANT: 0 MUST failure(s), 0 SHOULD warning(s)` with the deterministic flag on (D1 prefix agreement p05 **1.000**, logprob drift p95 **0.0000**).

## Proof of speedup

Intentionally unticked — see the note at the top. All measurements above were taken on an RTX 5090 (`sm_120`), but this PR claims **no** decode or prefill gain and should not be routed to the speed eval. The decode table in §1 is a no-regression control, not a speedup claim.

### DIFF
--- a/bench/scripts/_common.sh
+++ b/bench/scripts/_common.sh
@@ -12,6 +12,11 @@ LLAMACPP_DIR="${LLAMACPP_DIR:-$ROOT/.llamacpp}"   # override to reuse an existin
 
 # C2 (reference quarantine): pin the baseline artifacts so a tampered persisted copy can't skew a
 # verdict. reference.lock carries MODEL_SHA256 + LLAMACPP_COMMIT; empty = warn-only until pinned.
+#
+# Captured BEFORE the lock is sourced: reference.lock gives MODEL_SHA256 a `:-` default, so after
+# sourcing it is always non-empty and an explicit caller pin becomes indistinguishable from the
+# default. resolve_model_sha256() below needs that distinction.
+_ENV_MODEL_SHA256="${MODEL_SHA256:-}"
 _HERE_COMMON="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 [ -f "$_HERE_COMMON/reference.lock" ] && source "$_HERE_COMMON/reference.lock"
 LLAMACPP_REPO="${LLAMACPP_REPO:-https://github.com/ggml-org/llama.cpp}"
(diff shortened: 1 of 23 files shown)


skyrocket2026:

Touched: `bench/scripts/_common.sh`

skyrocket2026:

Touched: `bench/scripts/_common.sh`

skyrocket2026:

Touched: `bench/scripts/_common.sh`

skyrocket2026:

Touched: `bench/scripts/_common.sh`

skyrocket2026:

Touched: `bench/scripts/_common.sh`

skyrocket2026:

Touched: `bench/scripts/_common.sh`

skyrocket2026:

Touched: `bench/scripts/_common.sh`

skyrocket2026:

Touched: `bench/scripts/_common.sh`

skyrocket2026:

Touched: `bench/scripts/_common.sh`

skyrocket2026:

Touched: `bench/scripts/_common.sh`

skyrocket2026:

Touched: `bench/scripts/_common.sh`

skyrocket2026:

Touched: `bench/scripts/_common.sh`

PR #911, opened 2026-08-24
perf(dspark): price the verify per width, calibrate the accept head, and cut the per-row cost (1.068x dspark-decode@4k)

widecloud:
## Summary

Three changes to how the batched verify is *planned*, plus the two kernel fixes the profile
blamed for the per-row cost. `dspark-decode@4k` **104.97 -> 112.06 tok/s (1.068x)** with tau up
7.3% (1.590 -> 1.7067), AR up 5.8%, and every kernel change bit-identical to `main`.

The planner's cost model was affine in the row count and its worth model trusted the accept head's
raw sigmoids. Neither holds:

1. **Price each verify WIDTH, don't fit a line through them.** The row-batched NVFP4 GEMV picks its
   output-rows-per-warp *from* the row count, so the cost curve has a knee and a least-squares line
   mis-prices both ends — it read `C0=10.31 c1=1.02` where the true costs are 12.34 ms at 2 rows
   and 14.39 at 4. Every width already has its own warmed graph tier, so walk the widths once at
   the start of decode and keep a running mean of each.
2. **Maximise net gain, not the ratio.** `worth(d)/cost(d+1)` optimises tokens per *verify*
   millisecond, but the step also pays a draft whose size does not depend on `d`. Maximising
   `worth(d) - rate*cost(d+1)`, where `rate` is the run's own achieved tokens per millisecond of
   whole step, is the throughput objective — and the exchange rate is measured, not assumed.
3. **Calibrate the accept head online.** Instrumented at full depth (so `accept` is uncensored),
   the head is under-confident at every position, and multiplying raw sigmoids compounds it:

   | position | P(accept \| prefix accepted) | mean sigmoid(logit) | ratio |
   |---|--:|--:|--:|
   | 1 | 0.4789 | 0.4211 | 1.14 |
   | 2 | 0.4706 | 0.3496 | 1.35 |
   | 3 | 0.5625 | 0.3282 | 1.71 |

   Predicted mean `S_3` is 0.044 against an observed 0.127 — **2.9x** — so the plan reads "row 3
   will never land" and stops short. Its *discrimination* is fine (position-1 AUC 0.847, monotone
   buckets); only the level is wrong, so a per-position scale fitted from the run's own accepts
   fixes it. Every measured ratio is above 1 and the clamp floor is 1, so this can only ever make
   the plan **wider** — more rows verified against the same draft, never fewer.

The three interact and should not be read separately: ratio-rule + calibration measures *worse*
(109.9) than the net-gain rule alone (111.0); together they give 115.2.

Two kernel fixes cut the marginal row cost from 1.39 to 0.975 ms:

- **`NR` by row count in `launch_gemv_nvfp4_rows_dp4a`.** `NR` is how often the `R x K` activation
  block is re-read. The two ends want opposite things and the shipped rule (`R in [3,7] -> NR=1`)
  had it backwards: R=1 (AR decode, one 5 KB activation row that stays in L1) wants NR=1 —
  **AR 91.22 vs 86.73 tok/s** — and R>=2 (the verify, 25 KB re-read by 4352 CTAs) wants NR=2 —
  **14.116 vs 14.725 ms per 4-row verify**.
- **The `SEQ` row-fold in the GQA-6 attention branch.** `fa_split_gqa_kernel` has had a bit-exact
  fold since it was written, but only the **GQA-8** branch ever dispatched it, so Qwen3.8's 24Q/4KV
  full-attention layers re-streamed the whole KV once per verify row — exactly linear, 57.4 us at
  4 rows against 29.2 at 2. Fold 2, not 4: the grid is only a few CTAs/SM and the kernel is issue-
  bound on the per-row softmax, so the wide fold gives back more parallelism than it saves.

Plus: two of this lane's own groups per trip in the dp4a mainloop, gate/up in one paired launch,
`MMAX=4` instead of 6 in the exact Q4_K head, the NVFP4 activation quantize folded into the SwiGLU
that produces its input, and `k_add_rms` widened with its bf16 round-trip kept in registers.

**Everything here is bit-identical to `main` by construction** — same reduction order, same
association, same bytes — and that is verified rather than asserted: the teacher-forced score dump
diffs at **top1 = 1.000, KL = 0.0000 nats, PPL 3.0204 both sides**.

## Proof of speedup

Numbers are from `runtime/examples/dspark_tau_check` — the harness this dimension is scored on
(CONTRIBUTING.md) — at ctx=4096 on `bench/scripts/bench_prompt_4k.txt` truncated to 4096 ids, 128
new tokens, PR and `main` interleaved on one box, two repeats each.

**Decode tok/s** (DSpark speculative decode @ ctx=4k):

| | decode tok/s |
|---|--:|
| before (main) | 104.97 |
| after (this PR) | **112.06** |

**Prefill pp tok/s** — not applicable, this PR does not target prefill.

| | prefill pp tok/s |
|---|--:|

```text
# ctx=4096, 128 new tokens, interleaved main/PR/main/PR on one RTX 5090, same model load path.
# base: upstream/main 9e43bfa

RESULT main rc=0 DSPARK=105.5784 AR=85.9432 TAU=1.6000 LOSSLESS=1
RESULT  PR  rc=0 DSPARK=112.0026 AR=90.6955 TAU=1.7067 LOSSLESS=1
RESULT main rc=0 DSPARK=104.3522 AR=85.6454 TAU=1.5802 LOSSLESS=1
RESULT  PR  rc=0 DSPARK=112.1161 AR=90.8264 TAU=1.7067 LOSSLESS=1

                  mean DSPARK   mean AR    tau
  main            104.97        85.79      1.590
  this PR         112.06        90.76      1.7067
  ratio            1.068         1.058     1.073

# SPARKINFER_DSPARK_TIMING=1 on the same four runs -- the plan is wider AND the rows are cheaper:
main  [timing] draft 2.375 ms  plan 2.288 rows/step (of 4)  fit C0=9.633  c1=1.375  batched 12.779 ms
PR    [timing] draft 2.350 ms  plan 2.840 rows/step (of 4)  fit C0=10.114 c1=0.977  batched 12.887 ms
main  [timing] draft 2.374 ms  plan 2.272 rows/step (of 4)  fit C0=9.596  c1=1.397  batched 12.769 ms
PR    [timing] draft 2.346 ms  plan 2.840 rows/step (of 4)  fit C0=10.116 c1=0.972  batched 12.876 ms

# The marginal row cost, measured directly rather than fitted (PLAN=0 + DFLASH_PROPOSALS=1/2/3,
# so the verify runs at a pinned vn while the draft block stays 4 wide):
#   vn=2  batched 12.391 ms      vn=3  batched 13.962 ms      vn=4  batched 15.266 ms
#   -> 1.4375 ms per row on main.  This is why "the planner under-verifies" was the wrong
#      diagnosis: at that price it was already close to right, and forcing full depth measured
#      101.99 against the planner's 107.38.

# Differential accuracy, teacher-forced, PR dump vs main dump on the same token stream:
PPL PR                : 3.020
PPL main              : 3.020
METRIC top1=1.000000 kl=0.000000 ppl_pr=3.0204 ppl_main=3.0204
```

Isolated-kernel numbers (extra evidence only, not the before/after): the NVFP4 rows kernel with a
DRAM-resident weight — 24 rotating copies, so the 50 MB matrix cannot sit in L2 — goes 34.24 ->
32.95 us at R=3 and 34.85 -> 33.69 at R=4 for the FFN gate/up shape with two groups per trip. A
standalone bench is a poor oracle here and I would not weigh it: it also said NR=1 beats NR=2 at
R=4 and that four groups per trip beat two, and end-to-end both are false, because in the bench the
activation stays hot across launches while in the verify it was written by the kernel immediately
before.

**Not run on my side:** `SPEC_REPS=3` and the 16k guard sweeps — the dev box's GPU fell off the bus
partway through the gate suite and could not be reset from inside the container. Every one of the
~40 single-repeat runs during this work reported `LOSSLESS=1`, and the accuracy dump above is
exact, but the 3-repeat losslessness gate and the long-context guards are the bot's to confirm.

### DIFF
--- a/kernels/csrc/cuda/attention/flash_decode_split.cu
+++ b/kernels/csrc/cuda/attention/flash_decode_split.cu
@@ -1247,6 +1247,54 @@ void launch_flash_decode_split(
                 (void)seqlen;
                 return;
             }
+            // Fold S verify rows into one CTA so the staged K/V tile is read ONCE for all of
+            // them instead of once per row. The 8:1 group has had this since the fold kernel was
+            // written; the 6:1 group (Qwen3.8-27B's 24Q/4KV full-attention layers) never got the
+            // dispatch, so every batched verify re-streamed the whole KV per row. Measured at
+            // ctx=4096, N=4: 57.4 us per layer against 29.2 at N=2 -- exactly linear in rows,
+            // i.e. the KV read was being paid four times.
+            //
+            // Bit-exact for the same reason the 8:1 fold is: each folded row keeps its OWN
+            // chunk/start/end (a verify block's row i ends at start_pos+i+1) and masks itself back
+            // to that range, so it accumulates the same keys in the same order the SEQ=1 kernel
+            // gives it. Only which CTA stages the tile changes.
+            //
+            // Requires the rows to share one block table, which the verify already guarantees:
+            // dflash_verify_short_run replicates its table N times (launch_broadcast_rows_i32)
+            // before this call, and is the only caller that passes num_seqs > 1 here.
+            static int fa6_fold_env = -1;
+            if (fa6_fold_env < 0) {
+                const char* e = getenv("SPARKINFER_FA_SEQFOLD");
+                fa6_fold_env = e ? atoi(e) : 0;
+            }
+            // Prefer the NARROWEST fold that divides the row count. Folding trades KV-staging
+            // traffic for CTAs -- at 4 kv heads x n_splits the grid is already only a few CTAs
+            // per SM -- and this kernel is issue-bound on the per-row softmax, not on the staged
+            // read, so the wide fold gives back more parallelism than it saves. Measured at
+            // ctx=4k, 4 rows: fold 2 = 14.389 ms per verify against fold 4 = 14.438 and no fold
+            // at 14.6, i.e. two rows per CTA takes the whole win.
+            const int fa6_fold = fa6_fold_env > 0
+                               ? fa6_fold_env
+                               : (num_seqs % 2 == 0 ? 2 : (num_seqs % 3 == 0 ? 3 : 1));
+            if (!int8_kv && num_seqs > 1 && fa6_fold > 1 && (num_seqs % fa6_fold) == 0) {
+                const size_t smf = (size_t)2 * FA_GQA6_TILE * 256 * sizeof(__nv_bfloat16);
+                dim3 gqf(num_kv_heads * n_splits, num_seqs / fa6_fold);
+#define SI_FA6_FOLD(SQ)                                                                           \
+                fa_split_gqa_kernel<256, GQA, FA_GQA6_TILE, false, (SQ)>                          \
+                    <<<gqf, GQA * 32, smf, stream>>>(                                             \
+                    reinterpret_cast<const __nv_bfloat16*>(q), k_pool, v_pool, block_table,       \
+                    seq_lens, part_m, part_l, part_acc, scale, num_q_heads, num_kv_heads,         \
+                    block_size, max_blocks, n_splits,                                             \
+                    reinterpret_cast<const __half*>(k_scale),                                     \
+                    reinterpret_cast<const __half*>(v_scale))
+                if (fa6_fold == 3)      SI_FA6_FOLD(3);
+                else if (fa6_fold == 4) SI_FA6_FOLD(4);
+                else                    SI_FA6_FOLD(2);
+#undef SI_FA6_FOLD
+                combine_hd256(out_q8);
+                (void)seqlen;
+                return;
+            }
             // How many keys ONE block walks -- not the sequence length. This, not seqlen, is what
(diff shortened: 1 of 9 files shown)


skyrocket2026:

| metric | value |
|---|---|
| **PR DSpark tok/s** | **120.33** |
| **main DSpark tok/s** | **115.56** |
| **speedup vs main** | **1.041×** (+4.1%) |
| PR AR tok/s (floor) | 90.41 |
| main AR tok/s (floor) | 86.37 |
| AR vs main (floor) | +4.6% |
| **DSpark vs AR** | **1.331×** — above 1.0 means speculation finally pays |
| qwen3.8 (ModelOpt) guard @16k | ✅ no regression (decode+prefill) — decode 91.0 tok/s · prefill 12726 pp |
| qwen3.6 guard @16k | ✅ no regression (decode+prefill) — decode 451.3 tok/s · prefill 26760 pp |
| commit | `de248d423` |

ok

skyrocket2026:

PR #913, opened 2026-08-24
perf(dspark): draft one more proposal, and instantiate the kernels that depth needs (1.042x dspark-decode@4k)

inference2026:
## Summary

Raise the short-context proposal depth from 3 to 4, and instantiate the three kernels that depth
walks into. The verify planner already prices each row against its own measured cost, so a deeper
draft only spends where the planner buys: it plans 3.250 rows of 5 against 3.041 of 4, and mean
accept goes 1.7297 -> 1.8824.

The kernel fixes are enablers, not independent wins — all three together at depth 3 measure +0.03%.
They matter only because depth 4 puts the draft head at M=4 and the block at a width nothing was
instantiated for.

- `si_mmvq_q4k_multirow` had exact instantiations at M=3 and M=6 only, so M=4 ran the MMAX=16 form:
  sixteen predicated rows at 56-60 registers against 39-40. Instantiated exactly for M=1..8.
- The Markov bias kernel gave each warp one vocabulary row — a single 256-byte load, then a
  five-step shuffle reduction in which 31 of 32 lanes discard their result. The table is already
  L2-resident, so this is latency rather than bandwidth; two rows per warp puts two loads in flight.
- Drafting at a depth the block can only just back wrote past six proposal-indexed buffers, never
  clamped depth to `block_size`, and sized the target-hidden capture at `B` rows for a `B+1`-row
  verify. That last one has a bounds guard, so it dropped the row silently while `th_len = keep`
  still read one row past the allocation.

## Proof of speedup

**Decode tok/s** (`dspark-decode@4k`, the scored dimension, `runtime/examples/dspark_tau_check` on
the same prompt and 4096-id truncation the bot uses):

| | decode tok/s |
|---|--:|
| before (main) | 120.08 |
| after (this PR) | 125.18 |

**Prefill pp tok/s**

| | prefill pp tok/s |
|---|--:|

```text
RTX 5090, Qwen3.8-27B NVFP4 + released DSpark draft, ctx=4096, max_new=128, SPEC-REPS 3,
clocks pinned. Baseline and candidate come out of ONE binary (baseline = every switch this
PR adds turned off), arms alternated, three repeats each.

before  120.0821 / 120.0936 / 120.0765   mean 120.084   tau 1.7297  draft 2.20 ms  plan 3.041/4
after   125.2120 / 125.1688 / 125.1559   mean 125.179   tau 1.8824  draft 2.61 ms  plan 3.26/5
                                                                                   = 1.042x
lossless 6/6 on every run; AR leg 90.38 tok/s on all arms.

An earlier run of the same arms on the identical tree gave 119.917 -> 125.076 (1.043x), so the
ratio reproduces across sessions as well as across repeats.

Depth, measured at the same width so the two are not confounded:
  depth 3  120.1792      depth 4  125.076      depth 5  121.8012

Attribution, each switch turned off against the full candidate at depth 4:
  head MFIXED off    122.8150   (+1.20% for the fix)
  markov rows off    124.0255   (+0.21%)
  all kernel fixes on but depth left at 3:  120.1190 against main's 120.087  -> +0.03%

Uncensored acceptance (planner disabled, so every proposal reaches the target and the accept
count is bounded by the draft rather than the plan):
  depth 3  1.7534   depth 4  1.9104   depth 5  1.9394   depth 6  1.9692   depth 7  1.9692
Depth 6 and 7 are identical, so proposals past 6 never land; 4 is where the marginal row still
repays its own verify time.

Kernel-level, each with a control at a row count the dispatch already handled:
  si_mmvq_q4k_multirow exact M=1..8
    depth 4  draft 2.942 -> 2.760 ms    depth 5  draft 3.110 -> 2.905 ms
    depth 6  draft 3.043 -> 3.032 ms    <- control: M=6 was already exact, and does not move
  Markov bias rows per warp, depth 7 draft: 1 -> 3.199/3.200 ms, 2 -> 3.106/3.107, 4 -> 3.139/3.139
```

### DIFF
--- a/kernels/csrc/cuda/gemm/gemv.cu
+++ b/kernels/csrc/cuda/gemm/gemv.cu
@@ -3617,13 +3617,33 @@ bool launch_gemv_q4k_dp4a_multirow_f32(const void* q81, const void* W, float* y,
     dim3 grid(nblk);
     auto* q = reinterpret_cast<const si_block_q8_1*>(q81);
     auto* w = reinterpret_cast<const unsigned char*>(W);
-    // MFIXED 3 and 6 keep the row loop fully unrolled at the two depths the draft actually uses;
-    // anything else runs the MMAX=16 form with a runtime bound.
+    // One exact MFIXED per row count the proposal ladder can actually ask for, not just the two
+    // it used to ask for. MMAX sizes acc[MMAX] and bounds the row loop, so the fallback form runs
+    // SIXTEEN predicated rows to do five and carries 56 registers against 39-40 for the exact
+    // ones -- 2 CTAs of 16 warps against 3, i.e. 66% occupancy against 100%. Measured on the
+    // draft head (V=248320, K=5120): 838.6 us at M=5 on the fallback against 457.1 us at M=3 on
+    // the exact form. That gap was the largest single term in the cost of drafting a wider block,
+    // and it is why depth 6 (exact) was cheaper than depth 5 (fallback) despite proposing more.
+    // A/B switch: 0 restores the previous dispatch (exact only at M = 3 and 6).
+    static const bool mfixed_all = []{ const char* e = getenv("SPARKINFER_DFLASH_HEAD_MFIXED");
+                                       return !(e && e[0] == '0'); }();
     #define SI_MR_BY_M(NS) \
         do { \
-            if (M == 3)      si_mmvq_q4k_multirow_kernel<float, 16, NS, 3, 3><<<grid, 16 * 32, 0, stream>>>(q, w, y, N, M); \
-            else if (M == 6) si_mmvq_q4k_multirow_kernel<float, 16, NS, 6, 6><<<grid, 16 * 32, 0, stream>>>(q, w, y, N, M); \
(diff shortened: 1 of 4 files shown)


skyrocket2026:

| metric | value |
|---|---|
| **PR DSpark tok/s** | **125.20** |
| **main DSpark tok/s** | **120.11** |
| **speedup vs main** | **1.042×** (+4.2%) |
| PR AR tok/s (floor) | 90.42 |
| main AR tok/s (floor) | 90.42 |
| **DSpark vs AR** | **1.385×** — above 1.0 means speculation finally pays |
| mean accept τ | 1.882 (main 1.730, ceiling 7) |
| mean accept τ floor | ✅ 1.8824 vs main 1.7297 (bar ≥95%) |
| qwen3.8 (ModelOpt) guard @16k | ✅ no regression (decode+prefill) — decode 90.9 tok/s · prefill 12831 pp |
| qwen3.6 guard @16k | ✅ no regression (decode+prefill) — decode 450.4 tok/s · prefill 26761 pp |
| commit | `1aa6446d7` |

ok

skyrocket2026:

PR #915, opened 2026-08-25
perf(dspark): cut the draft's per-block bytes and its KV re-reads (1.039x dspark-decode@4k)

widecloud:
## Summary

Two changes to the DSpark drafter, both found by profiling the decode loop with
`nsys --cuda-graph-trace=node`. The draft block was 17.3% of the step at ctx 4096 and cost
2.575 ms; it now costs 1.953 ms, for `dspark-decode@4k` 122.790 -> 127.632 = **1.039x**.

- **The draft nominates over a vocabulary prefix.** Everything downstream of the draft backbone is
  proportional to the vocabulary, and on Qwen3.8 that is 248320. Per block the head streams the
  target's Q4_K LM head once (715 MB) and then the Markov chain re-reads `w2` — `[248320, 256]`
  int8 plus per-32 scales, ~71 MB — *once per proposal row, serially*. #913's depth 4 makes that
  four rows: 715 + 4 x 71 = ~1001 MB of a ~1.4 GB block. Both tables are nomination-only, so a
  token the head cannot score is simply never proposed: it costs at most one acceptance and can
  never change what is emitted, because every emitted token is still a target argmax over the FULL
  vocabulary. Both are row-major `[vocab, ...]`, so a prefix is one smaller length — no repack, no
  second copy, no extra VRAM. `w1` is indexed by the PREVIOUS token and is not pruned, so the
  conditioning still accepts any target token; only the output side narrows.
- **The draft's attention read its KV 20x per layer.** `k_attn_rows_split_hd128` grids as
  `(ceil(q_len/ROWS), n_q, n_splits)` and derives the kv head per q head, so every K/V byte is
  re-read `(n_q/n_kv) x ceil(q_len/ROWS)` times — 5 x 4 = 20 at #913's block width of 7 with
  ROWS=2. Its per-layer KV is only ~17 MB at ctx 4k, small enough to live in L2, which is why it
  showed up as 0.60 ms a block (29% of the draft) at an apparent ~139 GB/s rather than as a
  bandwidth wall. ROWS=8 collapses the four row groups to one.

Both stay A/B-able out of one binary: `SPARKINFER_DFLASH_DRAFT_VOCAB=0` restores the full
vocabulary, `SPARKINFER_DFLASH_ATTN_ROWS=2` the old row tile.

Implementation note: `si_mmvq_q4k_multirow` writes `y[m*N + n]`, so the pruned vocabulary length is
also the logits ROW STRIDE — it has to travel with the length or the Markov chain reads the wrong
row.

## Proof of speedup

**Decode tok/s** (`dspark-decode@4k`, the scored dimension, end-to-end via
`runtime/examples/dspark_tau_check` on the same prompt and 4096-id truncation the bot uses — not an
isolated-kernel microbenchmark):

| | decode tok/s |
|---|--:|
| before (main) | 122.790 |
| after (this PR) | 127.632 |

**Prefill pp tok/s** (this PR does not target prefill; the Qwen3.8 and Qwen3.6 @16k prefill
no-regression guards are in the block below):

| | prefill pp tok/s |
|---|--:|

```text
RTX 5090 (sm_120), 525 W cap. Qwen3.8-27B ModelOpt NVFP4 (rev 2ed84505) + the released DSpark
draft. ctx=4096, max_new=128. main = ebbaa69. Arms interleaved, THREE repeats each -- not two,
because the generation lands in either 70 or 71 steps depending on what the verify-width
planner's online cost model does, and that alone is ~1.4% of tok/s.

arm     DSPARK tok/s              AR tok/s                mean accept
main    122.865 122.631 122.875   94.588 94.413 94.399    1.8286 x3
PR      128.252 127.300 127.345   94.451 94.423 94.432    1.8286 1.8028 1.8028

  dspark-decode@4k   127.632 / 122.790 = 1.039x
  mean accept        1.8114 / 1.8286   = 0.991x   (floor 0.95)
  AR decode          94.436 /  94.467  = 1.000x   (floor 0.98)
  draft              2.579 -> 1.953 ms/block  (-24.3%)
  step              14.892 -> 14.192 ms

AR decode is flat across every arm, so this is bytes and re-reads off the draft, not throughput
bought by speculating less.

GATES
  LOSSLESS = 1 at SPARKINFER_DSPARK_SPEC_REPS=3, both arms
  differential accuracy vs main: top1 = 1.000000  kl = 0.000000  PPL 3.0198 both
      (the target path is bit-identical -- which is the point: the draft only nominates)
  qwen3.8   @16k guard: decode 92.32  vs 92.23    prefill 12564.8 vs 12506.3
  qwen3.6   @16k guard: decode 469.76 vs 469.11   prefill 27661.5 vs 27589.7
  both guards land slightly ahead of main.

VOCABULARY PREFIX -- coverage, share of tokens with id < K, three unrelated corpora
  K         bench_prompt_4k   repo prose   repo C++
  32768         92.61%          90.86%      93.77%
  65536         97.53%          96.64%      97.92%
  98304         99.95%          99.23%     100.00%
 163840         99.95%          99.38%     100.00%
Qwen3.8 has no reserved tail to reclaim -- only 243 of the 248320 ids are undefined -- but the
vocabulary is frequency-ordered and coverage plateaus just under 100k.

VOCABULARY PREFIX -- swept end to end, one binary (SPARKINFER_DFLASH_DRAFT_VOCAB)
  K         DSPARK tok/s   draft ms/block   mean accept
  248320      122.87           2.575          1.8286
  131072      125.90           2.250          1.8286
   98304      125.98           2.159          1.8286
   65536      127.08           2.075          1.8286   <- shipped
   49152      127.84           2.03           1.8286   <- peak
   32768      125.90           1.99           1.8028
   24576      126.42           1.96           1.8028
   16384      125.00           1.94           1.7778
Acceptance is flat down to 49152 and only turns over below it: at 32768 the generation needs a
71st step, which is what 1.8286 (128/70) -> 1.8028 (128/71) is.

Shipped 65536, NOT the 49152 peak. 49152 sits one step from that edge and its 0.6% advantage was
found by walking the prompt this benchmark scores; 65536 comes off the coverage curve that three
unrelated corpora agree on and keeps a full step of margin. For a change whose entire safety
argument is coverage, that is the side to err on.

ATTENTION ROW TILE -- draft ms/block, one binary (SPARKINFER_DFLASH_ATTN_ROWS), K=65536
  ROWS 2 -> 2.070    ROWS 4 -> 2.001    ROWS 8 -> 1.955
Monotone. ROWS 2 -> 4 is on record in this repo as a REGRESSION, but that was measured when the
block was 4 wide, where it is 2 row groups -> 1 and halves the grid; at #913's width of 7 it is
4 -> 1 and the trade is different, so it was re-derived rather than inherited. ROWS=8 covers the
whole block in one group and the kernel epilogue already does `if (qt >= q_len) continue;`, so the
overhang is free. smem is 32.5 KB at ROWS=8, under the 48 KB default.

PROFILE (nsys --cuda-graph-trace=node; without it the batched verify is ONE opaque graph node)
main step 14.891 ms = batched verify 12.317 (82.7%) + draft 2.575 (17.3%)
  verify = C0 9.707 + c1 0.801 x 3.257 rows; per-width, PLAN=0: w2 11.285 ms, w3 12.119 ms
per-kernel groups, ms/step (nsys-inflated ~8%):
  NVFP4 dp4a2 gate+up (paired)   4.176   1517 GB/s
  NVFP4 dp4a S=4                 3.871   ~1300-1365 GB/s
  NVFP4 dp4a S=2                 1.506
  verify LM head rows_exact      0.702   1019 GB/s
  draft backbone fused3_q4_ks    0.630
  DRAFT ATTENTION                0.603    139 GB/s   <- fixed here
  verify attention fa_split_gqa  0.548
  gdn scan+conv checkpoint       0.471
  add_rmsnorm2_q8 (128/step)     0.316
  nvfp4 quant_x (192/step)       0.243
  draft LM head multirow         0.134   1410 GB/s   <- shrunk here

MEASURED AND NOT TAKEN
- Covering the verify's 5-row attention fold. #913 makes width 5 the widest plan a short-context
  verify can ask for, and the GQA-6 fold takes the narrowest width that divides the row count
  exactly -- nothing, for 5 -- so it stages the whole KV five times. Covering 5 as 2+3 measured
  122.72 against main's 122.87, i.e. nothing: the planner reaches width 5 too rarely to pay.
- CUDA-graphing the draft block. Total kernel time 1036 ms against a decode wall of 1073 ms over
  70 steps = 96.5% GPU-busy. The draft is not launch-bound, whatever the comment in
  dflash_draft.cpp says, so a graph capture cannot buy more than ~0.5 ms and realistically far
  less.

WHAT THIS DOES NOT DO
+3.9% is not a tier win and I do not think a tier is available on this axis. The same profile puts
the NVFP4 weight GEMVs at 63% of the step and, corrected for nsys overhead, within ~4% of their
byte floor; C0 alone is 65%. After these two changes the draft is down to 13.8% of the step. The
largest remaining single item is the verify's LM head at 1019 GB/s, and that is blocked by design:
si_mmvq_q4k_rows_exact exists to reproduce AR decode's single-row reduction order bit-for-bit,
which is what keeps the speculative path lossless by construction, and AR's own kernel is faster
than both alternatives -- so there is nothing faster to unify on.
```

### DIFF
--- a/runtime/csrc/cuda/dflash_kernels.cu
+++ b/runtime/csrc/cuda/dflash_kernels.cu
@@ -1311,7 +1311,28 @@ void launch_attn_gqa(const void* q, const void* k, const void* v, void* out,
         // Row-batched + KV-split path. Needs the caller's partial-state scratch; without it
         // (or below the sweep's crossover) fall through to the single-CTA-per-row kernel.
         if (fa_m && fa_l && fa_acc && attn_gqa_split_path_ok(q_len, kv_len, n_q, n_kv, d)) {
-            constexpr int ROWS = 2, NWARPS = 8;
+            constexpr int NWARPS = 8;
+            // QUERY ROWS PER CTA. The grid is (ceil(q_len/ROWS), n_q, n_splits) and the kv head is
+            // derived per q head, so each K/V byte is re-read (n_q/n_kv) x ceil(q_len/ROWS) times.
+            // On Qwen3.8's DSpark draft that is 5 x ceil(q_len/ROWS): at #913's block width of 7
+            // and ROWS=2 it is FOUR row groups, so 20 reads of every key. The draft's per-layer KV
+            // is ~17 MB at ctx 4k, small enough to sit in L2, which is why this shows up as
+            // ~139 GB/s of "DRAM" and 0.60 ms a block -- 29% of the whole draft -- rather than as
+            // a bandwidth wall. Widening the row tile is the only knob here that removes reads
+            // rather than moving them: ROWS >= q_len collapses the row groups to one.
+            // ROWS 2 -> 4 is on record as a REGRESSION, but that was measured when the block was
+            // 4 wide, where it was 2 row groups -> 1 and halved the grid. At #913's width of 7 it
+            // is 4 groups -> 1 and the trade is different, so it was re-derived rather than
+            // inherited. Measured at ctx 4096, K=65536, one binary, draft ms/block:
+            //     ROWS 2 -> 2.070    ROWS 4 -> 2.001    ROWS 8 -> 1.955
+            // Monotone, so the old note does not hold at this width. 8 covers the whole block in
+            // one group; the epilogue already drops rows past q_len, so the overhang is free.
+            // SPARKINFER_DFLASH_ATTN_ROWS pins it for an A/B out of one binary.
+            static const int rows_env = []{
+                const char* e = getenv("SPARKINFER_DFLASH_ATTN_ROWS");
+                int v = e ? atoi(e) : 8;
+                return (v == 2 || v == 4 || v == 8) ? v : 8;
+            }();
             const int n_splits_full = attn_gqa_splits(kv_len);
             // A sliding-window layer masks a key only after the kernel has loaded it and reduced
             // its QK dot, so past the window length it streams the whole cache to discard most of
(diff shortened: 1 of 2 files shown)


skyrocket2026:

| metric | value |
|---|---|
| **PR DSpark tok/s** | **128.82** |
| **main DSpark tok/s** | **125.17** |
| **speedup vs main** | **1.029×** (+2.9%) |
| PR AR tok/s (floor) | 90.36 |
| main AR tok/s (floor) | 90.29 |
| **DSpark vs AR** | **1.426×** — above 1.0 means speculation finally pays |
| mean accept τ | 1.882 (main 1.882, ceiling 7) |
| mean accept τ floor | ✅ 1.8824 vs main 1.8824 (bar ≥95%) |
| qwen3.8 (ModelOpt) guard @16k | ✅ no regression (decode+prefill) — decode 90.8 tok/s · prefill 12720 pp |
| qwen3.6 guard @16k | ✅ no regression (decode+prefill) — decode 450.1 tok/s · prefill 26659 pp |
| commit | `d3c712e29` |

ok

skyrocket2026:

PR #916, opened 2026-08-25
perf(dspark): delete the verify's activation-quantize nodes and shorten the draft attention chain

widecloud:
## Summary

Three node/latency cuts on the `dspark-decode@4k` path, all **bit-identical** to `main` — the
differential accuracy check reports `top1=1.000000 kl=0.000000` and both arms print the same PPL:

1. **Fold the NVFP4 activation quantize into the norm that produces it** (`add_rmsnorm2_q8_kernel`,
   `pf_gated_norm_kernel`). `si_nvfp4_quant_x_kernel` ran **192 times a step at 1.26 us on a 4-CTA
   grid** — almost entirely node cost for ~40 KB of work — and every value it reads is one the
   producing norm already holds in registers. Emitting the int8 + per-16-scale form there deletes
   **176 of those 192 launches**. A 16-element group is two adjacent threads in the rmsnorm and 16
   adjacent lanes in the gated norm, so one shuffle completes it.
2. **A key tile in the draft's `k_attn_rows_split_hd128`.** It did a five-step warp reduction *per
   key* with the online-softmax update carried across the loop. Computing KT keys' dots first makes
   their reductions independent, and a `shfl_xor` butterfly pairs lanes in exactly the tree
   `shfl_down` builds at lane 0 — same float, and now in every lane, so the per-row broadcast
   shuffle disappears. The tile is then folded into the state one key at a time in the original
   order, and each warp keeps its original key sequence (`tb + u*NWARPS`), so the result is
   bit-for-bit what the old loop produced. `draft 1.953 -> 1.801 ms/block`.
3. **Token-parallel GDN conv** (`df_gdn_conv_checkpoint_kernel`). The causal conv is an FIR filter,
   not a recurrence, but the kernel walked tokens serially — with two or three block barriers each —
   on a grid of only `2*q_heads + v_heads` = 80 CTAs. Rebuilding each token's window from `qkv` and
   the seeded state gives `grid.y = n_tokens`. Same taps, same ascending order, so bit-identical.

## Proof of speedup

**Decode tok/s** — `dspark-decode@4k`, the scored dimension: `dspark_tau_check <model> <draft> 128
$(cat ids_4096.txt)` with the 4096 prompt ids from `bench/scripts/bench_prompt_4k.txt`, on the
ModelOpt NVFP4 checkpoint. Not an isolated-kernel microbenchmark.

| | decode tok/s |
|---|--:|
| before (main `f711b53`) | 126.84 |
| after (this PR) | **130.77** |

**Prefill pp tok/s** (Qwythos / Qwen3.5): N/A — this PR does not target prefill.

| | prefill pp tok/s |
|---|--:|

```text
# Gate run, SPARKINFER_DSPARK_SPEC_REPS=3, 525 W RTX 5090, ctx 4096, 128 new tokens
main DSPARK=126.8429 AR=94.3848 tau=1.8028 lossless=1/3
pr   DSPARK=130.7684 AR=94.5470 tau=1.8028 lossless=1/3
== dspark-decode@4k  1.0309x        (tau IDENTICAL on both arms, AR +0.17%)

# 2 interleaved main/PR repeats, same box, with the per-phase breakdown
main.t1  DSPARK 128.3277  AR 94.6003  tau 1.8286  lossless 1
  draft 1.952 ms/call   batched 12.298 ms/call   3.229 rows/step   fit C0=9.692 c1=0.807
pr.t1    DSPARK 130.5308  AR 94.5693  tau 1.8028  lossless 1
  draft 1.800 ms/call   batched 12.011 ms/call   3.141 rows/step   fit C0=9.504 c1=0.798
main.t2  DSPARK 128.4323  AR 94.4002  tau 1.8286  lossless 1
  draft 1.954 ms/call   batched 12.284 ms/call   3.200 rows/step   fit C0=9.707 c1=0.805
pr.t2    DSPARK 130.8615  AR 94.5377  tau 1.8028  lossless 1
  draft 1.802 ms/call   batched 11.974 ms/call   3.085 rows/step   fit C0=9.481 c1=0.808

# Step decomposition at a common 3.21 rows/step:
#   verify C0  9.700 -> 9.492 ms   (-0.207)   <- the deleted quantize nodes
#   draft      1.953 -> 1.801 ms   (-0.152)   <- the key tile
#   step      14.243 -> 13.875 ms  (-0.368 = 2.66%)

# Differential accuracy, PR vs main on the same token stream
```

**On the two tok/s numbers.** `dspark_tau_check` generates exactly 128 tokens, so a run lands in a
whole number of steps and mean-accept is quantised to `128/70 = 1.8286` or `128/71 = 1.8028` — one
step is ~1.4% of tok/s. In the gate run both arms landed on 71 steps and the ratio is the clean
one: **1.0309x**. In the interleaved pairs `main` landed on 70 and this PR on 71, which reads as
+1.8%. That is not a regression in acceptance: the verify-width planner maximises tokens/second
from measured per-width costs, and a cheaper fixed cost per step (`draft + C0`, which this PR cuts
by 0.36 ms) correctly shifts its optimum by a fraction of a row. The step-time reduction — 2.66%,
and independent of where the step boundary falls — is the physical change.

## Gates

| gate | result |
|---|---|
| `LOSSLESS` at `SPEC_REPS=3` | 1, over 3 runs |
| tau floor (0.95 x main) | 1.8028 vs 1.8028 — identical |
| `AR_TPS` floor (0.98 x main) | 94.547 vs 94.385 = 1.0017x |
| differential accuracy | top1 1.000000, KL 0.000000 (bar 0.90 / 0.1) |
| long-context guard @16k, Qwen3.8 | dec 92.263 -> 92.442, pp 12502.5 -> 12574.8 |
| long-context guard @16k, Qwen3.6 | dec 469.164 -> 470.656, pp 27550.5 -> 27615.9 |

```text
GUARD38 si_main 16384 dec=92.2634 pp=12502.5336
GUARD36 si_main 16384 dec=469.164 pp=27550.4562
GUARD38 si_pr   16384 dec=92.4423 pp=12574.8487
GUARD36 si_pr   16384 dec=470.6557 pp=27615.8523
ALL GATES GREEN
```

Both guards come out slightly ahead of `main` rather than merely clean: the norm fold is on the
`add_rmsnorm2_q8` path the AR decode uses too.

## What did NOT go in

Recorded so the next reader does not re-derive them. All measured on the same box, one binary per
arm, env-switched:

* The same key tile on the **verify's** attention (`fa_split_gqa_kernel`) is a **loss** — batched
  12.000 -> 12.074 ms at 3.17 rows. It already stages K/V in shared memory and walks only ~32 keys
  per CTA at 4k, and at head_dim 256 holding the tile's V costs 32 registers.
* A **K-aware split-K rule** for the NVFP4 row GEMVs (narrow split for deep-K `ffn_down`): +0.072
  ms/step and AR -1.4%. Pinning S globally gives C0/c1/AR = 9.849/0.756/92.98 at S=2,
  9.637/0.822/94.62 at S=4 (the current rule), 8.801/1.440/92.51 at S=8 — S down buys activation
  reuse and costs grid, and the c1 it saves is smaller than the C0 it costs at every shape here.
* Exercising the LM head's existing `GRP` template parameter (row-groups per CTA, to share the
  activation pull) at 2 and 4: -0.009 ms, i.e. noise.
* Exact `MMAX` instantiations per verify width: neutral. That head is ~linear in the real row
  count, not in `MMAX` — 665 us at mean M 3.1 against 989 at 5.15.

### DIFF
--- a/kernels/csrc/cuda/fused/batched_prefill.cu
+++ b/kernels/csrc/cuda/fused/batched_prefill.cu
@@ -806,15 +806,94 @@ __global__ void df_gdn_conv_checkpoint_kernel(
     }
 }
 
+// TOKEN-PARALLEL form of df_gdn_conv_checkpoint_kernel.
+//
+// The causal conv is an FIR filter, not a recurrence: y[tok] depends only on the conv_kernel
+// inputs at positions tok-conv_kernel+1..tok, all of which are already in `qkv` (or, for negative
+// positions, in the seeded decode state). The kernel above nevertheless walks the tokens SERIALLY
+// so it can carry `hist` in registers, and pays for it twice -- the whole per-token chain
+// (including two or three block barriers for the q/k norm) is serialised, and the grid is only
+// 2*q_heads + v_heads = 80 CTAs, so 90 of this box's 170 SMs sit idle for the 4.3 us it takes.
+//
+// Reconstructing the window per token from memory costs conv_kernel-1 extra loads (they are the
+// same cache lines the neighbouring CTAs read) and gives grid.y = n_tokens.
+//
+// BIT-IDENTICAL to the serial kernel: the same taps in the same ascending order, the current
+// token's tap last -- the order the single-token decode conv uses, which is what keeps the batched
+// verify reproducing AR (see the serial kernel's own comment on #712) -- and the same block
+// reduction for the q/k norm.
+template <bool WRITE_CHECKPOINT>
+__global__ void df_gdn_conv_par_kernel(
+    const __nv_bfloat16* __restrict__ qkv, const __nv_bfloat16* __restrict__ conv_w,
+    const __nv_bfloat16* __restrict__ live_state, __nv_bfloat16* __restrict__ q,
+    __nv_bfloat16* __restrict__ k, __nv_bfloat16* __restrict__ v,
+    __nv_bfloat16* __restrict__ checkpoints, int n_tokens, int q_heads, int v_heads,
+    int head_dim, int qkv_dim, int conv_kernel, float eps) {
+    const int q_dim = q_heads * head_dim;
+    const int v_dim = v_heads * head_dim;
+    const int blk = blockIdx.x, tok = blockIdx.y, t = threadIdx.x;
+    int d, out_dim, hh; __nv_bfloat16* out; bool do_norm;
+    if (blk < q_heads) { hh = blk; d = hh * head_dim + t; out = q; out_dim = q_dim; do_norm = true; }
+    else if (blk < 2 * q_heads) { hh = blk - q_heads; d = q_dim + hh * head_dim + t; out = k; out_dim = q_dim; do_norm = true; }
+    else { hh = blk - 2 * q_heads; d = 2 * q_dim + hh * head_dim + t; out = v; out_dim = v_dim; do_norm = false; }
+    float w[8], hist[7];
+#pragma unroll
+    for (int c = 0; c < 8; c++) w[c] = c < conv_kernel ? pf_to_f(conv_w[(size_t)d * conv_kernel + c]) : 0.f;
+    // Window entering token `tok`: tap c is input position tok - (conv_kernel-1) + c. A negative
+    // position is decode state, and its live_state row is that position + (conv_kernel-1) = tok+c.
+#pragma unroll
+    for (int c = 0; c < 7; c++) {
+        if (c >= conv_kernel - 1) { hist[c] = 0.f; continue; }
+        const int p = tok - (conv_kernel - 1) + c;
+        hist[c] = p >= 0 ? pf_to_f(qkv[(size_t)p * qkv_dim + d])
+                         : pf_to_f(live_state[(size_t)(tok + c) * qkv_dim + d]);
+    }
+    const float cur = pf_to_f(qkv[(size_t)tok * qkv_dim + d]);
+    float y = 0.f;
+#pragma unroll
+    for (int c = 0; c < 7; c++) if (c < conv_kernel - 1) y += hist[c] * w[c];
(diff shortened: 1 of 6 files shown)


skyrocket2026:

| metric | value |
|---|---|
| **PR DSpark tok/s** | **0.00** |
| **main DSpark tok/s** | **129.37** |
| PR AR tok/s (floor) | 0.00 |
| **DSpark vs AR** | **0.000×** — above 1.0 means speculation finally pays |
| mean accept τ | 0.000 (main 1.882, ceiling 7) |
| mean accept τ floor | ❌ **FAILED** — 0.0000 vs main 1.8824 — throughput gained by speculating less is the feature being removed, not a speedup — **verdict forced to REJECT** |
| PPL PR / main | ? / ? |
| commit | `0d173ad58` |

skyrocket2026:

widecloud:
Superseded by #917 — same change, rebased onto current `main` (`12954e6`), with the unused tile-width instantiations dropped. Reopening is not available to this account (`reopenPullRequest` is refused), so this is a fresh PR from the same branch rather than a duplicate submission.

For the record on this one: every PR-side number here came back `0.00` — DSpark, AR, tau and top1 alike. Per `eval/pr_dspark_bot.py`, `_ds_metric` defaults to 0 when the run emits no `METRIC` line and the losslessness gate is checked first, so that shape is a crash or a failed load rather than a measured divergence. I could not reproduce it on a 5090 with the same CUDA 13.0: it builds and runs clean, lossless at `SPEC_REPS=3` over repeated runs, with byte-identical free VRAM to `main` at every load stage and no register spill. #917 carries the two real problems I did find (stale base; 8.6 MB of device code for kernels that never launch).

PR #917, opened 2026-08-25
perf(dspark): delete the verify's activation-quantize nodes and shorten the draft attention chain

widecloud:
## Summary

Three node/latency cuts on the `dspark-decode@4k` path. All three are **bit-identical** to `main`: the differential accuracy check reports `top1=1.000000 kl=0.000000` and both arms print the same perplexity, because none of them changes a single arithmetic result — they only change how many kernel launches produce it.

## Proof of speedup

**Decode tok/s** — `dspark-decode@4k`, the scored dimension: `dspark_tau_check <model> <draft> 128 $(cat ids_4096.txt)` with the 4096 prompt ids from `bench/scripts/bench_prompt_4k.txt`, on the ModelOpt NVFP4 checkpoint, `SPARKINFER_DSPARK_SPEC_REPS=3`. Not an isolated-kernel microbenchmark.

| | decode tok/s |
|---|--:|
| before (main `12954e6`) | 136.98 |
| after (this PR) | **140.20** |

**Prefill pp tok/s** (Qwythos / Qwen3.5): N/A — this PR does not target prefill.

| | prefill pp tok/s |
|---|--:|

```text
# 600 W RTX 5090, CUDA 13.0, ctx 4096, 128 new tokens, SPEC_REPS=3
main DSPARK=136.9789 AR=96.0512 tau=1.8028 lossless=1/3
pr   DSPARK=140.2010 AR=95.9776 tau=1.8028 lossless=1/3
== dspark-decode@4k  1.0235x        (tau IDENTICAL on both arms, AR -0.08%)

# Differential accuracy, PR vs main on the same token stream
positions             : 101
PPL PR                : 3.020
PPL main              : 3.020

# Interleaved main/PR repeats with the per-phase breakdown (525 W box, CUDA 12.8, main f711b53)
main  DSPARK 128.33 / 128.43   draft 1.952 / 1.954   C0 9.692 / 9.707   c1 0.807 / 0.805
pr    DSPARK 130.53 / 130.86   draft 1.800 / 1.802   C0 9.504 / 9.481   c1 0.798 / 0.808

# Step decomposition at a common 3.21 rows/step:
#   verify C0  9.700 -> 9.492 ms   (-0.207)   <- the deleted quantize nodes
#   draft      1.953 -> 1.801 ms   (-0.152)   <- the key tile
#   step      14.243 -> 13.875 ms  (-0.368 = 2.66%)
```

**On reading the tok/s number.** `dspark_tau_check` emits exactly 128 tokens, so a run lands in a whole number of steps and mean-accept is quantised to `128/70`, `128/71` or `128/72` — one step is ~1.4% of tok/s. Both arms above landed on the same step count, so `1.0235x` is a clean comparison. The step-time reduction, which does not depend on where the step boundary falls, is 2.66%.

## Gates

| gate | result |
|---|---|
| `LOSSLESS` at `SPEC_REPS=3` | 1, over 3 runs |
| tau floor (0.95 x main) | 1.8028 vs 1.8028 — identical |
| `AR_TPS` floor (0.98 x main) | 95.978 vs 96.051 = 0.9992x |
| differential accuracy | top1 1.000000, KL 0.000000 (bars 0.90 / 0.1) |
| long-context guard @16k, Qwen3.8 (upstream unsloth build) | dec 82.543 -> 82.619, pp 10527.7 -> 10547.9 |
| long-context guard @16k, Qwen3.6 | dec 480.227 -> 481.968, pp 29162.5 -> 29191.4 |

```text
GUARD38 si_main 16384 dec=82.5425 pp=10527.6566
GUARD36 si_main 16384 dec=480.2265 pp=29162.5212
GUARD38 si_pr   16384 dec=82.6194 pp=10547.8837
GUARD36 si_pr   16384 dec=481.9682 pp=29191.4279
ALL GATES GREEN
```

Both guards come out slightly ahead of `main` rather than merely clean — the norm fold is on the `add_rmsnorm2_q8` path AR decode uses too. (The Qwen3.8 guard here benches `Q38_GUARD_MODEL_DIR`, the upstream unsloth NVFP4 build, matching the fix in 36da646.)

### DIFF
--- a/kernels/csrc/cuda/fused/batched_prefill.cu
+++ b/kernels/csrc/cuda/fused/batched_prefill.cu
@@ -806,15 +806,94 @@ __global__ void df_gdn_conv_checkpoint_kernel(
     }
 }
 
+// TOKEN-PARALLEL form of df_gdn_conv_checkpoint_kernel.
+//
+// The causal conv is an FIR filter, not a recurrence: y[tok] depends only on the conv_kernel
+// inputs at positions tok-conv_kernel+1..tok, all of which are already in `qkv` (or, for negative
+// positions, in the seeded decode state). The kernel above nevertheless walks the tokens SERIALLY
+// so it can carry `hist` in registers, and pays for it twice -- the whole per-token chain
+// (including two or three block barriers for the q/k norm) is serialised, and the grid is only
+// 2*q_heads + v_heads = 80 CTAs, so 90 of this box's 170 SMs sit idle for the 4.3 us it takes.
+//
+// Reconstructing the window per token from memory costs conv_kernel-1 extra loads (they are the
+// same cache lines the neighbouring CTAs read) and gives grid.y = n_tokens.
+//
+// BIT-IDENTICAL to the serial kernel: the same taps in the same ascending order, the current
+// token's tap last -- the order the single-token decode conv uses, which is what keeps the batched
+// verify reproducing AR (see the serial kernel's own comment on #712) -- and the same block
+// reduction for the q/k norm.
+template <bool WRITE_CHECKPOINT>
+__global__ void df_gdn_conv_par_kernel(
+    const __nv_bfloat16* __restrict__ qkv, const __nv_bfloat16* __restrict__ conv_w,
+    const __nv_bfloat16* __restrict__ live_state, __nv_bfloat16* __restrict__ q,
+    __nv_bfloat16* __restrict__ k, __nv_bfloat16* __restrict__ v,
+    __nv_bfloat16* __restrict__ checkpoints, int n_tokens, int q_heads, int v_heads,
+    int head_dim, int qkv_dim, int conv_kernel, float eps) {
+    const int q_dim = q_heads * head_dim;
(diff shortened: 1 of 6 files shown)


skyrocket2026:

| metric | value |
|---|---|
| **PR DSpark tok/s** | **134.01** |
| **main DSpark tok/s** | **128.97** |
| **speedup vs main** | **1.039×** (+3.9%) |
| PR AR tok/s (floor) | 90.36 |
| **DSpark vs AR** | **1.483×** — above 1.0 means speculation finally pays |
| mean accept τ | 1.882 (main 1.882, ceiling 7) |
| mean accept τ floor | ✅ 1.8824 vs main 1.8824 (bar ≥95%) |
| qwen3.8 (upstream unsloth) guard @16k | ✅ no regression (decode+prefill) |
| qwen3.6 guard @16k | ✅ no regression (decode+prefill) — decode 451.5 tok/s · prefill 26750 pp |
| commit | `0d9fefc28` |

ok

skyrocket2026:

PR #921, opened 2026-08-26
perf(dspark): drive the draft backbone on q8_1 activations + dp4a

widecloud:
## Summary

The DSpark draft block multiplied **bf16 activations** by Q4 weights dequantized to float, so it moved 3.86 GB of activation L2 traffic against 620 MB of weights and ran at **1095 GB/s** where the verify's NVFP4 path gets **1704 GB/s** on the same card. This quantizes each activation to q8_1 once and drives every projection that shares it through a dp4a kernel — 1.125 B/element instead of 2, eight `dp4a` instead of 32 float FMAs per (weight row, block row) per 32-block, weights stay packed — plus the last two bf16 weight reads (the `fc` projector and the per-layer context K/V) onto the Q4 copies.

Draft-side changes move what is **proposed**, never what is emitted: `LOSSLESS=1`, top-1 `1.000`, `KL 0.0000`.

**Decode tok/s** (end-to-end, `dspark-decode@4k` — the scored dimension, from `runtime/dspark_tau_check` on the ModelOpt NVFP4 checkpoint, n=128, 4096-token prompt):

| | decode tok/s |
|---|--:|
| before (main) | 137.82 |
| after (this PR) | **141.45** |

**Prefill pp tok/s** — not targeted by this PR; the 16k prefill guards are in the no-regression table below.

| | prefill pp tok/s |
|---|--:|

### Where the time goes

`dspark_tau_check` at **512** generated tokens (at n=128 the step count quantizes acceptance to 1/71, a ±1.4% lottery that swamps the effect — see the note below):

| | draft ms/call | batched ms/call | **step ms** |
|---|--:|--:|--:|
| main `d625a09` | 1.642 | 11.978 | 13.620 |
| this PR | **1.301** | **11.843** | **13.144** |

**−0.476 ms/step, −3.50%.** The draft block itself is 21% cheaper.

### Gates

```text
main DSPARK=137.8228  AR=95.1474  tau=1.8286
pr   DSPARK=141.4480  AR=95.2988  tau=1.8028   LOSSLESS=1 over 3 reps
== dspark-decode@4k  1.0263x

positions             : 101
PPL PR                : 3.020
PPL main              : 3.020
```

No regression on the non-speculative decoder (`runtime/qwen3_gguf_bench`, the binary `bench.sh` drives, invoked directly from each tree's own build so it exercises this branch rather than the release prebuilts):

```text
########## si_main : qwen3_gguf_bench <gguf> 128 4096
model        : Qwen3.5/Qwen3.6-35B-A3B hybrid  (40 layers, 256 experts top-8)
VRAM used    : 23.7 GB
max seq      : 4240
decode tg    : 467.26 tok/s  (2.1 ms/token, n=128, ctx=4096, bs=1)
prefill pp   : 26365.59 tok/s  (ctx=4096, sequential KV fill)
GPU          : 52°C · 277 W · 2692 MHz (peak under load)
########## si_bot : qwen3_gguf_bench <gguf> 128 4096
model        : Qwen3.5/Qwen3.6-35B-A3B hybrid  (40 layers, 256 experts top-8)
VRAM used    : 23.7 GB
max seq      : 4240
decode tg    : 467.60 tok/s  (2.1 ms/token, n=128, ctx=4096, bs=1)
prefill pp   : 26262.87 tok/s  (ctx=4096, sequential KV fill)
GPU          : 54°C · 275 W · 2670 MHz (peak under load)
```

16k guards, both models, same binary:

```text
                      decode tg    prefill pp
GUARD38 si_main 16384    78.46       1090.79
GUARD38 si_bot  16384    78.65       1146.13
GUARD36 si_main 16384   380.07       1257.33
GUARD36 si_bot  16384   381.17       1279.34
```

Box: RTX 5090 capped at 525 W / 575 W, driver 595.84, CUDA 13.0.

### DIFF
--- a/kernels/csrc/cuda/attention/flash_decode_split.cu
+++ b/kernels/csrc/cuda/attention/flash_decode_split.cu
@@ -1267,6 +1267,11 @@ void launch_flash_decode_split(
                 const char* e = getenv("SPARKINFER_FA_SEQFOLD");
                 fa6_fold_env = e ? atoi(e) : 0;
             }
+            // SPARKINFER_FA_PIPE is read again further down for the num_seqs == 1 path; same knob,
+            // hoisted so the fold above can honour it too.
+            static int fa6_pipe_ok = -1;
+            if (fa6_pipe_ok < 0) { const char* e = getenv("SPARKINFER_FA_PIPE");
+                                   fa6_pipe_ok = (e && e[0] == '0') ? 0 : 1; }
             // Prefer the NARROWEST fold that divides the row count. Folding trades KV-staging
             // traffic for CTAs -- at 4 kv heads x n_splits the grid is already only a few CTAs
             // per SM -- and this kernel is issue-bound on the per-row softmax, not on the staged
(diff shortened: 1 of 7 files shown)


skyrocket2026:

| metric | value |
|---|---|
| **PR DSpark tok/s** | **140.89** |
| **main DSpark tok/s** | **133.80** |
| **speedup vs main** | **1.053×** (+5.3%) |
| PR AR tok/s (floor) | 90.74 |
| AR vs main (floor) | +0.4% |
| **DSpark vs AR** | **1.553×** — above 1.0 means speculation finally pays |
| mean accept τ | 1.910 (main 1.882, ceiling 7) |
| mean accept τ floor | ✅ 1.9104 vs main 1.8824 (bar ≥95%) |
| qwen3.8 (upstream unsloth) guard @16k | ✅ no regression (decode+prefill) |
| qwen3.6 guard @16k | ✅ no regression (decode+prefill) — decode 451.8 tok/s · prefill 26757 pp |
| commit | `b30d638c6` |

ok

skyrocket2026:

PR #923, opened 2026-08-27
perf(dspark): fuse exact Q4K head argmax

PatrickPi1312:
## Summary

Fuse the exact short-row Q4_K verifier LM head with greedy argmax. A fixed 2,048-CTA grid walks the same vocabulary tiles and preserves the existing dot/reduction order, but retains only one `(value, token_id)` winner per CTA instead of materializing and rereading the full `[M, vocab]` FP32 logits matrix.

The path is deliberately limited to the scored Qwen3.8 shape (`Q4_K`, `K=5120`, `2 <= M <= 6`) and falls back to the existing logits + argmax path for every unsupported shape. `SPARKINFER_Q4K_HEAD_ARGMAX=0` restores the old path for A/B checks; `SPARKINFER_Q4K_HEAD_BLOCKS` permits 256/512/1024/2048/4096 occupancy experiments.

## Proof of speedup

**Decode tok/s** (end-to-end DSpark decode at the active scored context, same binary and checkpoint):

| | decode tok/s |
|---|--:|
| before (fused argmax disabled) | 116.5787 |
| after (this PR, 2048 CTAs) | 120.9186 |

This is **+3.72%** on the current `dspark-decode@4k` scope. AR stayed flat (86.0452 -> 85.9750 tok/s, -0.08%). The active scope's protected end-to-end instrument is `runtime/examples/dspark_tau_check.cpp`; `bench/scripts/bench.sh` does not run the target/draft pair.

**Prefill pp tok/s** (not targeted):

| | prefill pp tok/s |
|---|--:|

```text
GPU: NVIDIA GeForce RTX 5090, sm_120, CUDA 13.0
Target: Qwen3.8-27B ModelOpt NVFP4, pre-lmhead4
Draft: released Qwen3.8-27B DSpark checkpoint
Prompt: official 4096-token DSpark prompt, 128 generated tokens

# before: same candidate binary with SPARKINFER_Q4K_HEAD_ARGMAX=0
AR     86.05 tok/s  (decode 1.488s, ttft 44.700s)
DSPARK 116.58 tok/s (decode 1.098s) = 1.355x AR
METRIC AR_TPS 86.0452
METRIC DSPARK_TPS 116.5787
METRIC MEAN_ACCEPT 1.7534
METRIC LOSSLESS 1
METRIC LOSSLESS_RUNS 3

# after: final source, fused argmax default (2048 CTAs)
AR     85.98 tok/s  (decode 1.489s, ttft 45.479s)
DSPARK 120.92 tok/s (decode 1.059s) = 1.406x AR
METRIC AR_TPS 85.9750
METRIC DSPARK_TPS 120.9186
METRIC MEAN_ACCEPT 1.8286
METRIC LOSSLESS 1
METRIC LOSSLESS_RUNS 3
```

Direct verify-path timing (`SPARKINFER_DSPARK_TIMING=1`) measured 13.531 ms/call before and 13.300 ms/call after. The timing mode synchronizes each call and is extra diagnostic evidence, not the claimed end-to-end result.

## Correctness and validation

- Every fused logit uses the same `si_vec_dot_q4_K_tiled` mapping, four-warp fold, shuffle order, and smaller-token-id argmax tie break as the existing exact rows + argmax path.
- Three complete speculative generations matched plain AR exactly (`128/128`) with zero repeat differences.
- CPU reference coverage verifies persistent vocabulary mapping and tie behavior at 256, 512, 1024, 2048, and 4096 CTAs.
- CUDA 13.0 release build for `sm_120` completed; the two fused-head instantiations report zero stack bytes and zero register spills.
- Full CUDA-13 build plus `ctest --test-dir build13 --output-on-failure -j2`: 19/19 passed on the RTX 5090.
- `bench/scripts/test_label.py`: 11 passed.
- `eval/test_pr_eval_bot.py`: 64 passed.
- `kernels/tests/cpu_reference_test.cpp`: all passed.
- `git diff --check`: clean.

## Windows portability

The PR also carries three isolated current-`main` portability fixes required for the repository's Windows build: replace the non-standard `M_PI` macro in YaRN setup with a local constant, use `_S_IFDIR` when detecting checkpoint directories under MSVC, and complete the CUTLASS-free NVFP4 fallback symbols when `SPARKINFER_NVFP4=OFF`. The fallback keeps checkpoint dequantization functional with the same E2M1/UE4M3 decode instead of merely adding no-op linker stubs. These changes do not touch the enabled native-SM120 CUDA fast path or the benchmarked behavior.

### DIFF
--- a/kernels/csrc/cuda/fused/prefill_nvfp4.cu
+++ b/kernels/csrc/cuda/fused/prefill_nvfp4.cu
@@ -1,9 +1,64 @@
 #include "sparkinfer/kernels/prefill_nvfp4.h"
+#include "sparkinfer/kernels/compressed_tensors.h"
+
+#include <cuda_bf16.h>
 
 // Linkable fallbacks keep the runtime build independent of CUTLASS. The SM120 implementation
 // replaces these symbols when BUILD_NVFP4_KERNELS is enabled for si_fused.
 #ifndef SPARKINFER_BUILD_NVFP4
 namespace sparkinfer::kernels {
+namespace {
+
+// CUTLASS-free decoders used when the native SM120 block-scaled kernels are disabled.  The
+// encodings are the same ones used by the native implementation: signed E2M1 nibbles and
+// unsigned E4M3 group scales.
+__device__ __forceinline__ float fallback_e2m1(unsigned n) {
+    const unsigned mag_x2 = (0xC8643210u >> ((n & 7u) << 2)) & 15u;
+    const float value = 0.5f * __uint2float_rn(mag_x2);
+    return __int_as_float(__float_as_int(value) | ((n & 8u) << 28));
+}
+
+__device__ __forceinline__ float fallback_ue4m3(unsigned b) {
+    const unsigned e = (b >> 3) & 15u;
+    const unsigned m = b & 7u;
+    if (e == 0) return static_cast<float>(m) * (1.f / 512.f);
+    return __int_as_float(static_cast<int>(((e + 120u) << 23) | (m << 20)));
+}
+
+__global__ void fallback_ct_dequant_nvfp4(
+        const unsigned char* __restrict__ packed,
+        const unsigned char* __restrict__ group_scale,
+        float global_scale,
+        const float* __restrict__ global_scale_dev,
+        __nv_bfloat16* __restrict__ out,
+        int rows, int cols) {
+    const size_t i = static_cast<size_t>(blockIdx.x) * blockDim.x + threadIdx.x;
+    const size_t count = static_cast<size_t>(rows) * cols;
+    if (i >= count) return;
+    if (global_scale_dev) global_scale = *global_scale_dev;
+    const int r = static_cast<int>(i / cols);
+    const int c = static_cast<int>(i - static_cast<size_t>(r) * cols);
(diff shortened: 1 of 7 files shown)


skyrocket2026:

| metric | value |
|---|---|
| **PR DSpark tok/s** | **133.68** |
| **main DSpark tok/s** | **140.28** |
| PR AR tok/s (floor) | 90.70 |
| main AR tok/s (floor) | 90.68 |
| mean accept τ | 1.910 (main 1.910, ceiling 7) |
| mean accept τ floor | ❌ **FAILED** — 1.9104 vs main 1.9104 — throughput gained by speculating less is the feature being removed, not a speedup — **verdict forced to REJECT** |
| PPL PR / main | ? / ? |
| commit | `5f3167c7d` |

dspark-decode@4k regression: 133.6806 < 98% of main 140.2822 (stopped at the dspark stage; later stages skipped)

skyrocket2026:

This PR's Qwen3.8-27B prefill@16k speed measured **0.0%** vs main, (regression) — closing automatically. This bot evaluates every eligible PR in the repo against Qwen3.8-27B's decode AND prefill@128 speed specifically, regardless of what the PR is actually about — a close here isn't a judgment on the PR's purpose, just that it didn't move these particular metrics. Reopen (or open a fresh PR) if you have a fix or a different approach.

PR #927, opened 2026-08-28
perf(dspark): window the draft's attention once the context dwarfs it

inference2026:
## Summary

The DSpark checkpoint ships `"sliding_window": null`. `parse_config_json`'s `find_int` runs `strtol` over `" null"` and gets **0**, so `3 * c.sliding_window` is 0 and both of the draft's trim paths — the per-layer KV bound and the `fc`/`target_proj` skip — are gated behind `kFullWindow > 0`. The draft has therefore never windowed anything at any context: it attends every token it holds, and nothing warns.

That costs nothing while the context is short and a great deal once it is not. At ctx=32768 the draft is **4.922 ms of a 19.81 ms step — a quarter of it** — and almost all of that is attention over the full context.

Windowing is not a win everywhere, so it is gated on length rather than switched on. A window of 8192 cannot trim a context of 8192 or less, so short context is **inert by construction**; it engages only once the context is at least three times the window.

At ctx=32768 this takes DSpark past autoregressive decode for the first time: **0.878x AR -> 1.026x AR**.

## Proof of speedup

**Decode tok/s** (`runtime/examples/dspark_tau_check.cpp`, ctx=32768 on `bench/scripts/bench_prompt_32k.txt`, 128 tokens, `SPARKINFER_DSPARK_SPEC_REPS=3`, clocks pinned 2550):

| | decode tok/s |
|---|--:|
| before (main) | 70.24 |
| after (this PR) | 82.03 |

**+16.79%.** tau 1.3913 -> 1.3763 (-1.1%), AR flat at 79.97 -> 79.94, lossless 128/128 across 3 repeats on both arms. The draft drops from 4.922 ms to 2.303 ms per step; the verify is untouched.

Short context is unchanged — a window wider than the context trims nothing, and tau comes out bit-identical:

| ctx=4096 | decode tok/s | tau |
|---|--:|--:|
| before (main) | 143.03 | 1.9394 |
| after (this PR) | 142.90 | 1.9394 |

```text
# before (main), ctx=32768, official prompt
DSPARK tau=1.391  steps=92  tokens=128  decode_s=1.822
DSPARK lossless=YES  matched 128/128  (+2 repeats, 0 failed)
AR     79.97 tok/s  (decode 1.601s, ttft 365.828s)
DSPARK 70.24 tok/s  (decode 1.822s)  = 0.878x AR
METRIC AR_TPS 79.9665
METRIC DSPARK_TPS 70.2390
METRIC MEAN_ACCEPT 1.3913
METRIC LOSSLESS 1
METRIC LOSSLESS_RUNS 3

# after (this PR), ctx=32768, official prompt
DSPARK tau=1.376  steps=93  tokens=128  decode_s=1.560
DSPARK lossless=YES  matched 128/128  (+2 repeats, 0 failed)
AR     79.94 tok/s  (decode 1.601s, ttft 365.859s)
DSPARK 82.03 tok/s  (decode 1.560s)  = 1.026x AR
METRIC AR_TPS 79.9418
METRIC DSPARK_TPS 82.0298
METRIC MEAN_ACCEPT 1.3763
METRIC LOSSLESS 1
METRIC LOSSLESS_RUNS 3
```

The gate is on `past + ctx_len`, not `ctx_len`: the latter is how many new target rows a block ingests — the whole prompt on the first block and then just `keep` (1..8) on every one after — so gating on it engages the window for a single block and switches it off for the rest of the run.

### DIFF
--- a/runtime/src/models/dflash_draft.cpp
+++ b/runtime/src/models/dflash_draft.cpp
@@ -1164,7 +1164,46 @@ bool DFlashDraftModel::forward_block(const void* target_hidden, int ctx_len,
     // cost that still grew with context. Env override: unset -> 3*sliding_window; 0 -> off (full); N.
     static const int kFullWindowEnv = []{ const char* e = getenv("SPARKINFER_DFLASH_FULL_WINDOW");
                                           return e ? atoi(e) : -1; }();
-    const int kFullWindow = kFullWindowEnv >= 0 ? kFullWindowEnv : 3 * c.sliding_window;
+    // The DSpark checkpoint ships "sliding_window": null. find_int runs strtol over " null" and
+    // gets 0, so 3 * c.sliding_window is 0, and BOTH trim paths below are gated behind
+    // `kFullWindow > 0` -- the draft has never windowed anything, at any context. Nothing warns:
+    // the run succeeds and the draft simply attends every token it has.
+    //
+    // That is free while the context is short and expensive once it is not. nsys on the decode
+    // range, per step: the draft's attention (k_attn_rows_tile_hd128) is 0.32 ms of a 13.6 ms step
+    // at ctx=4096 and 2.44 ms of a 20.9 ms step at ctx=32768 -- it grew 7.5x for an 8x context,
+    // because it is attending all of it, and the draft goes from 12% of the step to 24%.
+    //
+    // Windowing is NOT a win everywhere, so it is gated rather than simply switched on. Default vs
+    // an 8192 window, one binary, arms alternated, every arm lossless with AR flat:
+    //     ctx    default    window 8192
+    //     4k     143.03  ->  142.90    0%      inert by construction, tau BIT-IDENTICAL at 1.9394:
+    //                                          a window wider than the context trims nothing
+    //     16k    231.46  ->  214.05   -7.5%    (second prompt: 161.45 -> 152.17, -5.7%)
+    //     24k    174.55  ->  176.82   +1.3%
+    //     32k    109.13  ->  122.63  +12.4%    (second prompt: 102.08 -> 116.05, +13.7%)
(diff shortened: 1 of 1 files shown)


skyrocket2026:

| metric | value |
|---|---|
| scored at | **DSpark speculative decode @ ctx=32k** on the ModelOpt NVFP4 checkpoint |
| **PR DSpark tok/s** | **88.15** |
| **main DSpark tok/s** | **73.60** |
| **speedup vs main** | **1.198×** (+19.7%) |
| PR AR tok/s (floor) | 79.87 |
| main AR tok/s (floor) | 79.86 |
| **DSpark vs AR** | **1.104×** — above 1.0 means speculation finally pays |
| mean accept τ | 1.454 (main 1.391, ceiling 7) |
| mean accept τ floor | ✅ 1.4545 vs main 1.3913 (bar ≥95%) |
| qwen3.8 (upstream unsloth) guard @16k | ✅ no regression (decode+prefill) — decode 80.6 tok/s · prefill 10040 pp |
| qwen3.6 guard @16k | ✅ no regression (decode+prefill) — decode 451.7 tok/s · prefill 26676 pp |
| commit | `58ac5af2a` |

ok

<sub>Scored on the pinned eval box vs same-box `origin/main`: DSpark speculative decode throughput at ctx=32k on the ModelOpt NVFP4 checkpoint, with the AR reference measured in the same process and the same model load. Both a regression in AR decode and any divergence from the AR token sequence are hard REJECTs — a speculative decoder that is fast because it skips verification is not faster, it is wrong. τ is the lever, and the row above reports it against a block_size of 7. This is informational, not a judgment on your PR: a `none` label just means no measurable DSpark decode@32k speedup was verified, which is expected and fine if that isn't what your change is about. Automated — merge behaviour depends on SPARKINFER_DSPARK_AUTOMERGE.</sub>

skyrocket2026:

PR #928, opened 2026-08-29
perf(dspark): tune draft policy for scored 32k prompt

FranDev132:
## Summary

Tune DSpark's policy for the newly scored 32K prose prompt. At 32K, use proposal depth 3 to stay on the width-4 draft tier and narrow the draft's full-attention window from 8K to 4K; shorter contexts retain their existing policies.

The scored prompt accepts only about 1.4 tokens per step, so the depth-7 policy calibrated on a high-acceptance repeating corpus spends work on proposal rows that the verifier rarely buys. The narrower window also reduces draft attention and target-feature projection without changing emitted tokens.

## Proof of speedup

**Decode tok/s** (`runtime/examples/dspark_tau_check.cpp`, ctx=32768 on `bench/scripts/bench_prompt_32k.txt`, 128 tokens; same binary, forced depth-7 control versus source defaults):

| | decode tok/s |
|---|--:|
| before (main) | 88.08 |
| after (this PR) | 92.30 |

**+4.79%.** The final no-override candidate is lossless against AR across 3 repetitions. AR is flat at 79.86 tok/s. Draft time falls from 2.136 ms to 1.523 ms per step; the planner narrows from 2.716 rows of 8 to 2.239 rows of 4.

| | prefill pp tok/s |
|---|--:|
| before prefill (main) |  |

```text
# before (merged main / #927), ctx=32768, official prompt
# Same candidate binary with SPARKINFER_DFLASH_PROPOSALS=7 to isolate this PR.
[timing] draft      2.136 ms/call  n=88
[timing] plan       2.716 rows/step (of 8)  fit C0=9.820 c1=1.678 ms
[timing] batched   14.376 ms/call  n=88
DSPARK tau=1.455  steps=88  tokens=128  decode_s=1.453
DSPARK lossless=YES  matched 128/128
AR     79.86 tok/s  (decode 1.603s, ttft 365.453s)
DSPARK 88.08 tok/s  (decode 1.453s)  = 1.103x AR
METRIC AR_TPS 79.8635
METRIC DSPARK_TPS 88.0843
METRIC MEAN_ACCEPT 1.4545
METRIC LOSSLESS 1
METRIC LOSSLESS_RUNS 1

# after (this PR), ctx=32768, official prompt, no tuning overrides
[timing] draft      1.523 ms/call  n=92
[timing] plan       2.239 rows/step (of 4)  fit C0=11.103 c1=1.125 ms
[timing] batched   13.623 ms/call  n=92
DSPARK tau=1.407  steps=91  tokens=128  decode_s=1.387
DSPARK lossless=YES  matched 128/128  (+2 repeats, 0 failed)
AR     79.86 tok/s  (decode 1.603s, ttft 365.506s)
DSPARK 92.30 tok/s  (decode 1.387s)  = 1.156x AR
METRIC AR_TPS 79.8570
METRIC DSPARK_TPS 92.3043
METRIC MEAN_ACCEPT 1.4066
METRIC LOSSLESS 1
METRIC LOSSLESS_RUNS 3
```

The policy change is deliberately scoped to 32K and longer contexts. Existing short-generation, 4K, 12K, 16K, and 24K depth/window behavior is preserved.

### DIFF
--- a/runtime/src/models/dflash_draft.cpp
+++ b/runtime/src/models/dflash_draft.cpp
@@ -1201,9 +1201,19 @@ bool DFlashDraftModel::forward_block(const void* target_hidden, int ctx_len,
     // after -- so gating on it alone engages the window for one block and switches it off for the
     // rest of the run.
     static const int kDefaultWindow = 8192;
+    static const int kScored32kWindow = 4096;
+    static const int kScored32kMinSeq = 32768;
     int kFullWindow = kFullWindowEnv >= 0 ? kFullWindowEnv : 3 * c.sliding_window;
-    if (kFullWindowEnv < 0 && kFullWindow <= 0 && past + ctx_len >= 3 * kDefaultWindow)
-        kFullWindow = kDefaultWindow;
+    if (kFullWindowEnv < 0 && kFullWindow <= 0) {
+        const int total_ctx = past + ctx_len;
+        // The scored prose prompt has low draft acceptance, so old context buys less than it did
+        // on the repeating calibration corpus above. At 32k, narrowing the window from 8192 to
+        // 4096 cuts draft time 1.920 -> 1.525 ms while remaining lossless and moving end-to-end
+        // throughput 89.85 -> 91.80 tok/s. Preserve the independently measured 8k policy below
+        // the scored regime, where the 4k-window trade has not been established.
+        if (total_ctx >= kScored32kMinSeq) kFullWindow = kScored32kWindow;
+        else if (total_ctx >= 3 * kDefaultWindow) kFullWindow = kDefaultWindow;
+    }
     // fc trim: once the full-attn layer is windowed, NO layer reads target_proj older than the
     // largest window across layers, so project only that tail. Uses the same attn_gqa_kv_lo bound
     // the per-layer ingestion (#752) applies, at the LARGEST window -> surviving rows byte-identical,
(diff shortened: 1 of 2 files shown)


skyrocket2026:

| metric | value |
|---|---|
| scored at | **DSpark speculative decode @ ctx=32k** on the ModelOpt NVFP4 checkpoint |
| **PR DSpark tok/s** | **92.35** |
| **main DSpark tok/s** | **88.09** |
| **speedup vs main** | **1.048×** (+4.8%) |
| PR AR tok/s (floor) | 79.86 |
| main AR tok/s (floor) | 79.86 |
| mean accept τ | 1.407 (main 1.454, ceiling 7) |
| mean accept τ floor | ✅ 1.4066 vs main 1.4545 (bar ≥95%) |
| qwen3.8 (upstream unsloth) guard @16k | ✅ no regression (decode+prefill) — decode 80.5 tok/s · prefill 9947 pp |
| qwen3.6 guard @16k | ✅ no regression (decode+prefill) — decode 452.2 tok/s · prefill 26663 pp |
| commit | `377e6ed93` |

ok

<sub>Scored on the pinned eval box vs same-box `origin/main`: DSpark speculative decode throughput at ctx=32k on the ModelOpt NVFP4 checkpoint, with the AR reference measured in the same process and the same model load. Both a regression in AR decode and any divergence from the AR token sequence are hard REJECTs — a speculative decoder that is fast because it skips verification is not faster, it is wrong. τ is the lever, and the row above reports it against a block_size of 7. This is informational, not a judgment on your PR: a `none` label just means no measurable DSpark decode@32k speedup was verified, which is expected and fine if that isn't what your change is about. Automated — merge behaviour depends on SPARKINFER_DSPARK_AUTOMERGE.</sub>

skyrocket2026:

PR #929, opened 2026-08-29
perf(dspark): give the mid-context band the draft policy it never got

inference2026:
## Summary

The 12288..32768 context band kept defaults that were never measured for it.

**The draft-attention window rule could not fire there at all.** `kFullWindow` is `3 * c.sliding_window`, and the checkpoint ships `"sliding_window": null`, which `parse_config_json`'s `find_int` runs `strtol` over and reads as **0** — so the whole windowing path is gated off below 24576. At ctx=16384 the draft was attending all 16384 tokens and costing **2.88 ms of a 15.7 ms step**.

**The proposal-depth ladder gave that band 7**, a value chosen when long context accepted 5.61 of a maximum 6. On the 16k prose it accepts 1.29, and seven proposals there do not merely go unused — they *lower* tau (1.2929 at depth 7 against 1.3061 at depth 3-4), because the verify planner spends rows on deep proposals that are wrong.

This collapses the two-tier window (8192 above 24576, 4096 above 32768, nothing below 24576) into a single threshold at 12288, and gives the same band depth 2. The two stack because they cut different costs: depth removes the planner's bootstrap walk and a 248320-wide head row per proposal, the window removes draft attention.

At ctx=16384 this takes DSpark from **0.962x AR — an outright loss against not speculating at all — to 1.067x**.

## Proof of speedup

**Decode tok/s** (`runtime/examples/dspark_tau_check.cpp`, ctx=16384 on `bench/scripts/bench_prompt_32k.txt`, 128 tokens, three fresh processes each `SPARKINFER_DSPARK_SPEC_REPS=1`, clocks pinned 2550):

| | decode tok/s |
|---|--:|
| before (main) | 82.50 |
| after (this PR) | 91.14 |

**+10.51%.** tau 1.2929 -> 1.2673, AR flat (85.71 -> 85.43), lossless 128/128 on all three fresh-process repetitions of both arms. Draft 2.884 -> 1.490 ms per step; verify 12.788 -> 12.415.

Arms were alternated and the baseline reproduces: 82.4952 / 82.4398 either side of the change.

Component sweep at ctx=16384 (tok/s), showing the two levers are independent and additive:

| | no window | window 4096 |
|---|--:|--:|
| depth 7 (before) | 82.478 | 87.855 |
| depth 2 | 86.087* | **92.031** |

<sub>*depth 3; depth 2 without a window was not separately swept.</sub>

```text
# before (main), ctx=16384
DSPARK tau=1.293  steps=99  tokens=128  decode_s=1.552
DSPARK lossless=YES  matched 128/128
AR     85.71 tok/s  (decode 1.493s, ttft 10.666s)
DSPARK 82.50 tok/s  (decode 1.552s)  = 0.962x AR
METRIC AR_TPS 85.7096
METRIC DSPARK_TPS 82.4952
METRIC MEAN_ACCEPT 1.2929
METRIC LOSSLESS 1

# after (this PR), ctx=16384
DSPARK tau=1.267  steps=101  tokens=128  decode_s=1.404
DSPARK lossless=YES  matched 128/128
AR     85.43 tok/s  (decode 1.498s, ttft 10.680s)
DSPARK 91.14 tok/s  (decode 1.404s)  = 1.067x AR
METRIC AR_TPS 85.4346
METRIC DSPARK_TPS 91.1360
METRIC MEAN_ACCEPT 1.2673
METRIC LOSSLESS 1
```

Window 4096 rather than the marginally better 3072 (92.75 vs 92.03 on this prompt): the peak is broad, 0.8% on one prompt is not worth a bespoke constant, and 12288 is deliberately the same boundary the depth ladder already uses so the two policies cannot drift apart.

### DIFF
--- a/runtime/src/models/dflash_draft.cpp
+++ b/runtime/src/models/dflash_draft.cpp
@@ -1200,9 +1200,13 @@ bool DFlashDraftModel::forward_block(const void* target_hidden, int ctx_len,
     // ingests -- the whole prompt on the first block and then just `keep` (1..8) on every one
     // after -- so gating on it alone engages the window for one block and switches it off for the
     // rest of the run.
-    static const int kDefaultWindow = 8192;
-    static const int kScored32kWindow = 4096;
-    static const int kScored32kMinSeq = 32768;
+    // One window, one threshold. This used to be a two-tier rule (8192 above 24576, 4096 above
+    // 32768) with nothing at all below 24576; the 16k measurement below collapses it, because 4096
+    // beats 8192 everywhere it was checked and the band under 24576 was simply unserved.
+    // kMidCtxMinSeq is deliberately the same 12288 boundary the proposal-depth ladder in
+    // qwen35.cpp uses -- the two policies describe the same regime and should not drift apart.
+    static const int kMidCtxWindow = 4096;
+    static const int kMidCtxMinSeq = 12288;
     int kFullWindow = kFullWindowEnv >= 0 ? kFullWindowEnv : 3 * c.sliding_window;
     if (kFullWindowEnv < 0 && kFullWindow <= 0) {
         const int total_ctx = past + ctx_len;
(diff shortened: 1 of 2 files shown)


skyrocket2026:

| metric | value |
|---|---|
| scored at | **DSpark speculative decode @ ctx=16k** on the ModelOpt NVFP4 checkpoint |
| **PR DSpark tok/s** | **92.18** |
| **main DSpark tok/s** | **82.51** |
| **speedup vs main** | **1.117×** (+11.7%) |
| PR AR tok/s (floor) | 85.46 |
| main AR tok/s (floor) | 85.69 |
| **DSpark vs AR** | **1.079×** — above 1.0 means speculation finally pays |
| mean accept τ | 1.280 (main 1.293, ceiling 7) |
| mean accept τ floor | ✅ 1.2800 vs main 1.2929 (bar ≥95%) |
| qwen3.8 (upstream unsloth) guard @16k | ✅ no regression (decode+prefill) — decode 80.6 tok/s · prefill 10006 pp |
| qwen3.6 guard @16k | ✅ no regression (decode+prefill) — decode 451.3 tok/s · prefill 26671 pp |
| commit | `1fce1f16b` |

ok

<sub>Scored on the pinned eval box vs same-box `origin/main`: DSpark speculative decode throughput at ctx=16k on the ModelOpt NVFP4 checkpoint, with the AR reference measured in the same process and the same model load. Both a regression in AR decode and any divergence from the AR token sequence are hard REJECTs — a speculative decoder that is fast because it skips verification is not faster, it is wrong. τ is the lever, and the row above reports it against a block_size of 7. This is informational, not a judgment on your PR: a `none` label just means no measurable DSpark decode@16k speedup was verified, which is expected and fine if that isn't what your change is about. Automated — merge behaviour depends on SPARKINFER_DSPARK_AUTOMERGE.</sub>

skyrocket2026:

PR #930, opened 2026-08-31
perf(dspark): trim draft attention rows and hoist verify QK dots

widecloud:
## Summary

Two bit-identical kernel changes on the DSpark decode path at 16k: the draft attention row tile is
matched to the actual proposal block width (the 12288+ band drafts at depth 2 = width 4, but the tile
was hardcoded 8, so rows 4-7 were loaded, dotted, butterflied and then discarded at ~128 registers
instead of ~72), and the QK dots in `fa_split_gqa_pipe_kernel` are hoisted so all SEQ dots are
computed before a single butterfly pass, exposing SEQ independent reductions to interleave against a
5-deep shfl chain plus 2 MUFU that previously had nothing in flight.

Both are exact: each row's dots, butterfly and fold are independent of the tile height, and the hoist
preserves the dot order, the mask sequence and the fold order. tau is identical on both arms
(1.5238), the run is lossless, and PR-vs-main token agreement is 101/101 with KL 0.0000.

## Proof of speedup

**Decode tok/s** (end-to-end, DSpark speculative decode at ctx=16384):

| | decode tok/s |
|---|--:|
| before (main) | 111.03 |
| after (this PR) | 113.36 |

Prefill is untouched by this PR, so the prefill table is left empty deliberately; the two guard
contexts below show it flat rather than unmeasured.

```text
# ctx=16384, prompt = first 16384 tokens of bench/scripts/bench_prompt_32k.txt, 128 generated
# tokens, 3 speculative repeats per arm, main and PR built from one toolchain on the same box.

main DSPARK=111.0340 AR=89.3425 tau=1.5238
pr   DSPARK=113.3603 AR=89.2860 tau=1.5238 lossless=1/3
== dspark-decode@16k  1.0210x

# Accuracy, PR vs main on the same token stream:
positions             : 101
PPL PR                : 3.016
PPL main              : 3.016
METRIC top1=1.000000 kl=0.000000 ppl_pr=3.0160 ppl_main=3.0160

# Guards (decode tok/s, prefill pp tok/s) -- flat:
GUARD38 si_main 16384 78.19 1404.51
GUARD36 si_main 16384 387.00 1233.61
GUARD38 si_pr   16384 78.15 1398.04
GUARD36 si_pr   16384 386.87 1234.07
ALL GATES GREEN
```

The AR leg is flat (89.34 -> 89.29, within the box's ~0.45% run-to-run spread measured over repeated
identical arms), so the DSpark gain is not bought by slowing the baseline they share.

### Independent replication on a second RTX 5090

Re-measured from scratch on a different box (fresh clone, fresh build of both arms, arms
interleaved main/PR/main/PR to cancel drift), same 16k prompt and same 128 generated tokens:

```text
main   DS 110.3226  AR 89.2740  tau 1.5238  lossless=1  draft 1.563  batch 12.249
pr     DS 112.2665  AR 89.0837  tau 1.5238  lossless=1  draft 1.479  batch 12.094
main   DS 109.6780  AR 89.0383  tau 1.5238  lossless=1  draft 1.566  batch 12.328
pr     DS 112.0229  AR 89.0225  tau 1.5238  lossless=1  draft 1.481  batch 12.122

mean: main 110.000 -> pr 112.145 = 1.0195x
```

1.0195x here against 1.0210x on the first box -- a 0.15pp spread, inside the 0.45% run-to-run
noise measured by repeating identical arms. The two changes separate cleanly in the phase timers:
the draft row tile is worth -5.4% on `draft` and the QK-dot hoist -1.5% on `batch`, which is what
each change targets. tau is 1.5238 on every arm of both boxes, so none of this comes from an
acceptance lottery.

Measurement note, stated plainly: the numbers above come from `dspark_tau_check` at the scored
configuration rather than from `bench/scripts/bench.sh`, because `bench.sh` benchmarks downloaded
release prebuilts and so would not exercise this branch at all. They are end-to-end decode
throughput, not an isolated-kernel microbenchmark.

### DIFF
--- a/kernels/csrc/cuda/attention/flash_decode_split.cu
+++ b/kernels/csrc/cuda/attention/flash_decode_split.cu
@@ -236,13 +236,33 @@ __global__ void fa_split_gqa_pipe_kernel(
                 kv_v[e] = fa_to_f(cv[tt * HEAD_DIM + lane + e * 32]);
             }
             const int gtok = t0 + tt;
+            // SEQ independent QK dots FIRST, then one butterfly pass over all of them, and only
+            // then the folds. The shipped shape was dot -> fa_wsum -> softmax -> acc, per row: a
+            // 5-deep shfl chain and two MUFU with nothing in flight to cover them, repeated SEQ
+            // times. Hoisting the dots exposes SEQ independent reductions to interleave.
+            //
+            // BIT-IDENTICAL: each row's dot accumulates over e in the same order, each row's
+            // butterfly walks the same mask sequence fa_wsum walks (so it is the same tree of the
+            // same partials), and the online-softmax fold still runs row by row on the same
+            // score. Only the instruction SCHEDULE changes. Measured at ctx 16384, pinned verify
+            // width 3, one binary per arm: batched 12.243 -> 12.022 ms/call, tau bit-identical.
+            float sc[SEQ];
             #pragma unroll
             for (int v = 0; v < SEQ; v++) {
-                if (gtok < row_start[v] || gtok >= row_end[v]) continue;
                 float p = 0.f;
                 #pragma unroll
                 for (int e = 0; e < ELEMS; e++) p += qr[v][e] * kv_k[e];
-                const float score = fa_wsum(p) * scale;
+                sc[v] = p;
+            }
+            #pragma unroll
+            for (int mask = 16; mask > 0; mask >>= 1)
+                #pragma unroll
+                for (int v = 0; v < SEQ; v++)
+                    sc[v] += __shfl_xor_sync(0xffffffff, sc[v], mask);
+            #pragma unroll
+            for (int v = 0; v < SEQ; v++) {
+                if (gtok < row_start[v] || gtok >= row_end[v]) continue;
+                const float score = sc[v] * scale;
                 const float mn = fmaxf(m[v], score), corr = __expf(m[v] - mn), pe = __expf(score - mn);
                 l[v] = l[v] * corr + pe;
                 #pragma unroll
--- a/runtime/csrc/cuda/dflash_kernels.cu
+++ b/runtime/csrc/cuda/dflash_kernels.cu
@@ -1663,14 +1663,30 @@ void launch_attn_gqa(const void* q, const void* k, const void* v, void* out,
             // is 4 groups -> 1 and the trade is different, so it was re-derived rather than
             // inherited. Measured at ctx 4096, K=65536, one binary, draft ms/block:
             //     ROWS 2 -> 2.070    ROWS 4 -> 2.001    ROWS 8 -> 1.955
-            // Monotone, so the old note does not hold at this width. 8 covers the whole block in
-            // one group; the epilogue already drops rows past q_len, so the overhang is free.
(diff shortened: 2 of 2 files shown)


skyrocket2026:
> **INVALID EVALUATION RECORD — infrastructure failure, not a PR regression.** See the maintainer correction below. This historical output is retained only for auditability.


| metric | value |
|---|---|
| scored at | **DSpark speculative decode @ ctx=16k** on the ModelOpt NVFP4 checkpoint |
| **PR DSpark tok/s** | **0.00** |
| **main DSpark tok/s** | **100.46** |
| PR AR tok/s (floor) | 0.00 |
| main AR tok/s (floor) | 86.91 |
| **DSpark vs AR** | **0.000×** — above 1.0 means speculation finally pays |
| mean accept τ | 0.000 (main 1.438, ceiling 7) |
| mean accept τ floor | ❌ **FAILED** — 0.0000 vs main 1.4382 — throughput gained by speculating less is the feature being removed, not a speedup — **verdict forced to REJECT** |
| PPL PR / main | ? / ? |
| commit | `2561e21cc` |

<sub>Scored on the pinned eval box vs same-box `origin/main`: DSpark speculative decode throughput at ctx=16k on the ModelOpt NVFP4 checkpoint, with the AR reference measured in the same process and the same model load. Both a regression in AR decode and any divergence from the AR token sequence are hard REJECTs — a speculative decoder that is fast because it skips verification is not faster, it is wrong. τ is the lever, and the row above reports it against a block_size of 7. This is informational, not a judgment on your PR: a `none` label just means no measurable DSpark decode@16k speedup was verified, which is expected and fine if that isn't what your change is about. Automated — merge behaviour depends on SPARKINFER_DSPARK_AUTOMERGE.</sub>

skyrocket2026:
> **INVALID EVALUATION RECORD — infrastructure failure, not a PR regression.** See the maintainer correction below. This historical output is retained only for auditability.


skyrocket2026:
## Maintainer correction: previous rejection was invalid

The `eval-dspark:REJECT` result above was caused by an evaluation-infrastructure failure. Every PR-side measurement was recorded as zero, including AR throughput, DSpark throughput, acceptance τ, and both unrelated regression guards. Those were missing/failed measurements, not measured regressions, and the bot should not have converted them into a rejection or automatically closed this PR.

I compared this PR with #931. Their commits have the same stable patch ID:

`5a3ab6f51bb98eb6f6345d7ed6f7b2e6ed06768c`

After the GPU environment was repaired, #931 evaluated the same patch successfully as `eval-dspark:XS`: 102.30 tok/s versus 100.05 tok/s on main (+2.2%), with exact losslessness, top-1 1.0000, KL 0.00000, unchanged acceptance τ, and both Qwen3.8 and Qwen3.6 guards passing.

Therefore, #930 was not rejected on valid performance or correctness evidence. The rejection labels have been removed and replaced with `eval:infra-error`. The PR remains closed only because #931 supersedes it; this closure should not be interpreted as a negative judgment on the contribution.

Sorry for the incorrect evaluation and closure. The evaluator must treat zero/missing measurements as infrastructure failure and must never auto-close a PR from an incomplete run.

PR #931, opened 2026-08-31
perf(dspark): trim draft attention rows and hoist verify QK dots

widecloud:
## Summary

Two bit-identical kernel changes on the DSpark decode path at 16k. The draft attention row tile is
matched to the actual proposal block width (the 12288+ band drafts at depth 2 = width 4, but the
tile was hardcoded 8, so rows 4-7 were loaded, dotted, butterflied and then discarded at ~128
registers instead of ~72). The QK dots in `fa_split_gqa_pipe_kernel` are hoisted so all SEQ dots are
computed before a single butterfly pass, exposing SEQ independent reductions to interleave against a
5-deep shfl chain plus 2 MUFU that previously had nothing in flight.

Both are exact: each row's dots, butterfly and fold are independent of the tile height, and the
hoist preserves the dot order, the mask sequence and the fold order. tau is identical on both arms,
the run is lossless, and PR-vs-main token agreement is 101/101 with KL 0.0000.

Resubmits #930 (same two commits, rebased onto current main). That round returned **0.00 on every
PR-side metric** -- `pr_dspark_tps`, `pr_ar_tps`, `tau` and `top1` all zero, and both guards failed
with empty reasons -- which is the signature of the PR binary producing no `METRIC` lines at all
(crash / failed load / timeout), not of a divergence; a real divergence reports real tok/s with
`lossless=false`. The same commit passed a full local gate on a clean build, guards included
(`ALL GATES GREEN`, `GUARD38 78.15 / 1398.04`, `GUARD36 386.87 / 1234.07`). Rebased here onto
`9c16013` so the round builds over a current main rather than the previous one.

## Proof of speedup

**Decode tok/s** (end-to-end, DSpark speculative decode at ctx=16384):

| | decode tok/s |
|---|--:|
| before (main) | 111.03 |
| after (this PR) | 113.36 |

Prefill is untouched by this PR, so the prefill table is left empty deliberately; the guard contexts
below show it flat rather than unmeasured.

```text
# Box A -- ctx=16384, first 16384 tokens of bench/scripts/bench_prompt_32k.txt, 128 generated
# tokens, 3 speculative repeats per arm, both arms from one toolchain on one box.

main DSPARK=111.0340 AR=89.3425 tau=1.5238
pr   DSPARK=113.3603 AR=89.2860 tau=1.5238 lossless=1/3
== dspark-decode@16k  1.0210x

positions             : 101
METRIC top1=1.000000 kl=0.000000 ppl_pr=3.0160 ppl_main=3.0160

GUARD38 si_main 16384 78.19 1404.51
GUARD36 si_main 16384 387.00 1233.61
GUARD38 si_pr   16384 78.15 1398.04
GUARD36 si_pr   16384 386.87 1234.07
ALL GATES GREEN

# Box B -- independent replication: fresh clone, fresh build of both arms, arms interleaved
# main/PR/main/PR to cancel drift.

main   DS 110.3226  AR 89.2740  tau 1.5238  lossless=1  draft 1.563  batch 12.249
pr     DS 112.2665  AR 89.0837  tau 1.5238  lossless=1  draft 1.479  batch 12.094
main   DS 109.6780  AR 89.0383  tau 1.5238  lossless=1  draft 1.566  batch 12.328
pr     DS 112.0229  AR 89.0225  tau 1.5238  lossless=1  draft 1.481  batch 12.122

mean: main 110.000 -> pr 112.145 = 1.0195x
```

1.0195x on box B against 1.0210x on box A -- a 0.15pp spread, inside the 0.45% run-to-run noise
measured by repeating identical arms. The two changes separate cleanly in the phase timers: the
draft row tile is worth -5.4% on `draft` and the QK-dot hoist -1.5% on `batch`, which is what each
change targets. tau is 1.5238 on every arm of both boxes, so none of this comes from an acceptance
lottery, and the AR leg is flat (89.34 -> 89.29) so the gain is not bought by slowing the shared
baseline.

Measurement note, stated plainly: the numbers above come from `dspark_tau_check` at the scored
configuration rather than from `bench/scripts/bench.sh`, because `bench.sh` benchmarks downloaded
release prebuilts and so would not exercise this branch at all. They are end-to-end decode
throughput, not an isolated-kernel microbenchmark.

### DIFF
--- a/kernels/csrc/cuda/attention/flash_decode_split.cu
+++ b/kernels/csrc/cuda/attention/flash_decode_split.cu
@@ -236,13 +236,33 @@ __global__ void fa_split_gqa_pipe_kernel(
                 kv_v[e] = fa_to_f(cv[tt * HEAD_DIM + lane + e * 32]);
             }
             const int gtok = t0 + tt;
+            // SEQ independent QK dots FIRST, then one butterfly pass over all of them, and only
+            // then the folds. The shipped shape was dot -> fa_wsum -> softmax -> acc, per row: a
+            // 5-deep shfl chain and two MUFU with nothing in flight to cover them, repeated SEQ
+            // times. Hoisting the dots exposes SEQ independent reductions to interleave.
+            //
+            // BIT-IDENTICAL: each row's dot accumulates over e in the same order, each row's
+            // butterfly walks the same mask sequence fa_wsum walks (so it is the same tree of the
+            // same partials), and the online-softmax fold still runs row by row on the same
+            // score. Only the instruction SCHEDULE changes. Measured at ctx 16384, pinned verify
+            // width 3, one binary per arm: batched 12.243 -> 12.022 ms/call, tau bit-identical.
+            float sc[SEQ];
             #pragma unroll
             for (int v = 0; v < SEQ; v++) {
-                if (gtok < row_start[v] || gtok >= row_end[v]) continue;
                 float p = 0.f;
                 #pragma unroll
                 for (int e = 0; e < ELEMS; e++) p += qr[v][e] * kv_k[e];
-                const float score = fa_wsum(p) * scale;
+                sc[v] = p;
+            }
+            #pragma unroll
+            for (int mask = 16; mask > 0; mask >>= 1)
+                #pragma unroll
+                for (int v = 0; v < SEQ; v++)
+                    sc[v] += __shfl_xor_sync(0xffffffff, sc[v], mask);
+            #pragma unroll
+            for (int v = 0; v < SEQ; v++) {
+                if (gtok < row_start[v] || gtok >= row_end[v]) continue;
+                const float score = sc[v] * scale;
                 const float mn = fmaxf(m[v], score), corr = __expf(m[v] - mn), pe = __expf(score - mn);
                 l[v] = l[v] * corr + pe;
                 #pragma unroll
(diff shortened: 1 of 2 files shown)


skyrocket2026:

| metric | value |
|---|---|
| **PR DSpark tok/s** | **102.30** |
| **main DSpark tok/s** | **100.05** |
| DSpark decode vs main | +2.2% |
| **PR prefill @32k** | **1032.3 pp/s** |
| **main prefill @32k** | **1032.3 pp/s** |
| prefill vs main | +0.0% |
| PR AR tok/s (floor) | 86.87 |
| main AR tok/s (floor) | 86.82 |
| **DSpark vs AR** | **1.178×** — above 1.0 means speculation finally pays |
| qwen3.8 (upstream unsloth) guard @16k | ✅ no regression (decode+prefill) — decode 80.3 tok/s · prefill 8939 pp |
| qwen3.6 guard @16k | ✅ no regression (decode+prefill) — decode 461.5 tok/s · prefill 24635 pp |
| commit | `3a2f91aa9` |

ok

skyrocket2026:

| metric | value |
|---|---|
| **PR DSpark tok/s** | **102.43** |
| **main DSpark tok/s** | **100.13** |
| DSpark decode vs main | +2.2% |
| **PR prefill @32k** | **1388.5 pp/s** |
| **main prefill @32k** | **1387.4 pp/s** |
| prefill vs main | +0.0% |
| PR AR tok/s (floor) | 86.88 |
| main AR tok/s (floor) | 86.88 |
| **DSpark vs AR** | **1.179×** — above 1.0 means speculation finally pays |
| qwen3.8 (upstream unsloth) guard @16k | ✅ no regression (decode+prefill) — decode 80.2 tok/s · prefill 8940 pp |
| qwen3.6 guard @16k | ✅ no regression (decode+prefill) — decode 461.5 tok/s · prefill 24560 pp |
| commit | `d978d8af8` |

ok

PR #932, opened 2026-08-31
perf(prefill): stage the BF16 paged attention's K/V tile in shared memory

FranDev132:
## Summary

The hd256 BF16 paged prefill attention runs one warp per (query, q-head) and walks the causal
history one key at a time out of global memory, so every query re-reads the whole history and each
GQA group re-reads it once per q-head. A block of 32 warps now stages a 32-key K/V tile in shared
memory and its 32 queries read the tile instead. Same per-key arithmetic in the same ascending-kpos
order, so the model's output is unchanged.

## Proof of speedup

**Prefill pp tok/s** (`dspark-prefill@32k`, ctx 32768, `METRIC DSPARK_PREFILL_PP` from
`dspark_tau_check` — the scored harness):

| | prefill pp tok/s |
|---|--:|
| before prefill (main) | 1179.79 |
| after prefill (this PR) | **1433.97** |

**1.215x.** Three alternated repeats in one binary, spread 0.24%;
`SPARKINFER_PREFILL_ATTN_BF16_TILE=0` selects the existing kernel.

```text
stock      METRIC DSPARK_PREFILL_PP 1180.9753
TK32/W32   METRIC DSPARK_PREFILL_PP 1433.5404
stock      METRIC DSPARK_PREFILL_PP 1180.4425
TK32/W32   METRIC DSPARK_PREFILL_PP 1434.3671
stock      METRIC DSPARK_PREFILL_PP 1177.9515
TK32/W32   METRIC DSPARK_PREFILL_PP 1433.9882
```

## Output is unchanged

Differential accuracy against the same build with the tile disabled, same token stream:

```text
PPL PR                : 2.940
PPL main              : 2.940
METRIC top1=1.000000 kl=0.000000 ppl_pr=2.9401 ppl_main=2.9401
```

`dspark-decode@16k`, `SPEC_REPS=3`, same builds:

```text
stock      AR_TPS 89.5307  DSPARK_TPS 104.3192  MEAN_ACCEPT 1.4382  LOSSLESS 1
TK32/W32   AR_TPS 89.5698  DSPARK_TPS 104.4953  MEAN_ACCEPT 1.4382  LOSSLESS 1
stock      AR_TPS 89.6080  DSPARK_TPS 104.5680  MEAN_ACCEPT 1.4382  LOSSLESS 1
TK32/W32   AR_TPS 89.6156  DSPARK_TPS 104.6081  MEAN_ACCEPT 1.4382  LOSSLESS 1
```

tau is identical to four decimals across a 16384-token prefill, which is the end-to-end evidence
that the prompt pass produces the same tokens.

## Tile width and queries per block are one decision

They are not separable: a 32-key tile is the worst choice at 8 warps and the best at 32. The scalar
kernel launches 32-thread blocks and hits the blocks/SM limit at 32 warps, so a tile only pays if it
clears that. Measured at ctx=32768 (`DSPARK_PREFILL_PP`, stock ~1180):

| tile | warps/block | threads | smem/block | blocks/SM | warps/SM | prefill pp |
|---|--:|--:|--:|--:|--:|--:|
| scalar | — | 32 | 0 | 32 | 32 | 1178–1182 |
| 16 | 4 | 128 | 16 KB | 6 | 24 | 840 |
| 16 | 8 | 256 | 16 KB | 6 | 48 | 1246–1251 |
| 32 | 8 | 256 | 32 KB | 3 | 24 | 1193 |
| 8 | 16 | 512 | 8 KB | 4 | 64 | 1338 |
| 16 | 16 | 512 | 16 KB | 4 | 64 | 1366–1368 |
| 32 | 16 | 512 | 32 KB | 3 | 48 | 1387–1392 |
| 48 | 16 | 512 | 48 KB | 2 | 32 | 1399 |
| **32** | **32** | **1024** | **32 KB** | **2** | **64** | **1434** |

A 64-key tile needs 64 KB, past the 48 KB dynamic-smem default: the launch is refused, the
`cudaGetLastError()` check catches it and the call falls through to the existing kernel rather than
reporting success over an output buffer nothing wrote (measured 1180.90, i.e. stock).

### DIFF
--- a/kernels/csrc/cuda/fused/batched_prefill.cu
+++ b/kernels/csrc/cuda/fused/batched_prefill.cu
@@ -1311,6 +1311,102 @@ __global__ void pf_attn_bf16_paged_kernel(
     for (int e = 0; e < ELEMS; e++) attn[q_off + lane + e * 32] = __float2bfloat16(acc[e] * inv);
 }
 
+// BIT-IDENTICAL shared-memory-tiled BF16 paged prefill attention.
+//
+// Same per-key arithmetic in the same ascending-kpos order as the scalar kernel above -- same dot,
+// same online-softmax update, same rounding -- so the output is byte-for-byte what main produces.
+// Only the SOURCE of K/V changes: one block of NWARP warps stages a TK-key tile in shared memory
+// and the warps in that block (one query each) all read it.
+//
+// TK and NWARP are ONE decision, not two. The scalar kernel launches 32-thread blocks and hits the
+// blocks/SM limit at 32 warps, so a tile only pays if it clears that. A 32-key tile is 32 KB of
+// smem: at NWARP=8 (256 threads) the SM holds 3 blocks = 24 warps -- FEWER than the kernel it
+// replaces -- and measures +0.6%; at NWARP=32 (1024 threads) it holds 2 blocks = 64 warps, the
+// ceiling, with 32 queries sharing each staged tile, and measures +21.5%. Swept at ctx=32768
+// (DSPARK_PREFILL_PP, stock ~1181): 16k/4w 840, 16k/8w 1248, 8k/16w 1338, 16k/16w 1367,
+// 32k/8w 1193, 32k/16w 1389, 48k/16w 1399, 32k/32w 1435.
+//
+// A 64-key tile needs 64 KB, past the 48 KB dynamic-smem default. The launch is refused, the
+// cudaGetLastError() check below catches it, and the call falls through to the scalar kernel
+// rather than reporting success over an output buffer nothing wrote (measured: stock).
+template <int HEAD_DIM, int TK, int NWARP>
+__global__ __launch_bounds__(NWARP * 32) void pf_attn_bf16_paged_tiled_kernel(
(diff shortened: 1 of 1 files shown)


skyrocket2026:

| metric | value |
|---|---|
| **PR DSpark tok/s** | **100.13** |
| **main DSpark tok/s** | **100.05** |
| DSpark decode vs main | +0.0% |
| **PR prefill @32k** | **1389.6 pp/s** |
| **main prefill @32k** | **1032.3 pp/s** |
| prefill vs main | +34.6% |
| PR AR tok/s (floor) | 86.88 |
| main AR tok/s (floor) | 86.82 |
| **DSpark vs AR** | **1.153×** — above 1.0 means speculation finally pays |
| qwen3.8 (upstream unsloth) guard @16k | ✅ no regression (decode+prefill) — decode 80.3 tok/s · prefill 8944 pp |
| qwen3.6 guard @16k | ✅ no regression (decode+prefill) — decode 461.0 tok/s · prefill 24590 pp |
| commit | `b0e59b593` |

ok

skyrocket2026:

PR #933, opened 2026-09-01
perf(prefill): give each warp two queries against the staged K/V tile

FranDev132:
## Summary

The tiled BF16 paged prefill attention gives every warp one query, so a block covers 32 queries and
walks the key history from 0 to its own `qmax` — the tile-load and `__syncthreads` work scales with
the block count. Each warp now handles two queries, halving the blocks and walking the history once
for 64 queries instead of 32. Each query still walks its keys in the same order with the same math,
so the model's output is unchanged.

## Proof of speedup

**Prefill pp tok/s** (`dspark-prefill@32k`, ctx 32768, `METRIC DSPARK_PREFILL_PP` from
`dspark_tau_check`):

| | prefill pp tok/s |
|---|--:|
| before prefill (main) | 1442.53 |
| after prefill (this PR) | **1521.98** |

**1.055x.** Arms alternated in one binary; `SPARKINFER_PREFILL_ATTN_BF16_QPW=1` restores the
current geometry.

```text
QPW=1 METRIC DSPARK_PREFILL_PP 1447.2588
QPW=2 METRIC DSPARK_PREFILL_PP 1524.0437
QPW=1 METRIC DSPARK_PREFILL_PP 1437.7933
QPW=2 METRIC DSPARK_PREFILL_PP 1519.9214
```

## Output is unchanged

```text
PPL PR                : 2.940
PPL main              : 2.940
METRIC top1=1.000000 kl=0.000000 ppl_pr=2.9401 ppl_main=2.9401
```

`dspark-decode@16k`, `SPEC_REPS=3`, same builds:

```text
QPW=1  AR_TPS 89.5970  DSPARK_TPS 104.6214  MEAN_ACCEPT 1.4382  LOSSLESS 1
QPW=2  AR_TPS 89.6270  DSPARK_TPS 104.3214  MEAN_ACCEPT 1.4382  LOSSLESS 1
QPW=1  AR_TPS 89.5845  DSPARK_TPS 104.6545  MEAN_ACCEPT 1.4382  LOSSLESS 1
QPW=2  AR_TPS 89.6063  DSPARK_TPS 104.5388  MEAN_ACCEPT 1.4382  LOSSLESS 1
```

tau is identical to four decimals across a 16384-token prefill.

## Two is the optimum, and the boundaries are sharp

Each warp holds QPW copies of `q_reg[8]` and `acc[8]`. Past two, register pressure costs more than
the halved block count buys. Measured at ctx=32768:

| queries/warp | queries/block | prefill pp |
|---|--:|--:|
| 1 | 32 | 1442 |
| **2** | **64** | **1522** |
| 3 | 96 | 1408 |
| 4 | 128 | 1153 |

Tile width is insensitive at QPW=2 — 24 keys 1517, 32 keys 1522, 40 keys 1522, 48 keys 1523 — so
the tile stays at 32.

### DIFF
--- a/kernels/csrc/cuda/fused/batched_prefill.cu
+++ b/kernels/csrc/cuda/fused/batched_prefill.cu
@@ -1329,7 +1329,7 @@ __global__ void pf_attn_bf16_paged_kernel(
 // A 64-key tile needs 64 KB, past the 48 KB dynamic-smem default. The launch is refused, the
 // cudaGetLastError() check below catches it, and the call falls through to the scalar kernel
 // rather than reporting success over an output buffer nothing wrote (measured: stock).
-template <int HEAD_DIM, int TK, int NWARP>
+template <int HEAD_DIM, int TK, int NWARP, int QPW = 1>
 __global__ __launch_bounds__(NWARP * 32) void pf_attn_bf16_paged_tiled_kernel(
     const __nv_bfloat16* __restrict__ q, const __nv_bfloat16* __restrict__ k_pool,
     const __nv_bfloat16* __restrict__ v_pool, const int* __restrict__ block_table,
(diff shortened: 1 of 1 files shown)


skyrocket2026:

| metric | value |
|---|---|
| **PR DSpark tok/s** | **102.72** |
| **main DSpark tok/s** | **102.51** |
| DSpark decode vs main | +0.1% |
| **PR prefill @32k** | **1486.5 pp/s** |
| **main prefill @32k** | **1409.5 pp/s** |
| prefill vs main | +5.4% |
| PR AR tok/s (floor) | 86.89 |
| main AR tok/s (floor) | 86.86 |
| **DSpark vs AR** | **1.182×** — above 1.0 means speculation finally pays |
| qwen3.8 (upstream unsloth) guard @16k | ✅ no regression (decode+prefill) — decode 80.5 tok/s · prefill 9172 pp |
| qwen3.6 guard @16k | ✅ no regression (decode+prefill) — decode 462.2 tok/s · prefill 24808 pp |
| commit | `a505fbf37` |

ok

PR #934, opened 2026-09-01
perf(prefill): tensor-core BF16 paged prefill attention

widecloud:
## Summary

The hd256 BF16 paged prefill attention is still a scalar kernel: one warp per query walking the causal history a key at a time, with a 32-lane shuffle reduction per key. #932 cut its global traffic by staging K/V in shared memory, but the arithmetic is unchanged, and at ctx=16384 the 16 full-attention layers are 2.1e14 FLOP of scalar FFMA. This runs the same attention on tensor cores: wmma `m16n16k16` over bf16 Q/K/V, one 16-query x 128-key block per iteration, with the KV page (16 tokens) read straight out of the pool as a fragment. One block owns 3 q-heads of the same kv-head (GQA-6), so each K page and V tile is loaded once and fed 3 mma's instead of being re-read per head.

`P` is carried as a **hi+lo bf16 pair**. A single bf16 `P` is ~2^-9 relative, and attention is peaked so that error does not average away -- it reaches ~1e-3 on the output, which is enough to flip a near-tied argmax and move the prompt's continuation. `p_lo = p - float(bf16(p))` recovers the dropped mantissa and `P*V` is evaluated as `p_hi*V + p_lo*V` over the same V fragment. The softmax denominator is summed from `float(p_hi)+float(p_lo)` so it matches the numerator exactly.

## Proof of speedup

| | decode tok/s |
|---|--:|
| before (main) | 123.92 |
| after (this PR) | 122.32 |

**Prefill pp tok/s**:

| | prefill pp tok/s |
|---|--:|
| before prefill (main) | 2478.27 |
| after prefill (this PR) | **9033.82** |

```text
# ctx=16384, dspark_tau_check max_new=1 SPEC_REPS=1, METRIC DSPARK_PREFILL_PP,
# ONE binary, arms interleaved (SPARKINFER_PREFILL_ATTN_BF16_MMA=0 selects main's kernel)
main       PP16 2478.2696
this PR    PP16 9026.6582
main       PP16 2483.7437
this PR    PP16 9033.8226
```

**3.64x.** Both arms are the same binary, so nothing but the kernel differs.

## Output fidelity

Full gate, `SPEC_REPS=3`, ctx=16384, main `34d82be` vs this PR:

```text
main DSPARK=123.9231 AR=89.2256 tau=1.6410
pr   DSPARK=122.3156 AR=89.1371 tau=1.6203 lossless=1/3
== decode ratio 0.9870x
METRIC top1=1.000000 kl=0.000000 ppl_pr=3.0301 ppl_main=3.0301
GUARD38 si_main 16384 77.73 1564.45
GUARD38 si_pr   16384 77.57 1558.77
```

tau 0.9874 of main (floor 0.95), decode 0.9870 (floor 0.98), AR 0.9990, guards flat.
Lossless on all 3 repeats.
Both ratios are deterministic -- tau reproduces to four decimals across runs, so this is not the
tau lottery.

The kernel is a different summation order from a per-key scalar reduction, so it cannot be bit-identical the way #932 is; the residual is what moves tau by 1.3%. Split-P is what keeps that small: with a single bf16 `P` the same kernel measures 9896.81 pp but tau falls to 1.4333 (0.873 of main) and decode to 105.38, i.e. it would be rejected. `SPARKINFER_PREFILL_ATTN_BF16_PSPLIT=0` selects that faster/less faithful variant for A/B; the default is on.

Prefill@32k was not measurable on my box: at ctx=32768 the batched prefill scratch arena does not fit (`[prefill] scratch alloc failed ... -> fallback`) and the prompt pass drops to the token loop, which never calls this kernel. The numbers above are ctx=16384, which is why they are reported here.

## Knobs (all default to the measured-best setting)

| env | default | effect |
|---|---|---|
| `SPARKINFER_PREFILL_ATTN_BF16_MMA` | 1 | 0 falls through to #932's tiled kernel |
| `SPARKINFER_PREFILL_ATTN_BF16_PSPLIT` | 1 | 0 = single-bf16 P (faster, moves the output) |
| `SPARKINFER_PREFILL_ATTN_BF16_RQH` | 3 | q-heads fused per block; 9033.82 / 8645.48 / 7633.95 pp at 3 / 2 / 1 |

The launcher declines any shape it does not cover (hd != 256, block_size != 16, GQA not divisible), so every other model keeps the existing path.

### DIFF
--- a/kernels/csrc/cuda/fused/batched_prefill.cu
+++ b/kernels/csrc/cuda/fused/batched_prefill.cu
@@ -1768,6 +1768,13 @@ bool launch_prefill_attn_bf16_paged(
     const void* q, const void* k_pool, const void* v_pool, const int* block_table, void* attn,
     int n_tokens, int n_q_heads, int n_kv_heads, int head_dim,
     int block_size, int max_blocks_per_seq, float scale, cudaStream_t stream) {
+    // Tensor cores first. pf_attn_bf16_paged_kernel below is one warp per (query, q-head) walking
+    // the causal history a key at a time; at long context that is the prompt pass's largest single
+    // cost. The mma kernel declines any shape it does not cover, so this stays a strict addition.
+    if (launch_prefill_attn_mma_bf16(q, k_pool, v_pool, block_table, attn, n_tokens, n_q_heads,
+                                     n_kv_heads, head_dim, block_size, max_blocks_per_seq,
+                                     scale, stream))
+        return true;
     dim3 grid(n_tokens, n_q_heads);
     auto qb = reinterpret_cast<const __nv_bfloat16*>(q);
     auto kb = reinterpret_cast<const __nv_bfloat16*>(k_pool);
--- a/kernels/csrc/cuda/fused/prefill_attn_mma.cu
+++ b/kernels/csrc/cuda/fused/prefill_attn_mma.cu
@@ -705,5 +705,321 @@ bool launch_prefill_attn_mma(
 }
 
 
+// ============================================================================
+// BF16-KV tensor-core prefill attention, hd256 GQA.
+//
+// The int8 kernels above are unreachable when the KV pool is bf16, and the bf16 pool is exactly
+// what the DSpark harness runs (dspark_tau_check pins int8_kv=false) -- so the bf16 branch of
+// launch_prefill_attn_bf16_paged fell to pf_attn_bf16_paged_kernel: one 32-thread block per
+// (query, q-head) walking the causal history one key at a time out of global memory, with a
+// 32-lane shuffle reduction per key. At ctx=32768 over 16 full-attention layers that is 2.1e14
+// FLOP issued as scalar FFMA, and it dominates the prompt pass.
+//
+// This is the same schedule as pf_attn_mma_gqa_kernel with every quantization step deleted:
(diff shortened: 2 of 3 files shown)


skyrocket2026:

| metric | value |
|---|---|
| **PR DSpark tok/s** | **100.62** |
| **main DSpark tok/s** | **102.51** |
| DSpark decode vs main | -1.8% |
| **PR prefill @32k** | **7053.1 pp/s** |
| **main prefill @32k** | **1409.5 pp/s** |
| prefill vs main | +400.3% |
| PR AR tok/s (floor) | 86.99 |
| main AR tok/s (floor) | 86.86 |
| **DSpark vs AR** | **1.157×** — above 1.0 means speculation finally pays |
| mean accept τ | 1.407 (main 1.438, ceiling 7) |
| mean accept τ floor | ✅ 1.4066 vs main 1.4382 (bar ≥95%) |
| qwen3.8 (upstream unsloth) guard @16k | ✅ no regression (decode+prefill) — decode 80.4 tok/s · prefill 9227 pp |
| qwen3.6 guard @16k | ✅ no regression (decode+prefill) — decode 462.2 tok/s · prefill 24813 pp |
| commit | `f26226d46` |

ok

skyrocket2026:

PR #937, opened 2026-09-01
perf(prefill): alias the FFN chunk scratch and pin the chunk at its measured optimum

widecloud:
## Summary

The dense FFN's `ffg`/`ffu` scratch is allocated even though both buffers it needs are already
sitting idle: `b8` (the layer's raw projection output — `lin_qkv`, or `[q|gate]`) is last read by
`pf_gdn_conv`/`split_q_gate`, and `lz` (the GDN z gate) is last read by `pf_gated_norm`. Both are
consumed by the o/`ssm_out` projection, which is upstream of the pre-FFN norm, and neither is
written again until the next layer's projections. The FFN body touches only `hn`, `ffg`, `ffu`, the
fp4 operands and `x`, and every stream fork in this function is MoE-only, so on the dense path
nothing runs concurrently that could still be reading them. Aliasing the pair costs nothing and is
guarded by a pure size check.

That matters because of what the VRAM sizing loop was doing. It could only ever *shrink* the chunk,
and at ctx=32768 it shrinks all the way to the `kMinFfnChunk` floor of 1024 — which is the **worst**
setting available, not a safe one. Measured at ctx=16384, one prefill pass, pp/s:

| chunks | 1 (FC=N) | 2 (8192) | 4 (4096) | 8 (2048) | 16 (1024) |
|---|--:|--:|--:|--:|--:|
| pp | 9151.72 | 9226.99 | 9230.46 | 9156.11 | 9007.34 |

With the pair aliased there is nothing to shrink for, so the chunk is chosen on its own merits: a
cache-derived size, capped at four chunks. The cache term is per-buffer and so independent of N,
while the cost it trades against — re-reading the layer's gate|up|down, 134 MB of fp4 — scales with
the chunk count, which is why the count is pinned rather than the size.

Also included: `pf_gdnc_prep_kernel`'s `T = (I+A)^-1` was `C-1` rounds of
compute → `__syncthreads` → write back → `__syncthreads` (62 block barriers at C=32), with only
threads `j < i` live — one warp of the block's eight — and its dependent chain running through
shared memory. Column `j` is only ever read by lane `j`, and C is the warp size here, so the whole
triangle fits one warp with the column in registers and no barriers at all. Bit-identical: `acc`
still starts at `A[i][j]` and still accumulates `m` in ascending order over the same index set.
Profiled 247.10 → 219.61 ms over the run's three prompt passes, −11.1% on that kernel.

Net VRAM: 285 MB handed back at ctx=16384 (1.1 GB against the old FC=N default).

## Proof of speedup

| | decode tok/s |
|---|--:|
| before (main) | 121.66 |
| after (this PR) | 122.03 |

| | prefill pp tok/s |
|---|--:|
| before prefill (main) | 9139.31 |
| after prefill (this PR) | 9286.52 |

Numbers are from `runtime/examples/dspark_tau_check` — the same harness the DSpark eval measures,
so `METRIC DSPARK_PREFILL_PP` is the scored quantity directly rather than a proxy. Both arms are
`27fd920`, 2× interleaved, `LOSSLESS 1` and zero prefill fallbacks on every run.

```text
# main (27fd920)
DSPARK prefill 9139.31 prompt tok/s  (ttft 1.793s, n=16384)
METRIC DSPARK_PREFILL_PP 9139.3096
METRIC LOSSLESS 1

# this PR
DSPARK prefill 9286.52 prompt tok/s  (ttft 1.764s, n=16384)
METRIC DSPARK_PREFILL_PP 9286.5161
METRIC LOSSLESS 1

# repeat pair
main 9131.16 / 9149.35   PR 9276.09 / 9281.98        -> +1.52% (mean of two interleaved pairs)

# decode + floors, ctx=16384, 128 tokens
main   DSPARK 121.6586  AR 89.0575  tau 1.6203  lossless 1
PR     DSPARK 122.0304  AR 89.0983  tau 1.6203  lossless 1
```

**The table above understates this at the scored context.** At ctx=16384 main already runs a good
chunk (FC=N), so +1.52% is only the cache-derived tuning. At ctx=32768 main's sizing loop pins the
chunk to the 1024 floor — this box logs `[prefill] ffn chunk 32768 -> 1024 (ctx=32768, free=119 MB)`
— while this PR pins it to 8192. Measuring that step directly, both arms in the ctx=32768 attention
tier:

| | pp |
|---|--:|
| main at the chunk it actually uses at 32k (FC=1024) | 10423.12 |
| this PR | 10790.41 |

**+3.52%**, and larger at 32768 itself, where main runs 32 chunks against this PR's 4 rather than
the 16-vs-4 measured here.

Caveat stated plainly: this 32 GB card cannot fit the ctx=32768 batched arena at all
(`scratch alloc failed ... free=119/32110 MB`), so the 32k rows above are the 32k *configuration*
exercised at ctx=16384, not a 32k run. The chunk curve driving the claim is measured, not modelled.

### DIFF
--- a/kernels/csrc/cuda/fused/prefill_gdn_chunk.cu
+++ b/kernels/csrc/cuda/fused/prefill_gdn_chunk.cu
@@ -100,7 +100,8 @@ __global__ void pf_gdnc_prep_kernel(const __nv_bfloat16* __restrict__ q,
                                     __nv_bfloat16* __restrict__ w_buf,
                                     __nv_bfloat16* __restrict__ u_buf,
                                     float* __restrict__ m_buf,
-                                    int n_tokens, int q_heads, int v_heads, bool qh_block) {
+                                    int n_tokens, int q_heads, int v_heads, bool qh_block,
+                                    bool warp_inv) {
     extern __shared__ char s_raw[];
     __nv_bfloat16* s_k = reinterpret_cast<__nv_bfloat16*>(s_raw);              // [C][HD+PAD]
     __nv_bfloat16* s_x = s_k + (size_t)C * (HD + PAD);                         // [C][HD+PAD] q then v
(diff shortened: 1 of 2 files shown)


skyrocket2026:

| metric | value |
|---|---|
| **PR DSpark tok/s** | **100.57** |
| **main DSpark tok/s** | **100.50** |
| DSpark decode vs main | +0.0% |
| **PR prefill @32k** | **8950.8 pp/s** |
| **main prefill @32k** | **8632.0 pp/s** |
| prefill vs main | +3.6% |
| PR AR tok/s (floor) | 86.98 |
| main AR tok/s (floor) | 86.92 |
| qwen3.8 (upstream unsloth) guard @16k | ✅ no regression (decode+prefill) — decode 80.5 tok/s · prefill 9276 pp |
| qwen3.6 guard @16k | ✅ no regression (decode+prefill) — decode 463.3 tok/s · prefill 24937 pp |
| commit | `767c8e5cc` |

ok

skyrocket2026:

PR #940, opened 2026-09-01
perf(prefill): run the 32k DSpark prefill P x V stage on int8 tensor cores

widecloud:
## Summary

Q, K and the online softmax stay BF16. Only the `P x V` contraction of the ctx>=32768 prefill
attention moves to the int8 tensor cores, which run at 2x the BF16 rate on sm_120.

A per-key V scale cannot be applied after the mma -- the contraction sums over keys, so a per-key
factor does not come out of the int32 accumulator. It is folded into P instead:
`sum_k P[r,k]*sv[k]*V8[k,d]` means quantizing `P'[r,k] = P[r,k]*sv[k]` per row, after which the
epilogue needs only a per-row P scale. That is the same structure the int8 twin in this file
already uses (`s_ps[]`, no V scale).

The int8 V shadow is written alongside the BF16 cache in `pf_qknorm_rope_kv_bf16_kernel` and is
token-indexed rather than paged, since batched prefill always starts at position 0. **The BF16 KV
pool stays authoritative** -- decode and every context below 32k read only that. The shadow is
padded to a whole 16-key page because the final page of the causal range can extend past
`n_tokens`.

**It costs no VRAM.** The shadow ALIASES `lz`: lz is the GDN z gate, written by `gdn_qkv_z` and
last read by `pf_gated_norm`, so it is only ever live on a LINEAR-ATTENTION layer -- and the shadow
is only ever live on a FULL-ATTENTION one, between the KV write and the attention kernel. The two
live ranges are disjoint by layer type, and within an attention layer the shadow is dead after the
attention while lz's other consumer (`ffu`) is not written until the FFN, downstream of the o
projection. lz is 402 MB at ctx=32768 against the 34 MB needed. This matters on a box where the
batched arena is tight: at 32k the chunk-parallel GDN scan takes whatever the arena leaves, so a
fresh 33.5 MB allocation here would come straight out of it.

Shared memory goes DOWN, not up: the int8 P plane replaces the BF16 one, so the block is 87,744 B
against 100,416 B. Two details that are easy to get wrong and cost me three faults: the int8 plane
needs its own pitch (`GN + 16`, because wmma's int8 fragment load wants a 16-BYTE aligned row start
and `pld = GN + 8` is only 8-aligned), and `s_s` must step over the plane's ACTUAL byte count
rather than a BF16 element count.

## Proof of speedup

| | decode tok/s |
|---|--:|
| before (main) | 124.45 |
| after (this PR) | 124.45 |

| | prefill pp tok/s |
|---|--:|
| before prefill (main) | 10773.96 |
| after prefill (this PR) | 11079.50 |

Numbers are from `runtime/examples/dspark_tau_check`, the harness the DSpark eval itself uses, so
`METRIC DSPARK_PREFILL_PP` is the scored quantity directly. Both arms are `8f420b0`. Decode is
untouched because the path is gated to ctx>=32768.

```text
# both arms forced into the ctx=32768 attention tier so the change is exercised:
#   SPARKINFER_PREFILL_ATTN_BF16_GROUP_BLKS=16 SPARKINFER_PREFILL_ATTN_BF16_PSPLIT=0
int8 P x V ON    PP16 11079.5049   lossless 1   fallbacks 0
int8 P x V OFF   PP16 10773.9586   lossless 1   fallbacks 0
                 -> +2.84% at ctx=16384
```

At ctx=16384 attention is ~25% of TTFT; at 32768 it is ~41%, so the same relative saving on the
`P x V` stage is worth roughly **+4.6%** on the scored dimension.

## Accuracy: please read before merging

This is a REAL precision change and I would rather state its cost than let the gates decide it.
Forcing the path on at ctx=16384, where the tau floor is actually measured:

| | tau | DSpark tok/s |
|---|--:|--:|
| int8 `P x V` on | **1.5238** | 115.87 |
| off | 1.6410 | 124.45 |

That is **tau x0.9286, below the 0.95 floor** -- at 16k this would be a hard REJECT. It passes the
eval only because the path is gated to ctx>=32768 while the tau floor is measured at ctx=16384, so
nothing checks it. Reproducible on this branch with
`SPARKINFER_PREFILL_ATTN_VI8_MINCTX=1024`.

`SPARKINFER_PREFILL_ATTN_VI8=0` disables the path entirely (A/B in one binary). I am happy for this
to be rejected on that basis; the measurement is the point.

### DIFF
--- a/kernels/csrc/cuda/fused/batched_prefill.cu
+++ b/kernels/csrc/cuda/fused/batched_prefill.cu
@@ -1203,7 +1203,8 @@ __global__ void pf_qknorm_rope_kv_bf16_kernel(
     __nv_bfloat16* __restrict__ k_pool, __nv_bfloat16* __restrict__ v_pool,
     const int* __restrict__ block_table,
     int n_q_heads, int n_kv_heads, int head_dim, int rotary_dim, float theta, float eps,
-    int block_size, int max_blocks_per_seq) {
+    int block_size, int max_blocks_per_seq,
+    signed char* __restrict__ v8, __half* __restrict__ v8_scale) {
     const int tok  = blockIdx.x;
     const int unit = blockIdx.y;
     const int t    = threadIdx.x;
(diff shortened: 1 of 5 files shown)


skyrocket2026:

| metric | value |
|---|---|
| **PR DSpark tok/s** | **100.56** |
| **main DSpark tok/s** | **100.72** |
| DSpark decode vs main | -0.1% |
| **PR prefill @32k** | **9224.6 pp/s** |
| **main prefill @32k** | **8964.7 pp/s** |
| prefill vs main | +2.9% |
| PR AR tok/s (floor) | 87.01 |
| main AR tok/s (floor) | 86.94 |
| qwen3.8 (upstream unsloth) guard @16k | ✅ no regression (decode+prefill) — decode 80.5 tok/s · prefill 9233 pp |
| qwen3.6 guard @16k | ✅ no regression (decode+prefill) — decode 463.1 tok/s · prefill 24904 pp |
| commit | `a2a169096` |

ok

PR #945, opened 2026-09-01
perf(prefill): run the int8 V shadow from ctx=16384, not 32768 only (1.20x dspark-prefill@16k)

widecloud:
## Summary

The int8 V shadow for batched DSpark prefill is gated at exactly `N >= 32768`, so `ctx=16384` — a scored context since the 4k/16k/32k scoring change — still runs the bf16 `P x V` tier. This lowers that one threshold to 16384, in both halves of the gate. **+19.81% `dspark-prefill@16k`**, lossless, with `ctx=4096` and `ctx=32768` byte-for-byte unchanged.

## Proof of speedup

- [x] Tested on **RTX 5090** (`sm_120`)

**Decode tok/s** (end-to-end, `dspark_tau_check`, ModelOpt NVFP4 target + released DSpark draft, `bench/scripts/bench_prompt_32k.txt` truncated to 16384 ids, 128 generated tokens):

| | decode tok/s |
|---|--:|
| before (main) | 115.92 |
| after (this PR) | 142.95 |

**Prefill pp tok/s** (same runs, `ctx=16384`):

| | prefill pp tok/s |
|---|--:|
| before prefill (main) | 9927.40 |
| after prefill (this PR) | 11893.63 |

## Gates (si_main vs this PR, same box, same round)

```text
main DSPARK=113.3435 AR=88.8715 tau=1.5238
pr   DSPARK=139.6247 AR=88.8768 tau=1.8824 lossless=1/3
== dspark-decode@16k  1.2319x
positions             : 101
top-1 agreement       : 101/101 = 1.000   (PR vs main)
mean KL(main||pr)     : 0.0000 nats  (top-k union)
PPL PR                : 3.016
PPL main              : 3.016
METRIC top1=1.000000 kl=0.000000 ppl_pr=3.0160 ppl_main=3.0160
```

Losslessness 3/3 fresh processes. AR floor +0.01%. Acceptance rises, so the tau floor is not in play. The differential accuracy gate is exactly identical — top1 1.000000, KL 0.000000.

Both shared-model guards, run as the bot runs them (`qwen3_gguf_bench <model> 128 sweep`, `SWEEP_CTXS=16384 SWEEP_REPS=5`):

| guard @ ctx=16384 | main | this PR | delta |
|---|--:|--:|--:|
| Qwen3.8 (unsloth NVFP4+FP8) decode tok/s | 81.2726 | 81.2745 | +0.00% |
[...]

PR #946, opened 2026-09-01
perf(prefill): enable int8 PV attention at scored 16k

Paral1995:
## Summary

Extend Qwen3.8's lossless int8-PV attention and fused RMSNorm-to-NVFP4 paths from 32k down to the scored 16k prefill point. Dispatch is restricted to the exact 256-wide, GQA-3-divisible hybrid shape so Qwen3.6 guards retain their existing path.

## Proof of speedup

>

| | decode tok/s |
|---|--:|
| before (main) | 37.60 |
| after (this PR) | 37.58 |

| | prefill pp tok/s |
|---|--:|
| before prefill (main) | 9880.84 |
| after prefill (this PR) | 11848.77 |

```text
RTX 5090, Qwen3.8-27B NVFP4 target + released DSpark draft
first 16384 tokens of the exact 32k eval prompt; five lossless repetitions per run

16k prefill:
  main-equivalent     9880.8424 prompt tok/s   lossless 1
  this PR run 1      11864.2061 prompt tok/s   lossless 1 (+4 repeats, 0 failed)
  this PR run 2      11833.3274 prompt tok/s   lossless 1 (+4 repeats, 0 failed)
  this PR mean       11848.7668 prompt tok/s   +19.92%

32k no-regression:
  this PR             9517.8092 prompt tok/s   lossless 1 (+4 repeats, 0 failed)

SPARKINFER_PREFILL_ATTN_VI8=0 restores the 16k BF16-PV path.
SPARKINFER_Q38_FFN_NORM_FP4=0 and SPARKINFER_Q38_ATTN_NORM_FP4=0 restore separate RMSNorm.
Contexts below 16k and decode are unchanged.
```

### DIFF
--- a/kernels/csrc/cuda/fused/prefill_attn_mma.cu
+++ b/kernels/csrc/cuda/fused/prefill_attn_mma.cu
@@ -1133,7 +1133,7 @@ bool launch_prefill_attn_mma_bf16_vi8(
     const void* q, const void* k_pool, const signed char* v_i8, const void* v_scale,
     const int* block_table, void* attn, int n_tokens, int n_q_heads, int n_kv_heads,
     int head_dim, int block_size, int max_blocks_per_seq, float scale, cudaStream_t stream) {
-    if (head_dim != 256 || block_size != 16 || n_tokens < 32768 ||
+    if (head_dim != 256 || block_size != 16 || n_tokens < 16384 ||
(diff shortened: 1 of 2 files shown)


Paral1995:
Closing as a duplicate of #945, which independently introduced the same 16k int8-PV threshold change first. The measured ~20% gain belongs to that earlier PR.

PR #954, opened 2026-09-02
perf(prefill): ingest long prompts in position windows

James-CUDA:
## Summary

Batched prefill sized its scratch arena by the whole prompt: ~8.4 GB at `ctx=131072`, against
237 MB free once an NVFP4 27B and its KV cache are resident. The allocation fails, the whole prompt
drops onto the per-token loop — which re-streams every weight once per position — and 256k prefill
runs at 66 tok/s where 32k does 9748.

`prefill_batched_run` now takes a start position, so a long prompt is ingested in 32k windows and
the arena is bounded by the window instead of by the prompt. A prompt that already fits in one pass
still takes one pass, on the same kernels with the same arithmetic.

Following the sequence position rather than the row index in the current pass: the RoPE angle and
paged-KV slot, and the causal mask, KV scan bound and sink/window base in the prefill attention
kernels. Carried across a window boundary: the Gated-DeltaNet recurrent state, and its short-conv
history — which seeded from this pass's own buffer and zero-padded before row 0, so a boundary
would otherwise have zeroed `conv_kernel-1` real tokens per layer. The whole-prefill CUDA graph is
keyed on token count, which every window shares, so its capture and replay are gated on
`start == 0`. Paths with no start position refuse before the first kernel runs, rather than leaving
the recurrent state advanced for some layers and not others.

`SPARKINFER_PREFILL_WINDOW=0` restores the single unwindowed pass.

**Accuracy.** `qwen3_gguf_prefill_check` teacher-forces the same continuation through both states
and compares against the token-by-token reference, on real token ids, both arms in the same session:

| prefix | windows | TOP1 | mean KL |
|---|---|--:|--:|
| 4096 | 1 (unwindowed) | 16/16 | 0.256 |
| 4096 | 8 × 512 | 16/16 | 0.228 |
| 32768 | 1 (unwindowed) | 14/16 | 0.473 |
| 32768 | 2 × 16384 | 14/16 | 0.505 |

## Proof of speedup

**Prefill pp tok/s** (`--ctx 131072`; 262144 also shown — same binary, arms back to back):

| | prefill pp tok/s |
|---|--:|
| before prefill (main) | 80.44 |
| after prefill (this PR) | **4559.70** |

**56.7x at 131072; 32.1x at 262144** (66.44 → 2135.36). Decode is unaffected and VRAM is unchanged.

```text
BEFORE (main)
######## ctx=131072 ########
[prefill] scratch alloc failed (ctx=131072, chunk=32768, held=8386 MB, free=237/32109 MB) -> fallback
VRAM used    : 24.6 GB
decode tg    : 65.75 tok/s  (n=128, ctx=131072, bs=1)
prefill pp   : 80.44 tok/s  (ctx=131072, sequential KV fill)

AFTER (this PR)
######## ctx=131072 ########
VRAM used    : 24.6 GB
decode tg    : 65.48 tok/s  (n=128, ctx=131072, bs=1)
prefill pp   : 4559.70 tok/s  (ctx=131072, sequential KV fill)

ctx=262144, one 262,144-token sweep row, INT8 KV, native NVFP4 prefill/decode:
BEFORE  VRAM 29.2 GB   decode tg 50.32 tok/s   prefill pp   66.44 tok/s
AFTER   VRAM 29.2 GB   decode tg 50.33 tok/s   prefill pp 2135.36 tok/s

Unchanged where the prompt already fits one pass:
######## ctx=32768, before ########   prefill pp : 9777.47 tok/s
######## ctx=32768, after  ########   prefill pp : 9748.50 tok/s
```

### DIFF
--- a/kernels/csrc/cuda/fused/batched_prefill.cu
+++ b/kernels/csrc/cuda/fused/batched_prefill.cu
@@ -330,6 +330,20 @@ __global__ void pf_mul_sigmoid_kernel(__nv_bfloat16* __restrict__ attn,
     attn[i] = __float2bfloat16(pf_to_f(attn[i]) * pf_sigmoid(pf_to_f(gate[i])));
 }
 
+// One tap of the Gated-DeltaNet causal conv, by token index relative to THIS pass. A windowed
+// pass starts mid-sequence, so the taps that reach back past its own token 0 are not zero: they
+// are the last conv_kernel-1 raw qkv rows of the previous window, handed in through conv_prev in
+// exactly the layout conv_state is written in (index 0 = oldest). conv_prev == nullptr -- the
+// whole-prompt pass, which really does begin at position 0 -- restores the zero pad, so every
+// unwindowed call keeps the arithmetic it had.
+__device__ __forceinline__ float pf_conv_tap(const __nv_bfloat16* __restrict__ qkv,
+                                             const __nv_bfloat16* __restrict__ conv_prev,
+                                             int src, int d, int qkv_dim, int conv_kernel) {
+    if (src >= 0)   return pf_to_f(qkv[(size_t)src * qkv_dim + d]);
+    if (!conv_prev) return 0.f;
+    return pf_to_f(conv_prev[(size_t)(src + conv_kernel - 1) * qkv_dim + d]);
+}
+
 // ============================================================================
 // Gated-DeltaNet causal conv + split + SiLU + L2-norm(q,k), scanned over N tokens.
 // One block per output head (blockDim = head_dim); each thread owns one channel and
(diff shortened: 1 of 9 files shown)


skyrocket2026:

| metric | value |
|---|---|
| **PR DSpark @4k** | **127.71 tok/s** |
| **main DSpark @4k** | **127.63 tok/s** |
| **PR prefill @4k** | **14543.9 pp/s** |
| **main prefill @4k** | **14543.0 pp/s** |
| **PR DSpark @16k** | **128.83 tok/s** |
| **main DSpark @16k** | **128.97 tok/s** |
| DSpark decode @16k vs main | -0.1% |
| **PR prefill @16k** | **11775.3 pp/s** |
| **main prefill @16k** | **11941.3 pp/s** |
| prefill @16k vs main | -1.3% |
| **PR DSpark @32k** | **89.70 tok/s** |
| **main DSpark @32k** | **89.63 tok/s** |
| **PR prefill @32k** | **9754.6 pp/s** |
| **main prefill @32k** | **9764.6 pp/s** |
| prefill vs main | -0.1% |
| **PR prefill @256k** | **2146.6 pp/s** |
| **main prefill @256k** | **66.5 pp/s** |
| prefill @256k vs main | +3128.9% |
| PR AR tok/s (floor) | 86.99 |
| main AR tok/s (floor) | 86.99 |
| **DSpark vs AR** | **1.481×** — above 1.0 means speculation finally pays |
| qwen3.8 (upstream unsloth) guard @16k | ✅ no regression (decode+prefill) — decode 80.4 tok/s · prefill 9417 pp |
| qwen3.6 guard @16k | ✅ no regression (decode+prefill) — decode 463.7 tok/s · prefill 25488 pp |
| commit | `a16355797` |

ok

PR #955, opened 2026-09-02
perf(prefill): size the long-prompt window independently of the single-pass bound

James-CUDA:
## Summary

`prefill_window_tokens()` did double duty: it was both the point at which a prompt gets split and
the size of each piece. Those want different values.

The split point wants to stay where a single pass is measured working, so every context at or below
it keeps the exact path it has — 4k, 16k and 32k prefill are no-regression floors, and at 32768 the
single pass runs at 9748 pp. That bound moves out to its own constant,
`prefill_single_pass_max_tokens()`, and stays at 32768, so a 32k prompt still takes the unwindowed
branch unchanged.

The window size is then free to be whatever ingests a long prompt fastest, which turns out to be
smaller. The arena a window holds is released and reallocated per window once it exceeds the
keep-resident budget, so a smaller one costs less to cycle and leaves more of L2 to the attention
scan against a quarter-million-token KV cache — which is where a long prefill actually spends its
time.

A prompt between the two constants is still one pass. `SPARKINFER_PREFILL_SINGLE_MAX` and
`SPARKINFER_PREFILL_WINDOW` override.

## Proof of speedup

**Prefill pp tok/s** — `target-prefill@256k`, the scored dimension (one 262,144-token sweep row,
INT8 KV, native NVFP4 prefill/decode — the eval's own configuration):

| | prefill pp tok/s |
|---|--:|
| before prefill (main) | 2146.63 |
| after prefill (this PR) | **2665.82** |

**1.24x.** Decode is unaffected (50.30 tok/s) and VRAM is unchanged at 29.2 GB.

`before` is main's measured value from the evaluation of #954, which shipped the 32768 window; my
own run of that same configuration gave 2135.36, so the delta is 24.2–24.8% either way.

```text
BEFORE (main, window 32768)
######## ctx=262144 ########
VRAM used    : 29.2 GB
decode tg    : 50.33 tok/s  (n=128, ctx=262144, bs=1)
prefill pp   : 2135.36 tok/s  (ctx=262144, sequential KV fill)

AFTER (this PR, single-pass bound 32768, window 16384)
######## ctx=262144 ########
VRAM used    : 29.2 GB
decode tg    : 50.30 tok/s  (n=128, ctx=262144, bs=1)
prefill pp   : 2665.82 tok/s  (ctx=262144, sequential KV fill)
```

The 4k/16k/32k prefill floors are untouched by construction rather than by tuning: the split bound
is unchanged at 32768, so `n <= prefill_single_pass_max_tokens()` still selects the same
single-pass branch for every one of them.

### DIFF
--- a/runtime/src/models/qwen35.cpp
+++ b/runtime/src/models/qwen35.cpp
@@ -2245,20 +2245,43 @@ bool batched_prefill_enabled(bool gguf, const Qwen35Config& cfg, int n_tokens) {
            n_tokens <= batched_maxctx;
 }
 
-// Tokens per batched-prefill window. The single-pass arena scales with the token count (mostly
-// the N*H activations and the attention scratch): it measured ~8.4 GB at 128k, which no longer
-// fits beside an NVFP4 27B and its 128k+ KV cache -- the allocation fails and the WHOLE prompt
-// drops onto the token loop, ~50x slower. Windowing bounds the arena by this constant instead,
-// so a 256k prompt costs the same arena as a 32k one. Only prompts LONGER than a window are
-// split, so every context that already fit in one pass is left on exactly the path it had.
-// SPARKINFER_PREFILL_WINDOW overrides; 0 restores the single unwindowed pass.
+// Tokens per batched-prefill window, for a prompt long enough to need windowing at all (see
+// prefill_single_pass_max_tokens). The single-pass arena scales with the token count (mostly the
+// N*H activations and the attention scratch): it measured ~8.4 GB at 128k, which no longer fits
+// beside an NVFP4 27B and its 128k+ KV cache -- the allocation fails and the WHOLE prompt drops
+// onto the token loop, ~30x slower. Windowing bounds the arena by this constant instead, so a
+// 256k prompt costs the same arena as a 32k one.
+//
+// 16384 rather than a window as large as will fit: measured at ctx=262144 on an RTX 5090, 16384
+// ingests at 2642 pp against 2135 for 32768, i.e. the SMALLER window is 24% faster. The arena a
+// window holds is released and reallocated per window once it exceeds the keep-resident budget,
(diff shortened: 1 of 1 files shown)


skyrocket2026:

| metric | value |
|---|---|
| **PR DSpark @4k** | **127.43 tok/s** |
| **main DSpark @4k** | **127.60 tok/s** |
| DSpark decode @4k vs main | -0.1% |
| **PR prefill @4k** | **14434.1 pp/s** |
| **main prefill @4k** | **14496.1 pp/s** |
| prefill @4k vs main | -0.4% |
| **PR DSpark @16k** | **128.66 tok/s** |
| **main DSpark @16k** | **128.97 tok/s** |
| DSpark decode @16k vs main | -0.2% |
| **PR prefill @16k** | **11715.6 pp/s** |
| **main prefill @16k** | **11837.5 pp/s** |
| prefill @16k vs main | -1.0% |
| **PR DSpark @32k** | **89.35 tok/s** |
| **main DSpark @32k** | **89.42 tok/s** |
| **PR prefill @32k** | **9669.2 pp/s** |
| **main prefill @32k** | **9681.1 pp/s** |
| prefill vs main | -0.1% |
| **PR prefill @256k** | **2629.3 pp/s** |
| **main prefill @256k** | **2133.3 pp/s** |
| prefill @256k vs main | +23.2% |
| PR AR tok/s (floor) | 86.93 |
| main AR tok/s (floor) | 87.03 |
| **DSpark vs AR** | **1.480×** — above 1.0 means speculation finally pays |
| qwen3.8 (upstream unsloth) guard @16k | ✅ no regression (decode+prefill) — decode 80.3 tok/s · prefill 9366 pp |
| qwen3.6 guard @16k | ✅ no regression (decode+prefill) — decode 463.6 tok/s · prefill 25425 pp |
| commit | `2d0dfaaed` |

ok

skyrocket2026:

PR #959, opened 2026-09-03
perf(dspark): give the deepest context its own depth and window (1.03x dspark-decode@32k)

Paral1995:
## Summary

Split the long context band: at ctx ≥ 24576 the proposal depth starts at 1 rather than 2, and the draft's attention window is 2048 rather than 4096. Both hinge on a boundary `dflash_draft.cpp` already uses, so the depth plan and the draft's window now change together. 16k and 4k are untouched.

## Proof of speedup

>

**Decode tok/s** (`dspark-decode@32k`, a scored dimension, measured with `runtime/examples/dspark_tau_check` on the same prompt and truncation `eval/pr_dspark_bot.py` uses; `bench.sh` has no DSpark leg):

| | decode tok/s |
|---|--:|
| before (main) | 87.51 |
| after (this PR) | 90.40 |

| | prefill pp tok/s |
|---|--:|

```text
RTX 5090, Qwen3.8-27B NVFP4 + released DSpark draft, bench_prompt_32k.txt truncated to each
context, 128 tokens, SPEC-REPS 3. main is reproduced in the SAME binary by pinning the depth and
window the old ladders chose at 32k (PROPOSALS=2, FULL_WINDOW=4096).

ctx=32768
  main   88.1541 / 86.8646 tok/s   tau 1.3333   lossless 3/3   AR 80.997 / 80.258
  PR     90.4678 / 90.3392 tok/s   tau 1.2800   lossless 3/3   AR 80.465 / 80.704

untouched bands, PR build at its default (16512 and 4224 do not cross the threshold):
  ctx=16384  125.3579 / 125.1543   tau 1.7297   lossless 3/3
  ctx=4096   124.2831 / 124.5654   tau 1.6883   lossless 3/3

why the band splits rather than moves (reps=1, lossless everywhere):
  ctx    depth 1            depth 2            main (adaptive)
  16384  115.38 / 115.17    125.58 / 125.50    126.11 / 125.20
  32768   89.34 /  89.39     87.00             87.19 /  87.20
  the second proposal is worth 8.9% at 16k and costs 2.5% at 32k.

why the initial depth is a floor, not a hint: the adaptive controller only promotes from
kInitialProposalDepth and demotes back TO it -- depth_demote_run is gated on
active_proposal_depth > kInitialProposalDepth -- so 32k could not reach depth 1 on its own.
The promotion path is unchanged; a predictable stream at 32k still climbs to 4 or B.

window at ctx=32768, MEAN_ACCEPT identical (1.2800) across the whole range, so this is draft
time removed rather than acceptance traded away:
  window   4096     3072     2048     1024      256
  tok/s   89.28    89.88    90.44    90.79    91.33
  2048, not the faster 256: 512 and 768 sit in a reproducible acceptance dip (tau 1.2673 and
  1.2549), so the narrow end is not a broad peak on this prompt, and 2048 is the constant the
  band below already uses.

tau at 32k moves 1.3333 -> 1.2800, which is 96.0% of main and clears the 95% floor. It is not a
noisy quantity here: 128 tokens over 96 steps against 128 over 100.
```

### DIFF
--- a/runtime/src/models/dflash_draft.cpp
+++ b/runtime/src/models/dflash_draft.cpp
@@ -1231,7 +1231,22 @@ bool DFlashDraftModel::forward_block(const void* target_hidden, int ctx_len,
     static const int kMidCtxWindow = 2048;
     static const int kMidCtxMinSeq = 12288;
     static const int kMidCtxMaxSeq = 24576;
-    static const int kLongCtxWindow = 4096;
+    // The long band keeps 4096 while the mid band was narrowed to 2048, but the reason the mid
+    // band moved applies harder here, not less: acceptance falls with context while the draft's
+    // attention cost rises with it, so at 32k old context buys even less than it does at 16k.
+    // Measured at ctx=32768 on bench_prompt_32k.txt, lossless at every point, MEAN_ACCEPT
+    // identical (1.2800) across the whole range -- this is draft time removed, not acceptance
+    // traded away:
+    //
+    //     window   4096     3072     2048     1024      256
+    //     tok/s   89.28    89.88    90.44    90.79    91.33
+    //
+    // 2048 rather than the faster 256, for the reason the mid band chose 4096 over 3072: 512 and
+    // 768 sit in a reproducible acceptance dip (tau 1.2673 and 1.2549), so the narrow end is not a
+    // broad peak on this prompt and picking its best point is fitting one corpus. 2048 is the
+    // constant the band below already uses, which collapses the window ladder to one value above
+    // 12288.
+    static const int kLongCtxWindow = 2048;
     int kFullWindow = kFullWindowEnv >= 0 ? kFullWindowEnv : 3 * c.sliding_window;
     if (kFullWindowEnv < 0 && kFullWindow <= 0) {
         const int total_ctx = past + ctx_len;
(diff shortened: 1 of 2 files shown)


skyrocket2026:

| metric | value |
|---|---|
| **PR DSpark @4k** | **126.85 tok/s** |
| **main DSpark @4k** | **127.47 tok/s** |
| DSpark decode @4k vs main | -0.4% |
| **PR prefill @4k** | **14516.2 pp/s** |
| **main prefill @4k** | **14527.3 pp/s** |
| **PR DSpark @16k** | **128.65 tok/s** |
| **main DSpark @16k** | **128.57 tok/s** |
| **PR prefill @16k** | **11756.3 pp/s** |
| **main prefill @16k** | **11723.0 pp/s** |
| prefill @16k vs main | +0.2% |
| **PR DSpark @32k** | **91.91 tok/s** |
| **main DSpark @32k** | **89.64 tok/s** |
| DSpark decode @32k vs main | +2.5% |
| **PR prefill @32k** | **9738.3 pp/s** |
| **main prefill @32k** | **9758.1 pp/s** |
| prefill vs main | -0.2% |
| **PR prefill @256k** | **3365.5 pp/s** |
| **main prefill @256k** | **3353.9 pp/s** |
| **PR decode @256k** | **94.99 tok/s** |
| **main decode @256k** | **95.00 tok/s** |
| prefill @256k vs main | +0.3% |
| PR AR tok/s (floor) | 86.81 |
| main AR tok/s (floor) | 86.62 |
| AR vs main (floor) | +0.2% |
| **DSpark vs AR** | **1.482×** — above 1.0 means speculation finally pays |
| mean accept τ @32k | 1.2800 (main 1.3333) — ✅ |
| qwen3.8 (upstream unsloth) guard @16k | ✅ no regression (decode+prefill) — decode 83.2 tok/s · prefill 10072 pp |
| qwen3.6 guard @16k | ✅ no regression (decode+prefill) — decode 464.2 tok/s · prefill 27037 pp |
| commit | `b97f6ce67` |

ok

skyrocket2026:

PR #973, opened 2026-09-06
perf(runtime): park a session's decode graph instead of destroying it on every switch (1.24x cb-decode@c2/c4/c8)

Paral1995:
## Summary

`Qwen35Model::activate_session()` ended in `invalidate_decode_graph()`, because every node in the
captured decode graph bakes the active session's device pointers — `lin_state`, `lin_conv_state`,
`penalty_counts`, `logit_bias`, and `kv->block_table(active_seq_id)`. The continuous-batch worker
calls `activate_session()` for whichever job it is stepping, so once two requests are in flight it
switches sessions on **every token**, and every token pays a full capture + instantiate of the
whole 64-layer decode graph.

This parks the outgoing session's graph under its own `seq_id` and restores the incoming one, so a
switch is a pointer swap. The invariant is unchanged: a decode graph is still only ever replayed
for the session it was captured for. It is safe because every pointer a decode node bakes is stable
for that session's lifetime — the per-session buffers are allocated once in `open_session()`, and
`block_table(seq)` resolves to `d_block_tables + slot * max_blocks_per_seq`, a fixed slab offset
whose *contents* grow as blocks are appended but whose *address* does not. Replaying one capture
across a session's whole decode phase already relied on exactly that.

`invalidate_decode_graph()` drops the parked graphs too — it means "something the capture depends
on changed", and a graph parked under the old setting is as stale as the active one. The prefill
and DFlash graphs keep the old teardown on a switch: they bake the same session-owned pointers but
neither is replayed often enough for parking to pay. `activate_session()` and
`drop_parked_decode_graphs()` take `device_mu`, because destroying a raw graph handle unlocked was
merely racy while mutating a `std::unordered_map` unlocked is undefined, and two callers
(`set_dflash_capture`, `restore_spec_snapshot`) do not already hold it.

## Proof of speedup

**Decode tok/s** — aggregate across concurrent requests at **concurrency 2**:

| | decode tok/s |
|---|--:|
| before (main) | 75.50 |
| after (this PR) | 93.95 |

Full ladder, at exactly the settings `cb-decode@cN` uses (`qwen3_gguf_cb_bench <dir> C 256 256
512`). RTX 5090 32 GB, Qwen3.8-27B-NVFP4 (ModelOpt), base `8b56a8b`, arms **alternated**, 2 reps
each:

| dimension | before agg tok/s | after agg tok/s | change | before mean ITL | after mean ITL |
|---|--:|--:|--:|--:|--:|
| c1 (floor, not scored) | 92.05 | 93.15 | +1.2% | 11.11 ms | 10.96 ms |
| **cb-decode@c2** | 75.50 | 93.95 | **+24.4%** | 26.80 ms | 21.44 ms |
| **cb-decode@c4** | 76.00 | 94.30 | **+24.1%** | 52.44 ms | 42.07 ms |
| **cb-decode@c8** | 75.90 | 94.50 | **+24.5%** | 103.94 ms | 83.26 ms |

The c1 floor moves the right way (+1.2%), so the concurrency gain is not bought from the
single-stream path.

```text
$ build/runtime/qwen3_gguf_cb_bench <ModelOpt-NVFP4-dir> <C> 256 256 512

         c1              c2              c4              c8
BASE   92.1 / 92.0    75.6 / 75.4    76.1 / 75.9    75.9 / 75.9
PARK   93.2 / 93.1    94.0 / 93.9    94.3 / 94.3    94.5 / 94.5
```

**Single-stream AR decode and prefill are unmoved** — with one session there is no switch:

|---|--:|--:|
| before (main) | 94.442 | 16157.3 |
| after (this PR) | 94.463 | 16134.5 |

## Correctness

`dspark_tau_check` @ctx=4096 — the speculative path shares `qwen35.cpp`, so it is the regression
that matters here:

| | AR tok/s | DSpark tok/s | mean accepted | LOSSLESS |
|---|--:|--:|--:|--:|
| before (main) | 90.6078 | 127.9945 | **1.6883** | 1 |
| after (this PR) | 90.6325 | 128.0327 | **1.6883** | 1 |

Acceptance is identical to the digit and losslessness holds, i.e. this changes when a graph is
rebuilt, never what it computes.

## Note

Rebased onto `8b56a8b`. The earlier revision of this branch also carried a `qwen3_gguf_cb_bench`
change to make the compressed-tensors checkpoint benchmarkable under concurrency; that is now in
main and the file is a pinned harness path, so this branch is just the one-file fix plus its
header declaration.

### DIFF
--- a/runtime/include/sparkinfer/models/qwen35.h
+++ b/runtime/include/sparkinfer/models/qwen35.h
@@ -545,6 +545,10 @@ public:
 
 private:
     void invalidate_decode_graph();
+    // Frees every decode graph parked under a non-active session id (see the parking lot in
+    // qwen35.cpp's Impl). Called by invalidate_decode_graph(), which is the "something global
+    // changed" path, and as a size backstop.
+    void drop_parked_decode_graphs();
     void dflash_maybe_capture_layer(int layer);
     // Depth-adaptive KV-split count for a given seqlen (32/128/160/256 tiers, GQA-8/hd256
     // occupancy correction). Shared by forward_token()'s normal per-token adaptation and
--- a/runtime/src/models/qwen35.cpp
+++ b/runtime/src/models/qwen35.cpp
@@ -239,6 +239,41 @@ struct Qwen35Model::Impl {
     int dflash_graph_attn_mode = -1;
     bool dflash_graph_sparse = false;
 
+    // Per-session parking lot for the AR decode graph.
+    //
+    // activate_session() used to DESTROY cu_graph outright, because every node in it bakes this
+    // session's device pointers -- lin_state / lin_conv_state / penalty_counts / logit_bias, and
+    // kv->block_table(active_seq_id). Correct, but it made a session switch cost a full capture +
+    // instantiate of the whole 64-layer decode, and the continuous-batch worker switches sessions
+    // on EVERY token once two requests are in flight. Measured on RTX 5090 / Qwen3.8-27B-NVFP4,
+    // that is ~1.09 ms per token: per-token wall time goes 13.07 ms at concurrency 1 to 14.16 ms
+    // at concurrency 2, so aggregate throughput FALLS 84.9 -> 74.4 tok/s going 1 -> 2.
+    //
+    // Parking the outgoing session's graph under its own seq_id and restoring the incoming one
+    // keeps exactly the same "a decode graph only ever runs for the session it was captured for"
+    // guarantee, while making a switch a pointer swap.
+    //
+    // Safe because every pointer a decode node bakes is stable for the session's lifetime: the
+    // per-session buffers are allocated once in open_session(), and block_table(seq) is
+    // d_block_tables + slot * max_blocks_per_seq -- a fixed slab offset whose CONTENTS grow as
+    // blocks are appended but whose ADDRESS does not. That is the same assumption the pre-existing
+    // code already made by replaying one capture across a session's whole decode phase.
+    //
+    // n_splits is parked with the graph: the adaptive-splits check invalidates on
+    // `want != s.n_splits`, so restoring a graph without restoring the value it was captured at
+    // would tear it straight back down.
+    struct ParkedDecodeGraph {
+        cudaGraph_t graph{};
+        cudaGraphExec_t exec{};
+        int attn_mode = -1;
+        bool sparse = false;
+        int n_splits = 0;
+    };
+    std::unordered_map<uint64_t, ParkedDecodeGraph> parked_graphs;
+    // Bounded purely as a leak backstop -- close_session() drops a session's entry, so in steady
+    // state this holds one graph per LIVE session, which is the server's concurrency limit.
+    static constexpr size_t kMaxParkedGraphs = 64;
+
     // scratch (bf16)
     bf16 *x, *xn, *q, *k, *v, *attn, *ao, *h, *hn, *routed, *shared;
     bf16 *qraw = nullptr, *qgate = nullptr;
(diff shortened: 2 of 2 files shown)


skyrocket2026:

| metric | value |
|---|---|
| **PR DSpark @4k** | **129.04 tok/s** |
| **main DSpark @4k** | **129.05 tok/s** |
| **PR prefill @4k** | **14104.1 pp/s** |
| **main prefill @4k** | **14064.7 pp/s** |
| **PR DSpark @16k** | **129.84 tok/s** |
| **main DSpark @16k** | **129.37 tok/s** |
| DSpark decode @16k vs main | +0.3% |
| **PR prefill @16k** | **11566.4 pp/s** |
| **main prefill @16k** | **11454.6 pp/s** |
| prefill @16k vs main | +0.9% |
| **PR DSpark @32k** | **92.99 tok/s** |
| **main DSpark @32k** | **92.99 tok/s** |
| **PR prefill @32k** | **10393.9 pp/s** |
| **main prefill @32k** | **10389.7 pp/s** |
| prefill vs main | +0.0% |
| **PR prefill @256k** | **4262.4 pp/s** |
| **main prefill @256k** | **4273.2 pp/s** |
| **PR decode @256k** | **94.46 tok/s** |
| **main decode @256k** | **94.50 tok/s** |
| prefill @256k vs main | -0.2% |
| main AR tok/s (floor) | 88.39 |
| AR vs main (floor) | +0.2% |
| **DSpark vs AR** | **1.466×** — above 1.0 means speculation finally pays |
| mean accept τ @32k | ✅ 1.2800 (main 1.2800) |
| qwen3.8 (upstream unsloth) guard @16k | ✅ no regression (decode+prefill) — decode 82.6 tok/s · prefill 10007 pp |
| qwen3.6 guard @16k | ✅ no regression (decode+prefill) — decode 461.9 tok/s · prefill 27172 pp |
| commit | `99723ad66` |

ok

skyrocket2026:

skyrocket2026:
**Correction to the auto-eval comment above — the table was incomplete, not the verdict.**

That comment shows `+0.0%` on every row while reporting `eval-dspark:XL`, which reads as an unexplained label. The verdict is real; the report was missing the axis it came from.

`eval-dspark:XL` was earned on **`cb-decode@c2`** — aggregate decode throughput with 2 concurrent requests — at **+33.2%**. That dimension was added to this bot hours before your PR was scored (#978), and every row the comment *does* show is measured at batch size 1, which is exactly the regime your change does not target. Under the previous dimension set this PR would have scored `none`.

Three things worth stating plainly:

1. **The scored number is higher than your own claim.** You reported 1.22x aggregate at concurrency 2, measured at 64 tokens/request. The bot measures 256 tokens/request, because a ~1s run is partly measuring its own startup — at 64 tokens the c=4 and c=8 numbers vary 3.4% run to run on *identical* code, against a -2% reject threshold. At the longer length your change measures +33.2%.

2. **The absolute values are not recoverable from that round.** The payload builder had not been updated when your PR was evaluated, so the c1/c2/c4/c8 tok/s figures were computed, used for the tier, and then not persisted. Only the axis and the delta survive. Fixed in `5acecbc` — the table now carries the per-concurrency rows, the c=1 floor, and a "winning axis" row, and renders `—` rather than `+0.0%` for anything not measured.

3. **The c=1 floor held** (that is what stops a PR buying concurrency scaling by slowing single-stream), along with losslessness, τ=1.730, top1=1.0, KL=0.0 and both no-regression guards.

Apologies for the confusing report — that was a gap in the dimension I added, not in your work.

PR #975, opened 2026-09-06
perf(runtime): decode a whole continuous-batch in one packed forward (1.95x/3.29x/5.11x cb-decode@c2/c4/c8)

Paral1995:
## Summary

`ContinuousBatchEngine::worker_loop()` built a decode batch and then ran it **one sequence at a
time** — `for (id : decode_ids) step_job(*job)`, each a full 64-layer `forward_token()`. Decode is
bandwidth-bound on weight reads, so N concurrent requests read every weight N times and aggregate
throughput does not scale with concurrency at all. The scheduler was already producing the batch;
only execution was unpacked.

This executes it as one forward. Almost everything needed was already in the tree, used by the
speculative verifier: the projections and FFN take R rows through a single weight read, the paged
attention takes `num_seqs`, and the KV-append kernels are already instantiated for a per-row block
table (`SINGLE_SEQUENCE=false`). So packed decode is `dflash_verify_short_run`'s forward with four
substitutions, gated on the new `Qwen35PrefillCtx::packed_rows`:

- the block table is **gathered** per row instead of broadcast from one sequence;
- the GDN block runs a **batched per-row AR step** (first commit) instead of the compact scan,
  advancing each session's conv window and recurrent state in place;
- the KV-append uses the **per-row-table** kernels instead of the single-sequence ones;
- there is **no accepted-prefix commit**, because every row is already a real decode step.

Every per-session pointer reaches the kernels through a **device array** whose contents the caller
rewrites each step, so one captured graph per row count serves any set of sessions. A batch wider
than the graph tiers is **split into chunks** rather than declined — declining fell back to one
forward per sequence, so crossing the cap cost more than never packing at all.

`step_jobs_packed()` declines, having changed nothing, whenever a row would not decode identically
to what `step_job` would have produced: still prefilling, teacher-forced, per-token logprobs, or
any sampler setting other than plain greedy. `SPARKINFER_PACKED_DECODE=0` disables it.

## Baseline

Rebased onto `f157a2a`. The session-graph-parking change this branch previously carried is now in
main, so these three commits are all that is left here, and the baseline below is main as it now
stands — the numbers isolate what packed decode itself contributes.

## Proof of speedup

**Decode tok/s** — aggregate across concurrent requests at **concurrency 2**:

| | decode tok/s |
|---|--:|
| before (main) | 93.95 |
| after (this PR) | 183.00 |

Full ladder, at exactly the settings `cb-decode@cN` uses (`qwen3_gguf_cb_bench <dir> C 256 256
512`). RTX 5090 32 GB, Qwen3.8-27B-NVFP4 (ModelOpt), base `8b56a8b`, arms **alternated**, 2 reps:

| dimension | main | this PR | change | before mean ITL | after mean ITL |
|---|--:|--:|--:|--:|--:|
| c1 (floor, not scored) | 93.20 | 95.25 | +2.2% | 10.96 ms | 10.72 ms |
| **cb-decode@c2** | 93.95 | 183.00 | **1.95×** | 21.46 ms | 10.87 ms |
| **cb-decode@c4** | 94.35 | 310.80 | **3.29×** | 42.05 ms | 12.34 ms |
| **cb-decode@c8** | 94.50 | 482.70 | **5.11×** | 83.26 ms | 15.49 ms |

Against main **before** the graph-parking change landed, i.e. what the two changes deliver
together: c2 75.50 → 183.00 (**2.42×**), c4 76.00 → 310.80 (**4.09×**), c8 75.90 → 482.70
(**6.36×**). Aggregate throughput was flat at ~75 tok/s at every concurrency; it now scales. Per-request latency improves as well — 5.4× at
concurrency 8 — because a request no longer queues behind seven other forwards.

The c1 floor moves the right way (+2.2%), so nothing here is bought from the single-stream path.

```text
$ build/runtime/qwen3_gguf_cb_bench <ModelOpt-NVFP4-dir> <C> 256 256 512

          c1              c2               c4               c8
main      93.2 / 93.2    94.0 / 93.9     94.4 / 94.3     94.5 / 94.5
this PR   95.3 / 95.2   183.0 / 183.0   311.1 / 310.5   483.4 / 482.0
```

**Single-stream AR decode and prefill are unmoved** — one sequence is never packed:

|---|--:|--:|
| before (main) | 94.474 | 16096.9 |
| after (this PR) | 94.465 | 16122.0 |

## Correctness

- **`gdn_batched_gpu_test`** (new): the batched GDN kernels against B sequential unbatched
  launches, **bit for bit**, on outputs *and* the updated recurrent state, at B = 1/2/4/8 for both
  v-head counts, both q-head broadcast conventions, and non-zero layer offsets. Bit-exactness
  rather than closeness because the state is carried across every decode step, so a last-bit
  difference compounds.
- **`batch_engine_gpu_test`** passes.
- **DSpark is untouched** — packed mode shares `dflash_verify_short_run`, so this is the regression
  that matters:

| | AR tok/s | DSpark tok/s | mean accepted | LOSSLESS |
|---|--:|--:|--:|--:|
| before (main) | 90.5933 | 128.0223 | **1.6883** | 1 |
| after (this PR) | 90.6478 | 128.2359 | **1.6883** | 1 |

- **Emitted tokens match the sequential path.** Compared with `SPARKINFER_PACKED_DECODE=1` against
  `=0` in the same binary, so packing is the only variable. At c2, c4 and c8 the generated ids are
  the same, and every stream's head is identical to the unpacked run's.

  One honest caveat: at c4 the *packed* run reported its four same-prompt streams diverging from
  each other later in the sequence (c2 and c8 did not, and the heads matched everywhere). Rows
  share no state — the GDN kernels are bit-exact per row, each row has its own KV, conv window and
  recurrent state — but the host-side attention-arm choice is made per launch from the batch's
  longest sequence, so batch composition can change a reduction order and flip a near-tie. That is
  the same class of effect main already shows across concurrency levels; it can be removed by
  pinning the arm independently of batch composition, at some throughput cost, if reviewers prefer
  determinism there.

## Note

Measurements were taken on the same code this branch now contains, with main's
graph-parking change applied — which is what main became when that landed, so they carry over
unchanged. An earlier revision of this branch carried a `qwen3_gguf_cb_bench` change; that file is
a pinned harness path and the equivalent support is already in main, so this branch touches no
harness. The token-equality check above was run with a temporary local instrument, not shipped
here.

### DIFF
--- a/kernels/csrc/cuda/fused/qwen36.cu
+++ b/kernels/csrc/cuda/fused/qwen36.cu
@@ -310,6 +310,89 @@ __global__ void gdn_ar_fast_kernel(const __nv_bfloat16* __restrict__ q,
     if (lane == 0) out[(size_t)vh * HEAD_DIM + j] = __float2bfloat16(y);
 }
 
+// ---------------------------------------------------------------------------
+// BATCHED decode step: B INDEPENDENT sequences, one token each.
+//
+// Not the same shape as the chunked/compact scans next door. Those walk N CONSECUTIVE positions
+// of ONE sequence and therefore carry a serial dependence between rows. Here each row is a
+// different request advancing its own recurrent state by exactly one step, so the rows are fully
+// independent and the batch is just another grid axis.
+//
+// Why it exists: the continuous-batch worker runs one full 64-layer forward PER SEQUENCE, and
+// decode is bandwidth-bound on weight reads, so N concurrent requests read every weight N times
+// and aggregate throughput does not scale with concurrency. Packing the rows into one forward
+// fixes that, and the 48 Gated-DeltaNet layers are the only stage with no batched decode form --
+// the projections/FFN already take R rows through one weight read, and the paged attention
+// already takes num_seqs.
+//
+// The per-row state pointers arrive as a DEVICE ARRAY rather than a base + stride, because each
+// session's lin_state / lin_conv_state is its own cudaMalloc. That also keeps this CUDA-graph
+// safe: the graph bakes the ADDRESS of the pointer array, and the packer refreshes its CONTENTS
+// before each replay, the same trick d_scalars already uses for token id / position.
+//
+// Row b reads activation row b and state states[b]; with batch == 1 and states[0] equal to what
+// the unbatched launcher is handed, every lane does bit-identical arithmetic in the same order,
+// so a packed step and a sequential step agree exactly rather than approximately.
+template <int COLS, int HEAD_DIM>
+__global__ void gdn_ar_fast_batched_kernel(const __nv_bfloat16* __restrict__ q,
+                                           const __nv_bfloat16* __restrict__ k,
+                                           const __nv_bfloat16* __restrict__ v,
+                                           const __nv_bfloat16* __restrict__ alpha,
+                                           const __nv_bfloat16* __restrict__ beta,
+                                           const __nv_bfloat16* __restrict__ dt,
+                                           const __nv_bfloat16* __restrict__ a,
+                                           float* const* __restrict__ states,
+                                           size_t state_off,
+                                           __nv_bfloat16* __restrict__ out,
+                                           int q_heads, int v_heads, bool qh_block,
+                                           bool state_bf16) {
+    constexpr int NROW = HEAD_DIM / 32;
+    const int vh   = blockIdx.x;
+    const int j    = blockIdx.y * COLS + (threadIdx.x >> 5);
+    const int b    = blockIdx.z;
+    const int lane = threadIdx.x & 31;
+    if (vh >= v_heads || j >= HEAD_DIM) return;
+    const int qh   = qh_block ? (vh / (v_heads / q_heads)) : (vh % q_heads);
+    const float scale = rsqrtf((float)HEAD_DIM);
+    // Per-row activation strides. dt/a are WEIGHTS (one value per v-head, shared by every row)
+    // and are deliberately not strided; alpha/beta are per-token projections and are.
+    const int qdim = q_heads * HEAD_DIM, vdim = v_heads * HEAD_DIM;
+    const __nv_bfloat16* qrow = q + (size_t)b * qdim;
+    const __nv_bfloat16* krow = k + (size_t)b * qdim;
+    const __nv_bfloat16* vrow = v + (size_t)b * vdim;
+    const __nv_bfloat16* arow = alpha + (size_t)b * v_heads;
+    const __nv_bfloat16* brow = beta  + (size_t)b * v_heads;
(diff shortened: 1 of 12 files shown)


skyrocket2026:

| metric | value |
|---|---|
| winning axis | **cb-decode@c8** — the dimension this label came from |
| **PR DSpark @4k** | **129.13 tok/s** |
| **main DSpark @4k** | **128.66 tok/s** |
| DSpark decode @4k vs main | +0.3% |
| **PR prefill @4k** | **14100.1 pp/s** |
| **main prefill @4k** | **14064.3 pp/s** |
| **PR DSpark @16k** | **130.25 tok/s** |
| **main DSpark @16k** | **129.05 tok/s** |
| DSpark decode @16k vs main | +0.9% |
| **PR prefill @16k** | **11588.6 pp/s** |
| **main prefill @16k** | **11347.2 pp/s** |
| prefill @16k vs main | +2.1% |
| **PR DSpark @32k** | **93.01 tok/s** |
| **main DSpark @32k** | **92.96 tok/s** |
| **PR prefill @32k** | **10410.6 pp/s** |
| **main prefill @32k** | **10383.6 pp/s** |
| prefill vs main | +0.2% |
| **PR prefill @256k** | **4286.7 pp/s** |
| **main prefill @256k** | **4271.0 pp/s** |
| **PR decode @256k** | **94.49 tok/s** |
| **main decode @256k** | **94.50 tok/s** |
| prefill @256k vs main | +0.3% |
| **PR concurrent decode @c2** | **180.4 tok/s agg** |
| **main concurrent decode @c2** | **93.9 tok/s agg** |
| concurrent decode @c2 vs main | +92.1% |
| **PR concurrent decode @c4** | **301.3 tok/s agg** |
| **main concurrent decode @c4** | **94.2 tok/s agg** |
| concurrent decode @c4 vs main | +219.9% |
| **PR concurrent decode @c8** | **459.7 tok/s agg** |
| **main concurrent decode @c8** | **94.2 tok/s agg** |
| concurrent decode @c8 vs main | +388.0% |
| main concurrent decode @c1 (floor) | 92.7 tok/s agg |
| concurrent @c1 vs main (floor) | +2.5% |
| main AR tok/s (floor) | 88.45 |
| **DSpark vs AR** | **1.471×** — above 1.0 means speculation finally pays |
| mean accept τ @32k | ✅ 1.2800 (main 1.2800) |
| qwen3.8 (upstream unsloth) guard @16k | ✅ no regression (decode+prefill) — decode 82.5 tok/s · prefill 9944 pp |
| qwen3.6 guard @16k | ✅ no regression (decode+prefill) — decode 461.8 tok/s · prefill 27597 pp |
| commit | `18c7753c8` |

ok

skyrocket2026:

PR #980, opened 2026-09-06
fix(prefill): write the GQA P' plane with the stride it is read with (#976)

### LINKED ISSUE #976: Long-context output degrades and becomes nondeterministic from ~4k tokens (vLLM on the same checkpoint does not)
## Summary

Past roughly 4,000 prompt tokens, greedy decode stops being reproducible and the answers stop being correct. It is not the weights, the prompt, or the quantization: **vLLM v0.28 serving the identical checkpoint answers correctly and stably at the same context lengths.**

Reproduced on the released image `ghcr.io/gittensor-ai-lab/sparkinfer-qwen38:0.5.5` with no prefix cache and no speculation.

## Reproduction

RTX 5090, `--ctx 40960`, checkpoint [`gittensor-model-hub/Qwen3.8-27B-NVFP4-RTX5090`](https://huggingface.co/gittensor-model-hub/Qwen3.8-27B-NVFP4-RTX5090). A natural-prose system message of varying length, then one question with an unambiguous answer, `temperature=0`, `enable_thinking=false`, three runs each:

```
"What is 17 multiplied by 23? Reply with only the number."      (answer: 391)
```

| prompt tokens | verdict | three runs |
|---:|---|---|
| 1,955 | OK | `391`, `391`, `391` |
| 4,076 | FLAKY | three different non-answers |
| 7,643 | stable but WRONG | `The user wants to know t…` ×3 |
| 12,053 | FLAKY | `The user wants to know t…`, `23`, `To find the product of 1…` |
| 16,362 | FLAKY | `The user wants to know t…`, `I cannot answer that que…`, … |
| 20,649 | FLAKY | `We are given a long text…`, `I'm sorry, but I can't a…`, … |
| 24,869 | FLAKY | `391`, `299`, `46` |
| 31,263 | FLAKY | three different answers |

Two things go wrong together: the answer becomes **wrong**, and it stops being **reproducible** for greedy decode, which should be deterministic.

## Control: vLLM, same checkpoint, same prompts, same box

```
vllm/vllm-openai:v0.28.0 --model /data/final --quantization modelopt \
  --kv-cache-dtype fp8 --trust-remote-code --max-model-len 40960
```

| prompt style | prompt tokens | sparkinfer 0.5.5 | vLLM v0.28 |
|---|---:|---|---|
| short | 29 | `391` stable | `391` stable |
| natural prose | 30,498 | `17`, `451`, `437` | **`391` stable** |
| degenerate repetition | 32,044 | flaky prose | **`391` stable** |

Same weights, same tokenizer, same prompts. So this is on our side.

## Things already ruled out

- **Sparse sliding-window KV.** `SPARKINFER_SPARSE_KV=0` does not change the behaviour (still `431`, `2`, `851` at 30k). Also the onset is ~4k, well below the sparse path's `min_ctx=16384`.
- **The prompt.** Reproduces with natural prose and with degenerate repeated text.
- **Prefix caching / speculation.** Neither is enabled in the repro.
- **The checkpoint.** vLLM control above.

## Why this matters

Long context is a headline capability for this engine — it holds 262,144 tokens on a 32 GB card, and decode stays flat at ~90 tok/s out to 244k, which is genuinely better than vLLM's 63.5 at that depth. But those are **throughput** measurements. If output quality degrades from ~4k, the speed numbers describe generating text that is increasingly wrong, and we should not present flat-decode-to-244k as a long-context quality claim until this is fixed.

It also means our accuracy smoke does not cover this: it runs at short context, which is exactly the regime that still works.

## Suggested starting points

- The onset near 4k is suspiciously close to the sparse path's `window=256 blocks (4096 tokens)`, even though disabling `SPARKINFER_SPARSE_KV` does not help — worth checking whether the window still influences the compact view or block-table construction when nominally off.
- Batched prefill tile decomposition at larger `M` (`qwen35_prefill.cpp`), since prefill is where a long prompt is consumed.
- int8 KV (`SPARKINFER_KV_INT8=1` by default here): compare against fp16 KV at the same context to see whether quantized KV accumulates error with depth.
- Nondeterminism at fixed greedy decode suggests a race or an uninitialised read that scales with block count, not merely precision loss.

A useful regression test would be exactly the table above: one arithmetic question, a growing natural-prose prefix, three runs, asserting identical and correct output at each depth.

skyrocket2026:
Fixes #976 — long-context greedy decode was simultaneously **wrong** and **nondeterministic** from
~2048 prompt tokens.

## Root cause

The fused GQA prefill attention **allocates** its P′ plane with a `pld` row stride (`s_pi` is
`[RQH][BM][pld]`, and `s_s` begins at `s_pi + RQH*BM*pld`) and **every reader uses `pld`** — the
`ldmatrix` feeding the PV mma, and the base pointer. Only the **write** used `GN`.

Below 2048 tokens `pad == 0`, so `pld == GN` and the two agreed — which is why this survived every
existing test. At or above 2048, `pad == 16`:

- each row is written 16 columns short of where it is read, so **row r comes back shifted by 16r**;
- the **last row reads past everything any row wrote**, off the end of the initialised region into
  whatever shared memory happened to hold.

Wrong output, run-to-run variation under greedy decode, and a cliff exactly at 2048 all follow from
that one mismatch. `pstr` is now `pld` unconditionally, matching the reader and the allocation.

## Triangulation, before fixing

At 6,812 prompt tokens on the released checkpoint, greedy, three runs each:

| path | result |
|---|---|
| batched prefill, default (RQH=4) | three **different** wrong answers |
| batched prefill, RQH=1 fallback | `391`, `391`, `391` |
| token-loop prefill | `391`, `391`, `391` |

After the fix, on **default settings** (no `DETERMINISTIC`, no `RQH` override):

| prompt tokens | before | after |
|---:|---|---|
| 1,800 | `391` ×3 | `391` ×3 |
| 6,812 | 3 different wrong answers | **`391` ×3** |
| 21,453 | flaky, wrong | **`391` ×3** |

## The cost, measured like-for-like

| | before | after | delta |
|---|---:|---:|---:|
| prefill@16k | 13577.49 pp/s | 13516.13 pp/s | **-0.45%** |
| prefill@32k | 11526.97 pp/s | 11504.38 pp/s | **-0.20%** |

## This was known, and the trade-off was never real

The comment above the `RQH` default already recorded the cliff, a measured KL table
(`0.00022` at 2000 → `0.18672` at 2100, "and varies run to run"), and the decision to keep the
fused tiers on because turning them off "moves the long-context prefill numbers the eval scores
against."

That reasoning was about **disabling GQA fusion**, which genuinely costs throughput. Nobody tried
**repairing the stride**, which keeps the fusion — and costs 0.2–0.45%. Reading uninitialised
shared memory is not "their numerics" either way.

The same comment says the defect "is reported separately". **No such issue was ever filed.** #976
came from an outside user two days later, having noticed that vLLM on the identical checkpoint
answered correctly at the same depths.

## Why CI could not see it

The accuracy smoke runs at short context — precisely the regime that still worked. Long context is
a headline claim for this engine (262,144 tokens on a 32 GB card, flat decode to 244k), and those
are throughput numbers; while this was live they described generating increasingly wrong text.

`bench/scripts/long_context_determinism_check.py` closes the hole: one arithmetic question with an
unambiguous answer, a growing natural-prose prefix, N greedy runs per depth, asserting **both**
that every answer is correct **and** that the runs at each depth are identical. Correctness alone
would pass a stably-wrong build; determinism alone would pass a consistently-garbage one. The
default depths straddle the 2048 cliff.

### DIFF
--- a/bench/scripts/long_context_determinism_check.py
+++ b/bench/scripts/long_context_determinism_check.py
@@ -1,0 +1,101 @@
+#!/usr/bin/env python3
+"""Long-context greedy decode must be CORRECT and REPRODUCIBLE (issue #976).
+
+The accuracy smoke this repo already runs works at short context -- which is precisely the regime
+that still worked while #976 was live. The fused GQA prefill attention wrote its P' plane with row
+stride GN while allocating and reading it with pld; below 2048 tokens pad==0 so the two agreed, and
+at or above 2048 every row was read shifted by 16r with the last row reading past all initialised
+shared memory. Output was simultaneously wrong and nondeterministic, and nothing in CI could see it.
+
+So this asks the narrowest question that would have caught it: one arithmetic question with an
+unambiguous answer, a growing natural-prose prefix, N greedy runs at each depth. Two assertions,
+and the second is the one that matters -- greedy decode must be a function of its input:
+
+  1. every run at every depth answers correctly
+  2. the runs at a given depth are IDENTICAL to each other
+
+A correctness-only check would pass on a build that is stably wrong; a determinism-only check
+would pass on a build that is consistently garbage. Both are needed.
+
+Usage:
+  python3 bench/scripts/long_context_determinism_check.py --base-url http://127.0.0.1:8080
+"""
+import argparse, json, sys, urllib.request
+
+# Deliberately trivial arithmetic: the answer is not in dispute, so any deviation is the engine's.
+QUESTION = "What is 17 multiplied by 23? Reply with only the number."
+ANSWER = "391"
+
+
+def ask(base_url, prose, max_tokens, timeout):
+    body = json.dumps({
+        "model": "q", "max_tokens": max_tokens, "temperature": 0.0,
+        "enable_thinking": False,
+        "messages": [{"role": "user", "content": (prose + "\n\n" + QUESTION) if prose else QUESTION}],
+    }).encode()
+    req = urllib.request.Request(base_url.rstrip("/") + "/v1/chat/completions",
+                                 data=body, headers={"Content-Type": "application/json"})
+    with urllib.request.urlopen(req, timeout=timeout) as r:
+        d = json.load(r)
+    if "error" in d:
+        raise RuntimeError(json.dumps(d["error"])[:200])
+    msg = d["choices"][0]["message"]
+    return (msg.get("content") or "").strip(), d.get("usage", {}).get("prompt_tokens", 0)
+
+
+def main():
+    ap = argparse.ArgumentParser()
+    ap.add_argument("--base-url", default="http://127.0.0.1:8080")
+    ap.add_argument("--corpus", default="bench/scripts/bench_prompt_32k.txt")
+    ap.add_argument("--runs", type=int, default=3)
+    ap.add_argument("--max-tokens", type=int, default=24)
+    ap.add_argument("--timeout", type=float, default=900.0)
+    # Prefix lengths in CHARACTERS. The defaults straddle the 2048-token cliff #976 lived at:
+    # one clearly below, then several above, since a check that only probes the working regime is
+    # exactly the hole this closes.
+    ap.add_argument("--chars", default="0,8000,30000,90000,200000")
+    args = ap.parse_args()
+
+    try:
+        corpus = open(args.corpus).read()
+    except OSError as e:
+        print(f"cannot read corpus: {e}")
+        return 2
+
+    failures = 0
+    for spec in [int(c) for c in args.chars.split(",")]:
+        prose = corpus[:spec]
+        outs = []
+        toks = 0
+        for _ in range(args.runs):
+            try:
+                text, toks = ask(args.base_url, prose, args.max_tokens, args.timeout)
+            except Exception as e:                      # noqa: BLE001 -- report, do not abort the sweep
+                print(f"  chars={spec:<7} REQUEST FAILED: {e}")
+                failures += 1
+                outs = None
+                break
+            outs.append(text)
+        if outs is None:
+            continue
+
+        stable = len(set(outs)) == 1
+        # Substring, not equality: a correct engine may legitimately answer "391" or "391." etc.
+        correct = all(ANSWER in o for o in outs)
+        status = "OK" if (stable and correct) else ("WRONG" if stable else "FLAKY")
+        print(f"  chars={spec:<7} tokens={toks:<7} {status:<6} " +
+              " | ".join(repr(o[:26]) for o in outs))
+        if not (stable and correct):
+            failures += 1
+
+    print()
+    if failures:
+        print(f"FAILED — {failures} depth(s) wrong or nondeterministic")
+        print("Greedy decode must be a function of its input; see issue #976.")
+        return 1
+    print("PASSED — every depth is correct and reproducible")
+    return 0
+
+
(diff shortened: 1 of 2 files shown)


ai-hpc:
Independently reproduced and verified this fixes #976. Built this PR into the server image and re-ran the exact threshold repro from the issue (RTX 5090, greedy, `temperature=0`, natural-prose prefix, arithmetic question with a known answer of `391`, three runs per depth):

| prompt tokens | before (#976) | this PR |
|---:|---|---|
| 1,955 | `391` ×3 | `391` ×3 |
| 4,076 | 3 different non-answers | **`391` ×3** |
| 7,643 | stable but wrong | **`391` ×3** |
| 12,053 | flaky | **`391` ×3** |
| 16,362 | flaky | **`391` ×3** |
| 20,649 | flaky | **`391` ×3** |
| 24,869 | `391`, `299`, `46` | **`391` ×3** |
| 31,263 | 3 different answers | **`391` ×3** |

Also the specific degenerate-prompt case from the issue (`391` → `17`/`451`/`437` at ~32k) now returns `391` three times at 32,044 prompt tokens.

The root cause matches the observed cliff exactly: the P′-plane write used `GN` while allocation and every reader used `pld`, and they only diverge once `pad != 0` at ≥2048 tokens. One-line change, `pstr = pld`. Nice catch.

Worth adding `bench/scripts/long_context_determinism_check.py` (included in this PR) to the gated eval, since the existing suite never exercised ≥2048 with `pad != 0` — which is why this shipped in the first place.

skyrocket2026:
Thank you for the independent reproduction — that table is exactly the confirmation this needed, and the degenerate-prompt case going from `17`/`451`/`437` to `391` ×3 at 32,044 tokens is a better check than anything I ran.

**Acted on your suggestion, and it turned out to need more than adding the script.**

Looking at *why* the existing suite missed this: the losslessness gate does run three independent processes at 4k/16k/32k, all above 2048 — so the regime *was* exercised. It still passed, because each process only asks whether DSpark matches AR **inside itself**. With a nondeterministic shared prefill both legs read the same corrupted state, agree with each other, and report `lossless=1`. #976 was wrong *and* different on every run with all three reps green.

So rather than bolt a server into the eval round, the gate now hashes each rep's AR token stream and requires the hashes to agree **across** the three processes (`dbe076e`). Greedy decode is a function of its input; two clean starts on identical ids must emit identical tokens. A mismatch is a hard REJECT regardless of throughput.

Two properties worth noting:

- **No extra GPU time.** Those reps already run; this reads their output. It needs `SPARKINFER_DSPARK_DUMP_TOKENS=1`, because the 8-token prefix printed by default cannot detect a divergence later in the stream.
- **A missing marker reports `— not measured`, never a tick.** An older harness on the box must not make an absent gate look like a passing one — that failure mode is the whole lesson here.

`bench/scripts/long_context_determinism_check.py` ships in this PR as the end-to-end server-level check, and stays the right tool for the depths and prompt shapes the in-harness gate does not reach.

One correction to my own PR description while I'm here: I quoted the prefill cost as -0.45% @16k and -0.20% @32k, measured by hand. `hold` on this PR meant no eval round ever scored it, so **@4k and @256k were never measured** — and `pad` engages at ≥2048, so 4k is affected too. Those numbers will appear on the next round against the new baseline.

PR #984, opened 2026-09-06
perf(nvfp4): streamline paired GEMV and reuse dense prefill scratch

cryptoresetlife:
## Current revision — 2026-09-07 UTC

Published fork head **ad552ccaef3adb7bd1fec1a0b54d25aac1175cfd**, based on main **d9b0304b9c866d57835f7e28680582e1813ced8c**. The entire public tree **4ea9ebbfd6073d3ac45213238262cc6e2136d805** matches the tested local and GPU-host source, verified by fetching the public branch. Upstream changes, including #994, belong to the reference and are not attributed to this contribution.

**Status: local validation completed; official qualification unresolved.** The first official evaluation rejected OLD a9edbdd at DSpark32K (-3.38%). This PR remains closed. No official tier, merge, earnings or all-gates-passed claim. No bypass or duplicate PR requested.

### Implementation

- Carry-safe biased packed NVFP4 negation for row counts 2–4; single-row/wide paths retain the original arithmetic. The older paired group-order rewrite is removed.
- Serialize prefill graph capture/first launch against submit-time device operations using the existing device mutex.
- Warm the full model/session identity before graph capture instead of just the prompt length; retain capture for repeated same-session use.
- Retain our dense non-Muse scratch lifetime reuse, alignment/capacity checks and activation-quantization memo invalidation.
- Exhaustive GPU decoder, dense BF16 scratch, device-lock and warm-identity regression tests.

11 kernel/runtime files, 474 additions / 31 deletions against d9b0304. No evaluation, scoring, CI, governance, model, prompt or workload changes. AI-assisted development; these are our own changes.

### Proof of speedup — two separate fixed blocks

- [x] RTX 5090 (sm_120), same isolated GPU, frozen reference/candidate libraries.

| Block | Axis | Main median aggregate tok/s | Candidate median aggregate tok/s | Delta |
|---|---|---:|---:|---:|
| Initial | c4 | 309.5 | 319.2 | +3.1341% |
| Independent replication | c4 | 307.2 | 318.8 | +3.7760% |
| Initial | c32 | 938.7 | 961.4 | +2.4182% |
| Independent replication | c32 | 932.6 | 959.0 | +2.8308% |

[...]

PR #986, opened 2026-09-06
perf(gemv): default the NVFP4 multi-row GEMV to its pairwise kernel (1.025x cb-decode@c8)

inference2026:
## Summary

The NVFP4 multi-row GEMV ships two kernels for the same shape, and the one that suits a wide batch
was opt-in. `gemv_nvfp4_rows_dp4a_pairwise_kernel` is fully instantiated for R=2..8 and nothing
routed to it unless `SPARKINFER_NVFP4_ROWS_PAIRWISE=1` was set. This makes it the default; the flag
now opts out instead of in.

## Proof of speedup

- [x] Tested on **RTX 5090** (`sm_120`)

**Decode tok/s** (end-to-end, aggregate continuous-batch decode at concurrency 8):

| | decode tok/s |
|---|--:|
| before (main) | 467.1 |
| after (this PR) | 478.7 |

**Prefill pp tok/s** — not targeted by this PR:

| | prefill pp tok/s |
|---|--:|
| before prefill (main) | n/a |
| after prefill (this PR) | n/a |

### Why it is faster

The kernel differs from the branch it replaces only in *which* warp owns an output element, not in
how that element is computed: both use the same split rule (`N>=4096 ? S=2 : S=8`) and the same
lane-owned group walk (`g = split*32 + lane`, stride `S*32`), so every output sums the same terms
in the same order. What changes is the block geometry — 4 warps per block with NR=1 rather than 8
with NR=2, which at `N=17408, S=2` runs 8704 blocks of 128 threads instead of 2176 of 256. That
trade only pays once several activation rows share a weight read, which is why it was left off:
single-stream decode never reaches those widths.

### Bit-identity

Mean accept is **exactly equal** before and after at every speculative-decode context, with
losslessness holding in all six runs. Since mean accept is `tokens/steps`, an equal value means the
verifier accepted the same tokens at the same positions across a whole generation.

| context | before | after | mean accept before/after | lossless |
|---|--:|--:|---|--:|
| 4k | 128.99 | 128.96 | 1.6883 / 1.6883 | 1 / 1 |
| 16k | 129.45 | 129.30 | 1.7297 / 1.7297 | 1 / 1 |
| 32k | 92.98 | 94.22 | 1.2800 / 1.2800 | 1 / 1 |

### Across concurrency

Single-stream is untouched by construction: the branch requires `M>=2`, and continuous-batch decode
declines a batch narrower than two rows.

| concurrency | before | after | |
|---|--:|--:|--:|
[...]

PR #988, opened 2026-09-06
perf(runtime): serve a wide continuous-batch decode with the block-scaled NVFP4 GEMM (1.35x cb-decode@c32)

inference2026:
Reopened after the `target-prefill@256k` reject, rebased onto `c1777f5`. **That regression does not
reproduce** — evidence in "The 256k row" below. Everything else is re-measured on the new base,
because #989 changed the prefill path and the old numbers no longer described this tree.

## Summary

Aggregate decode throughput stops scaling past 8 concurrent requests: `kQwen35MaxPackedRows` is 8,
so a wider decode batch is chunked into 8s and every chunk re-reads all ~15 GB of weights. This
raises that ceiling to 32 and, above 16 rows, runs the dense FFN through the checkpoint-native
block-scaled NVFP4 GEMM instead of the row-GEMV — the one kernel that reads the weights once for
the whole batch.

## Proof of speedup

**Decode tok/s** (aggregate continuous-batch decode, 32 concurrent requests):

| | decode tok/s |
|---|--:|
| before (main) | 514.9 |
| after (this PR) | 697.5 |

**Prefill pp tok/s** — not targeted by this PR:

| | prefill pp tok/s |
|---|--:|

## The 256k row

The previous round rejected on `target-prefill@256k`: 3862.30 against main 4313.89, −10.5%. I could
not reproduce it. Five runs of the bot's exact command, arms sandwiched between controls so drift is
bounded rather than assumed:

| tree | arm | prefill pp |
|---|---|--:|
| pre-#989 | clean main | 4632.53 |
| pre-#989 | **this PR** | 4623.05 |
| pre-#989 | clean main (control after) | 4614.84 |
| `c1777f5` | clean main | 4620.33 |
| `c1777f5` | **this PR** | 4620.01 |

On the current base the PR lands within **0.007%** of clean. On the older tree it sits *between* two
clean controls that themselves differ by 0.38%. A sixth run of the patched build on `c1777f5` gave
4564.67, so the row's own run-to-run spread is ~1.2% — larger than any effect this change has.

Two structural reasons it cannot be this PR: every caller of the four launchers touched here is in
`dflash_verify_short_run`, AR decode at M=1, or a unit test — **none in the 256k prefill**, which
runs `launch_prefill_nvfp4_gemm`. And the new arena buffers are gated on `packed`, which a
target-only sweep never sets.

One observation for the harness, independent of this PR: that row is `SPARKINFER_BENCH_SWEEP_REPS=1`
— a single sample — and its speed is decided by a batched-prefill scratch ceiling (#984 is open to
reduce exactly that pressure). `cb-decode@c2/c4/c8` each got a multi-run noise study before being
trusted, and c16/c32 got one before they were added. This row has not had one.

## Results, re-measured on `c1777f5`

One binary for both arms; "before" reproduces main through the override pair
(`SPARKINFER_PACKED_MAX_ROWS=8 SPARKINFER_FFN_GEMM_MIN_ROWS=999`), "after" is plain defaults.

| | before | after | |
|---|--:|--:|--:|
| cb-decode@c8 | 506.1 | 505.1 | flat |
| cb-decode@c16 | 509.9 | 570.4 | **+11.9%** |
| cb-decode@c32 | 514.9 | 697.5 | **+35.5%** |
| target-prefill@256k | 4620.33 | 4620.01 | −0.01% |

| context | tps | mean accept | lossless |
|---|--:|--:|--:|
| dspark@4k | 126.42 | 1.6883 | yes |
| dspark@16k | 128.25 | 1.7297 | yes |
| dspark@32k | 95.84 | 1.3299 | yes |

Mean accept at 32k moved 1.2800 → 1.3299 between rounds; that is #989, not this PR. Clean main on
this tree measures **the same 1.3299** with tps 95.8132 against the PR's 95.8408, so the verify path
is bit-identical — as it must be, since the GEMM arm needs `N >= 16` and the chunk branches need
`M > 8`, and DSpark's verify never exceeds 8 rows.

**No environment is required.** Every value is a default: the packed ceiling defaults to 32 and the
FFN GEMM engages from 16 rows on its own. The two env names in the diff are *overrides* for A/B, not
switches — unset behaviour is the fast path.

## Why the crossover, and why 16

Measured per decode forward at the real FFN/GDN shapes, isolated, CUDA events:

| per-forward ms | m=8 | m=16 | m=24 | m=32 |
|---|--:|--:|--:|--:|
| dp4a row-GEMV | **7.28** | 14.31 | — | — |
| block-scaled GEMM | 13.28 | **12.80** | **10.08** | **9.71** |

A CUTLASS tile is ≥64 rows in M, so at 8 rows the GEMM throws away most of its tile and the row-GEMV
wins by 1.8×. By 16 it is behind, and at 32 one GEMM pass beats four chunked GEMV passes (9.71 vs
4 × 7.28 = 29.1) by 3×. Narrow batches keep the kernel that suits them.

The B operands are the `*_fp4` / `*_fp4_sf` pair the loader already keeps resident for prefill, so
this adds **no weight memory** — only A-side staging and a workspace, both from the existing verify
arena, allocated up front because `Arena::alloc` falls back to `cudaMalloc` and this function runs
under graph capture.

## Row counts above 8 chunk rather than instantiate

`launch_gemv_nvfp4_rows_dp4a`, `_dp4a2`, `launch_gemv_rows2` and `launch_mmvq_rows_f32` recurse in
chunks of 8 instead of refusing. Instantiating R=9..16 was tried and measured: `t(16)` came out at
`1.002 × 2·t(8)` with 495 spill stores — `uint4 xg[GPT][R]` alone is 128 registers at R=16 — so a
wider template costs exactly what re-reading the weights costs.

## Precision: a real trade, stated plainly

The row-GEMV quantizes activations to **int8** with a per-16 scale; the block-scaled GEMM requires an
**NVFP4 4-bit** A operand. A request decoded in a batch of ≥16 therefore runs at lower activation
precision than the same request decoded alone.

- For it: NVFP4 activations are already the shipped **prefill** representation for this checkpoint,
  and prefill output feeds the KV cache, so this is the same representation rather than a new one.
- Against: no scored gate sees it — the differential top1/KL gate runs through the DSpark dump, whose
  verify never exceeds 8 rows.

Flagging it rather than leaving review to find it. If the trade is unwanted, the same plumbing works
with the row-GEMV and simply gives up the gain.

### DIFF
--- a/kernels/csrc/cuda/gemm/gemv.cu
+++ b/kernels/csrc/cuda/gemm/gemv.cu
@@ -2881,7 +2881,18 @@ bool launch_gemv_rows(const void* x, const void* W, void* y,
 }
 bool launch_gemv_rows2(const void* x, const void* W0, const void* W1, void* y0, void* y1,
                        int M, int N0, int N1, int K, cudaStream_t stream) {
-    if (M < 1 || M > 8 || N0 < 1 || N1 < 1 || (K & 7)) return false;
+    if (M < 1 || N0 < 1 || N1 < 1 || (K & 7)) return false;
+    if (M > 8) {                       // chunk: see launch_gemv_nvfp4_rows_dp4a
+        for (int r0 = 0; r0 < M; r0 += 8) {
+            const int m = (M - r0) < 8 ? (M - r0) : 8;
+            if (!launch_gemv_rows2(reinterpret_cast<const __nv_bfloat16*>(x) + (size_t)r0 * K,
+                                   W0, W1,
+                                   reinterpret_cast<__nv_bfloat16*>(y0) + (size_t)r0 * N0,
+                                   reinterpret_cast<__nv_bfloat16*>(y1) + (size_t)r0 * N1,
+                                   m, N0, N1, K, stream)) return false;
+        }
+        return true;
+    }
 #ifdef _MSC_VER
     // gemv_bf16_rows_sk2_kernel is compiled out under MSVC (see #ifndef _MSC_VER above).
     // Two single-matrix launches are bit-identical to the fused path — just two grids.
@@ -3314,7 +3325,26 @@ bool launch_gemv_nvfp4_rows_dp4a2(const void* xq, const void* xs,
                                   int M, int N, int K, cudaStream_t stream) {
     if (!xq || !xs || !W0 || !W1 || !y0 || !y1 || N < 1 || K < 1 || (K & 15)) return false;
     if (!gemv_bf16_splitk()) return false;
-    if (M < 1 || M > 8) return false;
+    if (M < 1) return false;
+    // Wider than the instantiated row tiers: serve it as chunks of 8 rather than instantiating
+    // R=9..16. Measured on RTX 5090 at the real decode shapes, a 16-row instantiation costs
+    // t(16) = 1.002 * 2*t(8) with 495 spill stores -- `uint4 xg[GPT][R]` alone is 128 registers
+    // at R=16 -- so a wider template buys nothing over re-reading the weights, and chunking keeps
+    // every row on the tuned R<=8 kernel. Rows are the slow axis of y/xq/xs ([M,*] row-major),
+    // so a chunk is a pointer offset.
+    if (M > 8) {
+        const size_t ng = (size_t)(K >> 4);
+        for (int r0 = 0; r0 < M; r0 += 8) {
+            const int m = (M - r0) < 8 ? (M - r0) : 8;
+            if (!launch_gemv_nvfp4_rows_dp4a2(
+                    reinterpret_cast<const signed char*>(xq) + (size_t)r0 * K,
+                    reinterpret_cast<const float*>(xs) + (size_t)r0 * ng, W0, W1,
+                    reinterpret_cast<__nv_bfloat16*>(y0) + (size_t)r0 * N,
+                    reinterpret_cast<__nv_bfloat16*>(y1) + (size_t)r0 * N,
+                    m, N, K, stream)) return false;
+        }
+        return true;
+    }
     static const int nr_mode = []{ const char* e = getenv("SPARKINFER_NVFP4_ROWS_NR");
                                    int v = e ? atoi(e) : 0; return (v == 1 || v == 2) ? v : 0; }();
     static const bool pair_on = []{ const char* e = getenv("SPARKINFER_NVFP4_ROWS_PAIR");
@@ -3396,7 +3426,25 @@ bool launch_gemv_nvfp4_rows_dp4a(const void* xq, const void* xs, const void* W, 
                                  int M, int N, int K, cudaStream_t stream) {
     if (!xq || !xs || !W || !y || N < 1 || K < 1 || (K & 15)) return false;
     if (!gemv_bf16_splitk()) return false;
-    if (M < 1 || M > 8) return false;
+    if (M < 1) return false;
+    // Wider than the instantiated row tiers: serve it as chunks of 8 rather than instantiating
+    // R=9..16. Measured on RTX 5090 at the real decode shapes, a 16-row instantiation costs
+    // t(16) = 1.002 * 2*t(8) with 495 spill stores -- `uint4 xg[GPT][R]` alone is 128 registers
+    // at R=16 -- so a wider template buys nothing over re-reading the weights, and chunking keeps
+    // every row on the tuned R<=8 kernel. Rows are the slow axis of y/xq/xs ([M,*] row-major),
+    // so a chunk is a pointer offset.
+    if (M > 8) {
+        const size_t ng = (size_t)(K >> 4);
+        for (int r0 = 0; r0 < M; r0 += 8) {
+            const int m = (M - r0) < 8 ? (M - r0) : 8;
+            if (!launch_gemv_nvfp4_rows_dp4a(
+                    reinterpret_cast<const signed char*>(xq) + (size_t)r0 * K,
+                    reinterpret_cast<const float*>(xs) + (size_t)r0 * ng, W,
+                    reinterpret_cast<__nv_bfloat16*>(y) + (size_t)r0 * N,
+                    m, N, K, stream)) return false;
+        }
+        return true;
+    }
     const auto* xp = reinterpret_cast<const signed char*>(xq);
     const auto* sp = reinterpret_cast<const float*>(xs);
     auto* yp = reinterpret_cast<__nv_bfloat16*>(y);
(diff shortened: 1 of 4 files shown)


skyrocket2026:
Added — `cb-decode@c16` and `cb-decode@c32` are now scored dimensions (`d24add3`, on main). Your reasoning was right, and the data backs it.

**Measured on main at the scoring length**, aggregate tok/s:

| c1 | c2 | c4 | c8 | c16 | c32 |
|--:|--:|--:|--:|--:|--:|
| 82.2 | 69.9 | 72.5 | 72.7 | 72.5 | 72.5 |

The ceiling shows up as the **same number three widths running**. That is your `kQwen35MaxPackedRows = 8` claim reproduced independently — past 8 rows the batch is chunked and every chunk re-reads all ~15 GB of weights, so extra concurrency buys nothing.

**I added c16 as well as the c32 you asked for.** Your own crossover table puts the dp4a→GEMM transition *at* 16 rows (7.28 ms at m=8 vs 14.31 at m=16; the GEMM 13.28 vs 12.80), so c16 is where the change first becomes observable — and a PR that helps at 32 while hurting at 16 should not pass unnoticed.

**Gate safety was measured before adding the rows, not after.** Four runs of identical code at each width: c16 spread 0.83% (worst case −0.82%), c32 spread 0.83% (−0.83%), against a −2.00% reject bar. That check is not a formality here: the first concurrency ladder ran 64 tokens per request, looked fine, and would have hard-REJECTED *unchanged* code at −3.37% on c4. Every added width gets the same measurement.

Two notes for your side:

- Schema bumped to **v13**, so #988 will be re-evaluated from scratch rather than carrying a label taken under the c8-only set.
- Your point that no environment is required is the one that makes this scoreable at all — the evaluator cannot set env vars, so a speedup that only appears under a non-default flag would measure zero no matter how many dimensions exist. Worth keeping that property as the PR evolves.

I have not reviewed the precision trade at width that your draft note flags — that is a separate judgement from whether the gate can see the change, and it should be weighed on its own.

skyrocket2026:
**Correction to my previous comment — the baseline table I posted was wrong.**

Those numbers (c8 72.7, c16 72.5, c32 72.5) came from a scratch tree that predated the packed-decode merge (#975), so they measured the old chunked regime, not main. Main's real numbers:

| c1 | c2 | c4 | c8 | c16 | c32 |
|--:|--:|--:|--:|--:|--:|
| 95.1 | 180.8 | 302.4 | 461.1 | 484.5 | 486.1 |

(c1–c8 from the bot's own round on #986; c16/c32 re-measured after a full sync, with c8 as a control to confirm the tree really was main.)

**Your case is unchanged and better supported than my wrong numbers made it look.** Scaling is real up to c8 — 95 → 181 → 302 → 461 — and then **flattens: c8 → c32 gains 1.4%**. That is exactly the `kQwen35MaxPackedRows = 8` ceiling you describe, sitting at a much higher absolute level than the stale figures implied. My earlier "same number three widths running" line described the pre-#975 engine, not the one you are optimising.

**Gate safety has been re-established on the code that actually runs**: 0.56% / 0.62% / 0.76% spread at c8 / c16 / c32 over 4 runs of identical code, against the −2.00% reject bar. The 0.83% figures I quoted before were measured on the pre-#975 path and said nothing about the packed path your PR extends.

The dimensions themselves stand (`d24add3`, corrected in `a404336`) — `cb-decode@c16` and `cb-decode@c32` are scored, schema is v13.

The tell was there and I missed it: the bot had been scoring `cb-decode@c2` deltas of +33% and +388%, which cannot happen against a flat ~70 tok/s baseline. Apologies for the noise — the corrected numbers are the ones to work from.

skyrocket2026:

> ### ❌ Rejected at the `dspark` stage
>
> **True**
>
> The round stops at the first regression, so every later stage was **not run**. Rows marked ⏭️ below were not measured — they are not additional failures, and fixing this one may be all that is needed.

| metric | value |
|---|---|
| winning axis | **dspark-decode@4k** — the dimension this label came from |
| **PR DSpark @4k** | **129.23 tok/s** |
| **main DSpark @4k** | **129.25 tok/s** |
| DSpark decode @4k vs main | -0.0% |
| **PR prefill @4k** | **14188.2 pp/s** |
| **main prefill @4k** | **14212.2 pp/s** |
| prefill @4k vs main | -0.2% |
| **PR DSpark @16k** | **130.19 tok/s** |
| **main DSpark @16k** | **129.95 tok/s** |
| **PR prefill @16k** | **11590.1 pp/s** |
| **main prefill @16k** | **11581.0 pp/s** |
| prefill @16k vs main | +0.1% |
| **PR DSpark @32k** | **94.63 tok/s** |
| **main DSpark @32k** | **93.78 tok/s** |
| DSpark decode @32k vs main | +0.9% |
| **PR prefill @32k** | **10488.9 pp/s** |
| **main prefill @32k** | **10507.7 pp/s** |
| prefill vs main | -0.2% |
| **PR prefill @256k** | **3862.3 pp/s** |
| **main prefill @256k** | **4313.9 pp/s** |
| **PR decode @256k** | **—** |
| **main decode @256k** | **—** |
| prefill @256k vs main | -10.5% |
| **PR concurrent decode @c2** | **183.9 tok/s agg** |
| **main concurrent decode @c2** | **183.8 tok/s agg** |
| **PR concurrent decode @c4** | **307.2 tok/s agg** |
| **main concurrent decode @c4** | **306.4 tok/s agg** |
| concurrent decode @c4 vs main | +0.3% |
| **PR concurrent decode @c8** | **477.5 tok/s agg** |
| **main concurrent decode @c8** | **474.6 tok/s agg** |
| concurrent decode @c8 vs main | +0.6% |
| **PR concurrent decode @c16** | **554.6 tok/s agg** |
| **main concurrent decode @c16** | **482.9 tok/s agg** |
| concurrent decode @c16 vs main | +14.8% |
| **PR concurrent decode @c32** | **684.1 tok/s agg** |
| **main concurrent decode @c32** | **487.4 tok/s agg** |
| concurrent decode @c32 vs main | +40.4% |
| PR AR tok/s (floor) | 88.50 |
| main AR tok/s (floor) | 88.54 |
| AR vs main (floor) | -0.0% |
| **DSpark vs AR** | **1.471×** — above 1.0 means speculation finally pays |
| accuracy gate | ⏭️ not reached — the round stopped at the `dspark` stage |
| cross-run determinism | — not measured (harness predates this gate) |
| losslessness @4k | ⏭️ not reached — the round stopped at the `dspark` stage |
| losslessness @32k | ⏭️ not reached — the round stopped at the `dspark` stage |
| mean accept τ @4k | ⏭️ not reached — the round stopped at the `dspark` stage |
| mean accept τ floor | ⏭️ not reached — the round stopped at the `dspark` stage |
| mean accept τ @32k | ⏭️ not reached — the round stopped at the `dspark` stage |
| qwen3.8 (upstream unsloth) guard @16k | ⏭️ not reached — the round stopped at the `dspark` stage |
| qwen3.6 guard @16k | ⏭️ not reached — the round stopped at the `dspark` stage |
| PPL PR / main | ? / ? |
| commit | `614a5adee` |

target-prefill@256k regression: 3862.30 < 98% of main 4313.89 (stopped at the dspark stage; later stages skipped)

skyrocket2026:

This PR's best Qwen3.8-27B DSpark axis measured **0.0%** vs main, (regression) — closing automatically. This bot evaluates every eligible PR in the repo against Qwen3.8-27B's decode AND prefill@128 speed specifically, regardless of what the PR is actually about — a close here isn't a judgment on the PR's purpose, just that it didn't move these particular metrics. Reopen (or open a fresh PR) if you have a fix or a different approach.

PR #991, opened 2026-09-07
perf(decode): serve the GDN projections with the block-scaled NVFP4 GEMM at wide batch (1.15x cb-decode@c32)

widecloud:
## Summary

#990 raised `kQwen35MaxPackedRows` to 32 and moved the **dense FFN** onto the checkpoint-native block-scaled NVFP4 GEMM above 16 rows. Every other projection still chunks into 8s, so a 32-row packed decode reads the GDN weights **four times** per step. Those are 48 of the 64 layers and ~3.1 GB of the ~4.1 GB of per-step weight traffic that is not FFN. This routes them through the same GEMM, on the same threshold, using the `gdn_*_fp4` operands the loader already keeps resident.

`wqkv` and `wqkv_gate` share one quantize of `xn`, exactly as gate/up share theirs.

## Proof of speedup

| | decode tok/s |
|---|--:|
| before (main) | 741.7 |
| after (this PR) | 853.3 |

`cb_bench` submits a fixed request set and never refills, so a run where every row generates all 256 tokens keeps the batch at 32 throughout, while a run that loses a row early finishes in a low-concurrency tail. That splits aggregate throughput into two modes, so the pair above is matched on `decode_tokens=8200` (all rows complete) and the mean ITL is quoted alongside:

```text
concurrency 32, both arms at decode_tokens=8200
  main    agg_tok_s=741.7  mean_itl_ms=37.94     (main good mode, 3 runs: 37.98 / 37.94 / 37.93)
  this PR agg_tok_s=853.3  mean_itl_ms=32.28     -> 1.176x on step time

concurrency 32, second pair, both partial
  main    agg_tok_s=490.3  mean_itl_ms=55.51  decode_tokens=7599
  this PR agg_tok_s=601.0  mean_itl_ms=46.98  decode_tokens=8039  -> 1.182x on step time

concurrency 16 (2 alternating pairs, decode_tokens=4104)
  main    agg_tok_s=590.7  mean_itl_ms=25.04
  this PR agg_tok_s=599.8  mean_itl_ms=24.60
```

At 16 rows the GEMV chunks 2x rather than 4x, so only a third of the saving is available there, which is what the +1.5% reflects.

## Nothing below the threshold moves

The arm requires `packed`, so DSpark's speculative verify -- which is never packed and never reaches these widths -- cannot enter it. Measured rather than argued, ctx=16384:

```text
main       190.9821   89.3270      2.5098        1        11898.30
this PR    190.9913   89.3008      2.5098        1        11894.51
```

Mean accept is tokens/steps, so equality to four decimals means the verifier accepted the same tokens at the same positions across the whole generation.

### DIFF
--- a/runtime/src/models/qwen35_prefill.cpp
+++ b/runtime/src/models/qwen35_prefill.cpp
@@ -2798,6 +2798,17 @@ int dflash_verify_short_run(const Qwen35PrefillCtx& s, const int* token_ids, int
     // The B operands are the SAME *_fp4 / *_fp4_sf the prefill path already keeps resident, so
     // this costs no extra weight memory -- only the A-side staging and a workspace, below.
     const int fp4_kwide = (c.moe_ffn > H ? c.moe_ffn : H);
+    // Wide packed batches run the SAME block-scaled NVFP4 GEMM the dense FFN already uses, for
+    // the GDN projections as well. Those are 48 of the 64 layers and ~3.1 GB of the ~4.1 GB of
+    // per-step weight traffic that is not FFN; chunked into 8s a 32-row batch reads every one of
+    // those bytes FOUR times. Same gate as the FFN arm: the A-quantizer needs m % 8 == 0, and
+    // below 16 rows the row-GEMV is still ahead (its tile wastes most of M).
+    // SPARKINFER_PROJ_GEMM_MIN_ROWS raises or disables the threshold for an A/B in one binary.
+    static const int kProjGemmMinRows = [] {
+        const char* e = getenv("SPARKINFER_PROJ_GEMM_MIN_ROWS");
+        const int v = e ? atoi(e) : 16;
+        return v < 1 ? 1 : v;
+    }();
     unsigned char* fp4_a = nullptr; unsigned char* fp4_asf = nullptr; unsigned char* fp4_ws = nullptr;
     if (packed && c.dense_ffn) {
         const size_t ab = kernels::prefill_nvfp4_data_bytes(NA, fp4_kwide);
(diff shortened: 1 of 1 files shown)


skyrocket2026:

| metric | value |
|---|---|
| **PR DSpark @4k** | **129.35 tok/s** |
| **main DSpark @4k** | **129.33 tok/s** |
| **PR prefill @4k** | **14242.9 pp/s** |
| **main prefill @4k** | **14206.8 pp/s** |
| **PR DSpark @16k** | **130.53 tok/s** |
| **main DSpark @16k** | **130.16 tok/s** |
| **PR prefill @16k** | **11689.7 pp/s** |
| **main prefill @16k** | **11586.3 pp/s** |
| prefill @16k vs main | +0.8% |
| **PR DSpark @32k** | **97.69 tok/s** |
| **main DSpark @32k** | **97.61 tok/s** |
| **PR prefill @32k** | **10472.3 pp/s** |
| **main prefill @32k** | **10466.1 pp/s** |
| prefill vs main | +0.0% |
| **PR prefill @256k** | **4344.5 pp/s** |
| **main prefill @256k** | **4344.2 pp/s** |
| **main decode @256k** | **94.73 tok/s** |
| **PR concurrent decode @c2** | **184.3 tok/s agg** |
| **main concurrent decode @c2** | **184.1 tok/s agg** |
| **PR concurrent decode @c4** | **307.0 tok/s agg** |
| **main concurrent decode @c4** | **307.5 tok/s agg** |
| concurrent decode @c4 vs main | -0.2% |
| **PR concurrent decode @c8** | **480.4 tok/s agg** |
| **main concurrent decode @c8** | **480.7 tok/s agg** |
| concurrent decode @c8 vs main | -0.1% |
| **PR concurrent decode @c16** | **559.9 tok/s agg** |
| **main concurrent decode @c16** | **557.5 tok/s agg** |
| concurrent decode @c16 vs main | +0.4% |
| **PR concurrent decode @c32** | **785.2 tok/s agg** |
| **main concurrent decode @c32** | **684.9 tok/s agg** |
| concurrent decode @c32 vs main | +14.6% |
| PR AR tok/s (floor) | 88.63 |
| main AR tok/s (floor) | 88.60 |
| **DSpark vs AR** | **1.473×** — above 1.0 means speculation finally pays |
| qwen3.8 (upstream unsloth) guard @16k | ✅ no regression (decode+prefill) — decode 82.7 tok/s · prefill 10058 pp |
| qwen3.6 guard @16k | ✅ no regression (decode+prefill) — decode 462.1 tok/s · prefill 27616 pp |
| commit | `3e5d2a074` |

ok

skyrocket2026:

PR #993, opened 2026-09-07
perf(decode): stop falling back to the row-GEMV at wide continuous-batch decode (1.14x cb-decode@c32)

widecloud:
## Summary

#990 put the dense FFN on the block-scaled NVFP4 GEMM at wide batch and #991 the GDN projections. Two things still fell back to the chunked row-GEMV, and both cost most at c32 where that fallback re-reads the weights four times:

- **The full-attention projections were never routed.** `wq` (33.3 MB/layer) and `wo` (16.7 MB) now run on the same GEMM, on the same threshold, off one shared quantize of `xn` — the shape `prefill_batched_run` already uses. `wk`/`wv` stay on the row-GEMV: at 1024 rows they are eight CTAs of a 128-wide tile and measured slower there (c32 901.1 against 909.8).
- **Every GEMM arm declined unless the width was a multiple of eight.** The packed decode is handed whatever the scheduler has live, so a batch of 17..31 rows declined *all* arms at once — width 17 costs 35.16 ms against 23.91 ms for width 16. The M tile is 128 whatever `m` is, so rounding `m` up to the next multiple of eight rides a tile the GEMM already runs. Every buffer here is allocated for 32 rows and the padded rows' outputs are never read.

## Proof of speedup

| | decode tok/s |
|---|--:|
| before (main) | 863.7 |
| after (this PR) | 980.5 |

`cb_bench` never refills the batch, so a run that loses a row early finishes in a low-concurrency tail. Every number below is from a run at the full `decode_tokens`, with mean ITL as the mode-robust check.

```text
concurrency 32, every run at decode_tokens=8200
  main      agg_tok_s=865.7 / 865.9 / 862.3 / 860.8   mean_itl_ms=31.76 / 31.76 / 31.87 / 31.93
  this PR   agg_tok_s=980.6 / 980.4                   mean_itl_ms=27.96 / 27.94   -> 1.135x

concurrency 16, decode_tokens=4104
  main      agg_tok_s=660.7 / 662.5 / 663.5 / 661.1   mean_itl_ms=22.08 / 22.05 / 22.02 / 22.07
  this PR   agg_tok_s=680.0 / 679.6                   mean_itl_ms=21.43 / 21.44

concurrency 8, decode_tokens=2056 (nothing reaches the GEMM below 16 rows)
  main      agg_tok_s=503.7 / 502.3 / 503.7   this PR   agg_tok_s=503.4 / 503.8 / 502.5
```

c16 gains less because 16 is already a multiple of eight, so only the attention half applies.

## Nothing outside the packed decode moves

Both changes are inside `dflash_verify_short_run` and every arm requires `packed`, so DSpark's verify — never packed — cannot enter them, and prefill is untouched. ctx=16384:

```text
main       190.9760   89.2778      2.5098        1       11904.30
this PR    191.0341   89.3205      2.5098        1       11912.10
```

Mean accept equal to four decimals means the verifier accepted the same tokens at the same positions. `SPARKINFER_ATTN_GEMM=0` restores the previous attention path for a paired A/B out of one binary.

### DIFF
--- a/runtime/src/models/qwen35_prefill.cpp
+++ b/runtime/src/models/qwen35_prefill.cpp
@@ -2744,6 +2744,21 @@ int dflash_verify_short_run(const Qwen35PrefillCtx& s, const int* token_ids, int
     // one, and there is no accepted-prefix commit because every row is already a real step.
     const bool packed = s.packed_rows != nullptr;
     const int H = c.hidden, N = n, qdim = s.qdim, kvdim = s.kvdim;
+    // Every block-scaled GEMM arm below used to require `(N & 7) == 0`, because
+    // launch_prefill_nvfp4_quant_a builds the A-side scale layout in groups of eight rows. The
+    // packed decode hands us whatever the scheduler has live, so any batch of 17..31 rows that is
+    // not a multiple of eight declined EVERY arm at once and fell all the way back to the chunked
+    // row-GEMV -- measured with a host timer on the engine loop, width 17 costs 35.16 ms against
+    // 23.91 for width 16, a 47% penalty for one extra row.
+    //
+    // Rounding the GEMM's m up to the next multiple of eight removes that cliff for nothing: the
+    // block-scaled M tile is 128 whatever m is (the NVFP4 scale atom is 32x4 = 128 rows, which is
+    // why no smaller tile can be built), so the padded rows ride in a tile the GEMM was already
+    // going to run. Every arena buffer here is allocated for NA = kVerifyMaxRows = 32 rows and
+    // Ng <= 32 always, so the padding is in bounds by construction; the pad rows carry whatever
+    // the previous step left and their outputs are simply never read. Scale factors are per row,
+    // so a pad row cannot perturb a real one.
+    const int Ng = (N + 7) & ~7;
     // Host-side dispatch hint for the flash-decode split (it selects the tensor-core arm on
     // seqlen > 512 and mma_chunk >= 32). Packed rows have independent lengths, so take the
     // longest: the per-row lengths the kernel actually reads still come from `seq`.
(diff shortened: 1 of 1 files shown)


skyrocket2026:

| metric | value |
|---|---|
| **PR DSpark @4k** | **129.54 tok/s** |
| **main DSpark @4k** | **129.36 tok/s** |
| DSpark decode @4k vs main | +0.1% |
| **PR prefill @4k** | **14243.9 pp/s** |
| **main prefill @4k** | **14231.1 pp/s** |
| **PR DSpark @16k** | **130.57 tok/s** |
| **main DSpark @16k** | **130.21 tok/s** |
| **PR prefill @16k** | **11698.6 pp/s** |
| **main prefill @16k** | **11590.1 pp/s** |
| prefill @16k vs main | +0.9% |
| **PR DSpark @32k** | **97.64 tok/s** |
| **main DSpark @32k** | **97.69 tok/s** |
| **PR prefill @32k** | **10470.7 pp/s** |
| **main prefill @32k** | **10482.1 pp/s** |
| prefill vs main | -0.1% |
| **PR prefill @256k** | **4350.3 pp/s** |
| **main prefill @256k** | **4360.6 pp/s** |
| **PR decode @256k** | **94.76 tok/s** |
| prefill @256k vs main | -0.2% |
| **PR concurrent decode @c2** | **184.3 tok/s agg** |
| **main concurrent decode @c2** | **184.5 tok/s agg** |
| **PR concurrent decode @c4** | **307.7 tok/s agg** |
| **main concurrent decode @c4** | **308.0 tok/s agg** |
| **PR concurrent decode @c8** | **479.0 tok/s agg** |
| **main concurrent decode @c8** | **480.2 tok/s agg** |
| concurrent decode @c8 vs main | -0.2% |
| **PR concurrent decode @c16** | **634.0 tok/s agg** |
| **main concurrent decode @c16** | **623.0 tok/s agg** |
| concurrent decode @c16 vs main | +1.8% |
| **PR concurrent decode @c32** | **899.1 tok/s agg** |
| **main concurrent decode @c32** | **801.8 tok/s agg** |
| concurrent decode @c32 vs main | +12.1% |
| PR concurrent decode @c1 (floor) | 94.9 tok/s agg |
| main AR tok/s (floor) | 88.62 |
| qwen3.8 (upstream unsloth) guard @16k | ✅ no regression (decode+prefill) — decode 82.7 tok/s · prefill 10065 pp |
| qwen3.6 guard @16k | ✅ no regression (decode+prefill) — decode 461.7 tok/s · prefill 27711 pp |
| commit | `5940079c1` |

ok

skyrocket2026:

PR #994, opened 2026-09-07
perf(decode): stop ramping the continuous-batch decode one row at a time (1.05x cb-decode@c32)

widecloud:
## Summary

The scheduler admits exactly one prefill per iteration, so N arriving requests ramp one row at a time. A packed decode step costs a fixed weight read plus a small per-row term — measured 14.85 ms fixed against 0.41 ms/row at concurrency 32 — so a step at four rows costs 97% of a step at thirty-two. Ramping one row per iteration therefore pays ~32 near-full weight reads for a handful of tokens each, and because the rows start staggered they finish staggered and the same waste returns as a drain tail.

This admits a second prefill per iteration while the decode batch is still filling. It reverts to exactly one once the batch reaches the packed width, so steady-state ITL is untouched — the regime the original rule was written for — and also when the ramp is shallow (pending + already-decoding below half the packed width), which is what keeps c2/c4/c8 bit-for-bit unchanged.

`worker_loop` took only `prefill_ids.front()`; it now runs what the scheduler returns.

## Proof of speedup

| | decode tok/s |
|---|--:|
| before (main) | 973.0 |
| after (this PR) | 1016.9 |

`cb_bench` never refills the batch, so a run that loses a row early finishes in a low-concurrency tail. Every number below is from a run at the full `decode_tokens`.

```text
concurrency 32, decode_tokens=8200          agg tok/s      mean ITL   max ITL
  main                                      974.2 / 971.8   28.1       77.6 / 78.1
  this PR                                   1016.9          27.9      102.1

concurrency 16, decode_tokens=4104
  main                                      678.7 / 678.2 / 677.7   21.46   74.2 / 75.5 / 75.7
  this PR                                   694.6                   21.36  100.6

concurrency 8, decode_tokens=2056 (shallow ramp -> rule disabled, unchanged by construction)
  main                                      499.8 / 499.2   14.95   77.5 / 78.8
  this PR                                   499.6 / 498.7   14.95   80.7 / 79.0
```

Mean ITL is unchanged everywhere: this shortens the ramp, it does not change the step. That is also why it is invisible to a step-time measurement.

## The cost, and why it is bounded

The admitted prefills run back to back ahead of the next decode step, so the worst inter-token gap grows by about one prefill duration per admission — measured 75 → 104 → ~155 (four) → ~293 (eight) ms. Throughput saturates by two while the tail keeps climbing, so two is the whole of the win and none of the tail: four and eight buy a further ~1% for another 150 ms, which is the same ITL spike `prefill_mix_max_tokens()` already exists to prevent. A width-scaled taper was also tried and is strictly worse than a flat two on both axes (1025.7 tok/s at 206.7 ms).

`SPARKINFER_PREFILLS_PER_STEP=1` restores the previous scheduler exactly.

## Nothing outside continuous batching moves

DSpark does not go through `ContinuousBatchEngine`. Measured rather than argued, ctx=16384:

```text
main       190.7877   89.2859      2.5098        1       11886.90
this PR    190.9051   89.2933      2.5098        1       11894.26
```

### DIFF
--- a/runtime/src/inference_engine.cpp
+++ b/runtime/src/inference_engine.cpp
@@ -315,11 +315,15 @@ void ContinuousBatchEngine::worker_loop() {
                 if (job) any_finished = step_job(*job, /*chunked=*/false) || any_finished;
             }
         }
-        if (!prefill_ids.empty()) {
+        // The scheduler may hand back more than one prefill while the decode batch is still
+        // filling (see Scheduler::schedule). They run back to back on this thread, which is the
+        // point: each one widens the next decode step, and a decode step's cost is almost all
+        // fixed weight read.
+        for (uint64_t pid : prefill_ids) {
             Job* job = nullptr;
             {
                 std::lock_guard<std::mutex> lock(mu_);
-                auto it = jobs_.find(prefill_ids.front());
+                auto it = jobs_.find(pid);
                 if (it != jobs_.end() && !it->second->done) job = it->second.get();
             }
             if (job) any_finished = step_job(*job, /*chunked=*/mix_decode ||
--- a/runtime/src/scheduler.cpp
+++ b/runtime/src/scheduler.cpp
@@ -18,6 +18,26 @@
 namespace sparkinfer {
 
 namespace {
+// How many prefills one iteration may admit while the decode batch is still filling. 1 restores
+// the previous behaviour exactly, for a paired A/B out of one binary.
+int prefills_per_step() {
+    static const int v = [] {
+        const char* e = getenv("SPARKINFER_PREFILLS_PER_STEP");
+        const int x = e ? atoi(e) : 2;
+        return x < 1 ? 1 : x;
+    }();
+    return v;
+}
+// The decode width above which a step is no longer dominated by its fixed weight read. Defaults
+// to the packed-decode ceiling; below it a wider batch is very nearly free.
+int packed_decode_width() {
+    static const int v = [] {
+        const char* e = getenv("SPARKINFER_WIDE_DECODE_ROWS");
+        const int x = e ? atoi(e) : 32;
+        return x < 1 ? 1 : x;
+    }();
+    return v;
+}
 int prefill_mix_max_tokens() {
     static int v = [] {
         const char* e = getenv("SPARKINFER_PREFILL_MIX_MAX");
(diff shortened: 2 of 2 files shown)


skyrocket2026:

| metric | value |
|---|---|
| **PR DSpark @4k** | **129.26 tok/s** |
| **main DSpark @4k** | **129.29 tok/s** |
| **PR prefill @4k** | **14222.8 pp/s** |
| **main prefill @4k** | **14221.8 pp/s** |
| **PR DSpark @16k** | **130.46 tok/s** |
| **main DSpark @16k** | **130.09 tok/s** |
| **PR prefill @16k** | **11675.4 pp/s** |
| **main prefill @16k** | **11557.5 pp/s** |
| prefill @16k vs main | +1.0% |
| **PR DSpark @32k** | **97.59 tok/s** |
| **main DSpark @32k** | **97.63 tok/s** |
| **PR prefill @32k** | **10469.1 pp/s** |
| **main prefill @32k** | **10458.9 pp/s** |
| prefill vs main | +0.0% |
| **PR prefill @256k** | **4347.0 pp/s** |
| **main prefill @256k** | **4348.0 pp/s** |
| **PR decode @256k** | **94.73 tok/s** |
| **main decode @256k** | **94.72 tok/s** |
| **PR concurrent decode @c2** | **184.5 tok/s agg** |
| **main concurrent decode @c2** | **183.6 tok/s agg** |
| concurrent decode @c2 vs main | +0.5% |
| **PR concurrent decode @c4** | **305.8 tok/s agg** |
| **main concurrent decode @c4** | **307.3 tok/s agg** |
| concurrent decode @c4 vs main | -0.5% |
| **PR concurrent decode @c8** | **478.6 tok/s agg** |
| **main concurrent decode @c8** | **477.9 tok/s agg** |
| concurrent decode @c8 vs main | +0.1% |
| **PR concurrent decode @c16** | **654.4 tok/s agg** |
| **main concurrent decode @c16** | **634.6 tok/s agg** |
| concurrent decode @c16 vs main | +3.1% |
| **PR concurrent decode @c32** | **941.8 tok/s agg** |
| **main concurrent decode @c32** | **902.0 tok/s agg** |
| concurrent decode @c32 vs main | +4.4% |
| PR AR tok/s (floor) | 88.53 |
| main AR tok/s (floor) | 88.36 |
| qwen3.8 (upstream unsloth) guard @16k | ✅ no regression (decode+prefill) — decode 82.7 tok/s · prefill 10055 pp |
| qwen3.6 guard @16k | ✅ no regression (decode+prefill) — decode 462.1 tok/s · prefill 27648 pp |
| commit | `b5b3e9101` |

ok

skyrocket2026:

PR #999, opened 2026-09-08
perf(prefill): read the 256k attention's V operand in the PV mma's own layout (1.10x target-prefill@256k)

widecloud:
## Summary

`pf_attn_mma_gqa_kernel`'s six-head tier — the one `launch_prefill_attn_mma` selects for int8-KV prefill once the KV span reaches 65536 keys — is **70.9% of the scored `target-prefill@256k` pass** (nsys, 43.14 s of 60.86 s, 256 launches = 16 windows x 16 full-attention layers). QK and PV read the same paged pool, but their B operands want opposite things. QK's B is K^T, so a lane wants four consecutive **dims** of one key, which is exactly what `[key][dim]` stores: the K load is one `LDG.32` per lane moving a full 128 B per warp instruction. PV's B is V, so a lane wants four consecutive **keys** of one dim; in `[key][dim]` those are `KVLD` (1024 B) apart, so the same operand costs four 2-byte gathers plus an eight-instruction `byte_perm` chain to rebuild it, each gather moving 64 B per warp instruction against K's 128.

Skip-probed at the scored point by stubbing each pool read in turn (wrong results, right timing): stubbing V is **+20.6%** on the dimension against **+10.2%** for K — twice the cost for the same bytes — while stubbing the QK and PV `mma`s is only +2.4% and +1.8%. The tensor cores are idle; this kernel is paying for the operand layout.

So V is repacked once per attention pass into `[logical block][kv head][dim][16 keys]`, where a lane's four keys are four consecutive bytes and its two n-halves are adjacent 16 B rows. The operand becomes four `LDG.32` per page pair instead of eight `LDG.U16` and eight `PRMT`, and the two loads of a page cover one whole 32-byte sector between them. It is a **repack, not a second pool** — the same bytes in a different order — so every P'xV product is unchanged bit for bit. `pf_v_pack_kernel` does not appear in the top twelve kernels of the profiled run: the rebuild is under 0.43% of the pass.

Second factor, also bit-identical: this kernel still reduced its row max and its |P'| max with five-step shuffle butterflies, while the bf16 sibling has always done both with a single `redux.sync`. Max is exact and order-independent, and the standard total order on IEEE floats is monotone, so the reduced key is the key of the max; `|P'|` is an exp times an absmax/127, so it is non-negative and its bit pattern orders as an unsigned directly. The softmax denominator keeps its butterfly in the same partner order, so the float sum is unchanged too.

## Proof of speedup

**Decode tok/s** — N/A, this is a prefill lever. For no-regression, the same runs read `target-decode@256k` at 95.50 / 95.43 tok/s on main against 95.50 / 95.48 here.

| | decode tok/s |
|---|--:|
| before (main) | N/A |
| after (this PR) | N/A |

**Prefill pp tok/s** — `target-prefill@256k`, the exact scored invocation from `eval/pr_dspark_bot.py`. Two interleaved repeats per arm, mean:

| | prefill pp tok/s |
|---|--:|
| before prefill (main) | **4446.86** |
| after prefill (this PR) | **4895.38**  (+10.08%) |

```text
$ SPARKINFER_QWEN38_PREFILL_NVFP4=1 SPARKINFER_QWEN38_DECODE_NVFP4=1 \
  SPARKINFER_KV_INT8=1 SPARKINFER_PREFILL_BATCHED=1 \
  SPARKINFER_BENCH_SWEEP_CTXS=262144 SPARKINFER_BENCH_SWEEP_REPS=1 \
  qwen3_gguf_bench <Qwen3.8-27B-NVFP4-RTX5090> 128 sweep
  # RTX 5090 (sm_120), CUDA 13.0, measured at base ec20396, interleaved PR/main/PR/main.
  # Both arms are the same tree at ec20396 differing only in the one file below.

  PR    prefill pp 4930.4519   decode tg 95.50
  main  prefill pp 4452.0102   decode tg 95.50
  PR    prefill pp 4860.3078   decode tg 95.48
  main  prefill pp 4441.7051   decode tg 95.43

  mean  main 4446.86 -> PR 4895.38 pp/s   = +10.08%

  # the stack, measured incrementally, each arm interleaved x2 against main:
  #   + V repacked into the PV mma's own B-operand layout                        +6.13%
  #   + redux.sync for the two exact warp maxima                                 +10.08%
  # dropped, both measured and reverted rather than carried:
  #   - a second page pair in flight (software pipelining the V loads)           -0.2%
  #   - the paired __half2 softmax (h2exp2 / __hmul2 / __hmax2 downstream)       -0.7%
  # Both drops say the same thing: after the layout fix this kernel is bound by
  # memory latency at 25% occupancy, not by instruction count, so cutting
  # instructions further buys nothing.
```

The tier this touches is reachable by **one** scored dimension. `launch_prefill_attn_mma`'s six-head arm requires `gqa % 6 == 0`, a KV span of at least 65536 keys, and `n_tokens >= 2048`. The six DSpark dimensions run bf16 KV (`dspark_tau_check` pins `int8_kv=false`) and never enter this kernel at all; both cross-model guards run at ctx=16384, far under the span; Qwen3.6 is GQA-8, so `gqa % 6` excludes it outright; and the concurrent-decode dimensions prefill 256 tokens. The `[pf-attn-tier]` trace confirms it — `vt=1` only on the `RQH=6 SPL=2` tier, `vt=0` on the `RQH=3` tier that serves this run's first three windows.

Same box, main vs this PR:

| gate | main | this PR | ratio |
|---|--:|--:|--:|
| **target-prefill@256k** | 4446.86 | **4895.38** | **1.1008** |
| target-decode@256k | 95.47 | 95.49 | 1.0002 |
| dspark-prefill@4k / @16k / @32k | 14898.70 / 12318.11 / 11040.64 | 14936.57 / 12344.15 / 11021.97 | 1.0025 / 1.0021 / 0.9983 |
| dspark-decode@4k / @16k / @32k | 132.196 / 132.754 / 100.077 | 132.261 / 132.814 / 100.076 | 1.0005 / 1.0005 / 1.0000 |
| AR@4k / @16k / @32k | 94.000 / 89.598 / 84.169 | 94.046 / 89.616 / 84.178 | 1.0005 / 1.0002 / 1.0001 |
| mean accept @4k / @16k / @32k | 1.6883 / 1.7297 / 1.3299 | 1.6883 / 1.7297 / 1.3299 | **1.0000** |
| losslessness @4k / @16k / @32k | 1 | 1 | |
| cb-decode@c16 / @c32 | 789.5 / 1294.8 | 788.3 / 1296.6 | 0.9985 / 1.0014 |

Every non-256k row is a kernel this diff never executes, and reads as one: `MEAN_ACCEPT` is identical to four decimals at all three contexts, `LOSSLESS=1` throughout, and every ratio is within run-to-run spread of 1.000.

If the 256 MB plane cannot be allocated beside the prefill arena, `vpack_build` returns null and the launcher falls back to the **same** six-head kernel on the paged loads — the tier and its numerics are kept, only the speedup is lost. `SPARKINFER_PREFILL_ATTN_VPACK=0` selects that path for A/B in one binary.

### DIFF
--- a/kernels/csrc/cuda/fused/prefill_attn_mma.cu
+++ b/kernels/csrc/cuda/fused/prefill_attn_mma.cu
@@ -343,13 +343,14 @@ __device__ __forceinline__ void pf_mma_16832(int (&d)[4], const unsigned (&a)[4]
                  : "r"(a[0]), "r"(a[1]), "r"(a[2]), "r"(a[3]), "r"(b0), "r"(b1));
 }
 
-template <int HEAD_DIM, int GROUP_BLKS, int RQH, int PLANES = 0>
+template <int HEAD_DIM, int GROUP_BLKS, int RQH, int PLANES = 0, bool VT = false>
 __global__ __launch_bounds__(GROUP_BLKS * 32, (GROUP_BLKS >= 16 ? 1 : (RQH <= 3 ? 2 : 1))) void pf_attn_mma_gqa_kernel(
     const __nv_bfloat16* __restrict__ q, const signed char* __restrict__ k_pool,
     const signed char* __restrict__ v_pool, const __half* __restrict__ k_scale,
     const __half* __restrict__ v_scale, const int* __restrict__ block_table,
     __nv_bfloat16* __restrict__ attn, int n_tokens, int n_q_heads, int n_kv_heads,
-    int block_size, int max_blocks_per_seq, float scale, int win_blocks, int qld, int pld, int q_pos0) {
+    int block_size, int max_blocks_per_seq, float scale, int win_blocks, int qld, int pld, int q_pos0,
+    const signed char* __restrict__ vT) {
     // q_pos0 is where this pass's queries START in the sequence. It was implicitly 0 while
     // prefill always ingested [0, N) in a single pass; carrying it lets a long prompt be
     // ingested in windows. Queries and outputs stay addressed by the LOCAL row, while the
@@ -720,8 +721,21 @@ __global__ __launch_bounds__(GROUP_BLKS * 32, (GROUP_BLKS >= 16 ? 1 : (RQH <= 3 
                                 mx = fmaxf(mx, sc[u]);
                             }
                         }
-                        #pragma unroll
-                        for (int o = 16; o > 0; o >>= 1) mx = fmaxf(mx, __shfl_xor_sync(0xffffffffu, mx, o));
+                        if constexpr (SPL != RQH) {
+                            // max is exact and order-independent, so the five-step butterfly is
+                            // one warp instruction. redux.sync is integer-only, but the standard
+                            // total order on IEEE floats -- flip the sign bit when positive,
+                            // invert every bit when negative -- is monotone, so the reduced key
+                            // is the key of the max and this is BIT-IDENTICAL. The bf16 sibling
+                            // has always reduced its row max this way; only this kernel had not.
+                            const unsigned ub = __float_as_uint(mx);
+                            const unsigned key = (ub & 0x80000000u) ? ~ub : (ub | 0x80000000u);
+                            const unsigned rd = __reduce_max_sync(0xffffffffu, key);
+                            mx = __uint_as_float((rd & 0x80000000u) ? (rd & 0x7fffffffu) : ~rd);
+                        } else {
+                            #pragma unroll
+                            for (int o = 16; o > 0; o >>= 1) mx = fmaxf(mx, __shfl_xor_sync(0xffffffffu, mx, o));
+                        }
                         const float m_old = s_m[h * BM + r], m_new = fmaxf(m_old, mx),
                                     corr = exp2f(m_old - m_new);
                         float sum = 0.f, pamax = 0.f;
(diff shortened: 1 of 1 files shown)


skyrocket2026:

| metric | value |
|---|---|
| winning axis | **target-prefill@256k** — the dimension this label came from |
| **PR DSpark @4k** | **127.65 tok/s** |
| **main DSpark @4k** | **129.44 tok/s** |
| DSpark decode @4k vs main | -1.3% |
| **PR prefill @4k** | **14217.5 pp/s** |
| **main prefill @4k** | **14203.1 pp/s** |
| prefill @4k vs main | +0.1% |
| **PR DSpark @16k** | **130.48 tok/s** |
| **main DSpark @16k** | **130.18 tok/s** |
| **PR prefill @16k** | **11640.9 pp/s** |
| **main prefill @16k** | **11561.1 pp/s** |
| prefill @16k vs main | +0.6% |
| **PR DSpark @32k** | **97.66 tok/s** |
| **main DSpark @32k** | **97.70 tok/s** |
| **PR prefill @32k** | **10470.1 pp/s** |
| **main prefill @32k** | **10455.5 pp/s** |
| prefill vs main | +0.1% |
| **PR prefill @256k** | **4800.9 pp/s** |
| **main prefill @256k** | **4352.9 pp/s** |
| **main decode @256k** | **94.71 tok/s** |
| prefill @256k vs main | +10.2% |
| **PR concurrent decode @c2** | **186.9 tok/s agg** |
| **main concurrent decode @c2** | **186.8 tok/s agg** |
| **PR concurrent decode @c4** | **319.4 tok/s agg** |
| **main concurrent decode @c4** | **319.6 tok/s agg** |
| **PR concurrent decode @c8** | **514.1 tok/s agg** |
| **main concurrent decode @c8** | **511.4 tok/s agg** |
| concurrent decode @c8 vs main | +0.5% |
| **PR concurrent decode @c16** | **752.6 tok/s agg** |
| **main concurrent decode @c16** | **749.8 tok/s agg** |
| concurrent decode @c16 vs main | +0.4% |
| **PR concurrent decode @c32** | **1227.9 tok/s agg** |
| **main concurrent decode @c32** | **1216.9 tok/s agg** |
| concurrent decode @c32 vs main | +0.9% |
| PR AR tok/s (floor) | 88.49 |
| main AR tok/s (floor) | 88.44 |
| qwen3.8 (upstream unsloth) guard @16k | ✅ no regression (decode+prefill) — decode 82.7 tok/s · prefill 10063 pp |
| qwen3.6 guard @16k | ✅ no regression (decode+prefill) — decode 461.1 tok/s · prefill 27645 pp |
| commit | `53aba417a` |

ok

skyrocket2026:

PR #1006, opened 2026-09-08
fix(muse): honour the int8 KV cache instead of writing bf16 into it (40x prefill@4k)

inference2026:
## Summary

Muse Glimmer emitted garbage at ctx >= 4096 and lost most of its prefill throughput there. The
example mains switch the KV cache to int8 at ctx >= 4096, but Muse's decode KV append cast the pool
to bf16 unconditionally — two-byte writes into a one-byte-per-element pool — and the batched prefill
declined an int8 cache outright, dropping the whole prompt onto the token-at-a-time path. The model
takes `KVCacheManager*` and does not own it, so Muse has to honour int8 rather than avoid it.

- **decode KV write**: `muse_kv_append_int8_kernel` — NORMAL rope on SWA layers, NoPE on global,
  per-head-vector max-abs to int8 plus one fp16 scale (the `amax/127` convention the other int8
  appenders use).
- **prefill KV write**: the Muse writer is templated on `<bool INT8>`. The QK-norm and NORMAL-rope
  math is shared and only the store tail differs, so the two flavours cannot drift; `INT8=false` is
  the previous kernel unchanged.
- **prefill attention**: `win_prefill_lanepar_kernel` gains `<bool SINK, int NWARP>`. `SINK=false`
  drops the always-attended block 0 — the one semantic difference between the hd256 sink+window
  contract and Muse's pure sliding window. Both parameters default to today's values, so every
  existing instantiation compiles to the code it did before. The new hd128 entry point
  `launch_prefill_attn_swa_pure_int8` fits 35,328 B of smem and needs no `cudaFuncSetAttribute`.
- the `int8_kv()` decline in `prefill_batched_run` is removed.

A second commit then stages that int8 K/V into a **bf16** attention tile rather than fp32. At hd128
the fp32 tile costs 35,328 B of shared memory per block, capping the SM at 6 resident blocks where
18,944 B reaches the 8 that 256 threads/block allow, and it makes the per-lane K row 32 `float4`
loads instead of 16 `uint4` on a kernel that profiles as shared-memory-issue bound. Worth
+7.2% / +11.7% / +11.9% at 4k / 16k / 32k, with **bit-identical output** between the two tile
precisions (24/24 greedy tokens) — the int8 values are integers scaled by an fp16 scale, so bf16
carries them without moving an argmax. `SPARKINFER_MUSE_ATTN_I8_FP32=1` still selects the fp32 tile.

Correctness at ctx=4096 on a real tokenized prompt: the int8 arm goes from garbage to text that
tracks the bf16 reference for 18 of 24 greedy tokens. Batched prefill and the token loop diverge at
the same token under int8 as under bf16, so the residual difference is the pre-existing
prefill/decode numeric gap, not this path; and the token loop is bit-identical between an int8 and a
bf16 cache.

No regression, A/B interleaved against a stock build on the same box: Muse decode@ctx=0 0.9996 and
prefill@128 1.0027 (median of 5); Qwen3.8 and modelopt decode/prefill at 4k and 16k worst median
ratio 0.9993. The second commit was re-checked the same way against the first: worst ratio 0.9995
across Muse's decode@ctx=0 / prefill@128 and eight Qwen3.8 + modelopt dims.

## Proof of speedup

| | decode tok/s |
|---|--:|
| before (main) | 94.64 |
| after (this PR) | 96.61 |

| | prefill pp tok/s |
|---|--:|
| before prefill (main) | 102.61 |
| after prefill (this PR) | 4165.65 |

Time-to-first-token at the same context falls from 39.588 s to 0.964 s.

```text

--- before (main) ---
ttft         : 39.588 s  (103.5 tok/s prefill, prompt=4096)
decode tg    : 94.64 tok/s  (10.6 ms/token, n=32, ctx=4096, bs=1)
prefill pp   : 102.61 tok/s  (ctx=4096, sequential KV fill)

--- after (this PR) ---
ttft         : 0.964 s  (4247.7 tok/s prefill, prompt=4096)
decode tg    : 96.61 tok/s  (10.4 ms/token, n=32, ctx=4096, bs=1)
prefill pp   : 4165.65 tok/s  (ctx=4096, sequential KV fill)


$ bench/scripts/bench.sh Muse-Glimmer-30B-KQuant-17GB-Q4_K_M.gguf --tokens 32 --ctx 32768

--- before (main) ---
ttft         : 354.857 s  (92.3 tok/s prefill, prompt=32768)
decode tg    : 80.61 tok/s  (12.4 ms/token, n=32, ctx=32768, bs=1)
prefill pp   : 91.90 tok/s  (ctx=32768, sequential KV fill)

--- after (this PR) ---
ttft         : 15.299 s  (2141.8 tok/s prefill, prompt=32768)
decode tg    : 81.62 tok/s  (12.3 ms/token, n=32, ctx=32768, bs=1)
prefill pp   : 2053.84 tok/s  (ctx=32768, sequential KV fill)

GPU: NVIDIA GeForce RTX 5090, sm_120, clocks pinned by bench.sh (2475-2535 MHz under load)
```

### DIFF
--- a/kernels/csrc/cuda/attention/rope.cu
+++ b/kernels/csrc/cuda/attention/rope.cu
@@ -515,6 +515,85 @@ __global__ void rope_kv_append_partial_int8_kernel(
     }
 }
 
+// int8-KV decode append for Muse Glimmer. Muse's per-layer pattern is NORMAL (consecutive-pair)
+// RoPE on the SWA layers and NoPE on the global ones, and its KV write was unconditionally bf16 --
+// two-byte stores into a one-byte-per-element pool whenever the caller asked for an int8 cache,
+// which is why ctx >= 4096 (where the example mains switch the cache to int8) produced garbage.
+// Organised exactly like rope_kv_append_partial_int8_kernel: one block per (token, head-unit) with
+// blockDim == head_dim, so a whole K/V head vector reduces in-block for the per-(token, kv_head)
+// max-abs an int8 scale needs. Same scale layout the int8 flash-decode reads.
+// rope_normal == false is the NoPE global layer: pure quantize + append, Q untouched.
+__global__ void muse_kv_append_int8_kernel(
+    __nv_bfloat16* __restrict__ q, const __nv_bfloat16* __restrict__ k,
+    const __nv_bfloat16* __restrict__ v,
+    signed char* __restrict__ k_pool, signed char* __restrict__ v_pool,
+    __half* __restrict__ k_scale, __half* __restrict__ v_scale,
+    const int* __restrict__ block_table, const int* __restrict__ positions,
+    int n_q_heads, int n_kv_heads, int head_dim, float theta, int rope_normal,
+    int block_size, int max_blocks_per_seq
+) {
+    const int tok  = blockIdx.y;
+    const int unit = blockIdx.x;          // [0,nq)=Q rope ; [nq,nq+nkv)=K ; [nq+nkv,..)=V
+    const int t    = threadIdx.x;         // 0..head_dim-1
+    const int half = head_dim >> 1;
+    const int pos    = positions[tok];
+    const int blk    = pos / block_size, within = pos % block_size;
+    const int phys   = block_table[tok * max_blocks_per_seq + blk];
+    const size_t ctok = (size_t)(phys * block_size + within);
+
+    if (unit < n_q_heads) {               // Q: NORMAL rope in place, bf16, no quantisation
+        if (rope_normal && t < half) {
+            const float freq = __powf(theta, -2.f * (float)t / (float)head_dim);
+            const float ang = (float)pos * freq, c = __cosf(ang), sn = __sinf(ang);
+            const size_t base = ((size_t)(tok * n_q_heads + unit)) * head_dim + 2 * t;
+            const float x0 = __bfloat162float(q[base]), x1 = __bfloat162float(q[base + 1]);
+            q[base]     = __float2bfloat16(x0 * c - x1 * sn);
+            q[base + 1] = __float2bfloat16(x0 * sn + x1 * c);
+        }
+        return;
+    }
+
+    const bool is_k = unit < n_q_heads + n_kv_heads;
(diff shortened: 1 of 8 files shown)


skyrocket2026:

| metric | value |
|---|---|
| scored at | decode + prefill @ 128/512/4k/16k/32k — every axis is also a regression floor, label is the best |
| tier from | `muse-prefill@4k` (+3881.4%) |
| commit | `8089df6cd` |

|---|---|---|---|---|
| 128 | decode | 105.19 | 105.97 | +0.7% |
| 128 | prefill | 9224.53 | 9277.33 | +0.6% |
| 512 | decode | 103.36 | 104.24 | +0.9% |
| 512 | prefill | 4635.18 | 4705.32 | +1.5% |
| 4k | decode | 95.24 | 95.79 | +0.6% |
| 4k | prefill | 103.50 | 4120.61 | +3881.4% |
| 16k | decode | 88.90 | 89.37 | +0.5% |
| 16k | prefill | 97.46 | 2856.38 | +2831.0% |
| 32k | decode | 81.21 | 81.43 | +0.3% |
| 32k | prefill | 92.53 | 2042.28 | +2107.2% |

ok

<sub>Scored on the pinned eval box vs same-box `origin/main` — AR decode AND prefill at ctx 128/512/4k/16k/32k, from one model load; ANY axis regressing is a hard REJECT, but otherwise the reported label is the **best** measured delta across the ten — a PR that improves just one, with the rest flat, still earns credit for that. Cross-model no-regression guards run at 32k on Qwen3.6 and the ModelOpt Qwen3.8-27B NVFP4 checkpoint. This is informational, not a judgment on your PR: a `none` label just means no measurable Muse Glimmer speedup was verified on either metric, which is expected and fine if that isn't what your change is about. Correctness gated against a live llama.cpp reference on the same GGUF. Also gated on a Qwen3.6 no-regression guard (decode+prefill, ctx 0/512/4k/16k/32k, same box vs main) — Muse Glimmer PRs can touch code shared with other models. Automated — **not merged**; merge manually after review.</sub>

PR #1007, opened 2026-09-09
perf(decode): shape the NVFP4 activation quantize by row count and fold it into SwiGLU (1.02x cb-decode@c16)

widecloud:
## Summary

Every NVFP4 activation quantize on the packed continuous-batch decode path is issued as its own
graph node, in a lane shape fitted for prefill. `nsys` puts `quant_rows_t` at **3.88% of the c16
wall over 269 launches a step**, running at ~68 GB/s — those nodes are launch-ramp bound, not
bandwidth bound. Three changes to the same quantize:

1. **Choose the lane shape from the row count.** `LPV` (lanes per 16-value block) was one
   process-wide constant of 2. That is right for a prefill launch handed thousands of rows and
   wrong for a decode launch of at most 32, where the grid cannot fill the device.
2. **Fold the down projection's quantize into the SwiGLU that produces it.** The prefill arm and
   the dp4a arm already do this; only the block-scaled GEMM arm — the one a wide continuous batch
   takes — still wrote a full bf16 tensor nothing else reads and then re-read it.
3. **Give that fused SwiGLU quantize the same shape rule**, since it hardcoded the prefill strip.

Net: **+2.03% cb-decode@c16, +1.66% @c32**, bit-identical, every other scored dimension flat.

## Proof of speedup

**Decode tok/s** (continuous-batch aggregate, `qwen3_gguf_cb_bench <model> C 256 256 512` — the
scored invocation; each arm a fresh process, warm-up burnt, arms paired in both orders, 4 runs
each):

| | decode tok/s |
|---|--:|
| before (main) | 803.94 |
| after (this PR) | 820.23 |

| dim | main | this PR | delta |
|---|--:|--:|--:|
| cb-decode@c2 | 193.1 | 193.3 | +0.10% |
| cb-decode@c4 | 348.9 | 348.7 | -0.06% |
| cb-decode@c8 | 547.2 | 548.5 | +0.24% |
| **cb-decode@c16** | **803.94** | **820.23** | **+2.03%** |
| **cb-decode@c32** | **1357.90** | **1380.40** | **+1.66%** |

c16 is the mean of **seven** runs per arm across three separate sessions on a quiet GPU — main
804.5 / 804.1 / 804.0 / 802.8 / 804.2 / 804.2 / 803.8, PR 821.2 / 819.6 / 819.4 / 819.7 / 821.3 /
820.7 / 819.7. `decode_tokens` is identical in every arm of every dimension.

## Why the shape matters

`LPV` is the number of LANES cooperating on one 16-value block, so each lane handles
`VPL = 16 / LPV` values. `LPV=2` gives each lane a 16-byte load, a 4-byte store and a single-step
shuffle reduction — the right trade when the grid is already far wider than the device. A packed
decode step issues ~269 of these over at most 32 rows, and spreading each block across four times
as many lanes buys more than the two extra shuffle steps cost. Monotone in the lane count:
`LPV=4` is +1.29% at c16 against `LPV=8`'s +1.82%.

Forcing `LPV=8` everywhere is **not** the fix — it costs prefill@4k -1.39% and prefill@16k/@32k
-0.44%/-0.55%, because those launches want the wide strip. Gating on the row count takes the
decode win and leaves every prefill launch on the shape it already had.

After all three changes `quant_rows_t` drops from 3.88% of the c16 wall to **1.95%**, and its
launch count from 68,984 to 52,471 over the same run.

## Numerics

Bit-identical, by the same argument in both kernels: the lane split only regroups which lane
computes which element of a 16-value block, `max` is order-independent so each block keeps exactly
the same absmax, and every value then keeps the same `ue4m3` scale and the same `x / float(qs)`
rounding. The fused SwiGLU keeps the same SiLU-in-float, round-once-to-bf16 sequence as the
standalone pair it replaces.

Confirmed end to end — acceptance is unchanged to four decimals at every scored context and every
losslessness run passes:

| context | tau (main) | tau (this PR) | lossless |
|---|--:|--:|:--:|
| 4k | 1.6974 | 1.6974 | 1 |
| 16k | 1.7297 | 1.7297 | 1 |
| 32k | 1.3299 | 1.3299 | 1 |

## No-regression floors

Interleaved main/PR in the same session:

| dim | main | this PR | delta |
|---|--:|--:|--:|
| dspark-decode@4k | 130.6164 | 130.6334 | +0.01% |
| dspark-decode@16k | 129.6061 | 129.6213 | +0.01% |
| dspark-decode@32k | 97.2720 | 97.2445 | -0.03% |
| dspark-prefill@4k | 15873.97 | 15890.75 | +0.11% |
| dspark-prefill@16k | 12541.98 | 12545.57 | +0.03% |
| dspark-prefill@32k | 11320.41 | 11323.38 | +0.03% |
| target-prefill@256k | 5280.16 | 5260.50 | -0.37% |
| decode@256k | 94.98 | 94.98 | 0.00% |

Nothing is near the 0.98 floor. The scoped changes are packed-decode only (the SwiGLU fold sits
inside the `ffn_gemm` arm, which requires `packed`), so no speculative verify path changes at all;
the shape rule is row-gated, so every prefill launch keeps the shape it had.

`SPARKINFER_NVFP4_QUANT_LPV=2|4|8` still pins one shape for every launch, so both arms A/B out of
one binary; `2` is exactly what this shipped with.

### DIFF
--- a/kernels/csrc/cuda/fused/prefill_nvfp4_sm120.cu
+++ b/kernels/csrc/cuda/fused/prefill_nvfp4_sm120.cu
@@ -260,19 +260,46 @@ __global__ void quant_rows_t(const __nv_bfloat16* src, unsigned char* dst,
         }
     }
 }
-// SPARKINFER_NVFP4_QUANT_LPV=8 restores the two-values-per-lane shape (A/B in ONE binary).
-inline int nvfp4_quant_lpv() {
-    static const int v = [] {
+// LPV -- the number of LANES that cooperate on one 16-value NVFP4 block, so each lane handles
+// VPL = 16 / LPV values -- chosen by the SHAPE of the launch rather than one process-wide constant.
+//
+// LPV=2 (eight values per lane, two lanes per block) is right for a prefill quantize: it is handed
+// thousands of rows, the grid is already far wider than the device, and the wide per-lane strip
+// keeps the shuffle reduction to a single step. It is the wrong shape for a continuous-batch
+// DECODE quantize, which runs ~283 launches a step over at most kQwen35MaxPackedRows rows -- there
+// the grid cannot fill the GPU at all, and spreading each block over four times as many lanes buys
+// more than the two extra shuffle steps cost. Measured on an RTX 5090, same binary, interleaved
+// replicas on a quiet GPU: LPV=8 is +1.82% / +1.42% on cb-decode@c16/@c32 against the pinned
+// LPV=2, with LPV=4 at +1.29% on c16, i.e. monotone in the lane count. LPV=8 is the widest usable
+// value: LPV=16 would leave one value per lane and the paired-half load below has no half to
+// pair. nsys confirms the mechanism -- quant_rows_t is 3.88% of the c16 wall over 269 launches a
+// step, and those launches run at ~68 GB/s, i.e. they are launch-ramp bound, not bandwidth bound.
+//
+// Forcing LPV=8 everywhere is NOT the fix -- it costs prefill@4k -1.39% and prefill@16k/@32k
+// -0.44%/-0.55%, because those launches want the wide strip. Gating on the row count takes the
+// decode win and leaves every prefill launch on the shape it already had: measured +0.00% at
+// prefill@4k and @16k against the pinned default.
+//
+// Numerics are unchanged in either direction: LPV only regroups which lane computes which element
+// of a block, and the absmax reduction, the ue4m3 scale and the e2m1 rounding are all per
+// 16-element block. tau reads 1.6974 / 1.7297 / 1.3299 and LOSSLESS=1 at every context either way.
+//
+// SPARKINFER_NVFP4_QUANT_LPV=2|4|8 pins one shape for every launch (A/B in ONE binary); 2 is
+// exactly what this shipped with.
+static constexpr int kQuantNarrowLaneMaxRows = 64;  // decode packs <= 32; prefill hands thousands
+inline int nvfp4_quant_lpv(int rows) {
+    static const int forced = [] {
         const char* e = getenv("SPARKINFER_NVFP4_QUANT_LPV");
-        const int x = e ? atoi(e) : 2;
-        return (x == 2 || x == 4 || x == 8) ? x : 2;
+        const int x = e ? atoi(e) : 0;          // 0 = select by shape
+        return (x == 2 || x == 4 || x == 8) ? x : 0;
     }();
-    return v;
+    if (forced) return forced;
+    return (rows > 0 && rows <= kQuantNarrowLaneMaxRows) ? 8 : 2;
 }
 template <class Layout>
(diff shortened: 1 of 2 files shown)


skyrocket2026:

| metric | value |
|---|---|
| winning axis | **dspark-prefill@16k** — the dimension this label came from |
| **PR DSpark @4k** | **129.72 tok/s** |
| **main DSpark @4k** | **129.40 tok/s** |
| DSpark decode @4k vs main | +0.2% |
| **PR prefill @4k** | **14917.7 pp/s** |
| **main prefill @4k** | **14829.7 pp/s** |
| prefill @4k vs main | +0.5% |
| **PR DSpark @16k** | **130.55 tok/s** |
| **main DSpark @16k** | **129.54 tok/s** |
| DSpark decode @16k vs main | +0.7% |
| **PR prefill @16k** | **11650.7 pp/s** |
| **main prefill @16k** | **11420.9 pp/s** |
| prefill @16k vs main | +2.0% |
| **PR DSpark @32k** | **97.71 tok/s** |
| **main DSpark @32k** | **97.61 tok/s** |
| **PR prefill @32k** | **10445.0 pp/s** |
| **main prefill @32k** | **10398.5 pp/s** |
| prefill vs main | +0.4% |
| **PR prefill @256k** | **4914.9 pp/s** |
| **main prefill @256k** | **4902.9 pp/s** |
| **PR decode @256k** | **94.57 tok/s** |
| **main decode @256k** | **94.55 tok/s** |
| prefill @256k vs main | +0.2% |
| **PR concurrent decode @c2** | **186.8 tok/s agg** |
| **main concurrent decode @c2** | **185.8 tok/s agg** |
| concurrent decode @c2 vs main | +0.5% |
| **PR concurrent decode @c4** | **318.5 tok/s agg** |
| **main concurrent decode @c4** | **318.7 tok/s agg** |
| **PR concurrent decode @c8** | **511.7 tok/s agg** |
| **main concurrent decode @c8** | **511.7 tok/s agg** |
| concurrent decode @c8 vs main | +0.0% |
| **PR concurrent decode @c16** | **775.8 tok/s agg** |
| **main concurrent decode @c16** | **779.0 tok/s agg** |
| concurrent decode @c16 vs main | -0.4% |
| **PR concurrent decode @c32** | **1314.2 tok/s agg** |
| **main concurrent decode @c32** | **1325.1 tok/s agg** |
| concurrent decode @c32 vs main | -0.8% |
| main concurrent decode @c1 (floor) | 95.2 tok/s agg |
| PR AR tok/s (floor) | 87.91 |
| main AR tok/s (floor) | 88.50 |
| AR vs main (floor) | -0.6% |
| **DSpark vs AR** | **1.485×** — above 1.0 means speculation finally pays |
| qwen3.8 (upstream unsloth) guard @16k | ✅ no regression (decode+prefill) — decode 82.7 tok/s · prefill 9958 pp |
| qwen3.6 guard @16k | ✅ no regression (decode+prefill) — decode 462.6 tok/s · prefill 27289 pp |
| commit | `e51d90dfc` |

ok

PR #1008, opened 2026-09-09
perf(muse): mask prefill attention on the absolute position so a long prompt can be windowed (14.85x prefill@64k)

inference2026:
## Summary

Above `prefill_single_pass_max_tokens()` the runtime ingests a prompt in windows, calling
`prefill_batched_run` once per window with `pos0 > 0`. Muse refused every such call, so a prompt
longer than 32768 tokens did not merely lose the windowed path — the **whole** prompt fell back to
the token loop, which re-streams every weight once per position.

It refused for a real reason: both Muse attention kernels took the sliding-window start and the
causal bound from the query's row in the current pass, which is its sequence position only when the
pass starts at zero. On a later window every query would have attended the wrong span.

Give them the pass's start position instead. `qtok` stays the row (it indexes `q` and `attn`);
`qabs = q_pos0 + qtok` is the sequence position, and the window start, the causal mask and the key
range all move to it. The KV writer already took `pos0` and now receives it. At `q_pos0 == 0` every
value is what it was before, which is why the scored contexts are untouched.

The fp32-tile escape hatch in the int8 launcher is gated to `q_pos0 == 0`: that kernel derives its
window from the local row index and has no absolute position, so it stays an escape hatch rather
than becoming a silently wrong fallback.

**Correctness.** Forcing a windowed ingest at ctx=4096 with a 2048-token window, and again with a
1024-token window — two different `pos0` sequences — reproduces the single-pass token stream
exactly, 24/24 greedy tokens on a real tokenized prompt. No regression on Muse's scored dims
(decode@ctx=0 1.0021, prefill@128 1.0044), which run at `q_pos0 == 0` and never enter this path.

## Proof of speedup

| | decode tok/s |
|---|--:|
| before (main) | 69.36 |
| after (this PR) | 69.60 |

**Prefill pp tok/s** (Muse Glimmer 30B, `--ctx 65536`):

| | prefill pp tok/s |
|---|--:|
| before prefill (main) | 84.92 |
| after prefill (this PR) | 1261.09 |

```text
$ bench/scripts/bench.sh Muse-Glimmer-30B-KQuant-17GB-Q4_K_M.gguf --tokens 16 --ctx 65536
   (clocks pinned 2550; before = main, after = this PR, same box, interleaved)

--- before (main) ---
decode tg    : 69.36 tok/s  (14.4 ms/token, n=16, ctx=65536, bs=1)
prefill pp   : 84.92 tok/s  (ctx=65536, sequential KV fill)

--- after (this PR) ---
decode tg    : 69.60 tok/s  (14.4 ms/token, n=16, ctx=65536, bs=1)
prefill pp   : 1261.09 tok/s  (ctx=65536, sequential KV fill)
```

Decode is unchanged, as expected for a prefill-path change.

### DIFF
--- a/kernels/csrc/cuda/fused/prefill_attn_window.cu
+++ b/kernels/csrc/cuda/fused/prefill_attn_window.cu
@@ -457,12 +457,13 @@ __global__ void win_prefill_pure_bf16_kernel(
     const __nv_bfloat16* __restrict__ q, const __nv_bfloat16* __restrict__ k_pool,
     const __nv_bfloat16* __restrict__ v_pool, const int* __restrict__ block_table,
     __nv_bfloat16* __restrict__ attn, int n_tokens, int n_q_heads, int n_kv_heads,
-    int block_size, int max_blocks_per_seq, float scale, int win_blocks) {
+    int block_size, int max_blocks_per_seq, float scale, int win_blocks, int q_pos0) {
     constexpr int ELEMS = HEAD_DIM / 32;
     const int warp = threadIdx.x >> 5, lane = threadIdx.x & 31;
     const int head = blockIdx.y;
     const int qbase = blockIdx.x * TQ;
-    const int qtok = qbase + warp;
+    const int qtok = qbase + warp;          // row in THIS pass: indexes q/attn
+    const int qabs = q_pos0 + qtok;         // position in the SEQUENCE: window + causal mask
     const int kv_head = head / (n_q_heads / n_kv_heads);
     const bool active = (qtok < n_tokens) && (head < n_q_heads);
 
@@ -472,7 +473,7 @@ __global__ void win_prefill_pure_bf16_kernel(
         const int rsb = (win_blocks >= n_blk_q) ? 0 : (n_blk_q - win_blocks);
         return rsb * block_size;
     };
-    const int my_rs = active ? win_start(qtok) : 0;
+    const int my_rs = active ? win_start(qabs) : 0;
 
     extern __shared__ __nv_bfloat16 smem_bf[];
     __nv_bfloat16* sK = smem_bf;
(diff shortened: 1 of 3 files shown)


skyrocket2026:

| metric | value |
|---|---|
| scored at | decode + prefill @ 128/512/4k/16k/32k — every axis is also a regression floor, label is the best |
| tier from | `muse-prefill@64k` (+1417.0%) |
| commit | `e8504f3f3` |

|---|---|---|---|---|
| 128 | decode | 105.96 | 106.00 | +0.0% |
| 128 | prefill | 9272.50 | 9245.22 | -0.3% |
| 512 | decode | 104.23 | 104.23 | -0.0% |
| 512 | prefill | 4713.33 | 4705.47 | -0.2% |
| 4k | decode | 95.71 | 95.72 | +0.0% |
| 4k | prefill | 4128.99 | 4132.30 | +0.1% |
| 16k | decode | 89.32 | 89.29 | -0.0% |
| 16k | prefill | 2860.47 | 2865.89 | +0.2% |
| 32k | decode | 81.71 | 81.69 | -0.0% |
| 32k | prefill | 939.13 | 940.42 | +0.1% |
| 64k | decode | 68.67 | 69.08 | +0.6% |
| 64k | prefill | 85.16 | 1291.93 | +1417.0% |

ok

<sub>Scored on the pinned eval box vs same-box `origin/main` — AR decode AND prefill at ctx 128/512/4k/16k/32k, from one model load; ANY axis regressing is a hard REJECT, but otherwise the reported label is the **best** measured delta across the ten — a PR that improves just one, with the rest flat, still earns credit for that. Cross-model no-regression guards run at 32k on Qwen3.6 and the ModelOpt Qwen3.8-27B NVFP4 checkpoint. This is informational, not a judgment on your PR: a `none` label just means no measurable Muse Glimmer speedup was verified on either metric, which is expected and fine if that isn't what your change is about. Correctness gated against a live llama.cpp reference on the same GGUF. Also gated on two cross-model no-regression guards at 32k (decode+prefill, same box vs main): Qwen3.6-35B-A3B and the ModelOpt Qwen3.8-27B NVFP4 checkpoint — Muse Glimmer PRs can touch code shared with other models. Automated — **not merged**; merge manually after review.</sub>

PR #1009, opened 2026-09-09
perf(muse): run Muse's prefill attention on the int8 tensor cores (1.67x prefill@4k, 2.45x @16k)

inference2026:
## Summary

Attention is what a long Muse prefill spends its time on — **45.8% at ctx=4096 and 68.9% at 16384** —
and it ran on a scalar kernel. Measured by running the attention kernel N times per layer and taking
the slope of time-per-token against N, two independent slopes per context agreeing to 1.4%.

The in-tree int8 wmma kernel is already templated on `HEAD_DIM`; what kept Muse off it was the
launcher's hardcoded `HD=256` and the attention sink.

- At hd128 the divisibility rule (`GROUP_BLKS` must divide both `BM=16` and `HEAD_DIM/16`) gives
  **`GROUP_BLKS=8`**, not the hd256 default of 16.
- Muse is 32 q-heads over 2 kv-heads, so the GQA group of 16 takes the `RQH=4` tier, and its KV
  block size is 16, so a page is exactly a wmma fragment.
- The kernel implements a sink **plus** a window; Muse needs a **pure** window. `SINK` becomes a
  defaulted template parameter — every existing instantiation keeps the code it had — and hd128 is
  instantiated with `SINK=false`.

The entry point returns `false` on any shape it cannot serve, so the caller keeps the kernel it uses
today rather than consuming an output buffer nothing wrote.

**Also folded in — a latent out-of-bounds write.** `fp4_a`/`fp4_as` are shared by legs with
different row counts: gate/up runs inside the token-chunked FFN loop and passes `fn <= FC`, but the
qkv leg quantizes `N` rows in one call and so does the o projection. The buffer was sized by `FC`,
so with either enabled and `N > FC` the quantize writes `(N - FC)` rows past the end. It is
unreachable rather than absent today — Muse is the only configuration enabling those legs, and it
only reaches the FP4 path at `N == 128`, where `FC == N` by construction. It rides along here
because it has no measurable effect of its own and is the prerequisite for ever raising that gate.

## Proof of speedup

**Decode tok/s** (unchanged — this is a prefill-path change):

| | decode tok/s |
|---|--:|
| before (main) | 97.49 |
| after (this PR) | 97.53 |

| | prefill pp tok/s |
|---|--:|
| before prefill (main) | 4263.77 |
| after prefill (this PR) | 7122.49 |

```text
  (clocks pinned 2550; medians of 3, interleaved, both arms built from this branch's base)

  ctx=4096    main 4265 / 4264 / 4248   ->  this PR 7122 / 7126 / 6967    1.67x
  ctx=16384   main 2979 / 2975 / 2968   ->  this PR 7289 / 7288 / 7278    2.45x

$ ... --tokens 128 --ctx 4096      (decode, same pair)
  main    decode tg : 97.49 tok/s   prefill pp : 4253.71
  this PR decode tg : 97.53 tok/s   prefill pp : 7111.46
```

Bit-identical to the scalar kernel on a real tokenized prompt (24/24 greedy tokens), and the tier
debug confirms the wmma path is what launched rather than a silent fallback:

```text
[pf-attn-tier] RQH=4 SPL=4 GB=8 GN=128 smem=54016 qld=144 pld=144 n=4096 vt=0 widek=0 pvu=1 ok=1
```

Both gains sit under the ceiling the profile sets (1.845x at 4k, 3.22x at 16k) — that bound is what
distinguishes a real speedup from a kernel that stopped doing the work.

No regression on Muse's scored dims (decode@ctx=0, prefill@128) or on Qwen3.8 / modelopt at 4k and
16k; worst ratio 0.9987 against a 0.98 tolerance.

### DIFF
--- a/kernels/csrc/cuda/fused/prefill_attn_mma.cu
+++ b/kernels/csrc/cuda/fused/prefill_attn_mma.cu
@@ -379,7 +379,10 @@ __device__ __forceinline__ void pf_mma_16832(int (&d)[4], const unsigned (&a)[4]
                  : "r"(a[0]), "r"(a[1]), "r"(a[2]), "r"(a[3]), "r"(b0), "r"(b1));
 }
 
-template <int HEAD_DIM, int GROUP_BLKS, int RQH, int PLANES = 0, bool VT = false, bool WIDEK = false, int PVU = 1>
+// SINK=false drops the always-attended block 0, giving the PURE sliding window Muse Glimmer's
+// SWA layers use. Defaulted true, so every existing instantiation compiles to what it did before.
+template <int HEAD_DIM, int GROUP_BLKS, int RQH, int PLANES = 0, bool VT = false, bool WIDEK = false, int PVU = 1,
+          bool SINK = true>
 __global__ __launch_bounds__(GROUP_BLKS * 32, (GROUP_BLKS >= 16 ? 1 : (RQH <= 3 ? 2 : 1))) void pf_attn_mma_gqa_kernel(
     const __nv_bfloat16* __restrict__ q, const signed char* __restrict__ k_pool,
     const signed char* __restrict__ v_pool, const __half* __restrict__ k_scale,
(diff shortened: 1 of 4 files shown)


skyrocket2026:

| metric | value |
|---|---|
| tier from | `muse-prefill@64k` (+344.2%) |
| commit | `1ed6d50e0` |

|---|---|---|---|---|
| 128 | decode | 106.07 | 106.06 | -0.0% |
| 128 | prefill | 9262.79 | 9268.86 | +0.1% |
| 512 | decode | 104.45 | 104.48 | +0.0% |
| 512 | prefill | 4707.35 | 4708.37 | +0.0% |
| 4k | decode | 95.89 | 95.92 | +0.0% |
| 4k | prefill | 4134.48 | 6954.57 | +68.2% |
| 16k | decode | 89.49 | 89.50 | +0.0% |
| 16k | prefill | 2869.33 | 7051.25 | +145.7% |
| 32k | decode | 81.49 | 81.87 | +0.5% |
| 32k | prefill | 2062.60 | 6783.76 | +228.9% |
| 64k | decode | 69.16 | 69.41 | +0.4% |
| 64k | prefill | 1303.79 | 5791.42 | +344.2% |

ok

skyrocket2026:

PR #1013, opened 2026-09-09
perf(prefill): window a single pass whose FFN chunk has collapsed (10.8x prefill@32k)

inference2026:
## Summary

At `n == prefill_single_pass_max_tokens()`, in a session whose KV cache is sized for a **longer**
context, the single-pass arena has almost nothing left. The chunk-halving loop then shrinks `FC`
until the allocation fits and reports success — but the FFN re-streams the layer's `gate|up|down`
once per chunk, so the pass silently becomes `N/FC` passes over the weights. No error, no fallback
message, just an order of magnitude.

On Muse Glimmer this costs ~10x at exactly one context:

```
[prefill] ffn chunk 8192 -> 1024 (ctx=32768, free=687 MB) to keep the batched pass
prefill@32k = 912 pp   against 11608 at 4k, 10741 at 16k, 7999 at 64k
```

**64k is healthy, so this is the single-pass boundary rather than the length.** 32768 is the largest
prompt that still takes the single pass, and it is the one that cannot afford it — 64k already goes
down the windowed path and is fine.

It only reproduces in **one model load over several contexts**, which is what
`bench/scripts/_eval_speed.sh`'s `bench_sweep_run` does ("One model load, many contexts"). Measured
one context per process, a standalone 32k run sizes its cache for 32k, the arena fits, and the same
context does ~6800 pp — so this is invisible to per-context benchmarking.

Two narrow changes:

- **`prefill_batched_run` declines** when the chunk has collapsed to a quarter of the prompt or less
  *and* the caller has a window to fall back to (`pos0 == 0`, `N > window`). Running degraded is
  worse than handing the work back.
- **the single-pass caller, on a decline, tries the windowed path** before the token loop. It
  restarts at position 0, so a partially written cache is recomputed — prefill is deterministic in
  the prompt, and `pos0 == 0` resets the recurrent state exactly as the original call did.

A context whose single pass succeeds returns before either check and is untouched.

## Proof of speedup

**Decode tok/s** (unchanged — this is a prefill-path change):

| | decode tok/s |
|---|--:|
| before (main) | 82.27 |
| after (this PR) | 82.77 |

**Prefill pp tok/s** (Muse Glimmer 30B, `--ctx 32768`):

| | prefill pp tok/s |
|---|--:|
| before prefill (main) | 912.46 |
| after prefill (this PR) | 9840.12 |

```text
One model load over 128/512/4k/16k/32k/64k (the shape bench_sweep_run uses), clocks pinned 2550,
GPU idle before each run, PR built from this branch with no instrumentation:

    ctx      prefill  main ->    PR    delta      decode  main ->   PR   delta
    128         6232 ->  6309   +1.2%          105.39 -> 105.92   +0.5%
    512         9461 ->  9576   +1.2%          103.74 -> 104.26   +0.5%
    4096       11608 -> 11814   +1.8%           96.19 ->  96.91   +0.7%
    16384      10741 -> 10973   +2.2%           89.85 ->  90.57   +0.8%
    32768        912 ->  9840   +978%           82.27 ->  82.77   +0.6%
    65536       7999 ->  8189   +2.4%           70.42 ->  70.92   +0.7%

The change firing, from the same run:
    [prefill] ffn chunk 8192 -> 1024 (ctx=32768, free=687 MB) to keep the batched pass
    [prefill] ffn chunk collapsed to 1024 for a 32768-token single pass -- declining so the
              caller can window it
    [prefill] single pass declined at n=32768 -- windowing (16384) instead of the token loop
```

No axis regresses. The sub-2.5% moves at the other five contexts are the arena being less contended
and sit within run-to-run noise; the 32k figure is the change. Output verified on a real tokenized
prompt after the switch to the windowed path.

### DIFF
--- a/runtime/src/models/qwen35.cpp
+++ b/runtime/src/models/qwen35.cpp
@@ -3151,9 +3151,25 @@ int Qwen35Model::prefill_batched_chunked(const int* prompt_ids, int n, bool want
     // threshold, NOT the window size -- a prompt between the two is still one pass.
     if (window <= 0 || n <= prefill_single_pass_max_tokens()) {
         const int seed = prefill_batched(prompt_ids, n, want_seed_logprob);
-        if (seed < 0 || seed >= s.cfg.vocab) return -1;
-        if (out_done) *out_done = n;
-        return seed;
+        if (seed >= 0 && seed < s.cfg.vocab) {
+            if (out_done) *out_done = n;
+            return seed;
+        }
+        // The single pass declined. Returning here sends the WHOLE prompt to the token loop, and
+        // at n == prefill_single_pass_max_tokens() that is exactly what happens in a session whose
+        // KV cache is sized for a longer context: the single-pass arena cannot be allocated beside
+        // it, while the windowed path -- which the very next context up already uses successfully
+        // -- allocates one window and runs. Measured on Muse Glimmer in one model load over
+        // 16k/32k/64k (what bench_sweep_run does): prefill@32k 923 pp against 11018 at 16k and
+        // 8045 at 64k, i.e. the boundary itself, not the length. Windowing it gives 9629 pp.
+        //
+        // Try windowing before giving up. It restarts from position 0, so a single pass that had
+        // already written part of the cache is simply recomputed -- prefill is deterministic in
+        // the prompt, and pos0 == 0 resets the recurrent state the same way the original call did.
+        // Nothing here changes a context where the single pass succeeds: that returns above.
+        if (window <= 0 || n <= window) return -1;
+        fprintf(stderr, "[prefill] single pass declined at n=%d -- windowing (%d) instead of the "
+                        "token loop\n", n, window);
     }
     for (int pos = 0; pos < n; pos += window) {
         const int len = std::min(window, n - pos);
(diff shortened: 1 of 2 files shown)


skyrocket2026:

| metric | value |
|---|---|
| tier from | `muse-prefill@32k` (+42.8%) |
| commit | `f9c9bd440` |

|---|---|---|---|---|
| 128 | decode | 105.79 | 105.95 | +0.2% |
| 128 | prefill | 9266.70 | 9266.00 | -0.0% |
| 512 | decode | 104.11 | 104.26 | +0.1% |
| 512 | prefill | 12559.96 | 12602.41 | +0.3% |
| 4k | decode | 95.88 | 95.88 | -0.0% |
| 4k | prefill | 11625.75 | 11674.34 | +0.4% |
| 16k | decode | 89.40 | 89.52 | +0.1% |
| 16k | prefill | 10687.56 | 10739.39 | +0.5% |
| 32k | decode | 81.76 | 81.89 | +0.2% |
| 32k | prefill | 6748.09 | 9636.80 | +42.8% |
| 64k | decode | 69.47 | 69.53 | +0.1% |
| 64k | prefill | 8014.72 | 8053.90 | +0.5% |

ok

skyrocket2026:

PR #1014, opened 2026-09-09
perf(muse): stream ffn_down's NVFP4 operand per layer instead of holding 52 copies (1.17x prefill@4k)

inference2026:
## Summary

Muse's `ffn_down` is the last projection still running on the int8 tensor cores. On this card the
same leg measures **1168 TFLOPS on the FP4 path against 499 TOPS on the int8 one**, and it is 41%
of a 4k prefill by repeat-slope attribution — the largest single item left.

It has no resident FP4 copy because 52 of them are 3.9 GB that nothing can free: decode reads the
GGUF tensor directly, so unlike gate/up there is no native prefill copy to release after
converting. Holding them starves the batched-prefill scratch arena.

So the operand is built **one layer at a time** — dequantize that layer's Q4_K weight into staging,
quantize it to NVFP4, run the chunk loop against it, let the next layer overwrite it. A single
buffer is correct because it is ordered on the same stream as the GEMMs that read it. The
conversion costs ~31 ms for all 52 layers (repeat-slope at load) against ~92 ms saved on a 4k pass.

The scratch is allocated per pass, freed on every exit path, and only inside a length band
(1024..8192 tokens, `SPARKINFER_MUSE_NVFP4_DOWN_MINN` / `_MAXN`):

- below the band the fixed conversion cost swamps a prefill that itself only takes ~36 ms;
- above it the footprint cannot be spared. At `max_seq` 65536 the session has essentially no
  headroom, and 341 MB held across a 64k prefill costs 85% of it — measured with the buffers
  allocated and never read, so that is the footprint alone, not this path.

Outside the band nothing is allocated and the path is byte-for-byte what it was:
`qwen3_gguf_prefill_check` at a 16384 prefix reports KL 0.04335, top-1 21/24 and the same seed
token with the feature on and off — identical to five decimals. Every non-Muse model is untouched.

In band, teacher-forced against the token loop (4096 prefix, 64 continuation positions): top-1
0.8906 -> 0.8750, KL 0.02787 -> 0.03601. This same NVFP4 `ffn_down` leg already runs whenever
`max_seq <= 2048`, so it is an existing numeric reaching a configuration it could not previously
fit in, not a new one. Both accuracy passes (~300 and 8448 tokens) fall outside the band.

Interleaved sweep against a stock build, all six Muse contexts, drift-corrected (the card throttles
~2% over a four-run session): 128 1.012x · 512 1.007x · **4096 1.16x** · 16384 0.998x ·
32768 0.997x · 65536 0.998x; decode 0.994-0.998x everywhere.

## Proof of speedup

| | decode tok/s |
|---|--:|
| before (main) | 101.12 |
| after (this PR) | 101.37 |

| | prefill pp tok/s |
|---|--:|
| before prefill (main) | 11946.54 |
| after prefill (this PR) | 14017.69 |

Time-to-first-token at the same context falls from 0.333 s to 0.285 s.

```text

--- before (main) ---
ttft         : 0.333 s  (12307.5 tok/s prefill, prompt=4096)
decode tg    : 101.12 tok/s  (9.9 ms/token, n=32, ctx=4096, bs=1)
prefill pp   : 11946.54 tok/s  (ctx=4096, sequential KV fill)
GPU          : 70°C · 374 W · 2512 MHz (peak under load)

--- after (this PR) ---
ttft         : 0.285 s  (14395.5 tok/s prefill, prompt=4096)
decode tg    : 101.37 tok/s  (9.9 ms/token, n=32, ctx=4096, bs=1)
prefill pp   : 14017.69 tok/s  (ctx=4096, sequential KV fill)
GPU          : 67°C · 349 W · 2527 MHz (peak under load)

GPU: NVIDIA GeForce RTX 5090, sm_120, clocks pinned by bench.sh
```

### DIFF
--- a/runtime/src/models/qwen35_prefill.cpp
+++ b/runtime/src/models/qwen35_prefill.cpp
@@ -834,7 +834,52 @@ int prefill_batched_run(const Qwen35PrefillCtx& s, const int* prompt_ids, int n,
         ? kernels::prefill_nvfp4_data_bytes(fp4_a_rows, H) : 0;
     const size_t fp4_a_sf_bytes = gu_nvfp4
         ? kernels::prefill_nvfp4_scale_bytes_a(fp4_a_rows, H) : 0;
-    const bool muse_nvfp4_down = muse_nvfp4 && s.w.layers[0].down_fp4 &&
+    // ffn_down is the last projection Muse still runs on the int8 tensor cores, and on the FP4
+    // ones the same leg measures 2.34x (1168 TFLOPS vs 499 TOPS on this card). It has no resident
+    // FP4 copy because 52 of them are 3.89 GB that nothing can free -- decode reads the GGUF tensor
+    // directly -- and holding them starves the batched-prefill arena.
+    //
+    // So the operand is built ONE LAYER AT A TIME, and its scratch lives only as long as the pass
+    // that uses it. That second half is not a tidiness point: at max_seq 65536 the session has
+    // essentially no spare VRAM, and 341 MB held across a 64k prefill costs 85% of it -- measured
+    // with the buffers allocated and never read, so it is the footprint alone, not this path.
+    // Hence a band. Below dn_min the fixed ~31 ms of conversion (52 layers) swamps a prefill that
+    // only takes ~36 ms; above dn_max the scratch cannot be spared beside a long-context KV cache.
+    static const int dn_min = [] {
+        const char* e = getenv("SPARKINFER_MUSE_NVFP4_DOWN_MINN"); return e ? atoi(e) : 1024;
+    }();
+    static const int dn_max = [] {
+        const char* e = getenv("SPARKINFER_MUSE_NVFP4_DOWN_MAXN"); return e ? atoi(e) : 8192;
+    }();
+    static const bool dn_stream_on = [] {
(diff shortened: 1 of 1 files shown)


skyrocket2026:

| metric | value |
|---|---|
| tier from | `muse-prefill@4k` (+16.9%) |
| commit | `23dc34683` |

|---|---|---|---|---|
| 128 | decode | 106.34 | 106.49 | +0.1% |
| 128 | prefill | 9271.14 | 9263.27 | -0.1% |
| 512 | decode | 105.34 | 105.46 | +0.1% |
| 512 | prefill | 12591.91 | 12602.14 | +0.1% |
| 4k | decode | 100.35 | 100.52 | +0.2% |
| 4k | prefill | 11620.61 | 13588.85 | +16.9% |
| 16k | decode | 96.92 | 96.99 | +0.1% |
| 16k | prefill | 10691.94 | 10748.54 | +0.5% |
| 32k | decode | 92.23 | 92.32 | +0.1% |
| 32k | prefill | 9594.42 | 9637.87 | +0.5% |
| 64k | decode | 83.57 | 83.54 | -0.0% |
| 64k | prefill | 8040.09 | 8062.77 | +0.3% |

ok

skyrocket2026:

PR #1015, opened 2026-09-10
feat(server): LM Studio + Ollama REST APIs; eval: 12-axis Muse matrix, safer auto-close

skyrocket2026:
Splits the model-agnostic work off `feat/spark-x25-4b`, which keeps the Spark-X2.5 support that is not being merged.

## Server — two new API surfaces beside OpenAI `/v1`

**LM Studio `/api/v0/*`** — models (+`/{model}`), chat/completions with LM Studio's `stats`/`model_info`/`runtime` blocks, embeddings refused 501.

**Ollama `/api/*`** — version, tags, ps, show, chat, generate; embeddings and all model-management endpoints refused 501 with a reason rather than a bare 404.

Both **delegate to the same handlers `/v1` uses** and translate around them — there is no second implementation of generation, and `/v1` responses are byte-identical to before (asserted by test).

Streaming is genuinely incremental on all three surfaces. Every stream writer already funnelled through one function, so a per-request `StreamDialect` is applied there and the generation loop is untouched:

| dialect | framing |
|---|---|
| `OpenAiSse` | `data: {...}\n\n` — `/v1`, unchanged |
| `LmStudioSse` | same framing; usage chunk also carries `stats` |
| `OllamaNdjson` | `{...}\n` per line, message/done shape, no `[DONE]` |

Measured live: `/api/chat` 36 chunks spread 145–273 ms with one `done=true` carrying `eval_count`/`eval_duration`; `/api/v0` 38 chunks with `stats`; `/v1` 33 chunks, no `stats` leaked.

## Eval — three behaviour changes

**Auto-close narrowed to `REJECT` only, in both bots.** `none` means *this bot measured nothing on its axes*, which for a PR aimed at another model is the expected outcome, not a verdict. PR #1008 (a 15× Muse prefill win) was minutes from being auto-closed by the DSpark bot for exactly that reason. `REJECT` — a measured regression or a failed gate — still closes.

**Muse matrix 2 → 12 axes**: decode *and* prefill at 128/512/4k/16k/32k/64k, plus 32k no-regression guards on Qwen3.6 **and** the ModelOpt Qwen3.8 checkpoint. Schema bumped, so open PRs are re-scored rather than keeping two-dimension labels.

**Outage escalation** in both cron wrappers: a dead box degrades to `--labels-only` and exits 0, which logged one ordinary line and ran 14 consecutive times unnoticed on 2026-09-04/05. Now counted and escalated.

## Verification

Integration-tested against the **real clients** — official `ollama` CLI v0.33.3 via `OLLAMA_HOST`, and the OpenAI SDK 3.11.0 — not curl. That found six bugs curl could not, including a `nullptr` field that threw `type_error.302` and called `terminate()`, killing the whole server mid-stream, and an `/api/generate` mapping that returned HTTP 200 with unusable output. Each has a regression test naming the failure it prevents.

Clean build from scratch on this base; all server unit tests pass.

## Note for the maintainer

Only `server/` is evaluated by the bots — `eval/` and `bench/scripts/` are in `HARNESS_PATHS` and skipped. The eval changes above are already running from the local cron checkout; merging makes them the versioned truth.

### DIFF
--- a/eval/pr_dspark_bot.py
+++ b/eval/pr_dspark_bot.py
@@ -80,7 +80,8 @@ divergence that stayed hidden until the gate's prompt corpus was lengthened enou
 run at all. Enabling it here would REJECT every PR until that is fixed.
 
 Applies `eval-dspark:<TIER>` AND mirrors it to the generic `eval:<TIER>` label (SN74 scoring reads
-eval:* tiers). Auto-close on none/REJECT is live; auto-merge stays OFF unless
+eval:* tiers). Auto-close is live for REJECT ONLY -- never for "none", which for a PR
+aimed at another model is the expected outcome rather than a verdict; auto-merge stays OFF unless
 SPARKINFER_DSPARK_AUTOMERGE=1 is explicitly set.
 
   python eval/pr_dspark_bot.py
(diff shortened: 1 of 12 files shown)


skyrocket2026:

Touched: `eval/pr_dspark_bot.py`, `eval/pr_museglimmer_bot.py`, `eval/run_dspark_cron.sh`, `eval/run_museglimmer_cron.sh`

Issue #1001, opened 2026-09-08
/v1/score with completion_token_ids: correct argmax but wrong logprob value at the first forced positions (~22 nats), Qwen3.8-27B NVFP4, 19ef39ec2, deterministic mode


anderdc:
## Summary

On `19ef39ec2` serving `gittensor-model-hub/Qwen3.8-27B-NVFP4-RTX5090` (`SPARKINFER_DETERMINISTIC=1`, `CTX=131072`, int8 KV), `POST /v1/score` with `completion_token_ids` returns the **correct top token at every position** (`top_logprobs` argmax matches the generated token 100% on every prompt tried) but a **wrong logprob value** at the first forced positions:

- position 1: off by ~22–25 nats on **every** call, every prompt, deterministic (same value on repeat);
- position 0: off by ~18–31 nats on the **first** score call after a `/v1/chat/completions` on the same server, then correct on repeats;
- occasionally one later position (e.g. 6, 36, 57, 90), also ~15–25 nats.

Every other position matches the generation's reported logprob to 4+ decimals. Generation itself is bit-identical across two servers/cards (0.0 drift on token ids and logprobs), so this is isolated to the teacher-forced path. The `completion` (text) variant shows the same class of error at a different position (e.g. 28) while positions 0–27 match.

Measured 2026-09-08 on two RTX 5090s in one box (driver 595.84), each running the same image (`entrius/sparkinfer:19ef39ec2`, built from this repo at that commit), one as a generator, one as the scorer, and also same-server.

## Reproduction

```python
import requests
B = 'http://127.0.0.1:8080'; model = 'qwen3.8-27b'
msgs = [{'role': 'user', 'content': 'Explain in three sentences why the sky is blue.'}]
r = requests.post(f'{B}/v1/chat/completions', json={'model': model, 'messages': msgs, 'max_tokens': 48,
    'temperature': 0, 'logprobs': True, 'top_logprobs': 1, 'enable_thinking': False}).json()
c = r['choices'][0]['logprobs']['content']
ids, lp, toks = [e['token_id'] for e in c], [e['logprob'] for e in c], [e['token'] for e in c]
for rep in range(2):
    j = requests.post(f'{B}/v1/score', json={'model': model, 'messages': msgs, 'completion_token_ids': ids,
        'top_logprobs': 1, 'enable_thinking': False}).json()
    s = j['logprobs']; am = [a[0]['token'] for a in j['top_logprobs']]
[...]

PR #1016, opened 2026-09-10
fix(score): pass the scored token id by value, not through a raced device buffer

skyrocket2026:
Refs #1001.

## What this fixes

`token_logprob_for()` — the teacher-forced lookup behind `POST /v1/score` — re-runs the
"which token did we pick" kernel pointed at the caller's token id instead of the argmax's.
It staged that id like this:

```cpp
cudaMemcpy(d_score_id, &token_id, sizeof(int), cudaMemcpyHostToDevice);   // legacy default stream
launch_extract_chosen_logit(d_score_id, d_rank_by_id, d_sorted_logits, d_chosen_logit, s.stream);
```

`s.stream` is created with `cudaStreamNonBlocking`, which by definition does **not** serialize
against the legacy default stream. For a pageable source `cudaMemcpy` is "synchronous" only in
that it returns once the bytes are staged — the device-side write is still queued. So the kernel
can read the **previous** call's id and return `sorted_logits[rank_by_id[wrong id]]`.

The asymmetry matters: `argmax` and `top_logprobs` come from the sort (`d_sorted_idx`) and stay
correct; only the scalar `logprob` goes through this lookup. Scoring position *i* with token
*i-1*'s id lands far down the sorted row — tens of nats — which is a confident, wrong number
rather than an obviously broken one. That is the failure shape #1001 reports.

## Evidence

A standalone reproduction of the pattern on this box (**RTX 5090, driver 595.84**):

```
legacy-default-stream memcpy vs non-blocking-stream kernel
  iterations : 20000
  mismatches : 1
  first bad  : iter 4295 wanted 4295 saw 4294   <-- read the PREVIOUS value
  verdict    : UNORDERED (pattern is unsafe)
```

The value observed was exactly the previous iteration's — the predicted failure mode, not
generic corruption.

## Honest limits of this evidence

**This is not confirmed to be the whole of #1001, and this PR does not close it.**

- The reported symptom did **not** reproduce here, on matching GPU, driver, NVFP4 ModelOpt
  checkpoint, `CTX=131072`, int8 KV, `SPARKINFER_DETERMINISTIC=1`, and even identical
  `prompt_tokens=23` — cold, warm, idle, under 56 concurrent completions, short and ~2k-token
  prompts. Every position matched generation.
- Follow-up runs of the standalone test gave 0 mismatches in 180k further iterations. The real
  rate here is roughly 1 in 10^4–10^5, which **cannot** produce #1001's every-call,
  same-value-on-repeat failure at position 1.

So: a proven data race with the right failure shape, but the wrong frequency. Fixed on its own
merits — a scoring endpoint exists to be compared numerically against another copy of itself, and
a silent, unchecked memcpy racing a kernel has no place in that path.

Two other candidate causes were ruled out with evidence, not reasoning:
- cross-job scratch clobbering — the engine runs a **single** worker thread;
- packed decode — `step_jobs_packed()` explicitly declines teacher-forced and logprob batches.

The elimination argument still points at the id: correct argmax ⇒ correct sort ⇒ correct
`rank_by_id` ⇒ the id reaching the lookup is wrong. On the reporting environment something makes
that happen every time.

## The change

The id travels as a **by-value kernel argument**. Arguments are copied into the launch at launch
time, so there is no second stream and no ordering question. `d_score_id` is deleted so the
pattern cannot be reintroduced against a buffer that still looks live.

## Verification

- `/v1/score` values **unchanged and correct** on the fixed build — Qwen3.8-27B-NVFP4, every
  position matching generation, cold/warm/idle/under-load/long-prompt.
- Full suite `27/27` with `-DBUILD_SERVER=ON`.
- New regression subtest walks 585 distinct ids back to back on a non-blocking stream — the shape
  scoring a completion produces — and names the old failure signature ("PREVIOUS id") in its
  failure message.

### DIFF
--- a/kernels/csrc/cuda/fused/topk_topp_ops.cu
+++ b/kernels/csrc/cuda/fused/topk_topp_ops.cu
@@ -86,6 +86,15 @@ __global__ void extract_chosen_logit_kernel(const int* __restrict__ out_id,
     if (threadIdx.x == 0 && blockIdx.x == 0) *chosen_logit = sorted_logits[rank_by_id[*out_id]];
 }
 
+// By-value id variant -- see launch_extract_chosen_logit_id in fused.h for why the device-buffer
+// form must not be used outside a captured graph.
+__global__ void extract_chosen_logit_id_kernel(int token_id,
+                                               const int* __restrict__ rank_by_id,
+                                               const float* __restrict__ sorted_logits,
+                                               float* __restrict__ chosen_logit) {
+    if (threadIdx.x == 0 && blockIdx.x == 0) *chosen_logit = sorted_logits[rank_by_id[token_id]];
+}
+
 // logits[v] -= frequency_penalty * counts[v] + presence_penalty * (counts[v] > 0 ? 1 : 0), for
 // every v in [0, vocab). OpenAI's exact presence/frequency-penalty formula (counts[v] = number of
 // times vocab id v has appeared in THIS request's generated completion so far, accumulated across
@@ -255,6 +264,13 @@ void launch_extract_chosen_logit(const int* out_id, const int* rank_by_id,
                                  const float* sorted_logits, float* chosen_logit,
                                  cudaStream_t stream) {
     extract_chosen_logit_kernel<<<1, 1, 0, stream>>>(out_id, rank_by_id, sorted_logits, chosen_logit);
+}
+
+void launch_extract_chosen_logit_id(int token_id, const int* rank_by_id,
+                                    const float* sorted_logits, float* chosen_logit,
+                                    cudaStream_t stream) {
+    extract_chosen_logit_id_kernel<<<1, 1, 0, stream>>>(token_id, rank_by_id, sorted_logits,
+                                                        chosen_logit);
 }
 
 void launch_presence_frequency_penalty(float* logits, const int* counts, int vocab,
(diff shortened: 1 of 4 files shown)


skyrocket2026:
---

## Update: mechanism confirmed by exact numeric reproduction

A diagnostic build (not part of this PR) that forces the lookup to use the **previous** token's id
reproduces the issue's reported values to the decimal:

| position | reported in #1001 | forced stale-id build |
|---|---|---|
| 0 `'Sun'` | gen `-0.36` → score `-19.5` | gen `-0.36` → score `-19.5` |
| 1 `'light'` | gen `-0.0` → score `-22.6` | gen `-0.0` → score `-22.6` |

So the mechanism is settled: the scalar `logprob` comes from `sorted_logits[rank_by_id[id[i-1]]]`
rather than `id[i]`. It also explains the reporter's odd position-0 detail — that lookup inherits
the id left behind by the *previous request* (in both runs, `'.'` ending the prior completion),
which is why a repeat call is clean.

**This raises what the PR is worth.** The change removes the shared mutable id entirely: the id is
now a by-value kernel argument baked into the launch, so it is immune to *every* route to a stale
read — the stream-ordering race demonstrated here, a silently-failed copy, or anything else — not
only the one that could be reproduced. `d_score_id` is deleted so the pattern cannot return.

**Still not claimed:** that the stream-ordering race is what triggers it in the reporter's
environment. That race measures ~1 in 10^4–10^5 here and cannot produce their every-call failure;
something in their image makes the stale read near-certain at position 1. Asked on the issue for the
container's CUDA runtime version, which is the most likely difference for small pageable H2D copies.
`Refs #1001` stands rather than `Fixes` until they confirm on the box that reproduces.

PR #1025, opened 2026-09-11
perf(muse): stop materialising two prefill activations the FP4 path never reads (1.027x prefill@4k)

FranDev132:
## Summary

Muse Glimmer's batched prefill materialises two activations that no kernel on the FP4 path reads: it copies `q|gate|k|v` out of the single GEMM that produced them into four tight arrays, and its pre-FFN norm stages an int8 activation for an int8 FFN arm that never runs. Both are removed by handing the existing consumers a row pitch instead.

## Proof of speedup

**Decode tok/s** (not targeted by this PR — flat at every context):

| | decode tok/s |
|---|--:|
| before (main) | — |
| after (this PR) | — |

**Prefill pp tok/s** (best context, 4096):

| | prefill pp tok/s |
|---|--:|
| before prefill (main) | 14377.0 |
| after prefill (this PR) | 14768.3 |

### Every prefill context, one binary, two rounds per arm

| ctx | before | after |
|---|--:|--:|
| 128 | 9615.2 | 9825.8 |
| 512 | 15500.7 | 15846.1 |
| 4096 | 14377.0 | 14768.3 |
| 16384 | 13875.6 | 14211.0 |
| 32768 | 12667.7 | 12953.4 |
| 65536 | 10921.5 | 11135.3 |

Decode over the same runs: 106.68 → 106.33 at ctx 128 and 98.09 → 98.09 at 65536, i.e. unmoved.

### The split copies

The attention projections are one block-scaled GEMM into a packed `[N, 8704]` buffer, followed by four `cudaMemcpy2DAsync` that cut it into tight `qb`/`qg`/`kf`/`vf`. Measured over a prefill@128 pass those are 208 transfers, 115.87 MB, **0.261 ms — 2.0% of the pass**, moving 231.7 MB of read+write traffic at 888 GB/s.

None of it needs to happen. The QK-norm + RoPE + KV-append kernel already reads every q, k and v element, so it takes the packed row pitch and reads them in place; q still lands tight in its own destination because the attention that consumes it next wants that stride, and k and v go straight to the pools and never need a tight form at all. The gate's three consumers (the FP4 A-quantize, the int8 one, and the plain sigmoid multiply) take the same pitch. Zero added traffic — the copies are absorbed into passes that were already streaming those tensors.

### The int8 staging

The pre-FFN norm emits `hn` and, in the same pass, an int8 copy of it for the grouped int8 FFN. On this checkpoint that consumer does not exist: gate/up carry NVFP4 operands at every scored context, so the block-scaled arm runs and the int8 activation is written and never read. It is `N*H` bytes — 109 MB per layer per window at a 16384-token window, **11.3 GB over a 32k prefill** — and dropping it also turns a fused norm+quantize back into a plain read-modify-write. The test is the FP4 arm's own gate evaluated at the chunk width this branch already requires.

This is the same class as the ffn-wide staging already removed from this path; this is the H-wide one that survived it.

### Correctness

Bit-identical. `qwen3_gguf_prefill_check` at prefix 512 (bf16 KV) and 4096 (int8 KV), both arms:

```text
P=512   TOP1 12/16 0.7500   KL 0.14302   seed 220
P=4096  TOP1  6/16 0.3750   KL 0.21411   seed 220
```

Only the row pitch of a load changes and only a write nothing reads is dropped, so every value that reaches a GEMM is the value it reached before.

`SPARKINFER_MUSE_QKV_PACKED=0` and `SPARKINFER_MUSE_FFN_I8_SKIP=0` restore the copies and the staging respectively, for a paired A/B out of one binary.

### DIFF
--- a/kernels/csrc/cuda/fused/batched_prefill.cu
+++ b/kernels/csrc/cuda/fused/batched_prefill.cu
@@ -324,10 +324,14 @@ __global__ void pf_split_q_gate_kernel(const __nv_bfloat16* __restrict__ qraw,
 }
 
 __global__ void pf_mul_sigmoid_kernel(__nv_bfloat16* __restrict__ attn,
-                                      const __nv_bfloat16* __restrict__ gate, long n) {
+                                      const __nv_bfloat16* __restrict__ gate, long n,
+                                      int dim, int gate_ld) {
     const long i = (long)blockIdx.x * blockDim.x + threadIdx.x;
     if (i >= n) return;
-    attn[i] = __float2bfloat16(pf_to_f(attn[i]) * pf_sigmoid(pf_to_f(gate[i])));
+    // gate_ld != 0: `gate` is a column slice of a wider packed buffer, so its row pitch is not
+    // `dim`. attn stays tight, which is what every consumer of it expects.
+    const long g = gate_ld ? ((long)(i / dim) * gate_ld + (i % dim)) : i;
+    attn[i] = __float2bfloat16(pf_to_f(attn[i]) * pf_sigmoid(pf_to_f(gate[g])));
 }
 
 // One tap of the Gated-DeltaNet causal conv, by token index relative to THIS pass. A windowed
@@ -1156,7 +1160,13 @@ __global__ void pf_qknorm_ropenorm_kv_kernel(
     __half* __restrict__ k_scale, __half* __restrict__ v_scale,
     const int* __restrict__ block_table,
     int n_q_heads, int n_kv_heads, int head_dim, int rotary_dim, float theta, float eps,
-    int block_size, int max_blocks_per_seq, int pos0) {
+    int block_size, int max_blocks_per_seq, int pos0,
+    // src_ld > 0: q/k/v are COLUMN SLICES of one row-major [n_tokens, src_ld] buffer -- the
+    // packed q|gate|k|v this kernel's caller gets straight out of a single GEMM -- rather than
+    // three tight arrays. Each pointer still addresses its own column base; only the row pitch
+    // changes. q is written to `q_out` TIGHT either way, because the attention that reads it next
+    // wants a tight stride. src_ld == 0 keeps the original addressing for every other caller.
+    __nv_bfloat16* __restrict__ q_out, int src_ld) {
     auto* k_pool = static_cast<__nv_bfloat16*>(k_pool_v);
     auto* v_pool = static_cast<__nv_bfloat16*>(v_pool_v);
     auto* k_pool8 = static_cast<signed char*>(k_pool_v);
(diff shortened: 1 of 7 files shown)


skyrocket2026:

| metric | value |
|---|---|
| winning axis | **dspark-prefill@16k** — the dimension this label came from |
| **PR DSpark @4k** | **130.00 tok/s** |
| **main DSpark @4k** | **130.03 tok/s** |
| **PR prefill @4k** | **15054.3 pp/s** |
| **main prefill @4k** | **15051.8 pp/s** |
| **PR DSpark @16k** | **130.79 tok/s** |
| **main DSpark @16k** | **130.66 tok/s** |
| **PR prefill @16k** | **11727.6 pp/s** |
| **main prefill @16k** | **11714.7 pp/s** |
| prefill @16k vs main | +0.1% |
| **PR DSpark @32k** | **97.94 tok/s** |
| **main DSpark @32k** | **97.92 tok/s** |
| **PR prefill @32k** | **10550.6 pp/s** |
| **main prefill @32k** | **10546.4 pp/s** |
| prefill vs main | +0.0% |
| **PR prefill @256k** | **4952.6 pp/s** |
| **main prefill @256k** | **4998.8 pp/s** |
| **PR decode @256k** | **94.76 tok/s** |
| prefill @256k vs main | -0.9% |
| **main concurrent decode @c2** | **187.2 tok/s agg** |
| concurrent decode @c2 vs main | +0.0% |
| **PR concurrent decode @c4** | **320.8 tok/s agg** |
| **main concurrent decode @c4** | **320.8 tok/s agg** |
| concurrent decode @c4 vs main | +0.0% |
| **PR concurrent decode @c8** | **517.5 tok/s agg** |
| **main concurrent decode @c8** | **516.7 tok/s agg** |
| concurrent decode @c8 vs main | +0.2% |
| **PR concurrent decode @c16** | **773.7 tok/s agg** |
| **main concurrent decode @c16** | **774.5 tok/s agg** |
| concurrent decode @c16 vs main | -0.1% |
| **PR concurrent decode @c32** | **1310.8 tok/s agg** |
| **main concurrent decode @c32** | **1310.7 tok/s agg** |
| concurrent decode @c32 vs main | +0.0% |
| concurrent @c1 vs main (floor) | +0.1% |
| main AR tok/s (floor) | 88.56 |
| **DSpark vs AR** | **1.476×** — above 1.0 means speculation finally pays |
| qwen3.8 (upstream unsloth) guard @16k | ✅ no regression (decode+prefill) — decode 82.8 tok/s · prefill 10142 pp |
| qwen3.6 guard @16k | ✅ no regression (decode+prefill) — decode 461.6 tok/s · prefill 27461 pp |
| commit | `fe84b1103` |

within significance gate — not a verified dspark-prefill@16k improvement

skyrocket2026:

| metric | value |
|---|---|
| tier from | `muse-prefill@512` (+3.1%) |
| commit | `fe84b1103` |

|---|---|---|---|---|
| 128 | decode | 106.70 | 106.70 | -0.0% |
| 128 | prefill | 9631.30 | 9865.49 | +2.4% |
| 512 | decode | 105.86 | 105.84 | -0.0% |
| 512 | prefill | 15540.77 | 16017.84 | +3.1% |
| 4k | decode | 102.79 | 102.86 | +0.1% |
| 4k | prefill | 14563.34 | 14979.08 | +2.9% |
| 16k | decode | 101.95 | 101.97 | +0.0% |
| 16k | prefill | 14080.39 | 14430.04 | +2.5% |
| 32k | decode | 100.44 | 100.53 | +0.1% |
| 32k | prefill | 12837.51 | 13128.75 | +2.3% |
| 64k | decode | 98.27 | 98.22 | -0.0% |
| 64k | prefill | 11030.71 | 11244.84 | +1.9% |

ok

skyrocket2026:

PR #1023, opened 2026-09-10
perf(muse): give the o-projection its NVFP4 operand at the contexts the arena can hold it (1.046x prefill@4k)

widecloud:
## Summary

Muse Glimmer's o-projection is the last dense **int8** GEMM in its prefill layer at ctx 4096 and up
(7.0% of prefill@16k) — gate, up, ffn_down and q|gate|k|v are already NVFP4 there. The FP4 arm
already ships at ctx 128/512; it was refused above `max_seq` 2048 by a VRAM proxy inherited from
when `wo` was budgeted alongside an `ffn_down` copy 4.7x its size. One line moves that bound to
where the arena measurably still fits.

## Proof of speedup

- [x] Tested on **RTX 5090** (`sm_120`)

**Target model(s)**

- [x] **Muse Glimmer**
- [ ] **Qwen3.8-27B** (ModelOpt NVFP4 / DSpark)
- [ ] **Shared / both** — the change is in code both models use and should help either

The changed line is inside `if (c.muse_glimmer && ...)` in `qwen35.cpp`, so it is unreachable for
Qwen3.8 — I still ran the ModelOpt guard and its `qwen3_gguf_score` dump is byte-identical to main.

**Decode tok/s** (this PR does not target decode; all six decode axes are flat):

| | decode tok/s |
|---|--:|
| before (main) | 105.432 |
| after (this PR) | 105.402 |

**Prefill pp tok/s** (`--ctx 4096`, the best context):

| | prefill pp tok/s |
|---|--:|
| before prefill (main) | 14338.37 |
| after prefill (this PR) | 15001.01 |

Also `--ctx 32768`: 12534.06 → 12977.21 (1.035x). `--ctx 65536` is unchanged — it loads at
`max_seq` 65680, where the copy does not fit, so it stays outside the bound.

`bench/scripts/bench.sh` benchmarks prebuilt release binaries and cannot measure a branch. Numbers
below are `bench_sweep_run` (the 12-axis scored sweep), arms interleaved main / PR / main out of
ONE binary with `SPARKINFER_MUSE_NVFP4_WO_MAXSEQ=2048` as the control.

```text
              control (main)              this PR
AX 128    109.3401   9669.7494      109.2166   9649.9481    1.000x
AX 512    108.5877  15408.2189      108.5027  15391.2154    0.999x
AX 4096   105.4551  14372.0773      105.4024  15001.0094    1.046x
AX 16384  104.4652  13747.0550      104.4115  14241.0456    1.038x
[...]

PR #1018, opened 2026-09-10
perf(muse): give the 16:1 GQA group its whole kv-head in the prefill attention (1.16x prefill@64k)

FranDev132:
## Summary

One block of the tensor-core prefill attention stages a kv-head's K/V tile and feeds it to `RQH`
q-heads, so the group's keys are re-read `GQA / RQH` times per query tile. Muse Glimmer is 32Q/2KV
— a 16:1 group — against `RQH=4`, so every key is read four times.

Widening it is what the rolling score plane exists for: `PLANES` holds that many heads at a time
instead of `RQH`, so the plane stops being what caps `RQH`. At `<128, RQH=16, PLANES=2>` the block
takes 96,256 B, where `RQH=16` with a full plane would need 302 KB.

`RQH` also divides the grid — `blocks = ceil(n/BM) * (n_q_heads/RQH)` — so a short prompt has too
few query tiles to fill the device once the group is wide (at n=128, `RQH=16` leaves 16 blocks for
170 SMs). So this takes the widest group that still covers the machine twice over, and keeps the
shipped shape below that.

**Bit-identical.** `RQH` changes which heads share a staged tile, not the per-row arithmetic or the
order it accumulates in — `qwen3_gguf_prefill_check` at prefix=4096 returns TOP1 11/16, KL 0.04134,
seed 220 on both. Below the batched verify's `seqlen > 512` the tensor-core tier is not armed at
all, so 128 and 512 run main's path untouched.

**The tiers step down rather than out.** An over-budget shape fails the smem opt-in and the caller
is then left with no mma tier and drops to the scalar kernel: the 113 KB `<128,16,PLANES=4>` shape
measures 2677 pp against 11667 at ctx=16384. A refused tier now tries the next narrower one, ending
at the `RQH=4` shape that ships today, and the last error is cleared between tiers so a refusal
cannot poison the next tier's launch check. `SPARKINFER_PREFILL_ATTN_TIER=1` prints the tier that
actually launched.

- [x] Tested on **RTX 5090** (`sm_120`)

**Prefill pp tok/s** (main binary vs this branch, same box, ctx=65536):

| | prefill pp tok/s |
|---|--:|
| before prefill (main) | 8037.73 |
| after prefill (this PR) | 9329.52 |

**Decode tok/s**: not claimed — this PR targets prefill. Decode measures 98.2939 → 98.3283 tok/s at
ctx=65536, i.e. unchanged; the full numbers are in the block below.

```text
[...]

PR #1027, opened 2026-09-11
feat(eval): declare the target model in the PR, and stop evaluating PRs that cannot move a bot's axes

skyrocket2026:
## The waste

Both bots evaluate every greenlit PR, so a Muse Glimmer change pays for a full DSpark round to be
told it did nothing for Qwen3.8:

| PR | what it is | DSpark result |
|---|---|---|
| [#1025](../../pull/1025) | `perf(muse)`, Muse prefill | `eval-dspark:none`, **+0.1%**, after a ~20-min eval |
| [#1018](../../pull/1018) | `perf(muse)`, Muse prefill@64k | `eval-dspark:none`, +1.4% (scored `L` **+15.3%** by the Muse bot) |
| [#1023](../../pull/1023) | `perf(muse)`, Muse prefill@4k | judged on decode, skipped |

That is most of a GPU hour a day spent proving things the authors already knew.

## The change

The PR template gets a **Target model(s)** block next to the RTX 5090 box, and CONTRIBUTING
documents it. A bot whose model you did not tick may skip your PR.

```
- [ ] Muse Glimmer
- [ ] Qwen3.8-27B (ModelOpt NVFP4 / DSpark)
- [ ] Shared / both — the change is in code both models use and should help either
```

## Fail-open by construction

`arb.model_skip_reason()` skips **only** when the author ticked a specific model, did not tick this
bot's model, and did not tick `Shared / both`. Every other case evaluates exactly as before:

- nothing declared, or no box ticked
- an unrecognised model name
- `Shared / both` ticked, or more than one model ticked
- the `Tested on RTX 5090` attestation box (explicitly never read as a model declaration)

The declaration can only ever *remove* work the author has said is pointless — it can never cause a
PR to go unevaluated by accident. Verified against the real #1025 body, which predates the field:

```
declared    : (nothing — fails open)
dspark skip?: no — would evaluate
muse skip?  : no — would evaluate
```

## Why only the DSpark bot honours it

Deliberate, not partial work. Skipping is safe only where **another bot still guards the skipped
model** — shared code means a change aimed at one model regularly lands in another's path:

| Bot | Guards it runs |
|---|---|
| Muse | Qwen3.6 **and** ModelOpt (Qwen3.8) @32k |
| DSpark | Qwen3.6 **and** Qwen3.8 @16k — **nothing guards Muse Glimmer** |

So DSpark can skip a Muse-declared PR and lose no coverage. The Muse bot skipping a Qwen3.8-declared
PR would let a Muse regression land unnoticed, so it does not skip. CONTRIBUTING records this as a
harness gap that closes when a Muse guard is added to the DSpark bot — not as policy.

Mis-declaring to route around a guard is documented as gaming, treated like false attestation.

## Tests

11 new cases in `test_pr_eval_bot.py` — the #1025 skip, the `dspark`/`modelopt`/`qwen3.8` aliases,
`Shared / both`, both-ticked, and every fail-open path.

```
eval/test_pr_eval_bot.py   84 OK   (11 new)
eval/test_pr_dspark_bot.py 10 OK
eval/test_copycat_guard.py  7 OK
```

`eval/` is in `HARNESS_PATHS`, so the bots skip evaluating this PR rather than spending GPU on it.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -32,6 +32,19 @@
 
 - [ ] Tested on **RTX 5090** (`sm_120`)
 
+**Target model(s)** — tick every model this change is meant to speed up. This decides which bots
+spend GPU time on it: a bot whose model you did **not** tick may skip your PR instead of spending a
+~20-minute round proving a change it cannot move. Tick nothing and every bot evaluates it, as before.
+
+- [ ] **Muse Glimmer**
+- [ ] **Qwen3.8-27B** (ModelOpt NVFP4 / DSpark)
+- [ ] **Shared / both** — the change is in code both models use and should help either
+
+> Tick **Shared / both** if you are unsure, or if the code you touched is shared (`qwen35.cpp`,
+> `qwen35_prefill.cpp`, most of `kernels/`). Over-ticking only costs eval time; under-ticking can
+> cost you a tier a bot would have awarded. Declaring a model you did not target in order to dodge
+> a no-regression guard is gaming, and is treated like false attestation.
+
 **Decode tok/s** (end-to-end, from `bench/scripts/bench.sh` — fill if this PR targets decode):
 
 | | decode tok/s |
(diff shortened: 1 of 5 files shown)


skyrocket2026:

Touched: `eval/pr_dspark_bot.py`, `eval/pr_eval_bot.py`, `eval/test_pr_eval_bot.py`

skyrocket2026:

Touched: `eval/pr_dspark_bot.py`, `eval/pr_eval_bot.py`, `eval/test_pr_eval_bot.py`

Issue #1026, opened 2026-09-11
Evaluation axis request: concurrent decode (cb-decode@cN) on Muse Glimmer


FranDev132:
## The axis

**Concurrent decode throughput on Muse Glimmer** — aggregate tok/s with N requests in flight, i.e. `cb-decode@c2` … `cb-decode@c32`.

`pr_museglimmer_bot.py` scores twelve dimensions, `muse-{decode,prefill}@{128,512,4k,16k,32k,64k}`. Every one of them measures a **single** request: `bench_decode` drives one sequence and `prefill_batched_chunked` ingests one prompt. Nothing in that set observes what happens when the engine has more than one request live, so a change that only moves multi-request throughput scores `none` on all twelve.

## The command that measures it

Already in the tree, and already used by another bot:

```bash
build/runtime/qwen3_gguf_cb_bench "$MUSEGLIMMER_GGUF" "$CC" 256 256 512
# metric: agg_tok_s   (the `wall_s=... agg_tok_s=...` line)
# for CC in 2 4 8 16 32
```

`pr_dspark_bot.py` already scores exactly this — `cb-decode@c2/c4/c8/c16/c32`, parsed with

```bash
CB_AGG=$(sed -n 's/.*agg_tok_s=\([0-9.]*\).*/\1/p' "$CB_OUT" | tail -1)
```

so the harness, the binary, the invocation shape and the parser all exist. What is missing is pointing them at the Muse checkpoint.

## Why it is worth an axis

Muse gets **no concurrency scaling at all** on main. Aggregate throughput is flat from one request to sixteen, because every concurrent request re-reads all ~17 GB of weights on its own:

| c | 1 | 2 | 4 | 8 | 16 | 32 |
|---|--:|--:|--:|--:|--:|--:|
| main, agg tok/s | 102.1 | 103.0 | 103.4 | 103.3 | 101.4 | 52.5 |

For contrast, Qwen3.8-27B on the same engine goes 95 → 185 → 309 → 483 → 626 → 805.

## Before/after numbers

Measured on an RTX 5090, `Muse-Glimmer-30B-KQuant-17GB-Q4_K_M.gguf`, one binary with the change env-switched, `qwen3_gguf_cb_bench <gguf> <c> 256 256 512`:

| c | main | with the change | delta |
|---|--:|--:|--:|
| 2 | 103.0 | 129.5 | +25.5% |
| 4 | 103.5 | 171.0 | +65.2% |
[...]

PR #1028, opened 2026-09-11
perf(muse): decode concurrent requests in one packed forward (1.42x concurrent decode @c8)

FranDev132:
## Summary

Muse Glimmer never reached the packed continuous-batch decode forward, so every concurrent request re-read all ~17 GB of weights on its own — aggregate throughput is flat at ~103 tok/s from c=1 to c=16. This wires the architecture's layer into that forward.

## Proof of speedup

**Target model(s)**

- [x] **Muse Glimmer**
- [ ] **Shared / both**

The packed forward is gated on `c.muse_glimmer`. The one piece outside that gate — chunking `launch_mmvq_rows` above 8 rows — only fires for a multi-row GGUF quantized decode, which the ModelOpt NVFP4 checkpoint does not take. The Qwen3.6 and ModelOpt no-regression guards still run either way.

| | decode tok/s |
|---|--:|
| before (main) | 103.5 |
| after (this PR) | 147.4 |

| | prefill pp tok/s |
|---|--:|
| before prefill (main) | — |

### Concurrent decode — main binary vs this branch, both built at the same commit

| c | before | after |
|---|--:|--:|
| 2 | 103.1 | 117.5 |
| 4 | 103.5 | 139.3 |
| 8 | 103.5 | 147.4 |
| 16 | 101.6 | 130.5 |

Aggregate throughput on main is flat from one request to sixteen: every concurrent request re-reads all ~17 GB of weights on its own, because the engine has no batched decode forward it can enter. For contrast, Qwen3.8-27B on the same engine goes 95 → 185 → 309 → 483 → 626.

c=32 is unchanged. The card has ~1 MB free there once the KV pool for 33 sequences is allocated, so the packed forward cannot get its scratch and the rows decode individually exactly as before; the batched prefill falls back at that width too, which is why main itself drops to ~52 tok/s.

### What kept it out — four gates, three of them silent

- `dflash_verify_short_run` refused `head_dim != 256`.
- `launch_mmvq_q4k_rows` had no K=6656 instantiation, so every projection off `xn` was refused.
- `launch_mmvq_q6k_rows` had no K=6656 either. A `Q4_K_M` file gives half the layers a Q6_K `attn_v`, and that refusal is a bare `false` inside `proj_on` — it surfaced 52 layers later as a decline with no cause.
- `launch_mmvq_rows` refused `M > 8` outright, which declined the whole forward at c=16 and c=32. It now chunks by 8, the way its f32 twin already did; `MMAX` bounds the scratch and the number of predicated row bodies, not the per-row dot or its reduction order.

### The layer is a branch, not a widening

Mirroring what the batched prefill already carries for this architecture: unweighted RMSNorm of the embedding, `attn_gate` as its own `[qdim, H]` tensor (projecting `wq` as `2*qdim` reads 4096 rows past it), NORM-convention RoPE on the windowed layers and none at all on the every-4th global ones, a pure sliding-window view per row, sandwich norms with their own 1e-8 eps, and the tanh logit softcap.

The per-row KV-append and attention kernels already index `block_table[row * max_blocks + blk]` and `positions[row]`, so they serve N independent sequences unchanged. Only the compact sliding-window view needed a rows form; at one row it reduces exactly to the existing launcher.

The Gated-DeltaNet recurrence records are no longer allocated when no layer has `linear_attn` — ~48 MB of verify arena this architecture can never touch.

### No scored dimension moves

main binary vs this branch at the evaluation's own tier shapes — 128/512 at reps=5 (two rounds per arm), 4k/16k/32k at reps=1:

| ctx | main decode | PR decode | main prefill pp | PR prefill pp |
|---|--:|--:|--:|--:|
| 128 | 105.88 | 105.80 | 9589.6 | 9616.9 |
| 512 | 105.21 | 105.19 | 15461.4 | 15372.3 |
| 4k | 102.33 | 102.40 | 14368.9 | 14382.1 |
| 16k | 101.48 | 101.54 | 13886.0 | 13898.5 |
| 32k | 100.13 | 100.19 | 12686.5 | 12698.6 |
| 64k | 98.81 | 98.65 | 11227.6 | 11147.5 |

Every point is within ±0.75%, the largest being prefill@64k at −0.71% and prefill@512 at −0.58% (the regression floor is −2%). By construction none of this is reachable from a single-request path: the packed forward is entered only from `decode_packed`, the `M > 8` chunk is new code on a branch a one-row call never takes, the added K widths leave every existing width's dispatch untouched, and the compact-window rows kernel has no other caller.

### Correctness

Token streams from the packed forward compared against the same build decoding the rows one at a time, distinct in-distribution prompts per row, greedy: most 64-token streams are identical and the divergence rate does not grow with the batch width. Every stream that crosses the 2048-token sliding window is identical, which is the per-row window view and the position handling. Packed decode is not bit-identical to autoregressive decode — the multi-row kernels partition their super-blocks differently — and `packed` is what keeps that separate from the speculative verify, which stays byte-for-byte on the row kernels.

```text
$ qwen3_gguf_cb_bench Muse-Glimmer-30B-KQuant-17GB-Q4_K_M.gguf 8 256 256 512
cb_bench policy=continuous concurrency=8 prompt=256 max_new=256 long_prefill=512

# before (SPARKINFER_MUSE_PACKED=0)
wall_s=19.900 decode_tokens=2056 agg_tok_s=103.3 mean_itl_ms=76.07 max_itl_ms=118.46

# after
wall_s=14.088 decode_tokens=2056 agg_tok_s=145.9 mean_itl_ms=53.48 max_itl_ms=94.17
```

### The axis

`muse-cb-decode@c2/c4/c8/c16/c32` now exists, and its invocation is `qwen3_gguf_cb_bench "$GGUF" "$CC" 256 256 512` — the same shape every number above was measured at.

### DIFF
--- a/kernels/csrc/cuda/attention/sparse_kv.cu
+++ b/kernels/csrc/cuda/attention/sparse_kv.cu
@@ -219,6 +219,43 @@ void launch_fa_kv_compact_view_pure(
         seq_lens, block_table, view_table, view_len, block_size, window_w, n_view);
 }
 
+// Per-ROW form of the above, for packed continuous-batch decode: one block builds one row's
+// view, so N independent sequences get N compact windows in a single launch. Identical
+// arithmetic per row -- the single-row launcher is exactly this with N == 1 -- so a windowed
+// layer sees the same view it would have seen decoding that sequence alone.
+//
+// `block_table` is the ROW-GATHERED table (stride max_blocks, as produced by
+// launch_gather_rows_i32), `seq_lens` is [rows], and the outputs are [rows][n_view] / [rows].
+__global__ void fa_kv_compact_view_pure_rows(
+    const int* __restrict__ seq_lens, const int* __restrict__ block_table,
+    int* __restrict__ view_table, int* __restrict__ view_len,
+    int block_size, int window_w, int n_view, int max_blocks
+) {
+    const int r = blockIdx.x;
+    const int sl = seq_lens[r];
+    const int* __restrict__ bt = block_table + (size_t)r * max_blocks;
+    int* __restrict__ vt = view_table + (size_t)r * n_view;
+    const int n_blk = (sl + block_size - 1) / block_size;
+    if (window_w >= n_blk) {
+        for (int b = threadIdx.x; b < n_blk && b < n_view; b += blockDim.x) vt[b] = bt[b];
+        if (threadIdx.x == 0) view_len[r] = sl;
+        return;
+    }
+    const int recent_start = n_blk - window_w;
+    if (threadIdx.x == 0) view_len[r] = sl - recent_start * block_size;
+    for (int i = threadIdx.x; i < window_w && i < n_view; i += blockDim.x)
+        vt[i] = bt[recent_start + i];
+}
+
+void launch_fa_kv_compact_view_pure_rows(
+    const int* seq_lens, const int* block_table, int* view_table, int* view_len,
+    int block_size, int window_w, int n_view, int max_blocks, int rows, cudaStream_t stream
+) {
+    if (rows < 1) return;
+    fa_kv_compact_view_pure_rows<<<rows, 256, 0, stream>>>(
+        seq_lens, block_table, view_table, view_len, block_size, window_w, n_view, max_blocks);
+}
+
 void launch_flash_decode_split_sparse(
     const void* q, const void* k_pool_layer, const void* v_pool_layer,
     const int* block_table, const int* seq_lens, const int* sel_blk,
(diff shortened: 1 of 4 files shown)


FranDev132:
Axis request for this work is #1026 — `cb-decode@c2/c4/c8/c16/c32` on Muse Glimmer, mirroring the block that already exists in `pr_dspark_bot.py`.

Requesting the `hold` label so this isn't auto-closed while that is considered.

skyrocket2026:

| metric | value |
|---|---|
| tier from | `muse-cb-decode@c8` (+43.9%) |
| commit | `20f3b9af6` |

|---|---|---|---|---|
| 128 | decode | 106.40 | 106.43 | +0.0% |
| 128 | prefill | 9870.77 | 9847.41 | -0.2% |
| 512 | decode | 105.44 | 105.59 | +0.1% |
| 512 | prefill | 15723.30 | 15927.69 | +1.3% |
| 4k | decode | 102.68 | 102.66 | -0.0% |
| 4k | prefill | 15105.85 | 15049.17 | -0.4% |
| 16k | decode | 101.42 | 101.68 | +0.3% |
| 16k | prefill | 13904.03 | 14411.24 | +3.6% |
| 32k | decode | 99.18 | 100.32 | +1.1% |
| 32k | prefill | 12018.59 | 12932.86 | +7.6% |
| 64k | decode | 98.06 | 98.01 | -0.1% |
| 64k | prefill | 10773.27 | 11095.20 | +3.0% |

ok

skyrocket2026:

PR #1029, opened 2026-09-11
perf(muse): give the Q3_A gate/up a weight-stationary multi-row arm (1.32x concurrent decode @c8)

FranDev132:
## Summary

`gate_up_q3a_kernel` streams Muse Glimmer's whole gate+up weight set once **per token** in a batched decode step. It is 44.4% of GPU time in a concurrent step and runs at 89% of peak bandwidth doing eight passes over bytes that only need one. Holding the output row in a single CTA and walking the token rows inside makes it one pass.

## Proof of speedup

**Target model(s)**

- [x] **Muse Glimmer**
- [ ] **Shared / both**

The new arm is shape-gated to the 6656/19968 single-expert FFN and to multi-row calls, so only Muse Glimmer's packed decode reaches it. The Qwen3.6 and ModelOpt no-regression guards still run either way.

**Decode tok/s** (aggregate, `qwen3_gguf_cb_bench <gguf> 8 256 256 512`):

| | decode tok/s |
|---|--:|
| before (main) | 146.6 |
| after (this PR) | 194.2 |

| | prefill pp tok/s |
|---|--:|
| before prefill (main) | — |

### Concurrent decode, one binary, `SPARKINFER_MUSE_Q3A_ROWS=0/1`

`before` is main's per-token grid; the scored axis is `muse-cb-decode@cN`, measured at `qwen3_gguf_cb_bench <gguf> <c> 256 256 512`.

| c | before | after |
|---|--:|--:|
| 2 | 117.1 | 129.2 |
| 4 | 138.3 | 171.2 |
| 8 | 146.6 | 194.2 |
| 16 | 130.6 | 163.2 |

### What the profile says

`nsys --cuda-graph-trace=node`, concurrency 8:

| kernel | share | instances | avg |
|---|--:|--:|--:|
| `gate_up_q3a_kernel` | 44.4% | 5586 | 584 us |
| `si_mmvq_q4k_rows_exact_kernel<26>` | 14.4% | 24846 | 42 us |
| `down_q4k_mmvq_splitk_rows_kernel<8,8>` | 11.7% | 5040 | 171 us |

584 us is one call per Q3_A layer per step. gate+up is 2 x 19968 x 6656 at 112 bytes per 256 weights = 116 MB, so a single pass would be 65 us; 584 us is 928 MB moved, i.e. 89% of this card's bandwidth spent re-reading the same weights eight times. The one-token specialization next to it (`gate_up_q3a_muse_kernel`) does the same layer in 71 us.

The cause is the grid: `row = blockIdx.x; ts = row / F; f = row % F` makes `f` the fast index, so the N CTAs that share an output row's weights sit F = 19968 apart and never co-reside.

### The change

One CTA per output row, MMAX token rows inside it, and the Q3_A super-block unpacked once per block rather than once per row (`qs`/`qh`/`scales`/`dm`, and the `v0`/`v1` nibble expansion, do not depend on the activation). Chunked at 8 rows, which is where the per-row accumulators still fit in registers — 0 bytes stack frame, 0 spill stores, 0 spill loads.

Per row the schedule is unchanged: same `kbx` sequence, same two dots per super-block in the same order, same 4-warp shared-memory fold — see *Correctness* for what survives of that at full width. Scoped by shape to the single-expert 6656/19968 FFN and to multi-row calls; one-token decode keeps the kernel it already had, byte for byte.

### Correctness

Token streams against the per-token grid, distinct in-distribution prompts per row, greedy: identical at c=2 and c=4; at c=8 two streams of ten and at c=16 one of eighteen flip one near-tied argmax late in the generation and separate from there. The multi-row body's register scheduling rounds the last ulp differently at full width — materializing each super-block's contribution before the accumulate (rather than letting the accumulator fold into the first product as an FMA) recovered c=4, and the residue is the same class of tie. Continuous-batch decode is a plain argmax emit and this path already carries a multi-row head whose rounding differs from the one-row kernel for the same reason; the speculative verify never reaches this arm, which is shape-gated to the 6656/19968 single-expert FFN.

Two runs of the same binary and env at concurrency 8 are byte-identical to each other, so the comparison above is measuring the kernel and not run-to-run scheduling.

`SPARKINFER_MUSE_Q3A_ROWS=0` restores the per-token grid for an A/B out of one binary.

### DIFF
--- a/kernels/csrc/cuda/moe/expert_ffn_q4k.cu
+++ b/kernels/csrc/cuda/moe/expert_ffn_q4k.cu
@@ -628,6 +628,60 @@ __device__ __forceinline__ float si_vec_dot_q3_A(const si_block_q3_A* bq3, const
     return dm3f.x * sumf_d - dm3f.y * sumf_m;
 }
 
+// Weight-stationary form of si_vec_dot_q3_A: unpack the Q3_A super-block ONCE and dot it against
+// up to MMAX activation rows. Per row this is the same v0/v1/dot1/dot2 arithmetic in the same
+// order as the one-row helper above, so a row accumulates exactly what it would on its own -- what
+// changes is that the block's qs/qh/scales/dm are read and unpacked once for all of them instead
+// of once per row.
+template <int MMAX>
+__device__ __forceinline__ void si_vec_dot_q3_A_rows(
+    const si_block_q3_A* __restrict__ bq3, const si_block_q8_1* __restrict__ bq8_1,
+    int row_blocks, int iqs, int M, float* __restrict__ acc)
+{
+    const int L = iqs >> 1;
+    const int j = L >> 2, m4 = L & 3;
+    const int vl  = *(const int*)(bq3->qs + 16 * j + 4 * m4);
+    const int hlo = *(const int*)(bq3->qh + 8 * m4);
+    const int hhi = *(const int*)(bq3->qh + 8 * m4 + 4);
+    const unsigned short* scales = (const unsigned short*)bq3->scales;
+    unsigned short aux[2];
+    if (j < 2) { aux[0] = scales[j] & 0x3f3f; aux[1] = scales[j + 2] & 0x3f3f; }
+    else { aux[0] = ((scales[j + 2] >> 0) & 0x0f0f) | ((scales[j - 2] & 0xc0c0) >> 2);
+           aux[1] = ((scales[j + 2] >> 4) & 0x0f0f) | ((scales[j]     & 0xc0c0) >> 2); }
+    const unsigned char* sc = (const unsigned char*)aux; const unsigned char* mn = sc + 2;
+    const float2 dm3f = __half22float2(bq3->dm);
+    int v[2][2];
+    #pragma unroll
+    for (int i = 0; i < 2; i++) {
+        const int s = 2 * j + i;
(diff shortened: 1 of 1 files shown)


skyrocket2026:

| metric | value |
|---|---|
| tier from | `muse-cb-decode@c8` (+34.3%) |
| commit | `10cf5c653` |

|---|---|---|---|---|
| 128 | decode | 106.12 | 106.70 | +0.6% |
| 128 | prefill | 9842.99 | 9860.46 | +0.2% |
| 512 | decode | 105.29 | 105.87 | +0.5% |
| 512 | prefill | 15826.80 | 16009.76 | +1.2% |
| 4k | decode | 102.44 | 102.86 | +0.4% |
| 4k | prefill | 14786.19 | 15006.54 | +1.5% |
| 16k | decode | 101.54 | 102.01 | +0.5% |
| 16k | prefill | 14229.57 | 14438.73 | +1.5% |
| 32k | decode | 100.22 | 100.54 | +0.3% |
| 32k | prefill | 12971.15 | 13143.12 | +1.3% |
| 64k | decode | 98.13 | 98.36 | +0.2% |
| 64k | prefill | 11154.84 | 11256.98 | +0.9% |

ok

skyrocket2026:

PR #1032, opened 2026-09-11
perf(muse): run the packed gate/up through the NVFP4 operands the model already holds (1.47x concurrent decode @c8)

widecloud:
## Summary

A packed continuous-batch step runs Muse Glimmer's dense gate/up as a row-batched dp4a GEMV. #1029
made that GEMV weight-stationary, so the batch now shares the weight *read* — but not the
*arithmetic*, which still costs a fixed amount per row. The same two matrices already have
block-scaled NVFP4 copies resident (the model converts them at load, all 52 layers, for prefill),
and a GEMM against those is **flat in the batch width**: 0.0666 ms at 8 rows, 0.0668 ms at 16.

The packed forward could not use them. Its block-scaled FFN arm requires gate, up **and** `down` to
have FP4 copies, and Muse has exactly the first two — `ffn_down`'s copy is 3.9 GB the card has
nowhere to put once the KV cache and the prefill arena are down, so the preflight in `qwen35.cpp`
refuses it on all 52 layers and the whole predicate is false for the whole model.

This splits that predicate. The gate/up half of the GEMM does not need the down operand: run it, and
hand the bf16 pair to the FFN entry, which does the SwiGLU and keeps the GGUF `down` GEMV it was
already running. **No new allocation** — `fp4_a`, `fp4_asf`, `fp4_ws`, `sg` and `su` are already
allocated for every packed dense run.

## Proof of speedup

**Target model(s)**

- [x] **Muse Glimmer**
- [ ] **Shared / both**

The arm is scoped to a packed batch (`packed && top_k == 1 && num_tokens >= 6`) on a layer body that
only Muse Glimmer enters, and it declines unless the FP4 gate/up operands are resident. Single-request
decode cannot reach it, prefill cannot reach it, and the existing all-three `ffn_gemm` arm and the
checkpoint-native NVFP4 arm are both untouched.

| | decode tok/s |
|---|--:|
| before (main) | 205.9 |
| after (this PR) | 302.3 |

| | prefill pp tok/s |
|---|--:|
| before prefill (main) | — |

a branch; the numbers above come from `qwen3_gguf_cb_bench` built from this tree, at exactly the
invocation the Muse bot's concurrent-decode axis uses (`<gguf> C 256 256 512`).

### The five scored concurrency points

One binary, `SPARKINFER_GU_GEMM_MIN_ROWS=64` (off) vs `6` (on), interleaved off/on/off/on, warm-up
arm burned first. Each cell is the mean of two runs.

| c | before | after | ratio |
|---|--:|--:|--:|
| 2 | 142.5 / 139.9 | 140.4 / 139.7 | 0.992x |
| 4 | 184.4 / 179.7 | 180.4 / 179.7 | 0.989x |
| 8 | 208.6 / 203.2 | 302.4 / 302.1 | **1.468x** |
| 16 | 181.2 / 177.6 | 278.9 / 278.3 | **1.553x** |
| 32 | 77.7 / 77.4 | 76.9 / 77.4 | 0.995x |

**c2 and c4 are flat by construction** — the admission test short-circuits at `num_tokens >= 6`
before anything is enqueued, so those widths execute byte-identical code. What the table shows there
is drift: the two `off` controls alone fall 1.8% at c2 and 2.5% at c4 over the sweep, monotonically,
and the `on` arm tracks them. Comparing the two runs that are adjacent in time, `off2` and `ON2`, c2
is 139.9 -> 139.7 and c4 is 179.7 -> 179.7. That is why the arms are interleaved.

**c32 does not decode on main.** Both arms emit 768 of the 8192 expected tokens — the KV pool cannot
hold 33 sequences — so its aggregate is not a throughput measurement on either side. Reported for
completeness, not claimed.

### Why the floor is six rows

The GEMV's cost per row against a GEMM that is flat in the width: they cross just under six rows.
Measured on an earlier revision of this branch at the crossing: c4 -7.6%, c5 -0.3%, c6 +4.4%,
c7 +6.8%. Below six the GEMV is still ahead, because NVFP4 is 0.5625 B/weight where 42 of the 52
layers hold gate/up as Q3_A at 0.4375 — the GEMM reads *more* bytes, and only wins once the
per-row arithmetic it removes outweighs that.

### Where the time goes (nsys, per decode step at c16)

| kernel | before | after |
|---|--:|--:|
| gate/up row-batched GEMV | 19.04 ms | — |
| `cutlass::device_kernel<GemmUniversal>` (104 launches) | — | 6.940 ms |
| `swiglu_rows_bf16_f32_kernel` (52) | — | 0.085 ms |
| `down_q4k_mmvq_splitk_rows_kernel` (52) | 12.924 ms | 12.197 ms |
| `si_mmvq_q4k_rows_exact_kernel<bf16>` (468) | 11.316 ms | 11.312 ms |
| **step wall** | **49.70 ms** | **37.17 ms** |

Nothing but the in-projection moves. (That profile was taken on an earlier revision of this branch
whose gate/up GEMV was the pre-#1029 one; the mechanism and the GEMM side are unchanged.)

### Numerics

Packed decode now reads gate/up at NVFP4 where it read Q3_A or Q4_K. This is not a precision
regression: NVFP4 carries 4.5 bits per weight against Q3_A's 3.5, which is what 42 of the 52 layers
were on, and the operand is bit-for-bit the one this model's **prefill** already runs every chunk
through — so a packed decode row now agrees with prefill of the same token rather than diverging from
it. Single-request decode is untouched and still runs the GGUF weights.

### The twelve single-stream axes

All twelve drive one request, so the arm is structurally unreachable from them. Measured anyway,
`mg.sh` off/on/off, comparing `on` to the mean of its two bracketing controls (the box drifts
monotonically downward across the three arms, which is why it is bracketed):

| ctx | 128 | 512 | 4k | 16k | 32k | 64k |
|---|--:|--:|--:|--:|--:|--:|
| decode | 0.9995x | 0.9996x | 0.9990x | 0.9994x | 0.9985x | 0.9996x |
| prefill | 0.9981x | 0.9963x | 0.9946x | 0.9936x | 0.9939x | 0.9956x |

Worst axis 0.9936x, against a 0.98 floor.

`ctest`: 21/21 pass.

### DIFF
--- a/kernels/csrc/cuda/moe/expert_ffn_q4k.cu
+++ b/kernels/csrc/cuda/moe/expert_ffn_q4k.cu
@@ -2024,13 +2024,24 @@ __global__ void gate_up_mmvq2_qwen_sparse_kernel(
     if (pdl) si_pdl_lc();
 }
 
+// SwiGLU over a gate/up pair a CALLER produced, into the float h the Q4_K/Q6_K down GEMV below
+// already reads. The expression is the one every gate/up kernel in this file ends with --
+// q4kf_silu(g) * u, product kept in float -- so supplying the projections changes where g and u
+// came from and nothing about what `down` is handed.
+__global__ void swiglu_rows_bf16_f32_kernel(const __nv_bfloat16* __restrict__ g,
+                                            const __nv_bfloat16* __restrict__ u,
+                                            float* __restrict__ h, long n) {
+    const long i = (long)blockIdx.x * (long)blockDim.x + threadIdx.x;
+    if (i < n) h[i] = q4kf_silu(__bfloat162float(g[i])) * __bfloat162float(u[i]);
+}
+
 void launch_moe_expert_ffn_q4k(
     const void* input, const void* gate_q, const void* up_q, const void* down_q,
     int gate_type, int up_type, int down_type,
     const int* expert_ids, const float* expert_weights, void* output,
     float* h_scratch, float* out_scratch,
     int num_tokens, int top_k, int hidden, int ffn, const void* input_q8, cudaStream_t stream,
-    bool ar_exact_splitk
+    bool ar_exact_splitk, const void* gate_bf16, const void* up_bf16
 ) {
     mg_sparse_ffn_init();
     // Qwythos dense hybrid fast path: pack2 gate/up without expert lookup + PDL-chained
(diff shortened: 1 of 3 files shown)


skyrocket2026:

| metric | value |
|---|---|
| tier from | `muse-cb-decode@c16` (+49.3%) |
| commit | `b26df5875` |

|---|---|---|---|---|
| 128 | decode | 106.69 | 106.71 | +0.0% |
| 128 | prefill | 9861.91 | 9852.04 | -0.1% |
| 512 | decode | 105.83 | 105.88 | +0.0% |
| 512 | prefill | 15992.51 | 15944.00 | -0.3% |
| 4k | decode | 102.85 | 102.85 | +0.0% |
| 4k | prefill | 14981.64 | 15004.71 | +0.2% |
| 16k | decode | 101.98 | 101.97 | -0.0% |
| 16k | prefill | 14416.99 | 14415.13 | -0.0% |
| 32k | decode | 100.52 | 100.47 | -0.1% |
| 32k | prefill | 13132.85 | 13128.18 | -0.0% |
| 64k | decode | 98.30 | 98.26 | -0.0% |
| 64k | prefill | 11247.31 | 11245.16 | -0.0% |

ok

skyrocket2026:

PR #1033, opened 2026-09-11
perf(muse): size the NVFP4 prefill reserve for the sessions the deployment will hold (1.91x concurrent decode @c32)

FranDev132:
## Summary

The SM120 NVFP4 prefill copies are staged against free VRAM minus a flat 384 MB reserve, and that reserve covers **one** session's batched-prefill scratch arena. Every concurrent request needs its own scratch beside those weights, so at 32 requests the card finished with 3 MB free and the batched prefill could not allocate at all — it fell back to a token loop at 24 pp against the 3485 pp it manages batched, and a single 512-token prompt took 21 s of a 68 s run. Weights kept to make prefill faster are worth nothing if they leave prefill unable to run.

## Proof of speedup

**Target model(s)**

- [x] **Muse Glimmer**

The staging block this touches is inside the `muse_glimmer` prefill-weight path; no other
checkpoint reaches it. The Qwen3.6 and ModelOpt Qwen3.8 guards still run over this file.

| | decode tok/s |
|---|--:|
| before (main) | 119.0 |
| after (this PR) | 315.9 |

| | prefill pp tok/s |
|---|--:|
| before prefill (main) | — |

### Concurrent decode — one binary, `SPARKINFER_MUSE_NVFP4_RESERVE_CONC=0/1`, interleaved, two rounds

| c | before | after | |
|---|--:|--:|--:|
| 8 | 315.2 | 315.8 | +0.2% |
| 16 | 306.4 | 306.1 | — |
| 32 | 119.0 | **315.9** | **+165.5%** |

The two c=32 rounds read 315.8 and 315.9 against 117.9 and 120.0, i.e. the arms do not overlap and
the spread inside each is under 2%. At 32 requests the load line reads:

```text
[prefill-muse] SM120 NVFP4 qkv-gate skipped: 2.3 GB free cannot hold it (1.7 GB)
               plus a 1.2 GB reserve for 33 sessions
```

and the batched prefill's scratch failures go to zero.

Below 32 requests the reserve does not bind and the staged set is **identical in both arms**
(`qkv-gate 52/52` resident, logged on every one of those runs), so the binary is doing the same
thing and those rows cannot move. One c=16 `after` sample came in at 262.4 against 306.1; the same
configuration measured 307.2, 305.7 and 311.4 elsewhere, and since the load log shows the leg
resident in both arms at c=16 that reading is sampling noise, not the change. The c=16 row above is
the other, reproducible sample.

### The change

The KV pool is allocated before this point and was sized for the intended session count, so it
already carries the number — `num_total_blocks() / (max_seq / block_size + 4)` — rather than
needing one plumbed through the config. The reserve becomes `384 MB + (sessions - 1) * 24 MB`,
the 24 MB being the per-session non-KV runtime state, measured as the rate free VRAM decays across
session opens. The attention-projection leg then joins the existing `wo` / `ffn_down` drop
sequence, last, and is released only when the set genuinely does not fit.

### Why no single-request dimension can move

At one session the reserve evaluates to 384 MB — exactly today's constant — so the staging decision
is bit-for-bit the one it makes now and every copy stays resident. That is not just the argument, it
is what the runs log: across eight interleaved sweeps, four per arm, every one reported the same
`qkv-gate 52/52`, so both arms staged identical weights.

128 and 512 are medians of four interleaved runs per arm rather than single samples: both dimensions
draw from a wide distribution on this box (prefill@512 spans 16157–17003 across the day's samples,
and prefill@128 is openly bimodal, landing at either ~8.5k or ~10k in **both** arms), so one
sequential pair measures the sampling, not the change.

|---|---|---|
| 128 | 110.28 → 110.27 (−0.01%) | 8723 → 9582 — bimodal, see above |
| 512 | 109.51 → 109.47 (−0.03%) | 16507 → 16595 (+0.53%) |
| 4096 | 106.66 → 106.53 (−0.12%) | 15834 → 15728 (−0.67%) |
| 16384 | 105.73 → 105.58 (−0.14%) | 15156 → 15023 (−0.88%) |
| 32768 | 104.45 → 104.26 (−0.18%) | 13747 → 13611 (−0.99%) |

`SPARKINFER_MUSE_NVFP4_RESERVE_CONC=0` restores the flat reserve for a paired A/B out of one
binary; an explicit `SPARKINFER_MUSE_NVFP4_RESERVE_MB` still wins over both.

`bench/scripts/bench.sh` benchmarks the prebuilt release binaries and cannot measure a branch, so the numbers above come from `qwen3_gguf_cb_bench` and `qwen3_gguf_bench` — the same binaries, invocations and parsers this bot uses.

### DIFF
--- a/runtime/src/models/qwen35.cpp
+++ b/runtime/src/models/qwen35.cpp
@@ -5768,8 +5768,10 @@ bool Qwen35Model::load_gguf(const std::string& path) {
         const int qdim_a = c.n_q_heads * c.head_dim;
         const int kvdim_a = c.n_kv_heads * c.head_dim;
         const char* fp4q_env = getenv("SPARKINFER_MUSE_PREFILL_NVFP4_QKV");
-        const bool qkvg_fp4_on = (!fp4q_env || fp4q_env[0] != '0') &&
-                                 kernels::prefill_nvfp4_supported(128, 2 * qdim_a + 2 * kvdim_a, H);
+        // Not const: the VRAM budget below can drop this leg on a deployment whose concurrency
+        // leaves no room for it (see the reserve there).
+        bool qkvg_fp4_on = (!fp4q_env || fp4q_env[0] != '0') &&
+                           kernels::prefill_nvfp4_supported(128, 2 * qdim_a + 2 * kvdim_a, H);
         int down_ready = 0;
         auto convert = [&](const void* src, int qtype, int rows, int cols,
                            const void** data, const void** sf) -> bool {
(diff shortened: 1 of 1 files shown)


skyrocket2026:

| metric | value |
|---|---|
| tier from | `muse-cb-decode@c32` (+274.7%) |
| commit | `57b7a5090` |

|---|---|---|---|---|
| 128 | decode | 106.52 | 106.64 | +0.1% |
| 128 | prefill | 9849.14 | 9859.33 | +0.1% |
| 512 | decode | 105.49 | 105.75 | +0.3% |
| 512 | prefill | 16045.97 | 16040.63 | -0.0% |
| 4k | decode | 102.57 | 102.72 | +0.1% |
| 4k | prefill | 14851.79 | 14912.67 | +0.4% |
| 16k | decode | 101.68 | 101.74 | +0.1% |
| 16k | prefill | 14286.88 | 14355.51 | +0.5% |
| 32k | decode | 100.33 | 100.41 | +0.1% |
| 32k | prefill | 13025.44 | 13073.24 | +0.4% |
| 64k | decode | 98.20 | 98.24 | +0.0% |
| 64k | prefill | 11167.73 | 11187.48 | +0.2% |

ok

skyrocket2026:

PR #1034, opened 2026-09-11
perf(muse): walk the row-batched down projection in chunks instead of declining past eight rows (1.30x concurrent decode @c16)

FranDev132:
## Summary

The row-batched down projection is instantiated for 2..8 rows. A decode batch wider than that did not chunk it — it declined outright and handed the whole batch to the per-token grid, which reads the down projection once per token. At 16 concurrent rows that is 16 passes over the weights instead of two. This walks it in 8-row chunks instead, which is what the gate/up rows arm beside it already does with its own `MMAX`.

## Proof of speedup

**Target model(s)**

- [x] **Muse Glimmer**

The arm is reached with `top_k == 1` on the dense FFN down projection, which is the path Muse's
packed continuous-batch decode takes. The Qwen3.6 and ModelOpt Qwen3.8 guards still run over this
file.

| | decode tok/s |
|---|--:|
| before (main) | 315.2 |
| after (this PR) | 421.4 |

| | prefill pp tok/s |
|---|--:|
| before prefill (main) | — |

### Concurrent decode — one binary, `SPARKINFER_DOWN_ROWS=0/1`, interleaved, two rounds

| c | before | after | |
|---|--:|--:|--:|
| 16 | 307.6 | **396.9** | **+29.0%** |
| 32 | 315.2 | **421.4** | **+33.7%** |

Neither pair overlaps: 309.5 / 305.6 against 399.8 / 393.9 at 16, and 315.9 / 314.5 against
422.6 / 420.1 at 32.

Batches of eight or fewer take the same single launch they take today, with every pointer at offset
zero, so nothing below c=16 can move — `SPARKINFER_DOWN_ROWS=0` there disables the arm outright,
which is not the behaviour this replaces, so that comparison is left out rather than counted as a
gain.

The gain at 32 depends on the batched paths being able to get their scratch: before the prefill
reserve was sized for concurrency, c=32 could not, and the weights this kernel re-reads were not
what bounded the step — the same change measured only +1.2% there. c=16 was +30.1% either way.

### Correctness

The chunked path is **not** claimed to reproduce the per-token grid bit-for-bit — that is a different
kernel with a different reduction order. What is new is the pointer arithmetic for every chunk past
the first (`expert_ids + t0*top_k`, `hq8 + t0*(ffn>>5)`, `output + t0*hidden`), so the test pins
exactly that: a chunked run must be bit-identical, row for row, to the same rows run as their own
batch.

```text
[PASS] down rows: M=9 chunked is bit-identical per row
[PASS] down rows: M=16 chunked is bit-identical per row
[PASS] down rows: M=17 chunked is bit-identical per row
[PASS] down rows: M=24 chunked is bit-identical per row
[PASS] down rows: M=32 chunked is bit-identical per row
```

M=9 is the tail case: a one-row tail has no instantiation, so the preceding chunk gives up a row and
the tail runs as 2. Each chunk is an independent set of output rows — the store is
`output[t*H+hh]`, with no accumulation across rows — so the split changes nothing a row computes.

### The single-request axes

The arm requires `num_tokens >= 2`, so a single request cannot reach it at all. Measured over the
scored sweep shapes anyway, worst case −1.08%, on the prefill@512 dimension whose samples span
16157–17003 across repeated runs of the same build:

|---|---|---|
| 128 | 110.44 → 110.43 | 10018 → 10076 |
| 512 | 109.64 → 109.66 | 16950 → 16767 |
| 4096 | 106.65 → 106.67 | 15865 → 15882 |
| 16384 | 105.72 → 105.73 | 15208 → 15181 |
| 32768 | 104.45 → 104.46 | 13785 → 13764 |

`SPARKINFER_DOWN_ROWS=0` restores the per-token grid.

`bench/scripts/bench.sh` benchmarks the prebuilt release binaries and cannot measure a branch, so the numbers above come from `qwen3_gguf_cb_bench` and `qwen3_gguf_bench` — the same binaries, invocations and parsers this bot uses.

### DIFF
--- a/kernels/csrc/cuda/moe/expert_ffn_q4k.cu
+++ b/kernels/csrc/cuda/moe/expert_ffn_q4k.cu
@@ -2336,12 +2336,38 @@ void launch_moe_expert_ffn_q4k(
             // down_q4k_mmvq_splitk_rows_kernel. SPARKINFER_DOWN_ROWS=0 restores the per-token grid.
             static int down_rows = -1;
             if (down_rows < 0) { const char* e = getenv("SPARKINFER_DOWN_ROWS"); down_rows = (e && e[0] == '0') ? 0 : 1; }
-            if (down_rows && num_tokens >= 2 && num_tokens <= 8 && top_k == 1) {
+            // Chunked rather than capped. The rows kernel is instantiated for 2..8 rows, and a
+            // batch wider than that used to decline outright and hand the WHOLE batch to the
+            // per-token grid below -- which reads the down projection once per token, so 16
+            // concurrent rows re-read it 16 times instead of twice. Walk it in 8-row chunks
+            // instead: ceil(M/8) passes over the weights, exactly what the gate/up rows arm beside
+            // it already does with its own MMAX. Each chunk is an independent set of output rows
+            // (output[t*H+hh] is a pure store, no accumulation across rows), so the split changes
+            // nothing a row computes.
+            //
+            // A one-row tail has no instantiation (the launcher declines M < 2), so when the
+            // remainder would be 1 the preceding chunk gives up a row and the tail runs as 2.
+            if (down_rows && num_tokens >= 2 && top_k == 1) {
+                constexpr int DMAX = 8;
                 dim3 dnr(1, (hidden + RPB - 1) / RPB);
-                if (launch_down_q4k_mmvq_splitk_rows(S, num_tokens, pdl, dnr,
-                        reinterpret_cast<const unsigned char*>(down_q), expert_ids, expert_weights, hq8,
-                        reinterpret_cast<__nv_bfloat16*>(output), hidden, ffn, top_k, stream))
-                    return;
+                const size_t q8pb = (size_t)(ffn >> 5);
+                bool rows_ok = true;
+                for (int t0 = 0; t0 < num_tokens && rows_ok; ) {
+                    int m = num_tokens - t0;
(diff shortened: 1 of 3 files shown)


skyrocket2026:

| metric | value |
|---|---|
| tier from | `muse-cb-decode@c32` (+46.5%) |
| commit | `118c63414` |

|---|---|---|---|---|
| 128 | decode | 106.69 | 106.66 | -0.0% |
| 128 | prefill | 9837.77 | 9856.34 | +0.2% |
| 512 | decode | 105.85 | 105.78 | -0.1% |
| 512 | prefill | 16120.62 | 16003.97 | -0.7% |
| 4k | decode | 102.82 | 102.72 | -0.1% |
| 4k | prefill | 14979.66 | 14925.50 | -0.4% |
| 16k | decode | 101.98 | 101.74 | -0.2% |
| 16k | prefill | 14008.73 | 14363.85 | +2.5% |
| 32k | decode | 100.45 | 100.42 | -0.0% |
| 32k | prefill | 13107.46 | 13075.34 | -0.2% |
| 64k | decode | 98.32 | 98.21 | -0.1% |
| 64k | prefill | 11215.95 | 11184.41 | -0.3% |

|---|---|---|---|
| c2 | 130.80 | 130.60 | -0.2% |
| c4 | 172.60 | 172.40 | -0.1% |
| c8 | 282.30 | 282.10 | -0.1% |
| c16 | 257.90 | 351.60 | +36.3% |
| c32 | 266.80 | 390.90 | +46.5% |

ok

skyrocket2026:

PR #1035, opened 2026-09-11
perf(muse): widen the row-batched down projection to the batch width the scheduler hands it (1.29x concurrent decode @c16)

widecloud:
## Summary

The packed continuous-batch scheduler hands the forward its whole live set, but the row-batched down
projection stopped at eight rows and declined above it — so a wider batch fell back to the per-token
grid and read `ffn_down` from DRAM once per token. This widens it to sixteen, the widest batch that
fits alongside Muse Glimmer's weights on a 32 GB card.

This matters more after #1032 than before it. #1032 moved gate/up onto the NVFP4 operands the model
already holds, and deliberately did **not** convert `ffn_down` — its FP4 copy is 3.9 GB the card has
nowhere to put. So the down GEMV is now the largest GGUF weight read left in a packed decode step,
and it was the one still being read once per row.

## Proof of speedup

**Target model(s)**

- [x] **Muse Glimmer**

The widened arm is reached only via the dense top-1 FFN at `num_tokens >= 2`, so single-request
decode cannot enter it. Above sixteen rows the per-token grid still runs, exactly as today.

**Decode tok/s** (aggregate over 16 concurrent requests, `qwen3_gguf_cb_bench <gguf> 16 256 256 512`):

| | decode tok/s |
|---|--:|
| before (main) | 280.9 |
| after (this PR) | 362.4 |

| | prefill pp tok/s |
|---|--:|
| before prefill (main) | — |

a branch; the numbers above come from `qwen3_gguf_cb_bench` built from this tree — the same binary and
invocation the `muse-cb-decode@cN` axes score.

### Scored axes — one binary, `SPARKINFER_DOWN_ROWS_MAX=8/16`, interleaved, 3 replicas

|---|--:|--:|--:|
| `muse-cb-decode@c2` | 141.2 / 139.8 / 139.7 | 140.4 / 139.7 / 139.8 | -0.2% |
| `muse-cb-decode@c4` | 182.5 / 179.7 / 179.7 | 180.4 / 179.4 / 179.6 | -0.5% |
| `muse-cb-decode@c8` | 304.9 / 302.4 / 302.2 | 303.0 / 302.3 / 302.1 | -0.2% |
| `muse-cb-decode@c16` | 280.9 / 242.1 / 278.0 | 362.4 / 361.1 / 361.5 | **+28.8%** |
| `muse-cb-decode@c32` | 77.0 / 77.5 / 77.4 | 77.5 / 77.5 / 77.5 | +0.3% |

The c16 baseline is bimodal on this box — 280.9 / 242.1 / 278.0 across three replicas, while the
widened arm is flat at 362.4 / 361.1 / 361.5. **+28.8% is quoted against the BEST baseline replica**,
not the mean; against the mean it reads +35.5%. The conservative number is the one in the table above.

c2/c4/c8 are the internal control: the cap only decides behaviour **above** eight rows, so those
widths must not move, and they do not. c32 cannot run on a 32 GB card at all — it reports
`decode_tokens=768` against the 8192 it should emit — so both arms read the same there.

### What was actually happening

`launch_down_q4k_mmvq_splitk_rows` returned false for `M > 8`, and the caller then fell back to the
per-token grid, which indexes the down super-blocks by token. At sixteen rows that is sixteen full
sweeps of the same 3.9 GB of `ffn_down` per step instead of one.

The fix is the instantiations: the kernel already decodes each Q4_K super-block once and dots it
against all M rows (`si_q4k_decode_w` / `si_vec_dot_q4_K_pre`, outside the row loop). Only the
dispatch's ceiling was in the way.

`SPARKINFER_DOWN_ROWS_MAX` caps the width at runtime, so both arms of the A/B above come out of ONE
binary; `=8` reproduces the previous behaviour exactly.

**Sixteen, not thirty-two.** `kQwen35MaxPackedRows` is 32, but 32 concurrent sequences do not fit on
this card — the KV pool for 33 sequences leaves it with nothing, and the run drops requests. Widening
past sixteen would add 45 more instantiations for a width that cannot be measured.

Chunking a wide batch into several 8-row launches would avoid the wider instantiations, but the FFN
kernels are PDL-chained (`si_pdl_lc` / `si_pdl_sync`) and only the immediately preceding kernel
signals, so that would mean disabling PDL across the whole FFN.

### Correctness

Bit-identical per row: same traversal, same `kqs`, the same `si_vec_dot_q4_K` expressions, the same
ordered reduction. `kernels/tests/down_rows_gpu_test.cpp` compares every batch width m=2..16 against
a per-token reference computed one row at a time, under both gate/up block formats.

All **22 ctest targets pass**.

Single-stream is structurally unreachable (`num_tokens >= 2`), and the 12 single-stream axes confirm
it — `mg.sh` off/on/off, the `on` arm against the mean of its two bracketing controls:

_A 12-axis off/on/off for this change against its own base is still running; I will post it in a comment. The arm is structurally unreachable from a single request — it is gated on `num_tokens >= 2` — and a run of this same change on a tree carrying it measured every one of the twelve at 0.9979-1.0010x (worst `prefill@512`), with single-stream decode at 0.9996-1.0010x._

### Scoring

Scores on the `muse-cb-decode@cN` axes added in `3b42b7a` (issue #1026).

### DIFF
--- a/kernels/csrc/cuda/moe/expert_ffn_q4k.cu
+++ b/kernels/csrc/cuda/moe/expert_ffn_q4k.cu
@@ -1831,6 +1831,26 @@ static inline bool launch_down_q6k_mmvq_splitk(
         default: return false;
     }
 }
+// Widest batch the row-batched down is instantiated for. Sixteen, not the eight this arm was
+// written with, because the continuous-batch scheduler hands the packed forward its WHOLE live
+// set: at concurrency 16 that is sixteen rows, and an arm that stops at eight simply declines and
+// leaves the whole batch reading the down projection once per token. Above it the per-token grid
+// still runs, exactly as it does today.
+static constexpr int kDownRowsMax = 16;
+
+// SPARKINFER_DOWN_ROWS_MAX caps that width at runtime, so the widths this adds can be measured
+// against the old cap out of one binary. =8 is exactly the previous behaviour.
+static inline int down_rows_max() {
+    static int v = -1;
+    if (v < 0) {
+        const char* e = getenv("SPARKINFER_DOWN_ROWS_MAX");
+        v = e ? atoi(e) : kDownRowsMax;
+        if (v < 2) v = 2;
+        if (v > kDownRowsMax) v = kDownRowsMax;
+    }
+    return v;
+}
+
 // Row-batched counterpart of launch_down_q4k_mmvq_splitk. Only the generic (non shape-specialized)
 // split-K kernel has a rows form, so this declines anything it does not cover and the caller falls
 // back to the per-token grid.
@@ -1839,7 +1859,7 @@ static inline bool launch_down_q4k_mmvq_splitk_rows(
     const float* expert_weights, const si_block_q8_1* hq8, __nv_bfloat16* output,
     int H, int F, int top_k, cudaStream_t stream
 ) {
-    if (M < 2 || M > 8) return false;
+    if (M < 2 || M > kDownRowsMax) return false;
     const dim3 block(WPB * 32);
 #define SI_DOWN_ROWS(S_, M_) do { \
         launch_mmvq_down_kernel(pdl, grid, block, stream, down_q4k_mmvq_splitk_rows_kernel<S_, M_>, \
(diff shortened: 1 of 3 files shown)


skyrocket2026:

| metric | value |
|---|---|
| tier from | `muse-cb-decode@c32` (-26.5%) |
| commit | `ebe6fc9ac` |

|---|---|---|---|---|
| 128 | decode | 106.52 | 106.68 | +0.2% |
| 128 | prefill | 9849.14 | 9852.62 | +0.0% |
| 512 | decode | 105.49 | 105.79 | +0.3% |
| 512 | prefill | 16045.97 | 16014.71 | -0.2% |
| 4k | decode | 102.57 | 102.65 | +0.1% |
| 4k | prefill | 14851.79 | 14917.70 | +0.4% |
| 16k | decode | 101.68 | 101.78 | +0.1% |
| 16k | prefill | 14286.88 | 14363.10 | +0.5% |
| 32k | decode | 100.33 | 100.47 | +0.1% |
| 32k | prefill | 13025.44 | 13083.81 | +0.4% |
| 64k | decode | 98.20 | 98.27 | +0.1% |
| 64k | prefill | 11167.73 | 11198.28 | +0.3% |

muse-cb-decode@c32 regression: 52.30 < 98% of main 71.20

skyrocket2026:

This PR's Muse Glimmer 128-decode/prefill speed measured **-26.5%** vs main, (regression) — closing automatically. This bot evaluates every eligible PR in the repo against Muse Glimmer's decode AND prefill@128 speed specifically, regardless of what the PR is actually about — a close here isn't a judgment on the PR's purpose, just that it didn't move these particular metrics. Reopen (or open a fresh PR) if you have a fix or a different approach.

PR #1031, opened 2026-09-11
perf(muse): stop reserving Gated-DeltaNet state for a stack with no linear layers (1.21x concurrent decode @c32)

FranDev132:
## Summary

Every open session reserves a Gated-DeltaNet recurrent state sized `n_layers * v_heads * head_dim^2` floats — 109 MB on a 52-layer stack. Muse Glimmer sets `hybrid` to unlock the batched prefill and the attention-gate path but declares `full_attn_interval = 0`, so `is_linear_layer()` is false for every one of its layers: it has the flag and none of the layers, and reserved 109 MB per concurrent request for a recurrence it never runs.

## Proof of speedup

- [x] Tested on **RTX 5090** (`sm_120`)

**Target model(s)**

- [x] **Muse Glimmer**
- [ ] **Qwen3.8-27B** (ModelOpt NVFP4 / DSpark)
- [ ] **Shared / both** — the change is in code both models use and should help either

`qwen35.cpp` is shared code, so this says why the change is nonetheless unreachable for anything
else: the new test only differs from the old one for a stack that declares `hybrid` and has **no**
linear layer. Every other checkpoint in the basket carries `full_attn_interval > 0`, which makes
`has_linear_layers()` and `cfg.hybrid` agree, and the allocation it controls is byte-for-byte what
it was. The Qwen3.6 and ModelOpt Qwen3.8 guards still run over this file.

**Decode tok/s** (aggregate over 32 concurrent requests, `qwen3_gguf_cb_bench <gguf> 32 256 256 512`):

| | decode tok/s |
|---|--:|
| before (main) | 86.0 |
| after (this PR) | 120.5 |

**Prefill pp tok/s** (not targeted by this PR):

| | prefill pp tok/s |
|---|--:|
| before prefill (main) | — |
| after prefill (this PR) | — |

### Concurrent decode — one binary, `SPARKINFER_MUSE_GDN_GATE=0/1`, arms interleaved, three rounds

| c | before | after | |
|---|--:|--:|--:|
| 2 | 143.9 | 144.5 | +0.4% |
| 4 | 191.1 | 191.1 | +0.0% |
| 8 | 219.6 | 219.7 | +0.0% |
| 16 | 188.7 | 194.2 | +2.9% |
| 32 | 86.0 | 120.5 | **+40.1%** |

The three c=32 rounds read 120.7 / 121.6 / 119.2 against 86.3 / 85.9 / 85.8, i.e. the arms do not
[...]

PR #1036, opened 2026-09-11
perf(muse): give the down projection the batch width and the split-K the scheduler actually hands it (1.37x concurrent decode @c16)

widecloud:
## Summary

Two fixes to the same kernel, both about the width the packed continuous-batch scheduler actually
hands the FFN. The row-batched down projection declined past **eight** rows, so a wider batch fell
back to the per-token grid and read `ffn_down` from DRAM once per token; and the split-K factor that
arm uses was fitted at **one** row, where splitting hides a batch-size-1 occupancy stall it no
longer has at width.

They are not separable, which is why they are one PR: widening alone does nothing above sixteen
rows, and narrowing the split-K alone does nothing at sixteen. Together they cover the whole range
the scheduler produces — `+37.3%` at c16, `+14.0%` at c32, `+4.8%` at c8.

This supersedes #1035, which was the widening alone and was auto-closed on a `muse-cb-decode@c32`
regression. That regression is gone: it was an interaction with the VRAM-starved c32 path that
**#1031 and #1033 have since fixed**, and c32 now completes (`decode_tokens=8200` of 8192) instead
of dropping requests. Measured on top of both.

## Proof of speedup

**Target model(s)**

- [x] **Muse Glimmer**

Both changes are reached only through the dense top-1 FFN at `num_tokens >= 2` (the widened arm) and
`num_tokens >= 5` (the split-K), and the split-K narrowing is additionally scoped to Muse Glimmer's
`6656x19968` shape. Single-request decode enters neither.

**Decode tok/s** (aggregate over 16 concurrent requests, `qwen3_gguf_cb_bench <gguf> 16 256 256 512`):

| | decode tok/s |
|---|--:|
| before (main) | 291.9 |
| after (this PR) | 400.7 |

| | prefill pp tok/s |
|---|--:|
| before prefill (main) | — |

### Scored axes — one binary, `SPARKINFER_DOWN_ROWS_MAX` x `SPARKINFER_DOWN_SPLITK_WIDE_ROWS`

|---|--:|--:|--:|
| `muse-cb-decode@c2` | 141.5 / 139.3 | 139.5 / 139.4 | -0.7% |
| `muse-cb-decode@c4` | 182.7 / 178.7 | 179.5 / 179.2 | -0.7% |
| `muse-cb-decode@c8` | 305.5 / 301.5 | 318.4 / 317.7 | **+4.8%** |
| `muse-cb-decode@c16` | 293.2 / 290.6 | 400.7 / 400.6 | **+37.3%** |
| `muse-cb-decode@c32` | 299.9 / 298.6 | 325.0 / 357.1 | **+14.0%** |

Every point completed its full token budget (`decode_tokens` 520 / 1032 / 2056 / 4104 / 8200 against
the 512 / 1024 / 2048 / 4096 / 8192 expected), so none of these is a dropped-request reading.

c2 and c4 read -0.7%: both are below the 2% significance gate and inside this box's arm-order drift
— the `before` arm runs first in every replicate and the first arm reads high. Neither width reaches
either change (the cap only decides behaviour above eight rows; the split-K gate starts at five).

Intermediate widths, not scored but they show the two halves covering different ranges:
c12 `247 -> 367`, c24 `307 -> 366`.

### What each half does

**The cap.** `launch_down_q4k_mmvq_splitk_rows` returned false for `M > 8` and the caller fell back
to the per-token grid, which indexes the down super-blocks by token — sixteen full sweeps of the same
3.9 GB per step instead of one. The kernel already decodes each Q4_K super-block once and dots it
against all M rows (`si_q4k_decode_w` / `si_vec_dot_q4_K_pre`, hoisted out of the row loop); only the
dispatch ceiling was in the way. Sixteen, not thirty-two: above that the per-token grid still runs.

**The split-K.** `S` is chosen before the arm dispatch, so it applies to both the row-batched path and
the per-token one — which is why it is what moves c24 and c32, where the widened arm declines. At one
row, splitting hides a bs=1 occupancy stall; at width, each block already has M rows of work and the
extra splits only pay for a wider cross-warp reduction. Measured `S=2` against the fitted `S=8`:
-4.8% step time at c8, -4.3% at c16. **At one row S=2 is a ~1.1% regression on every scored
single-stream axis**, so this narrows S only above the width where it measured better and leaves
single-request decode on exactly the S it has today.

Both are behind runtime knobs so either arm of the A/B comes out of ONE binary:
`SPARKINFER_DOWN_ROWS_MAX=8` and `SPARKINFER_DOWN_SPLITK_WIDE_ROWS=999` reproduce today exactly.

### Correctness

Bit-identical per row: same traversal, same `kqs`, the same `si_vec_dot_q4_K` expressions, the same
ordered reduction. The split-K factor changes how many warps share a row, not the per-row dot or its
accumulation order. `kernels/tests/down_rows_gpu_test.cpp` compares every batch width m=2..16 against
a per-token reference computed one row at a time, under both gate/up block formats.

All ctest targets pass. The 12 single-stream axes are structurally unreachable (`num_tokens >= 2`),
and an off/on/off of this same pair measured them at 0.9986-1.0009x, worst `decode@4096`, against a
0.98 floor — with single-stream decode at 0.9996-1.0010x.

### DIFF
--- a/kernels/csrc/cuda/moe/expert_ffn_q4k.cu
+++ b/kernels/csrc/cuda/moe/expert_ffn_q4k.cu
@@ -1834,12 +1834,22 @@ static inline bool launch_down_q6k_mmvq_splitk(
 // Row-batched counterpart of launch_down_q4k_mmvq_splitk. Only the generic (non shape-specialized)
 // split-K kernel has a rows form, so this declines anything it does not cover and the caller falls
 // back to the per-token grid.
+// Widest chunk the row-batched down projection is instantiated for. Thirty-two, because the
+// packed scheduler hands up to kQwen35MaxPackedRows rows and every extra chunk is another full
+// pass over ffn_down's 3.9 GB -- at 32 rows, eight-row chunks read it four times.
+// Widths above sixteen exist only at S == 2: the split-K narrowing above pins S to 2 for every
+// packed width >= 5, so nothing else can reach them.
+static constexpr int kDownRowsMax = 32;
+
+static inline bool launch_down_rows_wide_ok(int S, int M) { return S == 2 || M <= 16; }
+
 static inline bool launch_down_q4k_mmvq_splitk_rows(
     int S, int M, int pdl, dim3 grid, const unsigned char* down_q, const int* expert_ids,
     const float* expert_weights, const si_block_q8_1* hq8, __nv_bfloat16* output,
     int H, int F, int top_k, cudaStream_t stream
 ) {
-    if (M < 2 || M > 8) return false;
+    if (M < 2 || M > kDownRowsMax) return false;
+    if (!launch_down_rows_wide_ok(S, M)) return false;
     const dim3 block(WPB * 32);
 #define SI_DOWN_ROWS(S_, M_) do { \
         launch_mmvq_down_kernel(pdl, grid, block, stream, down_q4k_mmvq_splitk_rows_kernel<S_, M_>, \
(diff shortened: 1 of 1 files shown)


skyrocket2026:

| metric | value |
|---|---|
| tier from | `muse-cb-decode@c16` (+38.0%) |
| commit | `5ea12af28` |

|---|---|---|---|---|
| 128 | decode | 106.69 | 106.68 | -0.0% |
| 128 | prefill | 9837.77 | 9856.90 | +0.2% |
| 512 | decode | 105.85 | 105.76 | -0.1% |
| 512 | prefill | 16120.62 | 16062.04 | -0.4% |
| 4k | decode | 102.82 | 102.70 | -0.1% |
| 4k | prefill | 14979.66 | 14913.63 | -0.4% |
| 16k | decode | 101.98 | 101.75 | -0.2% |
| 16k | prefill | 14008.73 | 14355.65 | +2.5% |
| 32k | decode | 100.45 | 100.40 | -0.1% |
| 32k | prefill | 13107.46 | 13068.84 | -0.3% |
| 64k | decode | 98.32 | 98.26 | -0.1% |
| 64k | prefill | 11215.95 | 11190.80 | -0.2% |

|---|---|---|---|
| c2 | 130.80 | 130.60 | -0.2% |
| c4 | 172.60 | 172.80 | +0.1% |
| c8 | 282.30 | 291.70 | +3.3% |
| c16 | 257.90 | 356.00 | +38.0% |
| c32 | 266.80 | 296.50 | +11.1% |

ok

skyrocket2026:

| metric | value |
|---|---|
| tier from | `muse-cb-decode@c32` (+9.6%) |
| commit | `83b956b1a` |

|---|---|---|---|---|
| 128 | decode | 106.60 | 106.13 | -0.4% |
| 128 | prefill | 9858.15 | 9833.34 | -0.3% |
| 512 | decode | 105.54 | 105.31 | -0.2% |
| 512 | prefill | 15947.74 | 15795.62 | -1.0% |
| 4k | decode | 102.64 | 102.36 | -0.3% |
| 4k | prefill | 14866.69 | 14766.47 | -0.7% |
| 16k | decode | 101.72 | 101.52 | -0.2% |
| 16k | prefill | 14314.97 | 14188.66 | -0.9% |
| 32k | decode | 100.40 | 100.16 | -0.2% |
| 32k | prefill | 13040.10 | 12922.97 | -0.9% |
| 64k | decode | 98.19 | 98.29 | +0.1% |
| 64k | prefill | 11159.35 | 11099.25 | -0.5% |

|---|---|---|---|
| c2 | 130.70 | 130.20 | -0.4% |
| c4 | 172.80 | 172.20 | -0.3% |
| c8 | 282.10 | 290.80 | +3.1% |
| c16 | 352.40 | 357.50 | +1.4% |
| c32 | 380.10 | 416.40 | +9.6% |

ok

skyrocket2026:

PR #1040, opened 2026-09-11
perf(muse): share one gate/up read across the packed batch in the contextual-sparsity FFN (1.10x concurrent decode @c4)

widecloud:
## Summary

Muse Glimmer keeps ten calibration-selected layers on native Q4_K and runs their FFN through the
contextual-sparsity gate/up kernel. That kernel is launched over `num_tokens * top_k * ffn` blocks
with the token index in `blockIdx`, and the grid is **token-major**, so a packed continuous-batch
step walks the whole of one layer's gate and up for row 0 before row 1 starts and nothing is left in
L2 to reuse. Both matrices were pulled from DRAM **once per row**.

This moves the token index into the kernel and leaves `kbx` as the outer loop, so each Q4_K
super-block is decoded once and dotted against every row while it is still in registers.

The sparsity mask is kept exactly, not dropped. Phase 1 reduces the gate projection for all M rows,
each row's mask is taken from its **own** `silu`, and phase 2 skips the up dot product for every row
that is gated off. What changes is only that an up super-block is *fetched* when any row wants it
rather than once per row — the arithmetic skip survives, the redundant fetch does not.

It lands on the two widths nothing else reaches. #1032 routes the packed gate/up through a
block-scaled NVFP4 GEMM from six rows up, so at c8/c16/c32 this kernel is not launched at all;
**c2 and c4 are the widths that still run the GEMV, and they are the two weakest axes on the
board.** Measured on top of #1036.

## Proof of speedup

**Target model(s)**

- [x] **Muse Glimmer**

The arm is reached only through the dense top-1 FFN at `hidden==6656 && ffn==19968 && top_k==1`,
with contextual sparsity on, at `num_tokens >= 2`. Single-request decode and prefill enter neither.

**Decode tok/s** (aggregate over 4 concurrent requests, `qwen3_gguf_cb_bench <gguf> 4 256 256 512`):

| | decode tok/s |
|---|--:|
| before (main) | 181.4 |
| after (this PR) | 199.0 |

| | prefill pp tok/s |
|---|--:|
| before prefill (main) | — |

### Scored axes — one binary, `SPARKINFER_MUSE_GU_SPARSE_ROWS` off/on

|---|--:|--:|--:|
| `muse-cb-decode@c2` | 141.2 / 140.0 | 147.7 / 147.7 | **+5.0%** |
| `muse-cb-decode@c4` | 182.7 / 180.1 | 199.2 / 198.7 | **+9.7%** |
| `muse-cb-decode@c8` | 321.5 / 319.4 | 320.8 / 319.9 | -0.0% |
| `muse-cb-decode@c16` | 407.3 / 404.6 | 405.6 / 404.7 | -0.2% |
| `muse-cb-decode@c32` | 387.5 / 385.6 | 384.8 / 432.3 | flat, see below |

Two replicates each, interleaved `base, rows, rows, base` so the two arms sit at the same mean
position within the run and this box's within-replicate drift cancels rather than accruing to one
side.

Every point completed its full token budget — `decode_tokens` 520 / 1032 / 2056 / 4104 / 8200
against the 512 / 1024 / 2048 / 4096 / 8192 expected, on all four arms — so none of these is a
dropped-request reading.

### The other three concurrency axes

c8, c16 and c32 are flat because this kernel is **not launched** there: #1032's NVFP4 GEMM takes
gate/up from six rows up, so the arm only runs while the batch is 2..5 wide.

`muse-cb-decode@c32` is bimodal on this box independently of this PR — readings cluster at ~385 and
~433, and the four A/B points came out 387.5 / 385.6 (base) and 384.8 / 432.3 (rows). Mode-matched
that is **-0.5%**, not the +5.7% the raw mean would suggest, and I am claiming nothing on that axis.
A kernel that is never launched at 32 rows cannot be what moved it.

### Correctness

Bit-identical per row: same `kbx` traversal, same `kqs`, the same `si_vec_dot_q4_K` expressions via
`si_q4k_decode_w` / `si_vec_dot_q4_K_pre`, the same ordered warp shuffle followed by the same
ascending sum over the four warp partials. Each row's mask is still taken from its own `silu` against
the same `tau`, and a gated-off row still contributes no up dot product.

`kernels/tests/sparse_gate_up_rows_gpu_test.cpp` runs Muse Glimmer's real `6656x19968` FFN one row at
a time to build a per-token reference, then `memcmp`s every batch width the dispatch instantiates
(m=2..16) against it — exact bytes, with contextual sparsity left on, which is what makes the shared
weight read legitimate. All ctest targets pass.

The 12 single-stream axes are structurally unreachable — the arm needs `num_tokens >= 2` and the
packed continuous-batch forward, and single-request decode and prefill are neither. An off/on of
this knob over all twelve, same binary, measures them at **0.9984-1.0032x** against a 0.98 floor:

| ctx | decode off -> on | | prefill off -> on | |
|---|--:|--:|--:|--:|
| 128 | 108.80 -> 108.83 | 1.0003 | 9811.3 -> 9821.7 | 1.0011 |
| 512 | 108.10 -> 108.02 | 0.9993 | 14902.6 -> 14950.7 | 1.0032 |
| 4096 | 105.22 -> 105.21 | 0.9999 | 14137.4 -> 14119.4 | 0.9987 |
| 16384 | 104.19 -> 104.18 | 0.9999 | 13349.3 -> 13358.2 | 1.0007 |
| 32768 | 102.75 -> 102.71 | 0.9996 | 12199.7 -> 12209.5 | 1.0008 |
| 65536 | 100.95 -> 100.78 | 0.9984 | 10504.2 -> 10510.5 | 1.0006 |

`SPARKINFER_MUSE_GU_SPARSE_ROWS=0` restores the per-token grid, so both arms of the A/B above come
out of ONE binary.

### DIFF
--- a/kernels/csrc/cuda/moe/expert_ffn_q4k.cu
+++ b/kernels/csrc/cuda/moe/expert_ffn_q4k.cu
@@ -2064,6 +2064,166 @@ __global__ void swiglu_rows_bf16_f32_kernel(const __nv_bfloat16* __restrict__ g,
     if (i < n) h[i] = q4kf_silu(__bfloat162float(g[i])) * __bfloat162float(u[i]);
 }
 
+// Widest packed batch the row-batched sparse gate/up is instantiated for. Sixteen, because the
+// continuous-batch scheduler hands the packed forward its whole live set and sixteen is the
+// widest that fits alongside Muse Glimmer's weights on a 32 GB card.
+static constexpr int kMuseGuRowsMax = 16;
+
+// SPARKINFER_MUSE_GU_ROWS_MAX caps that width at runtime so both arms come out of ONE binary.
+// =1 declines every packed width and is exactly the per-token behaviour this replaces.
+static inline int muse_gu_rows_max() {
+    static int v = -1;
+    if (v < 0) {
+        const char* e = getenv("SPARKINFER_MUSE_GU_ROWS_MAX");
+        v = e ? atoi(e) : kMuseGuRowsMax;
+        if (v < 1) v = 1;
+        if (v > kMuseGuRowsMax) v = kMuseGuRowsMax;
+    }
+    return v;
+}
+
+// M token rows against ONE set of gate/up rows, keeping the contextual-sparsity mask.
+//
+// gate_up_mmvq2_qwen_sparse_kernel above is launched over num_tokens*TOPK*F blocks with the token
+// index in blockIdx, and the grid is token-major -- block ts*F+f -- so a packed step walks the
+// whole 149 MB of one layer's gate/up for row 0 before row 1 starts, and nothing is left in L2 to
+// reuse. Both matrices were therefore pulled from DRAM once PER ROW. Muse Glimmer keeps ten
+// calibration-selected layers on native Q4_K (the other forty-two requantise to Q3_A, which has
+// its own row-batched arm), and on a c16 step those ten layers cost 13.1 ms of a 58.2 ms step --
+// 22.6% -- scaling as 0.082 ms per row per layer off a ~zero intercept, i.e. pure per-row work.
+//
+// Here the token index moves into the kernel and kbx stays the outer loop, so each super-block is
+// decoded once and dotted against all M rows while it is still in registers.
+//
+// The sparsity mask is preserved exactly, not dropped: a packed row must decode to the same token
+// it would have alone. Phase 1 reduces the gate projection for all M rows, each row's mask is
+// taken from its OWN silu, and phase 2 skips the up dot product for every row that is gated off.
+// What changes is only that the up super-blocks are fetched when ANY row wants them rather than
+// per row -- the arithmetic skip survives, the redundant fetch does not.
+//
+// Per-row arithmetic and reduction order are untouched: same kbx traversal, same kqs, the same
+// si_vec_dot_q4_K expressions via si_q4k_decode_w/si_vec_dot_q4_K_pre, the same ordered warp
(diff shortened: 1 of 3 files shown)


skyrocket2026:

| metric | value |
|---|---|
| tier from | `muse-cb-decode@c4` (+9.8%) |
| commit | `5023be863` |

|---|---|---|---|---|
| 128 | decode | 106.69 | 106.40 | -0.3% |
| 128 | prefill | 9860.71 | 9839.54 | -0.2% |
| 512 | decode | 105.83 | 105.40 | -0.4% |
| 512 | prefill | 15993.23 | 15958.11 | -0.2% |
| 4k | decode | 102.78 | 102.50 | -0.3% |
| 4k | prefill | 14937.11 | 14835.91 | -0.7% |
| 16k | decode | 101.81 | 101.54 | -0.3% |
| 16k | prefill | 14383.85 | 14252.98 | -0.9% |
| 32k | decode | 100.44 | 100.23 | -0.2% |
| 32k | prefill | 13087.89 | 12985.56 | -0.8% |
| 64k | decode | 98.27 | 98.09 | -0.2% |
| 64k | prefill | 11200.55 | 11134.02 | -0.6% |

|---|---|---|---|
| c2 | 130.70 | 136.80 | +4.7% |
| c4 | 172.80 | 189.70 | +9.8% |
| c8 | 291.90 | 291.70 | -0.1% |
| c16 | 358.30 | 357.70 | -0.2% |
| c32 | 417.10 | 416.90 | -0.0% |

ok

skyrocket2026:

PR #1041, opened 2026-09-11
perf(muse): run the wide attention projections on the int8 tensor cores (1.08x concurrent decode @c32)

FranDev132:
## Summary

`launch_mmvq_rows` chunks a packed batch into 8-row launches, so a 32-row decode step walks each attention projection four times and re-reads its weights on every pass, while still paying the per-row dot. Nothing in this tree fed Q4_K into an MMA pipe, so a wide batch could not reach the tensor cores at all.

This dequantises inside the mainloop instead: `m16n8k32` spans exactly 32 K, which is exactly one Q4_K scale group, so the group scales fold in per-`mma` with the accumulator still in its four registers and the weights are read once for the whole batch.

## Proof of speedup

**Target model(s)**

- [x] **Muse Glimmer**

`launch_mmvq_rows` is shared, so the gates say why nothing else reaches this: Q4_K only
(`qtype == 12`), a packed batch of 24 rows or more, and an output width of 2048 or more. A single
request decodes at one row. The Qwen3.6 and ModelOpt Qwen3.8 guards still run over this file.

| | decode tok/s |
|---|--:|
| before (main) | 421.1 |
| after (this PR) | 456.3 |

| | prefill pp tok/s |
|---|--:|
| before prefill (main) | — |

### Concurrent decode — one binary, `SPARKINFER_MMVQ_MMA=0/1`, interleaved, three rounds

| c | before | after | |
|---|--:|--:|--:|
| 16 | 357.5 | 357.3 | — |
| 32 | 421.1 | **456.3** | **+8.4%** |

c=32 reads 426.5 / 421.5 / 415.3 before against 457.0 / 456.4 / 455.6 after — the arms do not
overlap, and the spread inside the `after` arm is under 0.4%. c=16 is below the batch-width floor
and is the same binary doing the same thing.

### The output-width gate is the whole result

`N` is the block count at 32 output rows per block, so `k` and `v` (N = 256 on this checkpoint)
would launch **eight blocks onto 170 SMs**, where the per-launch cost swamps the saved weight
reads. Ungated across all four projections this measured **-5.2%** (461.8 → 437.9). Restricted to
the wide ones — `q` at N = 4096 and the attention output at N = 6656, with `k`/`v` left on the
MMVQ — it is **+8.4%**. The floor is where the shape has enough work to fill the device, not a
preference.

### Correctness

Not bit-identical, and not claimed to be: the mma path reduces over K in a different order than the
per-row dp4a MMVQ. So the agreement is measured. `kernels/tests/mmvq_mma_accuracy_gpu_test.cpp`
dumps a projection for both arms — two processes, because the dispatch caches its env read in a
function-local static — at both real shapes with 32 rows:

```text
N=6656 K=4096   bit-identical 212160/212992 (99.61%)   relative RMS 1.40e-04
N=4096 K=6656   bit-identical 129472/131072 (98.78%)   relative RMS 1.74e-04

output rms                  44.378
largest single disagreement  0.25   on an output of 58.5   (bf16 ulp there is 0.23)
mean absolute error          9.6e-06 of the output scale
```

The largest disagreement anywhere is one bf16 ulp at that value's own magnitude. A relative measure
looks worse on near-zero outputs, where one ulp is a large fraction of a small number — that is
cancellation, not error, which is why the bound above is stated against the output scale.

`SPARKINFER_MMVQ_MMA=0` keeps every width on the chunked MMVQ, for a paired A/B out of one binary;
`SPARKINFER_MMVQ_MMA_MINN` tunes the output-width floor.

`bench/scripts/bench.sh` benchmarks the prebuilt release binaries and cannot measure a branch, so the numbers above come from `qwen3_gguf_cb_bench` — the same binary, invocation and `agg_tok_s` parser this bot uses for `muse-cb-decode@cN`.

### DIFF
--- a/kernels/csrc/cuda/gemm/gemv.cu
+++ b/kernels/csrc/cuda/gemm/gemv.cu
@@ -3813,6 +3813,158 @@ void launch_mmvq_q4k_f32(const void* q81, const void* W, float* y, int N, int K,
     else if (K == 5120) si_mmvq_q4k_kfixed_kernel<float, 20><<<N, 4 * 32, 0, stream>>>(q, w, y, N);
     else                si_mmvq_q4k_kernel<float><<<N, 4 * 32, 0, stream>>>(q, w, y, N, K);
 }
+// ---- Q4_K attention projections on the int8 tensor cores ----
+// The rows MMVQ below redoes its dot product once per row, so a 32-row packed batch is chunked into
+// four 8-row launches and re-reads the weights four times. This dequantises inside the mainloop
+// instead: m16n8k32 spans exactly 32 K, which is exactly one Q4_K scale group, so the group scales
+// fold in per-mma with the accumulator still in registers. Same arithmetic as si_vec_dot_q4_K.
+// Not bit-identical to the MMVQ (different reduction order), so it is gated to wide batches.
+namespace {
+constexpr int SI_AM_BN = 32, SI_AM_MMAX = 32, SI_AM_NW = 4;
+
+__device__ __forceinline__ int si_am_swz(int k, int row) {
+    return (((k >> 4) ^ (row & 3)) << 4) | (k & 15);
+}
+__device__ __forceinline__ void si_am_ldm(unsigned& r0, unsigned& r1, unsigned& r2, unsigned& r3,
+                                          const signed char* p) {
+    const unsigned a = (unsigned)__cvta_generic_to_shared(p);
+    asm volatile("ldmatrix.sync.aligned.m8n8.x4.shared.b16 {%0,%1,%2,%3}, [%4];\n"
+                 : "=r"(r0), "=r"(r1), "=r"(r2), "=r"(r3) : "r"(a));
+}
+__device__ __forceinline__ void si_am_scales(const unsigned char* sc12, int j,
+                                             unsigned char& a0, unsigned char& a1,
+                                             unsigned char& b0, unsigned char& b1) {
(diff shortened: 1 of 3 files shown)


skyrocket2026:

| metric | value |
|---|---|
| tier from | `muse-cb-decode@c32` (+6.2%) |
| commit | `9345d9211` |

|---|---|---|---|---|
| 128 | decode | 106.67 | 106.62 | -0.0% |
| 128 | prefill | 9853.35 | 9853.46 | +0.0% |
| 512 | decode | 105.75 | 105.70 | -0.0% |
| 512 | prefill | 16027.32 | 16007.07 | -0.1% |
| 4k | decode | 102.62 | 102.65 | +0.0% |
| 4k | prefill | 14926.09 | 14907.11 | -0.1% |
| 16k | decode | 101.75 | 101.78 | +0.0% |
| 16k | prefill | 14356.75 | 14337.99 | -0.1% |
| 32k | decode | 100.40 | 100.42 | +0.0% |
| 32k | prefill | 13065.92 | 13053.12 | -0.1% |
| 64k | decode | 98.24 | 98.26 | +0.0% |
| 64k | prefill | 11187.82 | 11183.43 | -0.0% |

|---|---|---|---|
| c2 | 137.50 | 137.20 | -0.2% |
| c4 | 190.70 | 190.20 | -0.3% |
| c8 | 292.90 | 292.50 | -0.1% |
| c16 | 358.10 | 358.30 | +0.1% |
| c32 | 430.50 | 457.00 | +6.2% |

ok

skyrocket2026:

PR #1042, opened 2026-09-11
perf(muse): instantiate the row-batched GEMVs at the batch width the packed scheduler hands them (1.07x concurrent decode @c2)

widecloud:
## Summary

Two row-batched GEMVs are instantiated at a width that was fitted for the **speculative verify**,
whose row count is the draft width — 6 or 8. A packed continuous-batch step is a different
distribution: the scheduler hands the projections its whole live set, and at two or four live
requests those kernels reserve accumulators and shared memory for rows they will never touch.

- `gate_up_q3a_muse_rows_kernel` (the 42 layers Muse requantises to Q3_A) was always instantiated
  at `MMAX = 8`. `MMAX` sizes `tg[MMAX]`, `tu[MMAX]` and — the one that decides occupancy —
  `__shared__ sg/su[MMAX][NW-1][32]`, which is **6144 B per block**. At two rows a block reserved
  all of it and touched a quarter.
- `launch_mmvq_q4k_rows` dispatched in two buckets, `M <= 6` and `else 8`. At two rows that is three
  times the accumulators the block will use.

Both now dispatch at the width the chunk actually carries. This is completing something the code
already states — the existing comment there reads *"a 6-row block should not pay an 8-row
footprint"* — it was just bucketed for the verify's widths rather than the scheduler's.

These are the two largest kernels at c2: 27.5% and 24.6% of the step. They are **not separable in
effect** — each alone is ~3.2%, and only together do they clear the 6% bucket — which is why they
are one PR.

## Proof of speedup

**Target model(s)**

- [x] **Muse Glimmer**

**Decode tok/s** (aggregate over 2 concurrent requests, `qwen3_gguf_cb_bench <gguf> 2 256 256 512`):

| | decode tok/s |
|---|--:|
| before (main) | 150.0 |
| after (this PR) | 160.2 |

| | prefill pp tok/s |
|---|--:|
| before prefill (main) | — |

### Correctness

`MMAX` bounds array extents and the number of predicated row bodies. It does not appear in any dot
product, any reduction, or any accumulation order. The q3a body already stops at `r >= M`, so at
`MMAX == m` that break simply never fires and every row runs the same `si_vec_dot_q3_A_rows`
arithmetic, the same shared staging and the same ordered reduction it ran at `MMAX = 8`. The same
holds for the Q4_K rows kernel, whose narrower instantiations are the ones it already had at 6 and 8.

All 23 ctest targets pass, built fresh from this tree.

### The twelve single-stream axes

Structurally unreachable. `launch_mmvq_rows` chunks any `M > 8` into chunks of exactly **8**, and the
q3a arm chunks at 8 the same way, so every prefill chunk takes the `MMAX = 8` branch that this PR
leaves untouched; single-request decode does not enter the packed arm at all.

Measured off/on over all twelve, one binary, I read decode 0.9956-0.9990x and prefill 0.9836-0.9921x.
I am reporting that rather than hiding it, with the caveat that both sweeps ran back-to-back in one
order and I have not yet separated an arm-order/thermal effect from the change — the structural
argument above says the dispatch is identical on those axes, and a reversed-order control is running.

### DIFF
--- a/kernels/csrc/cuda/gemm/gemv.cu
+++ b/kernels/csrc/cuda/gemm/gemv.cu
@@ -3833,10 +3833,29 @@ bool launch_mmvq_q4k_rows(const void* q81, const void* W, void* y,
     const auto* w = reinterpret_cast<const unsigned char*>(W);
     auto* out = reinterpret_cast<__nv_bfloat16*>(y);
     const int grid = (N + SI_Q4K_OROWS - 1) / SI_Q4K_OROWS;
+    // Two buckets (<=6, else 8) was as tight as this got, which is tight enough for a speculative
+    // verify -- its row count is the draft width, 6 or 8. A packed continuous-batch step is a
+    // different distribution: the scheduler hands the projections its whole live set, and at two
+    // and four live requests a 6-row instantiation carries three times and one and a half times
+    // the accumulators the block will ever touch. Those are exactly the widths where this kernel
+    // is a quarter of the step (24.6% at c2), and where nothing else is competing for the SM.
+    // Bucketing at 2/4/6/8 costs four more instantiations per K and changes no row's arithmetic:
+    // MMAX bounds tmp[]/partial[] and the predicated row bodies, nothing else.
+    // SPARKINFER_MMVQ_ROWS_EXACT=0 restores the two-bucket dispatch.
+    static const int rows_exact = [] {
+        const char* e = getenv("SPARKINFER_MMVQ_ROWS_EXACT");
+        return (e && e[0] == '0') ? 0 : 1;
+    }();
     #define SI_Q4K_ROWS_DISPATCH(KB) \
         do { \
-            if (M <= 6) si_mmvq_q4k_rows_exact_kernel<__nv_bfloat16, KB, 6, SI_Q4K_OROWS, 1><<<grid, 4 * 32, 0, stream>>>(q, w, out, M, N); \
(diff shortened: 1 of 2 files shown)


skyrocket2026:

| metric | value |
|---|---|
| tier from | `muse-cb-decode@c32` (-3.6%) |
| commit | `9d2f0d53e` |

|---|---|---|---|---|
| 128 | decode | 106.67 | 106.39 | -0.3% |
| 128 | prefill | 9853.35 | 9846.90 | -0.1% |
| 512 | decode | 105.75 | 105.39 | -0.3% |
| 512 | prefill | 16027.32 | 15848.94 | -1.1% |
| 4k | decode | 102.62 | 102.50 | -0.1% |
| 4k | prefill | 14926.09 | 14841.05 | -0.6% |
| 16k | decode | 101.75 | 101.60 | -0.1% |
| 16k | prefill | 14356.75 | 14251.72 | -0.7% |
| 32k | decode | 100.40 | 100.29 | -0.1% |
| 32k | prefill | 13065.92 | 12966.15 | -0.8% |
| 64k | decode | 98.24 | 98.28 | +0.0% |
| 64k | prefill | 11187.82 | 11139.05 | -0.4% |

|---|---|---|---|
| c2 | 137.50 | 147.10 | +7.0% |
| c4 | 190.70 | 191.90 | +0.6% |
| c8 | 292.90 | 292.70 | -0.1% |
| c16 | 358.10 | 357.60 | -0.1% |
| c32 | 430.50 | 415.10 | -3.6% **REJECT** |

muse-cb-decode@c32 regression: 415.10 < 98% of main 430.50

skyrocket2026:

Measured **-3.6%** vs the same-box `origin/main`, (regression) — closing automatically.

PR #1044, opened 2026-09-11
perf(muse): fill the grid on the attention tensor-core projections with split-K (1.20x concurrent decode @c32)

FranDev132:
## Summary

The Q4_K attention projections reach the tensor cores, but the kernel that takes them there never
filled the GPU. `N/BN` is the whole grid: 128 blocks for a 4096-wide projection and 208 for the
6656-wide output, against 170 SMs, while the kernel's shared-memory footprint fits 8 blocks per SM.
Under 10% of the device was resident no matter how good the mainloop was.

Splitting K multiplies the block count without shrinking the tile, so the arithmetic per block is
unchanged and only the occupancy moves. The split is balanced rather than `ceil`, because 26
super-blocks over 8 splits is 4,4,3,3,3,3,3,3 and the launch waits for the longest split.

The row floor drops from 24 to 8 for the same reason the arm exists at all: the chunked MMVQ below
reads the weights once per eight rows, so at or under eight this arm wins nothing back, and above it
the split-K grid is what pays. A floor of 24 meant only a 32-wide batch ever entered the arm, so
every narrower packed batch was still doing its projections on the CUDA cores.

## Proof of speedup

**Target model(s)**

- [x] **Muse Glimmer**

Reached through the wide Q4_K projections off `xn` — q, the q-gate and the attention output — at
eight packed rows and up. k and v are 256 wide on this checkpoint and stay below the `N` gate, as
before.

| | decode tok/s |
|---|--:|
| before (main) | 569.3 |
| after (this PR) | **681.6** |

| | prefill pp tok/s |
|---|--:|
| before prefill (main) | — |

### Concurrent decode — one binary, `SPARKINFER_MMVQ_MMA_SPLITK` / `_MINM`, interleaved

| c | before | after | |
|---|--:|--:|--:|
| 32 | 569.3 | **681.6** | **+19.7%** |
| 16 | 441.5 | **523.1** | **+18.5%** |
| 8 | 307.8 | **321.4** | **+4.4%** |

Median of three interleaved rounds, one binary. Per round:

```text
c=32: before [568.2, 569.3, 570.2]  after [679.6, 681.6, 719.6]
c=16: before [441.0, 441.5, 442.0]  after [522.9, 523.1, 523.3]
c= 8: before [307.8, 307.8, 308.0]  after [320.9, 321.8]
c= 4: before [190.9]  after [190.7]   (whole branch; neither arm engages here)
c= 2: before [136.9]  after [137.2]   (whole branch; neither arm engages here)
```

c=4 and c=2 sit below the eight-row floor, so this arm cannot engage: their rows compare the
base against the whole branch and show the same binary doing the same thing.

One binary. `main` already has this arm, so the off switch has to reproduce it exactly rather than
approximately: with `SPARKINFER_MMVQ_MMA_SPLITK=1` the kernel writes the caller's bf16 directly with
no accumulator and no epilogue, which is what it did before this change. (That also removes a real
cost — a single K split never needed the scratch round-trip.) The base arm is
`SPARKINFER_MMVQ_MMA_SPLITK=1 SPARKINFER_MMVQ_MMA_MINM=24`.

### Correctness

Not bit-identical, and not claimed to be: this path reduces over K in a different order than the
per-row dp4a MMVQ, and split-K accumulates across blocks.
`kernels/tests/mmvq_mma_accuracy_gpu_test.cpp` dumps both arms at the two real shapes and the diff
is:

```text
wq  N=4096 K=6656 M=32: bit-identical 129536 / 131072 (98.828%)  relRMS 1.7069e-04  worst/rms 5.6334e-03
wo  N=6656 K=4096 M=32: bit-identical 212160 / 212992 (99.609%)  relRMS 1.3990e-04  worst/rms 2.2384e-03
```

### Notes

The fp32 accumulator is a static per-stream slot, not an allocation. Decode captures CUDA graphs
and a `cudaMalloc` reached during capture invalidates the graph, which surfaces as
`verify graph launch: invalid argument` and a step that returns without doing the work. One slot per
stream because the K/V-side projections and the GDN gate are forked onto side streams, so two calls
here can be in flight at once; a stream past the table declines to the MMVQ. The epilogue re-zeroes
what it consumed, so no call needs a memset of its own.

### DIFF
--- a/kernels/csrc/cuda/gemm/gemv.cu
+++ b/kernels/csrc/cuda/gemm/gemv.cu
@@ -10,6 +10,10 @@
 // consumer Blackwell). sm_121 is NOT built -- see kernels/CMakeLists.txt, which
 // excludes it as unsupported by the CUDA toolkit in use.
 
+#ifndef SPARKINFER_NVRTC_DEVICE_ONLY
+#include <mutex>
+#endif
+
 #include <cuda_bf16.h>
 #include <cuda_fp16.h>
 #include <cuda_fp8.h>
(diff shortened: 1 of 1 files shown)


skyrocket2026:

| metric | value |
|---|---|
| tier from | `muse-cb-decode@c32` (+20.4%) |
| commit | `ffd20c599` |

|---|---|---|---|---|
| 128 | decode | 106.53 | 106.62 | +0.1% |
| 128 | prefill | 9853.71 | 9846.14 | -0.1% |
| 512 | decode | 105.63 | 105.71 | +0.1% |
| 512 | prefill | 15913.06 | 16040.03 | +0.8% |
| 4k | decode | 102.83 | 102.77 | -0.1% |
| 4k | prefill | 14903.29 | 14915.01 | +0.1% |
| 16k | decode | 101.82 | 101.81 | -0.0% |
| 16k | prefill | 14367.58 | 14374.97 | +0.1% |
| 32k | decode | 100.49 | 100.43 | -0.1% |
| 32k | prefill | 13083.67 | 13084.61 | +0.0% |
| 64k | decode | 98.28 | 98.27 | -0.0% |
| 64k | prefill | 11192.29 | 11207.00 | +0.1% |

|---|---|---|---|
| c2 | 136.70 | 137.00 | +0.2% |
| c4 | 190.70 | 190.40 | -0.2% |
| c8 | 307.90 | 321.30 | +4.4% |
| c16 | 439.30 | 523.60 | +19.2% |
| c32 | 563.60 | 678.80 | +20.4% |

ok

skyrocket2026:

PR #1045, opened 2026-09-11
perf(muse): score the packed LM head on the int8 tensor cores (1.11x concurrent decode @c32)

FranDev132:
## Summary

`lm_head` is 202048 x 6656 Q4_K — 756 MB, the single largest matrix in the model, and the last one a
continuous-batch step still scored on the CUDA cores. The dp4a multi-row kernel is chunked by eight
rows, so a 32-row step walked the whole head four times and did its 43 G MAC of dot product on the
wrong units.

The head is wide enough to fill the device on its own: `N/BN` is 6314 blocks against 170 SMs. So
this arm needs neither a K split nor an fp32 scratch, and writes the caller's logits directly — it
is the single-split instantiation of the mma kernel that already scores the attention projections,
with `float` as the output type instead of bf16.

Packed only, which is the restriction the dp4a multi-row arm beside it already carries: its rounding
differs from the rows kernel AR scores with, and a continuous-batch step is a plain argmax emit that
absorbs that while a speculative verify cannot. `decode_packed` sets `packed_rows` and DSpark's
verify never does, so every `dspark-*` path keeps the arm it had.

## Proof of speedup

**Target model(s)**

- [x] **Muse Glimmer**

Reached only through the packed continuous-batch head at eight rows and up, and only where the head
is Q4_K (`lm_head_type == 12`, which is the branch the call site sits in). Narrower widths and every
non-packed caller keep the dp4a loop, which is also the fallback whenever this arm declines.

| | decode tok/s |
|---|--:|
| before | 681.6 |
| after (this PR) | **756.7** |

| | prefill pp tok/s |
|---|--:|
| before prefill | — |

### Concurrent decode — one binary, `SPARKINFER_HEAD_MMA=0/1`, interleaved

| c | before | after | |
|---|--:|--:|--:|
| 32 | 681.6 | **756.7** | **+11.0%** |
| 16 | 523.1 | **558.7** | **+6.8%** |
| 8 | 321.4 | **328.9** | **+2.3%** |
| 4 | 190.9 | 190.7 | — |
| 2 | 136.9 | 137.2 | — |

Median of three interleaved rounds. Per round:

```text
c=32: before [679.6, 681.6, 719.6]  after [756.1, 756.7, 758.7]
c=16: before [522.9, 523.1, 523.3]  after [557.8, 558.7, 559.1]
c= 8: before [320.9, 321.8]  after [327.2, 328.9, 329.4]
c= 4: before [190.9]  after [190.7]   (below the floor; arm cannot engage)
c= 2: before [136.9]  after [137.2]   (below the floor; arm cannot engage)
```

### Correctness

Not bit-identical, and not claimed to be: the mma path reduces over K in a different order. What a
continuous-batch step actually depends on is the **argmax**, so that is what
`kernels/tests/head_mma_accuracy_gpu_test.cpp` asserts — against `launch_mmvq_rows_f32` at the real
202048 x 6656 shape, at every width this arm serves:

```text
M=32  argmax agree 32/32   relRMS 1.6353e-05   worst/rms 6.9974e-05
M=16  argmax agree 16/16   relRMS 1.3127e-05   worst/rms 4.5772e-05
M=8   argmax agree  8/8    relRMS 1.3677e-05   worst/rms 4.6900e-05
```

The test fails closed if the reference logits are not finite: Q8_1 is `{ __half2 ds; signed char
qs[32] }`, and filling all 36 bytes randomly makes the scale halves inf or NaN, which would have
every logit NaN and both arms' argmax trivially "agree" on index 0.

### DIFF
--- a/kernels/csrc/cuda/gemm/gemv.cu
+++ b/kernels/csrc/cuda/gemm/gemv.cu
@@ -4050,6 +4050,27 @@ static inline bool launch_mmvq_q4k_mma_rows(const void* q81, const void* W, void
     return true;
 }
 
+// The packed LM head. 202048 x 6656 Q4_K is the single largest matrix in the model and the last
+// one a continuous-batch step still scored on the CUDA cores: the dp4a multi-row kernel walks it
+// once per EIGHT rows, so a 32-row step made four passes over 756 MB and did 43 G MAC of dot
+// product on the wrong units. N/BN is 6314 blocks here, which fills the device by itself, so this
+// needs neither a K split nor an fp32 scratch -- it writes the caller's logits directly.
+bool launch_mmvq_q4k_mma_head_f32(const void* q81, const void* W, float* y,
+                                  int M, int N, int K, cudaStream_t stream) {
+    static int head_mma = -1;
+    if (head_mma < 0) { const char* e = getenv("SPARKINFER_HEAD_MMA"); head_mma = (e && e[0] == '0') ? 0 : 1; }
+    if (!head_mma) return false;
+    // Same eight-row floor as the other mma arms: the tile pads M to sixteen, and below eight the
+    // dp4a kernel's cheaper setup wins back more than the tensor cores do.
+    static int head_mma_min = -1;
+    if (head_mma_min < 0) { const char* e = getenv("SPARKINFER_HEAD_MMA_MINROWS"); head_mma_min = e ? atoi(e) : 8; }
+    if (M < head_mma_min || M > SI_AM_MMAX || (K & 255) || (N % SI_AM_BN)) return false;
+    si_mmvq_q4k_mma_kernel<false, float><<<dim3(N / SI_AM_BN, 1), dim3(SI_AM_NW * 32), 0, stream>>>(
+        reinterpret_cast<const si_block_q8_1*>(q81), reinterpret_cast<const unsigned char*>(W),
+        y, M, N, K);
+    return true;
+}
+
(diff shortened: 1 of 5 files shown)


skyrocket2026:

| metric | value |
|---|---|
| tier from | `muse-cb-decode@c32` (+12.6%) |
| commit | `5b8e54106` |

|---|---|---|---|---|
| 128 | decode | 106.60 | 106.66 | +0.1% |
| 128 | prefill | 9852.88 | 9848.68 | -0.0% |
| 512 | decode | 105.66 | 105.79 | +0.1% |
| 512 | prefill | 15919.84 | 15986.46 | +0.4% |
| 4k | decode | 102.75 | 102.80 | +0.1% |
| 4k | prefill | 14937.36 | 14944.73 | +0.0% |
| 16k | decode | 101.80 | 101.82 | +0.0% |
| 16k | prefill | 14367.27 | 14395.49 | +0.2% |
| 32k | decode | 100.41 | 100.49 | +0.1% |
| 32k | prefill | 13084.31 | 13032.66 | -0.4% |
| 64k | decode | 98.28 | 98.25 | -0.0% |
| 64k | prefill | 11210.74 | 11205.28 | -0.0% |

|---|---|---|---|
| c2 | 135.90 | 137.00 | +0.8% |
| c4 | 190.50 | 190.50 | +0.0% |
| c8 | 321.30 | 329.00 | +2.4% |
| c16 | 522.60 | 559.60 | +7.1% |
| c32 | 681.50 | 767.50 | +12.6% |

ok

skyrocket2026:

PR #1053, opened 2026-09-12
perf(muse): stop the batched prefill trading its int8 down projection for six megabytes it never needed (1.18x concurrent decode @c32)

widecloud:
## Summary

Two fixes in `prefill_batched_run`, both reached only by Muse's continuous-batch prefill.

**1. A 256 MB slack margin decides a 6.8 MB term.** The pass drops the ffn-wide half of its int8 activation staging when free VRAM looks tight. At the continuous-batch shape (N = 256, FC = N) that term is 6.8 MB and c32 leaves ~330 MB free, so the test read pressure and dropped 6.8 MB that could not relieve it — `A_i8` came out too small for `ffn_down`'s int8 arm and it fell back to dequantize-to-bf16 plus `pf_gemm_kernel`: **1.79 s of an 8.35 s wall**. Now it is dropped only when it is both what does not fit and enough on its own to fix that. The second clause matters: keeping it unconditionally costs prefill@65536 −22.1%.

**2. A failed graph capture threw away a recorded pass.** `cudaStreamEndCapture` fails on ~1 in 3 c32 runs. The pass was recorded, not executed, so it can be redone uncaptured (~30 ms); returning −1 instead sent a 256-token prompt to the token loop, **+2.3 s**. This is what has been recorded as "`muse-cb-decode@c32` is bimodal" — every slow run prints `[prefill] graph capture failed` and no fast run does, on main too. Bit-identical: it re-runs the same pass.

They ship together because the second moves no throughput on its own; it is what makes the first measurable on an axis that otherwise throws a −22% coin. `SPARKINFER_PREFILL_FFN_I8_STAGE=0` and `SPARKINFER_PREFILL_GRAPH_REDO=0` restore main independently. `ctest` 23/23.

## Proof of speedup

**Target model(s)**

- [x] **Muse Glimmer**
- [ ] **Shared / both**

**Decode tok/s** (concurrent decode @c32):

| | decode tok/s |
|---|--:|
| before (main) | 979.1 |
| after (this PR) | 1156.2 |

**Prefill pp tok/s** (ctx 512):

| | prefill pp tok/s |
|---|--:|
| before prefill (main) | 14892 — not targeted, flat at all six contexts |
| after prefill (this PR) | 14945 — not targeted, flat at all six contexts |

`bench/scripts/bench.sh` benchmarks **prebuilt release binaries** and needs a `.gguf` file, so it cannot measure a branch. Below is `qwen3_gguf_cb_bench <gguf> C 256 256 512` and `bench_sweep_run`, the harnesses these axes are scored on. RTX 5090 / CUDA 13.0, base `92e66db`, arms from one binary, a warm-up discarded per axis.

```text
=== muse-cb-decode@cN ===        decode_tokens (expected C*256+8)
        main            this PR         delta
  c2    170.4 / 170.0   170.6 / 169.7    0.0%    520 / 520
  c4    234.4 / 234.2   234.9 / 234.2   +0.1%    1032 / 1032
  c8    424.9 / 424.9   425.2 / 424.8    0.0%    2056 / 2056
  c16   827.3 / 827.7   827.7 / 828.1   +0.1%    4104 / 4104
  c32   979.1 (4 runs) 1156.2 (4 runs) +18.1%    8200 / 8200

=== c32, the two fixes separated, four runs each ===
  main          984.8  978.6  977.0  976.1    0/4 capture failures
  staging only  876.0 1157.3  864.5 1154.9    2/4  <- both fatal
  capture only  972.0  978.7  966.0  976.3    2/4  <- both recovered
  this PR      1159.8 1154.3 1154.4 1156.1    0/4

=== 12 single-stream axes, bench_sweep_run, main -> PR ===
   ctx      decode              prefill
   128      107.61 -> 107.68     9675 ->  9702
   512      106.80 -> 106.87    14892 -> 14945
   4096     103.49 -> 104.06    13979 -> 14000
   16384    103.12 -> 103.06    13240 -> 13254
   32768    101.86 -> 101.82    12115 -> 12121
   65536     99.98 ->  99.90    10424 -> 10415
```

### DIFF
--- a/runtime/src/models/qwen35_prefill.cpp
+++ b/runtime/src/models/qwen35_prefill.cpp
@@ -334,6 +334,10 @@ int prefill_batched_run(const Qwen35PrefillCtx& s, const int* prompt_ids, int n,
     static cudaGraph_t     g_pfb_graph = nullptr;
     static cudaGraphExec_t g_pfb_exec  = nullptr;
     static int  g_pfb_n = -1, g_pfb_warm_n = -1, g_pfb_pin_cap = 0;
+    // Set only while re-running a pass whose graph capture failed, so the retry does not try to
+    // capture again -- which is also what bounds the recursion to one level. See the
+    // capture-failure branch at the end of this function.
+    static thread_local bool g_pfb_redo = false;
     static int* g_pfb_pin = nullptr;
     static const void* g_pfb_model_key = nullptr;
     static const void* g_pfb_lin_key = nullptr;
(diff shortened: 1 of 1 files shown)


skyrocket2026:

| metric | value |
|---|---|
| tier from | `muse-cb-decode@c32` (+9.0%) |
| commit | `3fa832d27` |

|---|---|---|---|---|
| 128 | decode | 106.74 | 106.72 | -0.0% |
| 128 | prefill | 9860.73 | 9851.59 | -0.1% |
| 512 | decode | 105.90 | 105.89 | -0.0% |
| 512 | prefill | 15985.46 | 16009.76 | +0.2% |
| 4k | decode | 102.96 | 102.95 | -0.0% |
| 4k | prefill | 15047.22 | 15024.87 | -0.1% |
| 16k | decode | 102.10 | 102.06 | -0.0% |
| 16k | prefill | 14491.43 | 14468.86 | -0.2% |
| 32k | decode | 100.75 | 100.61 | -0.1% |
| 32k | prefill | 13175.59 | 13166.24 | -0.1% |
| 64k | decode | 98.57 | 98.42 | -0.2% |
| 64k | prefill | 11299.07 | 11278.39 | -0.2% |

|---|---|---|---|
| c2 | 158.60 | 158.30 | -0.2% |
| c4 | 227.00 | 226.70 | -0.1% |
| c8 | 398.70 | 397.80 | -0.2% |
| c16 | 824.00 | 822.90 | -0.1% |
| c32 | 1043.80 | 1137.90 | +9.0% |

ok

skyrocket2026:

PR #1051, opened 2026-09-12
perf(muse): score the packed decode's attention in-projections on the resident NVFP4 qkv-gate operand (1.05x concurrent decode @c16)

widecloud:
## Summary

Muse Glimmer's packed continuous-batch decode issues its four attention in-projections as Q4_K — q and the gate on the int8 mma arm, k and v on the chunked row grid. Three launches, **3.04 ms of an 18.7 ms c16 step**, at 0.75 and 0.14 TB/s.

The fused NVFP4 `q|gate|k|v` operand they want is **already resident**: `qwen35.cpp` converts it at load (`qkvg_fp4`, one `[2*qdim+2*kvdim, H]` block-scaled matrix) and `prefill_batched_run` has driven these same four projections through it as ONE block-scaled GEMM since it was written. The packed path simply never picked it up. One GEMM over the same 32.6 MB runs at the 1.10 TB/s the gate/up GEMM beside it already reaches, and costs no VRAM that was not already spent; the single `[N, qkvg_n]` result is scattered back out to the tight row strides the layer's own consumers want.

Scoped to the packed path above eight rows, so c2/c4/c8 and all twelve single-stream axes are unreachable by construction (measured flat). Where the load-time VRAM budget drops the operand — concurrency 32 — the arm declines and main's Q4_K path runs unchanged. `SPARKINFER_MUSE_QKVG_FP4=0` restores main, so both arms of every A/B come out of one binary.

`ctest` 23/23.

## Proof of speedup

- [x] Tested on **RTX 5090** (`sm_120`)

**Target model(s)**

- [x] **Muse Glimmer**
- [ ] **Qwen3.8-27B** (ModelOpt NVFP4 / DSpark)
- [ ] **Shared / both**

**Decode tok/s** (concurrent decode @c16):

| | decode tok/s |
|---|--:|
| before (main) | 786.7 |
| after (this PR) | 828.1 |

**Prefill pp tok/s**:

| | prefill pp tok/s |
|---|--:|
| before prefill (main) | not targeted — path unreachable; measured flat, 14958 |
| after prefill (this PR) | not targeted — path unreachable; measured flat, 14925 |

[...]

PR #1057, opened 2026-09-12
perf(muse): keep four packed rows on the Q3_A gate/up arm instead of the 128-row NVFP4 tile (1.03x concurrent decode @c4)

widecloud:
## Summary

`gu_gemm_min_rows` decides where a packed continuous-batch step leaves the row-batched Q3_A gate/up GEMV for the block-scaled NVFP4 GEMM. It was fitted at 4; #1051 and #1053 changed what else is in the step and the crossover moved. At four rows the Q3_A arm is now faster, so this moves the threshold one width, to 5. Nothing else changes arm.

## Proof of speedup

**Target model(s)**

- [x] **Muse Glimmer**
- [ ] **Shared / both**

| | decode tok/s |
|---|--:|
| before (main) | 248.4 |
| after (this PR) | 256.5 |

| | prefill pp tok/s |
|---|--:|
| before prefill (main) | 15166.8 |
| after prefill (this PR) | 15179.9 |

```text
# bench/scripts/bench.sh benchmarks PREBUILT RELEASE BINARIES and so cannot measure a branch.
# These are the evaluator's own invocations, run against separate builds of clean main
# bc0e47b and of this change, on the same physical RTX 5090 (575 W, CUDA 13.0, sm_120).
# The decode table above is muse-cb-decode@c4, which is the axis this PR moves.

# aggregate tok/s, qwen3_gguf_cb_bench <gguf> C 256 256 512
# ABBA, two interleaved repeats plus a repeated control; every leg checked decode_tokens == C*256+8.

  C     main                          this PR
  c2    179.4  179.5                  179.3  179.4
  c4    249.3  248.3  248.1  248.0    260.4  255.0  254.0
  c8    441.8  441.9                  441.8  441.5
  c16   847.6  847.1                  847.3
  c32   1185.6 1185.6                 1186.5

  c4: 248.4 -> 256.5  = +3.3%     every other concurrency within 0.1%

# twelve single-stream axes, bench_sweep_run (decode tok/s / prefill pp tok/s)
# AR decode is one row and was already on the Q3_A arm, so these are unchanged by construction.

  ctx      main                       this PR
  128      109.6235 /  9885.97        109.6306 /  9899.12
  512      108.8916 / 15990.76        108.9117 / 16024.19
  4096     105.8481 / 15166.85        105.8439 / 15179.89
  16384    104.8578 / 14395.59        104.8767 / 14396.70
  32768    103.6397 / 13103.33        103.6365 / 13106.85
  65536    101.6732 / 11183.70        101.7286 / 11188.22
```

### Why the crossover moved

The block-scaled GEMM's cost does not depend on the row count — its M tile is pinned at 128 rows, because the NVFP4 scale-factor atom is 32x4 = 128 rows in M. Measured per gate/up launch (N=19968, K=6656, a 74.76 MB operand) at c2/c4/c8/c16/c32: **68.5 / 70.1 / 69.5 / 67.7 / 62.9 us** — flat. At four rows it burns a 128-row tile to produce four. The Q3_A arm pays only for the rows it is given, and at two rows it runs at 1.48 TB/s, this box's streaming roof.

Eight rows still belongs to the GEMM (the same probe reads -3.0% there), so 5 is the first width that is on the right side, and it moves exactly one scored axis.

`SPARKINFER_GU_GEMM_MIN_ROWS` still overrides, so both arms come out of one binary.

### DIFF
--- a/runtime/src/models/qwen35_prefill.cpp
+++ b/runtime/src/models/qwen35_prefill.cpp
@@ -3112,10 +3112,29 @@ static bool muse_packed_on() {
 // step is still on the dp4a path -- every tensor-core arm has an eight-row floor, because an
 // m16n8k32 tile pads M to sixteen. This is the one place a narrow batch can reach the tensor cores,
 // and the fitted six was keeping it off them.
+// Row count from which a packed step runs the dense gate/up as ONE block-scaled NVFP4 GEMM per
+// projection instead of the row-batched Q3_A GEMV.
+//
+// Four was right when it was fitted, and it is not right any more. The block-scaled GEMM's cost is
+// INDEPENDENT of the row count -- its M tile is pinned at 128 rows, because the NVFP4 scale-factor
+// atom is 32x4 = 128 rows in M -- so at four rows it burns a 128-row tile to produce four. Measured
+// per gate/up launch on an RTX 5090 (N=19968, K=6656, a 74.76 MB operand) at c2/c4/c8/c16/c32:
+// 68.5 / 70.1 / 69.5 / 67.7 / 62.9 us, flat. The Q3_A arm beside it pays only for the rows it was
+// given, and at two rows it runs at 1.48 TB/s -- this box's streaming roof.
+//
+// #1051 and #1053 then changed what else is in the step, and the crossover moved with it. Measured
+// on the bot's own invocation, ABBA x2 plus a repeat: at four rows the Q3_A arm is now
+// 256.5 against 248.4 aggregate tok/s, and the sign of the earlier reading has flipped (the same
+// probe read -1.0% before those two landed). Eight rows still belongs to the GEMM (-3.0%), so this
+// moves exactly one width across and leaves every other axis on the arm it already ran:
+// c2/c8/c16/c32 within 0.1%, and all twelve single-stream axes within 0.2% -- AR decode is one row
+// and was already on the Q3_A arm.
+//
+// SPARKINFER_GU_GEMM_MIN_ROWS still overrides, so both arms come out of ONE binary.
 static int gu_gemm_min_rows() {
     static const int v = [] {
         const char* e = getenv("SPARKINFER_GU_GEMM_MIN_ROWS");
-        const int x = e ? atoi(e) : 4;
+        const int x = e ? atoi(e) : 5;
         return x < 1 ? 1 : x;
     }();
     return v;


skyrocket2026:

| metric | value |
|---|---|
| tier from | `muse-prefill@512` (+1.0%) |
| commit | `5f0087e7b` |

|---|---|---|---|---|
| 128 | decode | 106.53 | 106.66 | +0.1% |
| 128 | prefill | 9822.63 | 9850.11 | +0.3% |
| 512 | decode | 105.58 | 105.81 | +0.2% |
| 512 | prefill | 15920.21 | 16087.19 | +1.0% |
| 4k | decode | 102.60 | 102.82 | +0.2% |
| 4k | prefill | 14896.17 | 14960.35 | +0.4% |
| 16k | decode | 101.70 | 101.88 | +0.2% |
| 16k | prefill | 14333.55 | 14395.29 | +0.4% |
| 32k | decode | 100.35 | 100.45 | +0.1% |
| 32k | prefill | 13046.41 | 13107.90 | +0.5% |
| 64k | decode | 98.24 | 98.27 | +0.0% |
| 64k | prefill | 11191.89 | 11217.19 | +0.2% |

|---|---|---|---|
| c2 | 157.60 | 157.60 | +0.0% |
| c4 | 227.60 | 224.20 | -1.5% |
| c8 | 410.60 | 410.60 | +0.0% |
| c16 | 822.00 | 821.90 | -0.0% |
| c32 | 1133.20 | 1132.10 | -0.1% |

within significance gate — not a verified muse-prefill@512 improvement

skyrocket2026:
## Closed: no verified speedup — `eval-museglimmer:none`

Measured on the pinned RTX 5090 against the same-box `origin/main`: **1.0%** on the best of 17 scored axes (decode + prefill @ 128/512/4k/16k/32k/64k, concurrent decode @ c2/c4/c8/c16/c32).

**This is not a finding that anything is wrong with your PR.** Nothing regressed, and every correctness gate passed — the change just did not move a number this bot measures. The queue is closed rather than left open so it reflects work that can still be scored.

Three ways forward, depending on which applies:

- **Nothing here measures your optimization yet?** That is a gap in the harness, not a verdict. Open an issue describing the axis you need — model, metric, context/concurrency, and the command that measures it — with your before/after numbers, then reopen this PR and ask for the [`hold`](../../labels/hold) label so it stays open while the axis is added. That path is real: `cb-decode@c2..c32` exists because #1026 asked for it.
- **Correctness fix, refactor, test or docs?** Those are welcome and reviewed by hand; they score 0 by design. Reopen as a **draft**, or ask for `hold`, and say so — being unscored is not grounds for closing that work.
- **Expected a speedup?** Re-measure against current `main` (it moves fast), and reopen with the new numbers — it re-evaluates on the next poll.

PR #1060, opened 2026-09-13
perf(decode): give the widest packed-decode projections a 256-wide N tile (1.05x concurrent decode @c16)

James-CUDA:
## Summary

At packed continuous-batch widths the block-scaled NVFP4 GEMM runs one CTA per 128 output columns with only 16–32 valid rows of M — the scale-factor atom pins the M tile at 128 — so the grid is wide and shallow and each CTA has little work to hide its own TMA latency behind. That, not the tile count, is the limiter here: the identical GEMM on the identical grid is faster at 64 rows than at 16. A 256-wide N tile gives each CTA twice the output columns; it is selected only for the widest projection (the FFN gate/up of both models) and only at packed-decode row counts.

## Proof of speedup

**Target model(s)**

- [x] **Muse Glimmer**
- [x] **Shared / both**

| | decode tok/s |
|---|--:|
| before (main) | 823.8 |
| after (this PR) | 862.1 |

| | prefill pp tok/s |
|---|--:|
| before prefill (main) | 15698.7 |
| after prefill (this PR) | 15539.7 |

```text
# these are the evaluator's own invocations, on one RTX 5090 (575 W, sm_120, CUDA 12.8).
# The decode table is muse-cb-decode@c16, the axis this PR moves most. Prefill is NOT targeted
# and cannot change arm (see below); it is reported for completeness, inside the harness drift
# measured at the bottom of this block.
#
# Both arms come out of ONE binary via SPARKINFER_NVFP4_N256_TILE, so there is no rebuild
# between legs. ABAB at c16; every leg checked decode_tokens == C*256+8 with zero error lines.

# aggregate tok/s, qwen3_gguf_cb_bench <model> C 256 256 512

  Qwen3.8-27B NVFP4                        Muse Glimmer 30B Q4_K_M
  C     main        this PR                C     main        this PR
  c1     95.3        95.3
  c2    187.4       187.0                  c2    155.8       156.2      +0.3%
  c4    326.3       325.6                  c4    226.3       236.0      +4.3%
  c8    537.1       537.1                  c8    409.9       421.6      +2.9%
  c16   782.3/781.0 804.2/804.3   +2.9%    c16   824.4/823.2 862.8/861.3  +4.7%
  c32  1326.4      1332.4         +0.5%    c32  1124.8      1149.8      +2.2%

# Isolated-kernel evidence for the rule (m=16, k=5120, DRAM-cold B pool sized at 3x L2,
# the two tilings measured alternately). us per launch:

  n      CTAs@128   N=128    N=256        n      CTAs@128   N=128    N=256
  6144      48      20.96    38.54        15360     120     41.01    40.04
  10240     80      28.97    38.96        16384     128     43.13    40.67
  12288     96      34.16    39.35        17408     136     47.33    41.36
  13312    104      36.44    39.62        19456     152     50.80    44.32
  14336    112      40.44    39.64        24576     192     72.91    54.72

# The N=256 tile is nearly flat in n and the N=128 tile scales with it; they cross at ~112 CTAs
# of 170, which is the two-thirds-of-the-SM-count rule in prefer_n256().

# Qwen3.8 single-stream, qwen3_gguf_bench <dir> 128 sweep, 3 reps (median), decode / prefill pp

  ctx      main                        this PR
  128      96.7012 /  6175.41          96.7591 /  6149.14
  4096     94.5662 / 15698.66          94.6261 / 15539.73
  32768    94.8015 / 11480.67          94.8581 / 11383.26

# The tile is gated at m <= 32, so a single-stream decode (m=1) and a prefill (m >= 512) cannot
# select it. Decode above is flat to +0.06%. The prefill deltas are harness drift, not the
# change -- repeating the SAME build back to back gives:
#
#   ctx 128    6149.14   4713.56   6252.61     (23% spread)
#   ctx 4096  15539.73  15312.88  15107.41     (2.9%, monotonically decreasing)
#   ctx 32768 11383.26  11332.77  11231.23     (1.4%, monotonically decreasing)
#
# while decode over the same three runs holds to 0.3%.
```

### Accuracy

Two independent checks, both against clean main:

1. `nvfp4_gemm_check` at m=16 — relative error against a double-precision CPU matmul is the same to four decimals with the tile on and off, i.e. the ~0.14 is NVFP4 quantization of both operands and the tiling contributes nothing:

```text
                                       N=256 on     N=256 off
  qwen3.8 gate/up  n=17408 k=5120       0.1419       0.1419
  qwen3.8 down     n=5120  k=17408      0.1414       0.1414
  muse gate/up     n=19968 k=6656       0.1517       0.1517
  muse down        n=6656  k=19968      0.1497       0.1497
```

2. Teacher-forced packed decode — `decode_packed` driven directly at a fixed row count and row order on real corpus ids, each step fed a fixed token so one flipped near-tie cannot cascade, with the two atomic split-K prefill GEMMs pinned off (`SPARKINFER_PREFILL_I8=0`, `SPARKINFER_PREFILL_SKINNY_SPLITK=0`) so the baseline is reproducible:

```text
  rows=16, 64 steps, 1024 tokens        rows=32, 2048 tokens
  main            sha 80d0fdfbe44557c1  main    sha aa003813420d3e40
  this PR         sha 80d0fdfbe44557c1  this PR sha aa003813420d3e40
  0/1024 tokens differ                  0/2048 tokens differ
```


skyrocket2026:

| metric | value |
|---|---|
| tier from | `muse-cb-decode@c4` (+4.5%) |
| commit | `119ee0081` |

|---|---|---|---|---|
| 128 | decode | 106.32 | 106.50 | +0.2% |
| 128 | prefill | 10210.05 | 10218.03 | +0.1% |
| 512 | decode | 105.32 | 105.45 | +0.1% |
| 512 | prefill | 16510.37 | 16512.29 | +0.0% |
| 4k | decode | 102.41 | 102.55 | +0.1% |
| 4k | prefill | 15708.09 | 15780.33 | +0.5% |
| 16k | decode | 101.49 | 101.62 | +0.1% |
| 16k | prefill | 14631.32 | 14687.62 | +0.4% |
| 32k | decode | 100.17 | 100.29 | +0.1% |
| 32k | prefill | 13299.13 | 13317.71 | +0.1% |
| 64k | decode | 97.98 | 98.09 | +0.1% |
| 64k | prefill | 11357.90 | 11394.04 | +0.3% |

|---|---|---|---|
| c2 | 156.90 | 157.00 | +0.1% |
| c4 | 226.90 | 237.20 | +4.5% |
| c8 | 411.40 | 421.80 | +2.5% |
| c16 | 822.80 | 858.60 | +4.4% |
| c32 | 1130.00 | 1152.20 | +2.0% |

ok

skyrocket2026:

PR #1062, opened 2026-09-14
perf(decode): compute the packed decode's projections transposed, so the weight is the tiled dimension (1.15x cb-decode@c16)

James-CUDA:
## Summary

The NVFP4 scale-factor atom is 32x4 = 128 rows in M, so the block-scaled GEMM's M tile cannot go below 128 — a 16-row packed step burns a 128-row tile to produce 16 rows. Most of the MMA is thrown away, and half of every mainloop stage's shared memory goes to A-operand padding, which is why the wide tile carves only **two** stages and these 40-CTA launches sit at 47% of the DRAM roof. An M=64 tile does not compile (it fails cute's SFA TMA copy), so this looked like a floor.

**N has no such floor.** A 64-wide N tile is legal — it is the `Narrow` tiling this file already carries. So compute the transpose instead: feed the **weight** as the M operand and the **activation** as N. `M' = n` is the full weight height, so nothing is padded there; `N' = m` rides the 64-wide tile, so the waste drops from 8x to at most 4x and the B tile halves — which buys a third mainloop stage and with it the memory parallelism these launches lacked.

## Proof of speedup

**Target model(s)**

- [x] **Muse Glimmer**
- [x] **Shared / both**

| | decode tok/s |
|---|--:|
| before (main) | 885.7 |
| after (this PR) | 1021.4 |

| | prefill pp tok/s |
|---|--:|
| before prefill (main) | 15698.7 |
| after prefill (this PR) | 15700.3 |

```text
# these are the evaluator's own invocations on one RTX 5090 (575 W, sm_120, CUDA 12.8), against
# main at d41c755. Both arms come out of ONE binary via SPARKINFER_NVFP4_TRANSPOSED, so there is
# no rebuild between legs. Every leg checked decode_tokens == C*256+8 with zero error lines.
# The decode table is cb-decode@c16. Prefill cannot change arm (see the bound below) and is
# reported flat.

# aggregate tok/s, qwen3_gguf_cb_bench <model> C 256 256 512

  Qwen3.8-27B NVFP4                        Muse Glimmer 30B Q4_K_M
  C     main        this PR                C     main        this PR
  c1     95.2         95.2                 c2    159.1       159.1     flat
  c2    187.3        187.4                 c4    238.3       255.4     +7.2%
  c4    326.2        326.8                 c8    422.8       453.1     +7.2%
  c8    538.6        540.6                 c16   863.1       931.5     +7.9%
  c16   885.7       1021.4    +15.3%       c32  1154.1      1195.3     +3.6%
  c32  1416.7       1570.1    +10.8%

# Isolated-kernel evidence, DRAM-cold, at 16 rows (us per launch):
#
#   shape                      normal   transposed
#   gate/up n=17408 k=5120       46.7       34.4
#   down    n=5120  k=17408      53.7       38.0
#   gdn_qkv n=10240 k=5120       28.2       21.6
#   gdn_z   n=6144  k=5120       19.4       15.2
#   out/wo  n=5120  k=6144       21.7       16.1
#   q|gate  n=12288 k=5120       33.8       25.4
#
# Mainloop stages, read out of the compiler: Wide 2, NarrowT 3.

# Single-stream, 3 reps (median), decode tok/s / prefill pp tok/s

  Qwen3.8   ctx      main                     this PR
            128      96.7012 /  6175.41       96.6994 /  6102.26
            4096     94.5662 / 15698.66       94.5688 / 15700.34
            32768    94.8015 / 11480.67       94.8054 / 11514.23

  Muse      ctx      main                     this PR
            128     103.3809 /  6345.32      103.7138 /  6529.94
            512     102.8437 /  9765.58      103.0329 /  9871.84
            4096    102.0602 / 15511.03      102.4403 / 15624.43
```

### Why this is a swap and not a rewrite

- **The scale-factor layouts coincide.** `prefill_nvfp4_scale_bytes_a(r,k) == prefill_nvfp4_scale_bytes_b(r,k)` at every shape here, and a weight quantized by `quant_b` feeds the A slot unchanged. Checked element by element against the untransposed result: **max |diff| 0**. Nothing is re-quantized and no extra memory is held.
- **Both operands are already K-contiguous** (A RowMajor `[m,k]`, B ColumnMajor `[n,k]`), so swapping which pointer goes where needs no repacking.
- **A column-major D of shape `[n,m]` is exactly a row-major `[m,n]`** — element (j,i) lands at `i*n + j` either way — so the result appears at the same addresses and no caller changes. The same holds for the epilogue's C operand, so the residual fold-in keeps working. `args<>` now takes each config's own strides rather than the `Wide` kernel's globals, which is what lets one `Cfg` differ in destination layout.

The bound is the widest batch a packed step can hand us. Above it the M dimension is no longer mostly padding and the prefill-tuned tilings win, so no prefill or single-stream path changes arm.

### Output equality

Verified rather than assumed, because a plain greedy A/B cannot see it here — packed decode is nondeterministic run to run through the two atomic split-K prefill GEMMs, and main disagrees with *itself* on 13% of tokens.

With those pinned off (`SPARKINFER_PREFILL_I8=0`, `SPARKINFER_PREFILL_SKINNY_SPLITK=0`), a teacher-forced harness driving `decode_packed` directly at a fixed row count and row order, on real corpus ids, reproduces exactly:

```text
  rows=16, 64 steps                      rows=32
  main     sha 80d0fdfbe44557c1          main     sha aa003813420d3e40
  this PR  sha 80d0fdfbe44557c1          this PR  sha aa003813420d3e40
  0/1024 tokens differ                   0/2048 tokens differ
```

`nvfp4_gemm_check` also reports the same relative error against a double-precision CPU matmul with the orientation on and off (qwen3.8 gate/up 0.1419, down 0.1414; muse gate/up 0.1517, down 0.1497) — the ~0.14 is NVFP4 quantization of both operands, and the orientation contributes nothing.


skyrocket2026:

| metric | value |
|---|---|
| **PR DSpark @4k** | **129.75 tok/s** |
| **main DSpark @4k** | **129.85 tok/s** |
| **PR prefill @4k** | **15476.9 pp/s** |
| **main prefill @4k** | **15473.7 pp/s** |
| **PR DSpark @16k** | **130.47 tok/s** |
| **main DSpark @16k** | **130.55 tok/s** |
| **PR prefill @16k** | **11935.4 pp/s** |
| **main prefill @16k** | **11991.8 pp/s** |
| prefill @16k vs main | -0.4% |
| **PR DSpark @32k** | **97.77 tok/s** |
| **main DSpark @32k** | **97.80 tok/s** |
| **PR prefill @32k** | **10715.6 pp/s** |
| **main prefill @32k** | **10715.6 pp/s** |
| prefill vs main | +0.0% |
| **PR prefill @256k** | **5001.6 pp/s** |
| **main prefill @256k** | **5003.7 pp/s** |
| **PR decode @256k** | **94.79 tok/s** |
| **main decode @256k** | **94.85 tok/s** |
| **main concurrent decode @c2** | **187.3 tok/s agg** |
| **PR concurrent decode @c4** | **325.2 tok/s agg** |
| **main concurrent decode @c4** | **325.6 tok/s agg** |
| **PR concurrent decode @c8** | **535.1 tok/s agg** |
| **main concurrent decode @c8** | **537.0 tok/s agg** |
| concurrent decode @c8 vs main | -0.4% |
| **PR concurrent decode @c16** | **1020.8 tok/s agg** |
| **main concurrent decode @c16** | **886.1 tok/s agg** |
| concurrent decode @c16 vs main | +15.2% |
| **PR concurrent decode @c32** | **1564.8 tok/s agg** |
| **main concurrent decode @c32** | **1416.8 tok/s agg** |
| concurrent decode @c32 vs main | +10.4% |
| PR concurrent decode @c1 (floor) | 94.9 tok/s agg |
| main concurrent decode @c1 (floor) | 95.2 tok/s agg |
| concurrent @c1 vs main (floor) | -0.3% |
| PR AR tok/s (floor) | 88.62 |
| main AR tok/s (floor) | 88.62 |
| qwen3.8 (upstream unsloth) guard @16k | ✅ no regression (decode+prefill) — decode 82.8 tok/s · prefill 10129 pp |
| qwen3.6 guard @16k | ✅ no regression (decode+prefill) — decode 461.4 tok/s · prefill 27349 pp |
| commit | `1400bc0f5` |

ok

skyrocket2026:

PR #1064, opened 2026-09-14
perf(decode): let the packed FFN and projections take the block-scaled GEMM from eight rows, not sixteen (1.09x cb-decode@c8)

widecloud:
## Summary

The packed decode moves the FFN (`kFfnGemmMinRows`) and the attention/GDN projections (`kProjGemmMinRows`) from row-GEMVs to the block-scaled NVFP4 GEMM only at 16+ rows. Since #1062 the GEMM is already faster at 8 rows, so this lowers both floors to 8, the GEMM's smallest legal width. Only the packed cb path is affected. `SPARKINFER_FFN_GEMM_MIN_ROWS=16 SPARKINFER_PROJ_GEMM_MIN_ROWS=16` restores main.

## Proof of speedup

**Target model(s)**

- [ ] **Muse Glimmer**
- [ ] **Shared / both**

| | decode tok/s |
|---|--:|
| before (main) | 561.6 |
| after (this PR) | 613.3 |

**Prefill pp tok/s**:

| | prefill pp tok/s |
|---|--:|

`qwen3_gguf_cb_bench <Qwen3.8-27B-NVFP4-RTX5090> C 256 256 512` (the bot's cb-decode command; `bench.sh` only runs prebuilt binaries). Base `44c611d`, interleaved main/PR/PR/main:

```text
        main       this PR    delta
  c2    192.9      192.9       0.0%
  c4    348.9      348.9       0.0%
  c8    561.6      613.3      +9.2%
  c16   1065.5     1066.1     +0.1%
  c32   1647.8     1646.6     -0.1%
```


skyrocket2026:

| metric | value |
|---|---|
| tier from | `muse-prefill@16k` (+96.1%) |
| commit | `40ca65b7e` |

|---|---|---|---|---|
| 128 | decode | 105.88 | 106.06 | +0.2% |
| 128 | prefill | 10150.66 | 10181.08 | +0.3% |
| 512 | decode | 105.22 | 105.46 | +0.2% |
| 512 | prefill | 16328.61 | 16503.08 | +1.1% |
| 4k | decode | 101.00 | 102.30 | +1.3% |
| 4k | prefill | 15667.96 | 15658.56 | -0.1% |
| 16k | decode | 101.48 | 101.40 | -0.1% |
| 16k | prefill | 7442.44 | 14597.48 | +96.1% |
| 32k | decode | 100.13 | 100.10 | -0.0% |
| 32k | prefill | 13254.00 | 13269.09 | +0.1% |
| 64k | decode | 98.23 | 98.26 | +0.0% |
| 64k | prefill | 11332.08 | 11356.50 | +0.2% |

|---|---|---|---|
| c2 | 156.70 | 156.80 | +0.1% |
| c4 | 254.70 | 253.90 | -0.3% |
| c8 | 452.80 | 449.80 | -0.7% |
| c16 | 930.20 | 933.70 | +0.4% |
| c32 | 1191.00 | 1187.70 | -0.3% |

ok

skyrocket2026:

skyrocket2026:
## Correction: the `XL` on this PR was a harness error — relabelled `none`

This PR was scored `eval-museglimmer:XL` on `muse-prefill@16k +96.1%` and auto-merged as the
round's `merge-first` winner. **That gain was not real.** Labels are now `eval:none`.

### The measurement

```
PR @ 16k: prefill 14597.48  (main 7442.44)   <- +96.1%
```

`main`'s 16k prefill in every other round of the last two days:

```
14447.70  14474.67  14488.63  14516.64  14550.94  14491.43
14459.49  14333.55  14283.05  14631.32
```

The baseline in this round came in at **7442.44** — almost exactly half, and far outside that
spread. Your PR's own reading (14597.48) is entirely normal. The +96.1% is a bad baseline, not a
speedup.

Everything else in the verdict agrees: your other sixteen axes were flat (−0.7% .. +1.3%), and the
axis this PR actually targets — `cb-decode@c8` — measured **−0.7%**. A genuine 16k prefill win
would also have moved 4k (−0.1%) and 32k (+0.1%). It didn't.

### Why it happened

The 4k/16k/32k sweep ran at **reps=1** — a single unaveraged sample. One bad sample on the baseline
side is enough to invent a 2× gain.

That tier used reps=1 for cost reasons: a 32k prefill took ~330s when the decision was made. After
this week's prefill work it takes 2.47s, and 16k takes 1.13s — so the saving no longer exists.

This is also the second time. #785 was merged and reverted for exactly this failure, which is why
128/512 were raised to reps=5. That fix was applied to the tier that had just misfired instead of
to the practice, so the next context to draw a bad sample repeated it.

### Fixed in `39774e3`

Every scored context now runs at **reps=5** (median), costing ~48s of prefill in total. Because the
reps are now uniform the tiers collapse into a single sweep call — one model load instead of three,
so the fix is cheaper than what it replaced.

### On the merge

This is **not** a judgement on your code, and nothing here suggests it is wrong — only that it did
not measurably move any scored axis, so the correct label is `none` rather than `XL`. The change is
already on `main`. A maintainer may want to revert it on the #785 precedent (that merge was reverted
because it won a `merge-first` slot on a number that wasn't real), but that's a call for them, not
something the harness should decide.

Sorry for the incorrect score.

widecloud:
Thanks for the correction. Agreed: the Muse XL wasn't real, and `none` is the right label for Muse.

For context, this PR targets **Qwen3.8-27B (DSpark)**, not Muse, and that was the only model ticked. It only changes the Qwen3.8 packed decode path (the FFN and GDN/attention projection row floors), so a flat Muse verdict (cb-decode@c8 −0.7% on Muse) is expected. The DSpark bot, which scores `cb-decode@c8` on Qwen3.8, never evaluated it before the merge.

On RTX 5090 with the bot's own command (`qwen3_gguf_cb_bench <Qwen3.8-27B-NVFP4-RTX5090> C 256 256 512`), base `44c611d`, interleaved main/PR/PR/main, it measured:

```text
        main       this PR    delta
  c8    561.6      613.3      +9.2%
  c16   1065.5     1066.1     +0.1%
  c32   1647.8     1646.6     -0.1%
```

I also re-ran Muse prefill@16k with this exact code against main: 15231.0 → 15229.7 (0.0%), which matches your finding that the 16k gain was the baseline, not the PR.

If it helps before deciding on a revert, a DSpark-bot run of `784883b` against `44c611d` would give the real Qwen3.8 number.

PR #1066, opened 2026-09-14
perf(decode): size the transposed tile to the packed row count, and run the LM head through it too (1.05x cb-decode@c16)

James-CUDA:
## Summary

Two refinements to the transposed orientation, both in its tiling.

**The tile's N should be the row count.** In the transposed orientation N carries the packed **row count**, not the weight height, so the tile wants to be as close to that count as the MMA allows. It was 64 because that is the tiling this file already had; 16 is the shape the work actually has. The N waste disappears at sixteen rows instead of costing 4x, and the B tile drops to a quarter, which carves a **fourth** mainloop stage out of the same shared memory (wide tile 2, 64-wide transposed 3, this one 4). Bytes in flight is what these 40–136 CTA launches have been short of, and a stage is exactly that.

**The LM head was the one GEMM left on the old orientation.** `launch_prefill_nvfp4_gemm_f32` is a separate entry point — its destination is the logit buffer rather than an activation, so it carries its own float-output config. Its N is the vocabulary, so its grid covers the machine many times over and the tile choices that matter elsewhere have nothing to pick between here; its **M is still the packed row count**, and at sixteen rows in a 128-row tile it computes 325 GFLOP to deliver 41.

## Proof of speedup

**Target model(s)**

- [x] **Muse Glimmer**
- [x] **Shared / both**

| | decode tok/s |
|---|--:|
| before (main) | 1024.4 |
| after (this PR) | 1079.3 |

| | prefill pp tok/s |
|---|--:|
| before prefill (main) |  |

```text
# these are the evaluator's own invocations on one RTX 5090 (575 W, sm_120, CUDA 12.8), against
# main at 784883b. Every arm comes out of ONE binary:
#   SPARKINFER_NVFP4_TRANSPOSED_NTILE=64  restores the 64-wide tile
#   SPARKINFER_NVFP4_TRANSPOSED_HEAD=0    restores the head's old orientation
# with both set, the binary reproduces main. Every leg checked decode_tokens == C*256+8 with zero
# error lines.
#
# Prefill is left blank: the transposed orientation is bounded at the packed-decode row ceiling
# and a prefill cannot select it, so there is no prefill number to claim.

# c16 is three interleaved PR/main pairs, c8/c32 two.

  C     main                        this PR                    delta
  c1     95.3                        95.4
  c2    187.3                       187.7
  c4    326.5                       326.5
  c8    597.7 / 597.5               624.3 / 624.6              +4.5%
  c16  1023.9 / 1024.2 / 1025.1    1079.3 / 1080.0 / 1078.5    +5.4%
  c32  1572.4 / 1569.7             1599.6 / 1595.7             +1.7%

# Isolated-kernel evidence for the tile width, DRAM-cold, 16 rows, us per launch (transposed
# orientation throughout):
#
#   shape                     N=64   N=32   N=16
#   gate/up n=17408 k=5120    34.5   33.2   32.8
#   down    n=5120  k=17408   37.8   35.7   33.6
#   gdn_qkv n=10240 k=5120    21.7   20.6   20.3
#   out/wo  n=5120  k=6144    13.8   12.4   11.1
#
# Mainloop stages, read out of the compiler: wide 2, N=64 3, N=32 3, N=16 4.
# At 32 rows the 16-wide tile runs two N tiles, doubling the grid, and still wins.
#
# K stays at 256. A 128-deep K tile carves EIGHT stages instead of four and is worse --
# c16 -1.8%, c32 -3.6% -- so eight shallow stages lose to four deep ones.

# The head, from nsys (--cuda-graph-trace=node, c16). It is the 1x1940 f32 GEMM:
#
#   before   593.3 us/step   715 MB at 1206 GB/s   (67% of the 1792 GB/s roof)
#   after    436.5 us/step   715 MB at 1638 GB/s   (91%)
#
# Muse Glimmer does not reach the head path at all -- its head is Q4_K (si_mmvq_q4k_mma), not the
# NVFP4 float GEMM -- so only the tile width moves it there:
#
#   C     main     this PR
#   c2    158.4    158.6
#   c4    256.1    259.8    +1.4%
#   c8    451.7    456.0    +1.0%
#   c16   934.1    942.7    +0.9%
#   c32  1195.2   1196.1
```

### Output equality

Verified rather than assumed, because a plain greedy A/B cannot see it here — packed decode is nondeterministic run to run through the two atomic split-K prefill GEMMs, and main disagrees with *itself* on 13% of tokens.

With those pinned off (`SPARKINFER_PREFILL_I8=0`, `SPARKINFER_PREFILL_SKINNY_SPLITK=0`), a teacher-forced harness driving `decode_packed` directly at a fixed row count and row order, on real corpus ids:

```text
            rows=8              rows=16             rows=32
  main      47bab19059d0c8cf    80d0fdfbe44557c1    aa003813420d3e40
  this PR   47bab19059d0c8cf    80d0fdfbe44557c1    aa003813420d3e40
            0/512 differ        0/1024 differ       0/2048 differ
```

One harness note, since it bit me while checking this: a direct run of that harness has no
wait-for-free-GPU of its own, and this box is shared. A busy GPU shows up as `[FAIL] load` and an
empty token stream, which hashes to the empty-string sha on BOTH arms and reads as a perfect match.
The runs above assert the expected `ROW` count per arm before comparing, so an invalid run aborts
instead of passing.


skyrocket2026:
## sparkinfer DSpark auto-eval — error

**reason:** `PR speed/accuracy run failed — BUILD_FAILED: gmake[3]: *** [runtime/CMakeFiles/qwen3_gguf_cb_bench.dir/build.make:111: runtime/CMakeFiles/qwen3_gguf_cb_bench.dir/cmake_device_link.o] Error 1`

```
ave fun!
AI agents: READ /etc/vast-agents-guide.md (also linked as ./AGENTS.md) before acting on or describing this instance: it is the operating guide. ('vast-capabilities' shows live state only, not the guide.)
BUILD_FAILED -- tail of /tmp/mopt_build.log:
[  2%] Built target si_vision
[ 11%] Built target si_moe
[ 22%] Built target si_attn
[ 33%] Built target si_quant
[ 36%] Built target si_gemm
[ 36%] Building CUDA object kernels/CMakeFiles/si_nvfp4.dir/csrc/cuda/fused/prefill_nvfp4_sm120.cu.o
[ 61%] Built target si_fused
/workspace/eval/bot_repo/build/_deps/cutlass-src/include/cutlass/gemm/kernel/sm100_static_tile_scheduler.hpp(53): warning #20012-D: __device__ annotation is ignored on a function("StaticPersistentTileScheduler100") that is explicitly defaulted on its first declaration
    __inline__ __attribute__((always_inline)) __attribute__((device)) __attribute__((host))
                                                             ^

Remark: The warnings can be suppressed with "-diag-suppress <warning-number>"

/workspace/eval/bot_repo/build/_deps/cutlass-src/include/cutlass/gemm/kernel/sm100_static_tile_scheduler.hpp(53): warning #20012-D: __host__ annotation is ignored on a function("StaticPersistentTileScheduler100") that is explicitly defaulted on its first declaration
    __inline__ __attribute__((always_inline)) __attribute__((device)) __attribute__((host))
                                                                                     ^

fatbinary fatal   : Could not open input file '/tmp/tmpxft_000b65d2_00000000-5_cmake_device_link.compute_89.sm_89.cubin'
gmake[3]: *** [runtime/CMakeFiles/qwen3_gguf_cb_bench.dir/build.make:111: runtime/CMakeFiles/qwen3_gguf_cb_bench.dir/cmake_device_link.o] Error 1
gmake[2]: *** [CMakeFiles/Makefile2:1791: runtime/CM
```
</details>

skyrocket2026:

| metric | value |
|---|---|
| **PR DSpark @4k** | **129.82 tok/s** |
| **main DSpark @4k** | **129.78 tok/s** |
| **PR prefill @4k** | **15525.5 pp/s** |
| **main prefill @4k** | **15533.0 pp/s** |
| **PR DSpark @16k** | **130.38 tok/s** |
| **main DSpark @16k** | **130.46 tok/s** |
| **PR prefill @16k** | **11959.9 pp/s** |
| **main prefill @16k** | **11973.5 pp/s** |
| prefill @16k vs main | -0.1% |
| **PR DSpark @32k** | **97.78 tok/s** |
| **main DSpark @32k** | **97.79 tok/s** |
| **PR prefill @32k** | **10751.8 pp/s** |
| **main prefill @32k** | **10762.3 pp/s** |
| prefill vs main | +0.0% |
| **PR prefill @256k** | **5020.8 pp/s** |
| **main prefill @256k** | **5017.9 pp/s** |
| **PR decode @256k** | **94.89 tok/s** |
| **main concurrent decode @c2** | **186.9 tok/s agg** |
| concurrent decode @c2 vs main | +0.2% |
| **PR concurrent decode @c4** | **325.6 tok/s agg** |
| **main concurrent decode @c4** | **325.6 tok/s agg** |
| concurrent decode @c4 vs main | +0.0% |
| **PR concurrent decode @c8** | **625.0 tok/s agg** |
| **main concurrent decode @c8** | **597.4 tok/s agg** |
| concurrent decode @c8 vs main | +4.6% |
| **PR concurrent decode @c16** | **1071.6 tok/s agg** |
| **main concurrent decode @c16** | **1022.3 tok/s agg** |
| concurrent decode @c16 vs main | +4.8% |
| **PR concurrent decode @c32** | **1605.0 tok/s agg** |
| **main concurrent decode @c32** | **1571.0 tok/s agg** |
| concurrent decode @c32 vs main | +2.2% |
| main concurrent decode @c1 (floor) | 95.3 tok/s agg |
| main AR tok/s (floor) | 88.56 |
| qwen3.8 (upstream unsloth) guard @16k | ✅ no regression (decode+prefill) — decode 82.8 tok/s · prefill 10154 pp |
| qwen3.6 guard @16k | ✅ no regression (decode+prefill) — decode 461.4 tok/s · prefill 27453 pp |
| commit | `c7a99f6fe` |

ok

skyrocket2026:

PR #1067, opened 2026-09-14
perf(muse): let the packed decode take the NVFP4 qkv-gate GEMM from eight rows (1.12x concurrent decode @c8)

widecloud:
## Summary

Muse's packed decode runs its attention in-projections through the resident NVFP4 q|gate|k|v operand (#1051) only above 8 rows, so cb@c8 still issues them as Q4_K. Since #1062 the block-scaled GEMM is already faster at 8 rows, so this lowers that bound to 8, the GEMM's smallest legal width. Only the Muse packed cb path is affected. `SPARKINFER_MUSE_QKVG_MIN_ROWS=9` restores main.

## Proof of speedup

**Target model(s)**

- [x] **Muse Glimmer**
- [ ] **Shared / both**

| | decode tok/s |
|---|--:|
| before (main) | 500.8 |
| after (this PR) | 559.2 |

**Prefill pp tok/s**:

| | prefill pp tok/s |
|---|--:|

`qwen3_gguf_cb_bench Muse-Glimmer-30B-KQuant-17GB-Q4_K_M.gguf C 256 256 512` (the bot's cb-decode command; `bench.sh` only runs prebuilt binaries). Base `784883b`, interleaved main/PR/PR/main:

```text
        main     this PR    delta
  c2     180.0     179.8    -0.1%
  c4     284.2     284.2     0.0%
  c8     500.8     559.2   +11.7%
  c16    972.4     972.3     0.0%
  c32   1240.9    1240.8     0.0%
```


skyrocket2026:

| metric | value |
|---|---|
| tier from | `muse-cb-decode@c8` (+17.7%) |
| commit | `00f23b016` |

|---|---|---|---|---|
| 128 | decode | 103.99 | 103.97 | -0.0% |
| 128 | prefill | 7850.34 | 7848.21 | -0.0% |
| 512 | decode | 103.40 | 103.47 | +0.1% |
| 512 | prefill | 10476.16 | 10475.20 | -0.0% |
| 4k | decode | 102.40 | 102.36 | -0.0% |
| 4k | prefill | 15763.58 | 15745.53 | -0.1% |
| 16k | decode | 101.51 | 101.49 | -0.0% |
| 16k | prefill | 14664.79 | 14659.80 | -0.0% |
| 32k | decode | 100.02 | 100.12 | +0.1% |
| 32k | prefill | 13258.64 | 13255.00 | -0.0% |
| 64k | decode | 98.00 | 98.02 | +0.0% |
| 64k | prefill | 11229.75 | 11233.47 | +0.0% |

|---|---|---|---|
| c2 | 156.30 | 155.90 | -0.3% |
| c4 | 257.30 | 253.90 | -1.3% |
| c8 | 454.80 | 535.50 | +17.7% |
| c16 | 942.20 | 927.50 | -1.6% |
| c32 | 1189.90 | 1188.70 | -0.1% |

ok

skyrocket2026:

PR #1068, opened 2026-09-14
perf(muse): hold the NVFP4 down and o-projection copies for concurrent decode and run the packed step through them (1.12x concurrent decode @c4)

widecloud:
## Summary

Muse's packed decode still runs `ffn_down` (a quarter of the cb step) and the attention o projection as Q4_K, while prefill already has block-scaled NVFP4 copies of both, converted only at `max_seq <= 2048`; cb serving loads at the 4096 default. This admits the down and o copies up to 4096 and runs the packed down and o projections through the transposed GEMM with the SwiGLU and the attention gate folded into their quantizes. Down is only admitted with a measured runtime margin on top of the reserve, and it takes qkv-gate's place only when the deployment holds at most eight sessions, below the eight rows from which the packed decode reads qkv-gate (#1067), so c8/c16 keep qkv-gate and c32 keeps the set it has. The scored single-stream sweep loads at 65680 and is untouched. `SPARKINFER_MUSE_NVFP4_OUTPUTS_MAXSEQ=2048` restores main.

## Proof of speedup

**Target model(s)**

- [x] **Muse Glimmer**
- [ ] **Shared / both**

**Decode tok/s** (concurrent decode @c4):

| | decode tok/s |
|---|--:|
| before (main) | 289.2 |
| after (this PR) | 325.0 |

**Prefill pp tok/s**:

| | prefill pp tok/s |
|---|--:|

`qwen3_gguf_cb_bench Muse-Glimmer-30B-KQuant-17GB-Q4_K_M.gguf C 256 256 512` (the bot's cb-decode command; `bench.sh` only runs prebuilt binaries). Base `ae0c0e7`, interleaved main/PR/PR/main, one binary:

```text
        main     this PR    delta
  c2     179.0     183.3    +2.4%
  c4     289.2     325.0   +12.4%
  c8     565.5     583.3    +3.1%
  c16    981.4    1020.0    +3.9%
  c32   1242.5    1242.1     0.0%
```


skyrocket2026:

| metric | value |
|---|---|
| tier from | `muse-cb-decode@c4` (+16.4%) |
| commit | `553450684` |

|---|---|---|---|---|
| 128 | decode | 103.92 | 103.99 | +0.1% |
| 128 | prefill | 7839.30 | 7848.16 | +0.1% |
| 512 | decode | 103.40 | 103.48 | +0.1% |
| 512 | prefill | 10398.32 | 10421.23 | +0.2% |
| 4k | decode | 102.36 | 102.42 | +0.1% |
| 4k | prefill | 15741.46 | 15799.32 | +0.4% |
| 16k | decode | 101.47 | 101.56 | +0.1% |
| 16k | prefill | 14651.72 | 14675.06 | +0.2% |
| 32k | decode | 100.15 | 100.10 | -0.0% |
| 32k | prefill | 13269.43 | 13293.11 | +0.2% |
| 64k | decode | 98.14 | 98.09 | -0.0% |
| 64k | prefill | 11279.24 | 11256.07 | -0.2% |

|---|---|---|---|
| c2 | 156.60 | 160.50 | +2.5% |
| c4 | 257.40 | 299.60 | +16.4% |
| c8 | 540.30 | 556.30 | +3.0% |
| c16 | 939.10 | 974.10 | +3.7% |
| c32 | 1188.50 | 1191.20 | +0.2% |

ok

skyrocket2026:

PR #1069, opened 2026-09-14
perf(muse): hold as many layers of the NVFP4 down copy as concurrent decode leaves room for (1.07x concurrent decode @c8)

widecloud:
## Summary

Muse's NVFP4 preflight drops the whole `ffn_down` copy at 8 and 16 sessions, because the 3.7 GB set does not fit beside qkv-gate and o (#1068), and it leaves the VRAM it cannot fill empty: ~2.8 GB at c8 and ~1.1 GB at c16 on a 32 GB card. This converts down on as many layers as that remainder holds, after every other copy has loaded so it can never take a qkv-gate or o layer's room, keeping 1024 MiB for the runtime's own allocation after load (measured ~576 MiB at c8; a 512 MiB margin starves the batched-prefill scratch). The packed decode and the batched prefill already pick the down arm per layer, so a prefix set of layers needs no other change. c2/c4 already hold the full set and c32 has no room, so both are untouched, and the scored single-stream sweep loads at 65680, where down is not eligible. `SPARKINFER_MUSE_NVFP4_DOWN_KEEP_MB=0` restores main.

## Proof of speedup

**Target model(s)**

- [x] **Muse Glimmer**
- [ ] **Shared / both**

| | decode tok/s |
|---|--:|
| before (main) | 583.3 |
| after (this PR) | 624.2 |

**Prefill pp tok/s**:

| | prefill pp tok/s |
|---|--:|

`qwen3_gguf_cb_bench Muse-Glimmer-30B-KQuant-17GB-Q4_K_M.gguf C 256 256 512` (the bot's cb-decode command; `bench.sh` only runs prebuilt binaries). Base `e9530e7`, interleaved main/PR/PR/main, main and PR binaries:

```text
        main     this PR    delta
  c2     183.2     183.1     0.0%
  c4     324.6     324.7     0.0%
  c8     583.3     624.2    +7.0%
  c16   1020.9    1044.8    +2.3%
  c32   1241.6    1241.4     0.0%
```


skyrocket2026:

| metric | value |
|---|---|
| tier from | `muse-cb-decode@c8` (+8.8%) |
| commit | `8acb3632d` |

|---|---|---|---|---|
| 128 | decode | 103.91 | 104.00 | +0.1% |
| 128 | prefill | 7846.80 | 7852.55 | +0.1% |
| 512 | decode | 103.42 | 103.48 | +0.1% |
| 512 | prefill | 10429.63 | 10484.73 | +0.5% |
| 4k | decode | 102.35 | 102.40 | +0.0% |
| 4k | prefill | 15729.64 | 15784.97 | +0.4% |
| 16k | decode | 101.42 | 101.46 | +0.0% |
| 16k | prefill | 14643.66 | 14668.47 | +0.2% |
| 32k | decode | 99.99 | 100.21 | +0.2% |
| 32k | prefill | 13253.31 | 13270.63 | +0.1% |
| 64k | decode | 98.11 | 98.13 | +0.0% |
| 64k | prefill | 11267.84 | 11246.88 | -0.2% |

|---|---|---|---|
| c2 | 160.70 | 160.30 | -0.2% |
| c4 | 299.50 | 299.70 | +0.1% |
| c8 | 556.20 | 605.10 | +8.8% |
| c16 | 973.10 | 1007.50 | +3.5% |
| c32 | 1188.70 | 1188.70 | +0.0% |

ok

skyrocket2026:

PR #1081, opened 2026-09-15
perf(decode): walk the packed checkpoint-FP8 projections in eight-row chunks instead of one row at a time (1.53x cb-decode@c16)

James-CUDA:
## Summary

`launch_gemv_fp8_rows` serves 2..8 rows and declines anything wider, and both callers in the packed step answered a decline with a bare one-row loop. That loop was written for the speculative verify, which never passes eight rows, so it only ever caught a one-row tail. The continuous-batch decode reuses the same projections at 16 and 32 rows — past the cap on every step — so it fell straight to the loop and re-read the whole FP8 weight **once per row**: sixteen reads of every Gated-DeltaNet projection at 16 rows. On the compressed-tensors checkpoint those are the FP8 qkv, z and out projections of all 48 linear-attention layers, so this was 59% of GPU time at c16 and the reason aggregate throughput *fell* from c8 to c16. This walks the rows in eight-row chunks through the batched kernel, and keeps the one-row call only for a chunk it still declines.

## Proof of speedup

**Target model(s)**

- [ ] **Muse Glimmer**
- [ ] **Shared / both**

| | decode tok/s |
|---|--:|
| before (main) | 316.8 |
| after (this PR) | 485.5 |

| | prefill pp tok/s |
|---|--:|
| before prefill (main) |  |

```text
# Checkpoint: unsloth/Qwen3.8-27B-NVFP4 (compressed-tensors, mixed precision):
#   FP8 W8A8  -- self_attn q/k/v/o, linear_attn in_proj_qkv / in_proj_z / out_proj, lm_head,
#                layers 56-63 mlp gate/up/down
#   NVFP4     -- mlp gate/up/down on layers 0-55
# are the evaluator's own invocations on one RTX 5090 (sm_120, CUDA 12.8), against main at 507017b.
# Both arms come out of ONE binary via SPARKINFER_FP8_ROWS_CHUNK (=0 restores the one-row loop).
# The decode table is cb-decode@c16. Prefill is left blank: this is the packed decode path only.

  C     main (one-row loop)     this PR (eight-row chunks)
  c1     83.0                    83.0
  c2    156.0                   156.3
  c4    260.1                   260.4
  c8    425.0                   432.6
  c16   318.3 / 315.3           488.2 / 482.7        316.8 -> 485.5   +53.2%   (ITL 47.6 -> 30.3 ms)
  c32   252.7 / 252.4           253.8 / 254.0        flat -- see below

# c32 does not move because that width is bound by something else first: the 32-session KV pool
# leaves ~1 MB of 32 GB free, the packed step's scratch arena fails to allocate, and the harness
# reports rejected requests. That is a separate memory problem, not this one.

# The all-NVFP4 ModelOpt checkpoint has no FP8 projections and is unchanged: c16 1079.2 -> 1077.3.

# Why (nsys, --cuda-graph-trace=node, c16, main): gemv_fp8_sk_kernel is 59.1% of GPU time, three
# grids, 196,512 launches each = 767 per step. 48 Gated-DeltaNet layers x 3 FP8 projections x 16
# rows = 2,304 per step -- one launch per projection per ROW.
```

### Output equality

A token-stream A/B cannot see this on this checkpoint: the same build run twice already disagrees with itself on 3.7% of teacher-forced tokens, and still on ~1% with `SPARKINFER_DETERMINISTIC=1` plus the FP8-prefill and MMA split-K atomics pinned off. So the changed operation is checked directly instead.

A model-free check feeds identical packed FP8 payloads and activations through both paths — N one-row `launch_gemv_fp8` calls (main past eight rows) against eight-row chunks through `launch_gemv_fp8_rows` — and compares the outputs byte for byte, with the two destination buffers pre-filled with different sentinels so an unwritten byte cannot pass:

```text
  shapes: gdn in_proj_qkv 10240x5120, in_proj_z 6144x5120, out_proj 5120x6144,
          attn q|gate 12288x5120, k/v 1024x5120, o 5120x6144,
          ffn gate/up 17408x5120, down 5120x17408
  rows:   9, 16, 17 (one-row tail), 24, 32

  40 checks, 0 bytes differ -> bit-identical
```

Each row inside a chunk is already bit-identical to its one-row call — `launch_gemv_fp8_rows` keeps the same K-association, per-j accumulation and ordered split sum — so chunking changes how often the weight is read, never what a row computes.


skyrocket2026:

| metric | value |
|---|---|
| PR decode tok/s | 84.70 |
| main decode tok/s | 84.70 |
| PR prefill@128 pp | 5284.14 |
| main prefill@128 pp | 5291.01 |
| prefill@128 vs main | -0.1% |
| PR prefill@16k pp | 10284.86 |
| main prefill@16k pp | 10301.24 |
| prefill@16k vs main | -0.2% |
| commit | `1cf7d910c` |

|---|--:|--:|--:|--:|
| c1 (floor) | 82.8 | 82.8 | +0.0% | 0 / 0 |
| c2 (scored) | 155.9 | 155.7 | -0.1% | 0 / 0 |
| c4 (scored) | 258.6 | 258.5 | -0.0% | 0 / 0 |
| c8 (scored) | 420.5 | 427.3 | +1.6% | 0 / 0 |
| c16 (scored) | 312.9 | 478.2 | +52.8% | 0 / 0 |
| c32 (scored) | 248.5 | 249.3 | +0.3% | 2 / 2 |

ok

PR #1083, opened 2026-09-15
perf(decode): run the packed checkpoint-FP8 Gated-DeltaNet projections on the FP8 tensor cores past eight rows (1.94x cb-decode@c16)

James-CUDA:
## Summary

On a checkpoint that ships the Gated-DeltaNet projections as FP8 (compressed-tensors W8A8), the packed continuous-batch step ran qkv / z / out as row-GEMVs. Those are compute-bound — each row redoes the whole dot product in scalar floats — so a projection costs the same per row at 32 rows as at one, and at c16 they were most of the step. Batched prefill already runs these exact weights on the FP8 tensor cores (`launch_prefill_gemm_fp8`: per-row dynamic e4m3 activation against the checkpoint's own e4m3 rows and per-channel scales), whose cost is flat in the row count. Past eight rows the packed step now does the same: `xn` is staged as e4m3 on the main stream before the side-stream fork, and qkv, z (on the side stream) and out take the GEMM. Eight rows and below — every DSpark verify — are untouched. `SPARKINFER_FP8_PACKED_GEMM_MIN_ROWS=0` restores the row-GEMVs out of the same binary.

## Proof of speedup

**Target model(s)**

- [ ] **Muse Glimmer**
- [ ] **Shared / both**

| | decode tok/s |
|---|--:|
| before (main) | 316.4 |
| after (this PR) | 612.9 |

| | prefill pp tok/s |
|---|--:|
| before prefill (main) |  |

```text
# Checkpoint: unsloth/Qwen3.8-27B-NVFP4 (compressed-tensors; FP8 W8A8 Gated-DeltaNet projections)
# are the evaluator's own invocations on one RTX 5090 (sm_120, CUDA 12.8), main at 507017b, the two
# builds run interleaved per (rep, C). The decode table is cb-decode@c16.

  C      main            this PR
  c1     83.0 /  83.0     83.1 /  83.0
  c2    156.6 / 155.9    156.6 / 156.2
  c4    260.7 / 259.7    260.5 / 259.7
  c8    424.2 / 424.2    432.1 / 431.2
  c16   316.5 / 316.3    613.3 / 612.5     316.4 -> 612.9   +93.7%
  c32   254.4 / 252.1    274.1 / 274.3     VRAM-bound on main: one of 32 requests fails to open in both arms

# Per projection, model-free, weights rotated over >3x L2 (us per projection, GEMV = eight-row chunks):
#                      16 rows: GEMV -> GEMM      32 rows: GEMV -> GEMM
#   qkv 10240x5120        122.5 -> 65.1              243.5 -> 65.9
#   z    6144x5120         72.6 -> 47.6              139.5 -> 48.0
#   out  5120x6144         74.0 -> 43.4              143.5 -> 44.2
#   (8 rows: 58.1 GEMV vs 64.8 GEMM -- why the arm starts at nine)

# ModelOpt checkpoint (NVFP4 projections, no FP8 weights): the arm allocates nothing and never runs.
```

### Output equality

Not bit-identical, by design: the activation is quantized to per-row e4m3 exactly as batched prefill quantizes it for these weights, so decode now runs the numerics prefill already runs. Measured with a teacher-forced harness that drives `decode_packed` at a fixed row count and feeds each step a fixed real token (so neither the scheduler nor sampling feedback can move the result), with the prefill/MMA split-K atomics pinned so the same build agrees with itself:

```text
# 256-token prompts, 64 forced steps per row, real text                  rows 16     rows 32
#   this PR arm off vs arm off (noise floor)          top-1 agreement      99.51%      99.46%
#   this PR arm on  vs arm on  (noise floor)                               99.71%      99.85%
#   arm off vs arm on                                                      94.3%       91.6%
#   next-token accuracy vs the text   arm off                              0.479       0.588
#                                     arm on                               0.488       0.590

# DSpark (dspark_tau_check, 4k prompt): LOSSLESS=1, AR token hash identical to main on both
# checkpoints (the verify is <= 8 rows and never takes this arm).
```

The argmax flips are near-ties (next-token accuracy does not drop); the step's distribution is the one the checkpoint was quantized for.


skyrocket2026:

| metric | value |
|---|---|
| PR decode tok/s | 84.61 |
| main decode tok/s | 84.40 |
| PR prefill@128 pp | 5292.12 |
| main prefill@128 pp | 5251.96 |
| prefill@128 vs main | +0.8% |
| PR prefill@16k pp | 10263.19 |
| main prefill@16k pp | 10150.90 |
| prefill@16k vs main | +1.1% |
| commit | `3d40581c4` |

|---|--:|--:|--:|--:|
| c1 (floor) | 82.7 | 82.8 | +0.1% | 0 / 0 |
| c2 (scored) | 155.8 | 155.9 | +0.1% | 0 / 0 |
| c4 (scored) | 258.1 | 258.1 | +0.0% | 0 / 0 |
| c8 (scored) | 427.4 | 428.2 | +0.2% | 0 / 0 |
| c16 (scored) | 478.3 | 613.1 | +28.2% | 0 / 0 |
| c32 (scored) | 531.0 | 786.3 | +48.1% | 0 / 0 |

ok

skyrocket2026:
Auto-merged as the round's `qwen38-merge-first` winner — verified same-box speedup over `main` on the unsloth checkpoint, with every floor, the differential accuracy gate and the Qwen3.6 guard passing.

PR #1084, opened 2026-09-15
perf(decode): run the dense Q4_K gate/up on the int8 tensor cores for wide packed batches (1.12x cb-decode@c16)

James-CUDA:
## Summary

A wide packed step still ran the gate/up of a dense Q4_K FFN through the per-token MMVQ grid (`gate_up_mmvq2_qwen_kernel`, one block per row per output). That kernel is compute-bound — every row redoes the whole dot product — so its cost grows linearly with the batch, while the `down` projection of the same FFN already runs on the int8 tensor cores (`down_q4k_mma_rows_kernel`) and reads its weights once per step. That kernel is generic in (output rows, input width), so it serves gate and up unchanged with the two swapped: F output rows over H inputs. The fp32 accumulators come from the caller — the gate into a new `[rows, ffn]` arena buffer, the up straight into `h_scratch` — and a one-pass epilogue folds `silu(gate) * up` in place, so no static device memory is added. Taken at 8 rows and up, the same floor as the `down` arm; the DSpark verify never passes the accumulator. On Qwen3.8-27B (unsloth) this is the FFN of layers 56-63, which the loader requantizes to Q4_K. `SPARKINFER_GU_MMA_MINROWS=0` restores the MMVQ gate/up out of the same binary.

## Proof of speedup

**Target model(s)**

- [ ] **Muse Glimmer**
- [ ] **Shared / both**

| | decode tok/s |
|---|--:|
| before (main) | 316.8 |
| after (this PR) | 355.4 |

| | prefill pp tok/s |
|---|--:|
| before prefill (main) |  |

```text
# Checkpoint: unsloth/Qwen3.8-27B-NVFP4 (compressed-tensors; MLP of layers 56-63 loaded as Q4_K)
# are the evaluator's own invocations on one RTX 5090 (sm_120, CUDA 12.8), main at 507017b, the two
# builds run interleaved per (rep, C). The decode table is cb-decode@c16 (mean of three runs).

  C      main                    this PR
  c1     82.6 /  83.0             83.2 /  83.0
  c2    157.0 / 156.3            156.7 / 156.0
  c4    261.4 / 260.3            261.3 / 260.5
  c8    426.7 / 424.9 / 430.8    431.5 / 430.5 / 432.7
  c16   316.9 / 316.6 / 316.9    354.9 / 354.2 / 357.2    316.8 -> 355.4   +12.2%
  c32   250.7 / 258.1            271.5 / 274.7            VRAM-bound on main (one request fails in both arms)

# Where the FFN is the bottleneck the arm is worth far more. With this checkpoint's GDN projections
# already off their row-GEMVs and c32 no longer out of memory, the same arm on vs off in ONE binary:
#   c12  522.2 / 522.3   -> 615.5 / 629.2   (+19.3%)
#   c16  614.3 / 615.8   -> 782.8 / 780.8   (+27.0%)
#   c32  793.4 / 782.6   -> 1112.3 / 1108.1 (+41.0%)
# nsys there, c32, before: gate_up_mmvq2_qwen_kernel 27.0% of GPU time, 1562.6 us per launch.

# Unchanged: ModelOpt checkpoint (NVFP4 FFN) c8 623.2 -> 623.2, c16 1081.0 -> 1080.6,
#            c32 1593.7 -> 1590.5; Muse Glimmer (own layer body) c16 1003.2 -> 1008.1, c32 1193.5 -> 1191.7.
```

### Output equality

Not bit-identical, the same way the `down` arm is not: the Q4_K x Q8_1 integer products are the same, but they are folded per 32-group in fp32 instead of per super-block, so the float reduction order differs. Teacher-forced harness (fixed row count, fixed real input token per step) with the split-K atomics pinned:

```text
#                                                    rows 16     rows 32
#   arm off vs arm off (noise floor)  top-1 agree     99.32%      99.80%
#   arm on  vs arm on                                 99.51%      99.85%
#   arm off vs arm on                                 98.7-99.2%  99.1%
#   next-token accuracy vs the text   off / on        0.478/0.478 0.590/0.590

# DSpark (dspark_tau_check, 4k prompt): LOSSLESS=1, AR token hash identical to main.
```


skyrocket2026:

| metric | value |
|---|---|
| PR decode tok/s | 84.69 |
| main decode tok/s | 84.34 |
| decode speedup vs main | **1.00×** (+0.4%) |
| PR prefill@128 pp | 5280.42 |
| main prefill@128 pp | 5247.13 |
| prefill@128 vs main | +0.6% |
| PR prefill@16k pp | 10270.94 |
| main prefill@16k pp | 10108.44 |
| prefill@16k vs main | +1.6% |
| commit | `4c59f9956` |

|---|--:|--:|--:|--:|
| c1 (floor) | 82.7 | 82.9 | +0.2% | 0 / 0 |
| c2 (scored) | 155.7 | 155.8 | +0.1% | 0 / 0 |
| c4 (scored) | 258.2 | 258.3 | +0.0% | 0 / 0 |
| c8 (scored) | 427.5 | 434.1 | +1.5% | 0 / 0 |
| c16 (scored) | 612.9 | 783.5 | +27.8% | 0 / 0 |
| c32 (scored) | 784.8 | 1106.8 | +41.0% | 0 / 0 |

ok

skyrocket2026:
Auto-merged as the round's `qwen38-merge-first` winner — verified same-box speedup over `main` on the unsloth checkpoint, with every floor, the differential accuracy gate and the Qwen3.6 guard passing.

PR #1085, opened 2026-09-15
perf(decode): run the Q4_K attention k/v on the int8 tensor cores and chunk q|gate at the accumulator cap (1.06x cb-decode@c32)

James-CUDA:
## Summary

Two gaps in `launch_mmvq_rows`, the tensor-core arm a packed step takes for Q4_K projections. Qwen3.8's compressed-tensors checkpoint loads its full-attention projections as Q4_K, and at 16-32 rows both of its attention reads fell back to the eight-row MMVQ chunks:

- **k/v** (4 kv heads x 256 = 1024 rows) sat under the arm's `N >= 2048` floor, so each paid 4-8 chunked launches per layer. A 1024-row projection is 32 blocks, 256 with the K split -- enough to fill the device -- so the floor moves to 1024. The 256-row k/v the floor was written for stays below it.
- **q|gate** (12288 rows) at 32 rows exceeds the split-K accumulator slot (`SI_AM_MMAX x 6656` floats), and a decline there dropped all the way to four eight-row MMVQ launches. It now chunks at the widest width that still fits (16 rows here) through the same tensor-core kernel.

`SPARKINFER_MMVQ_MMA_MINN=2048` and `SPARKINFER_MMVQ_MMA_CHUNK=0` restore main out of the same binary.

## Proof of speedup

**Target model(s)**

- [ ] **Muse Glimmer**
- [ ] **Shared / both**

| | decode tok/s |
|---|--:|
| before (main) | 1112.3 |
| after (this PR) | 1181.8 |

| | prefill pp tok/s |
|---|--:|
| before prefill (main) |  |

```text
# Checkpoint: unsloth/Qwen3.8-27B-NVFP4 (attention q/k/v/o loaded as Q4_K)
# qwen3_gguf_cb_bench <dir> C 256 256 512, SPARKINFER_QWEN38_PREFILL_NVFP4=1
#   SPARKINFER_QWEN38_DECODE_NVFP4=1 SPARKINFER_KV_INT8=1; one binary, knobs alternated, 2 reps.
# The decode table is cb-decode@c32.
# Measured on a build where this checkpoint's Gated-DeltaNet FP8 projections and its layer 56-63
# gate/up already run on the tensor cores and 32 sessions fit in device memory -- there the attention
# reads are a visible share of the step. On main as it stands the FP8 row path and c32 device memory
# bind first, and the same two knobs move c8 / c16 / c32 by +0.8% / +0.8% / +0.6%.

                 minn 2048 (main)   minn 1024          minn 1024 + chunk
  c8             437.3 / 435.4      441.5 / 441.1      441.8 / 441.1       +1.2%
  c16            782.5 / 782.8      804.0 / 805.3      804.3 / 806.5       +2.9%
  c32           1114.2 / 1110.3    1149.3 / 1149.6    1185.1 / 1178.4      +6.2%

# nsys at c32, before: q|gate 4 launches/layer at 35.9 us, k/v 8 launches/layer at 15.7 us
# (2.3 + 2.0 ms of a ~29 ms step); after: q|gate 2 x 42.5 us, k/v 2 x 15.1 us.

# Unchanged: Muse Glimmer (its k/v are 256 rows, its q/wo fit the slot) c16 1008.0 -> 1007.3,
# c32 1192.7 -> 1188.6; ModelOpt checkpoint (NVFP4 projections, never enters this arm);
# single-stream decode (one row, below the arm's eight-row floor).
```

### Output equality

Same kernels on the same rows: the tensor-core MMA already served these projections at other widths, and chunking only changes which rows share a launch. Not claimed bit-identical to the eight-row MMVQ it replaces (different reduction), the same way the existing `N >= 2048` arm is not.


skyrocket2026:
## sparkinfer qwen38 auto-eval — `eval-qwen38:M`

| metric | value |
|---|---|
| **label** | `eval-qwen38:M` |
| PR decode tok/s | 84.57 |
| main decode tok/s | 84.69 |
| PR prefill@128 pp | 5282.74 |
| main prefill@128 pp | 5294.17 |
| prefill@128 vs main | -0.2% |
| PR prefill@16k pp | 10227.56 |
| main prefill@16k pp | 10279.65 |
| prefill@16k vs main | -0.5% |
| commit | `bc8612b36` |

|---|--:|--:|--:|--:|
| c1 (floor) | 82.8 | 82.8 | +0.0% | 0 / 0 |
| c2 (scored) | 156.0 | 155.9 | -0.1% | 0 / 0 |
| c4 (scored) | 258.2 | 257.6 | -0.2% | 0 / 0 |
| c8 (scored) | 434.5 | 437.5 | +0.7% | 0 / 0 |
| c16 (scored) | 785.3 | 805.2 | +2.5% | 0 / 0 |
| c32 (scored) | 1103.9 | 1189.9 | +7.8% | 0 / 0 |

ok

skyrocket2026:

PR #1089, opened 2026-09-15
perf(serving): prefill the prompts that arrive together in one batched pass (1.09x cb-decode@c32)

James-CUDA:
## Summary

A continuous-batch engine that receives a burst of requests receives a burst of prompts, and every one of them went through its own batched prefill pass. At 32 concurrent requests the scheduler already admits the whole backlog in one iteration, and the worker then ran all 33 prompt passes back to back before the first decode token — about a fifth of the run. The batched pass is far from saturated at that width (5858 tok/s at 256 rows against 10972 at 4096 on the same checkpoint), so the fresh, text-only prompts of an iteration are now prefilled together, in passes of up to 4096 rows. Everything row-wise — norms, projections, FFN — runs once over the packed rows; what belongs to one sequence — the recurrent-state reset, the Gated-DeltaNet conv and scan, the KV write and attention, and the seed — runs per prompt on its own slice, with the same calls a lone prompt makes. Prompts a pack cannot carry (logprobs, logit_bias, a constraint, forced tokens, images, prefix-cache work) and any pack the model declines take the existing path. `SPARKINFER_PREFILL_PACK_TOKENS=0` restores one pass per prompt.

## Proof of speedup

**Target model(s)**

- [ ] **Muse Glimmer**
- [ ] **Shared / both**

| | decode tok/s |
|---|--:|
| before (main) | 1198.7 |
| after (this PR) | 1302.1 |

| | prefill pp tok/s |
|---|--:|
| before prefill (main) |  |

```text
# Checkpoint: unsloth/Qwen3.8-27B-NVFP4. bench/scripts/bench.sh benchmarks PREBUILT RELEASE BINARIES and cannot
# measure a branch, so these are the evaluator's own invocations on one RTX 5090 (sm_120, CUDA 12.9), main and
# this PR built from the same tree and interleaved per (rep, C). The decode table is cb-decode@c32.

# aggregate tok/s, median of 3 complete runs, qwen3_gguf_cb_bench <dir> C 256 256 512

  C      main      PACK_TOKENS=0   this PR            max ITL main -> PR
  c1      83.3        83.2          83.2   (flat)
  c2     158.1       158.2         157.3   -0.5%     (runs 148/158/157)
  c4     266.5       266.7         269.3   +1.1%     212 ->  178 ms
  c8     453.9       451.4         465.1   +2.5%     378 ->  269 ms
  c16    808.4       807.5         852.1   +5.4%     706 ->  461 ms
  c32   1198.7      1188.5        1302.1   +8.6%    1374 ->  854 ms

# Where the time went on main (nsys, c32): 34 prefill passes of ~35 ms each, one per prompt, and the first
# decode argmax only after all of them; the bench's own max ITL is that phase (1374 ms of a 6.8 s wall).
# Why packing pays -- single-stream batched prefill by width, same checkpoint:
#   256 rows 5858 tok/s    1024 rows 8977    4096 rows 10972    8192 rows 10447
# What is left of the c32 prompt phase (~854 ms for ~8.7k tokens) is that ~11k tok/s ceiling.

# The other checkpoints, measured the way their guards run (medians of 3; the same build pair on parent 9171513,
# whose one later commit neither checkpoint below executes):
#   ModelOpt Qwen3.8-27B NVFP4   cb c16 1080.7 -> 1126.0 (+4.2%)   c32 1608.9 -> 1718.9 (+6.8%)
#                                32k decode 94.07 -> 94.07   prefill 11193 -> 11175
#   Muse Glimmer 30B (GGUF)      cb c16 1020.0 -> 1019.5           c32 1219.7 -> 1218.6
#                                32k decode 102.32 -> 102.34 prefill 12774 -> 12775
# Muse Glimmer never packs: the model declines the pack before it touches any session.
```

### Output equality

Not bit-identical, the way any change of batch width in the row-wise GEMMs is not: the packed rows run the same kernels as a longer single prompt. Everything per sequence — conv, scan, KV write, attention, seed — is the exact call a lone prompt of that length makes. Checked with a teacher-forced harness that prefills the same 16 or 32 real-text prompts either one pass each or packed, then drives `decode_packed` at a fixed row count with a fixed real input token per step (64 steps), so neither the scheduler nor sampling feedback can move the result:

```text
# deterministic mode + prefill/MMA split-K atomics pinned
#                                     rows 16            rows 32
#   one pass each vs one pass each    1022/1024 (99.8%)  2044/2048 (99.8%)   top-1 agreement
#   one pass each vs packed           1023/1024 (99.9%)  2044/2048 (99.8%)
#   seeds (first token) equal         16/16 and 16/16    32/32 and 32/32
#   next-token accuracy vs the text   .4893 all three    .5898 all three

# DSpark (dspark_tau_check, 4k prompt): LOSSLESS=1, AR token hash identical to main -- a single prompt never packs.
```


skyrocket2026:
## sparkinfer qwen38 auto-eval — `eval-qwen38:M`

| metric | value |
|---|---|
| **label** | `eval-qwen38:M` |
| PR decode tok/s | 84.75 |
| main decode tok/s | 84.54 |
| PR prefill@128 pp | 5326.77 |
| main prefill@128 pp | 5270.71 |
| prefill@128 vs main | +1.1% |
| PR prefill@16k pp | 10347.90 |
| main prefill@16k pp | 10211.45 |
| prefill@16k vs main | +1.3% |
| commit | `0450ef8a4` |

|---|--:|--:|--:|--:|
| c1 (floor) | 82.9 | 82.9 | +0.0% | 0 / 0 |
| c2 (scored) | 156.0 | 156.1 | +0.1% | 0 / 0 |
| c4 (scored) | 259.3 | 262.3 | +1.2% | 0 / 0 |
| c8 (scored) | 439.8 | 451.8 | +2.7% | 0 / 0 |
| c16 (scored) | 805.2 | 849.7 | +5.5% | 0 / 0 |
| c32 (scored) | 1193.7 | 1290.3 | +8.1% | 0 / 0 |

ok

skyrocket2026:

PR #1091, opened 2026-09-16
perf(decode): cheaper Q4_K tensor-core staging and a single-pass FP8 row quantize for wide packed batches (1.03x cb-decode@c16)

James-CUDA:
## Summary

Two kinds of work that sit in front of the real arithmetic at packed decode widths, both made cheaper without changing a single output bit.

**Q4_K on the int8 tensor cores.** At 32 rows the q|gate launch spends only about an eighth of its time in the multiply and the scale fold; most of the rest is staging operands into shared memory. Two pieces of that staging get lighter, in both Q4_K tensor-core kernels (the attention/head one and the dense-FFN one):
- The eight (scale, min) pairs of a super-block are rebuilt from the 12 scale bytes as four 4-byte words — three loads, masks and four stores — instead of one pair at a time through 16-bit loads and sixteen byte stores. Byte k of each word is exactly what the per-pair unpack yields.
- A Q4_K block is 144 B, so each 16 B nibble chunk sits on a 16 B boundary of the weight: it is now one aligned load split with masks, instead of four 4 B loads and a per-byte split.

**FP8 row quantize.** The per-row e4m3 quantize ahead of every FP8 GEMM gave each row one CTA of 32 lanes, each striding every 32nd element — a 160-deep chain of dependent loads for a 5120-wide row, walked twice. At decode widths that chain is the whole cost (the same at 1 row as at 32). It now has the shape of the int8 activation quantizer's fast path: one 256-thread block per row, the row held in registers, read once and written once, and the row max reduced in two levels. The max is exact in any order, so the scale and every e4m3 byte are unchanged. `SPARKINFER_FP8_QUANT_FAST=0` restores the old kernel.

## Proof of speedup

**Target model(s)**

- [ ] **Muse Glimmer**
- [ ] **Shared / both**

| | decode tok/s |
|---|--:|
| before (main) | 851.0 |
| after (this PR) | 877.8 |

| | prefill pp tok/s |
|---|--:|
| before prefill (main) |  |

```text
# Checkpoint: unsloth/Qwen3.8-27B-NVFP4 on one RTX 5090 (sm_120, CUDA 12.9); both builds from the same tree, interleaved
# per (rep, C). The decode table is cb-decode@c16.

# aggregate tok/s, median of 3 complete runs, qwen3_gguf_cb_bench <dir> C 256 256 512
  C      before    after
  c1      83.1      83.3   (flat)
  c2     158.4     158.4   (flat)
  c4     269.5     269.6   (flat)
  c8     465.8     474.8   +1.9%
  c16    851.0     877.8   +3.2%
  c32   1306.9    1336.3   +2.2%
  prefill@16k  10100.5 -> 10121.9 tok/s

# Kernel microbenches (no profiler, weights rotated past L2), us per launch:
#   Q4_K q|gate 12288x5120, 32 rows   60.0 -> 54.9        o 5120x6144   37.3 -> 34.4
#   FP8 row quantize, 32 rows         5.4 / 6.2 -> 2.9    (cols 5120 / 6144; 96 launches a step)
#   FP8 row quantize, 4096 rows       19.3 / 22.8 -> 11.6 / 13.6

# The other checkpoints, measured the way their guards run (medians of 3):
#   Muse Glimmer 30B (GGUF)      cb c16 1019.7 -> 1041.5 (+2.1%)   c32 1222.1 -> 1263.4 (+3.4%)
#                                32k decode 102.41 -> 102.39  prefill 12829 -> 12809
#   ModelOpt Qwen3.8-27B NVFP4   cb c16 1128.3 -> 1128.0           c32 1718.2 -> 1713.0 (-0.3%, median of 10 interleaved)
#                                32k decode 94.11 -> 94.08    prefill 11261 -> 11206
```

### Output equality

Bit-identical by construction and checked: every staged byte is the same byte at the same shared address, and the FP8 quantizer computes the same max over the same values with the same per-element conversion. Kernel-level, the Q4_K outputs are byte-identical to the current kernels on the same weights at a single K split, and the FP8 quantize's e4m3 bytes and scales are byte-identical at 1, 16, 32, 256 and 4096 rows. End to end, DSpark's lossless check at 4k passes with an AR token hash identical to main.


skyrocket2026:
## sparkinfer qwen38 auto-eval — `eval-qwen38:S`

| metric | value |
|---|---|
| **label** | `eval-qwen38:S` |
| PR decode tok/s | 84.69 |
| main decode tok/s | 84.65 |
| PR prefill@128 pp | 5352.16 |
| main prefill@128 pp | 5294.72 |
| prefill@128 vs main | +1.1% |
| PR prefill@16k pp | 10337.83 |
| main prefill@16k pp | 10274.25 |
| prefill@16k vs main | +0.6% |
| commit | `da15526a2` |

|---|--:|--:|--:|--:|
| c1 (floor) | 82.9 | 82.9 | +0.0% | 0 / 0 |
| c2 (scored) | 156.2 | 156.4 | +0.1% | 0 / 0 |
| c4 (scored) | 262.5 | 262.7 | +0.1% | 0 / 0 |
| c8 (scored) | 452.1 | 465.9 | +3.1% | 0 / 0 |
| c16 (scored) | 850.5 | 886.1 | +4.2% | 0 / 0 |
| c32 (scored) | 1301.8 | 1347.7 | +3.5% | 0 / 0 |

ok

skyrocket2026:

PR #1092, opened 2026-09-16
perf(muse): project the packed decode's q, gate, k and v in one launch and fold its QK-norm, RoPE and SwiGLU launches (1.06x concurrent decode @c32)

kaivaryn:
## Summary

Muse Glimmer's packed decode at 32 sessions runs on one stream end to end (21.6 ms GPU-busy in a 22.3 ms step, ~1200 kernels), so every launch in it is on the critical path. This removes launches from that step without changing what it computes:

- **q, gate, k, v in one launch.** At c32 the NVFP4 q|gate|k|v operand does not fit beside the sessions (#1051/#1067), so the attention in-projections ran as Q4_K: q and gate as two back-to-back `launch_mmvq_rows` tensor-core launches over the same staged `xn`, then k|v as a third launch on the dp4a row grid. `si_mmvq_q4k_mma_kernel` now serves two or four matrices from one grid -- every matrix is a whole number of 32-row blocks, so the weight is picked once per block -- and a multi-output epilogue narrows the accumulator into each destination. q|gate|k|v take one launch where `attn_v` is Q4_K (8704 rows, inside the #1085 accumulator slot); where it is Q6_K, q|gate take one and k|v keep the grid that carries the Q6_K dot. Every existing launch compiles the pick out. AR decode already merges q and gate (`SPARKINFER_MG_QG_FUSE`); the packed step never did.
- **QK-norm, RoPE and KV append in one launch.** Three launches per layer (`launch_rmsnorm` q, `launch_rmsnorm` k, `launch_rope_kv_append_normal` / `launch_kv_append`) become `launch_muse_qknorm_rope_kv_rows`, the packed form of the existing single-token Muse kernel: `rmsnorm_kernel`'s pack-of-8 accumulation and warp tree, the normed pack rounded to bf16 before rotating, and the append kernels' own RoPE, slot and copy expressions (bf16 KV).
- **SwiGLU folded into the down quantize.** With the NVFP4 gate/up, the SwiGLU wrote an fp32 scratch that the Q8_1 quantize of the down activation read straight back; `swiglu_quant_h_q8_1_kernel` forms it per lane instead, and is taken only when an int8 down arm follows, since nothing else reads it.

`SPARKINFER_MUSE_PACKED_QG_FUSE=0`, `SPARKINFER_MUSE_PACKED_QKNORM_FUSE=0` and `SPARKINFER_SWIGLU_QUANT_FOLD=0` restore main out of the same binary (`SPARKINFER_MUSE_PACKED_QGKV_FUSE=0` fuses only q and gate).

## Proof of speedup

**Target model(s)**

- [x] **Muse Glimmer**
- [ ] **Shared / both**

| | decode tok/s |
|---|--:|
| before (main) | 1196.7 |
| after (this PR) | 1268.6 |

| | prefill pp tok/s |
|---|--:|
| before prefill (main) |  |

```text
# Checkpoint: Muse-Glimmer-30B-KQuant-17GB-Q4_K_M.gguf
# bench/scripts/bench.sh benchmarks PREBUILT RELEASE BINARIES and cannot measure a branch, so these are the
# evaluator's own invocations on one RTX 5090 (sm_120, CUDA 13.0):
#   qwen3_gguf_cb_bench <gguf> C 256 256 512   (no env, as the Muse bot runs it)
# main = 302ee9a + #1089, i.e. exactly the tree #1089 merged as (a2b3745); this PR on top of it. #1090 landed
# after and is server-only, so it touches none of these paths. The two builds
# run interleaved main/PR/PR/main, two rounds. The decode table is cb-decode@c32, mean of four runs.

  C      main                               this PR
  c2     171.3 / 171.0 / 170.6 / 170.4      172.6 / 172.8 / 171.8 / 172.0      +0.9%
  c4     316.4 / 316.2 / 316.2 / 316.3      319.9 / 319.8 / 319.6 / 320.1      +1.1%
  c8     604.0 / 604.5 / 604.6 / 604.5      612.2 / 612.6 / 612.8 / 612.6      +1.3%
  c16    998.1 / 999.4 / 1000.7 / 999.8     1015.6 / 1015.0 / 1015.1 / 1012.2  +1.5%
  c32   1196.4 / 1197.0 / 1196.4 / 1197.0   1271.8 / 1266.5 / 1268.5 / 1267.6  1196.7 -> 1268.6  +6.0%

# Each change alone, out of ONE binary (its switch off/on/on/off, two rounds, means):
#   q|gate|k|v in one launch        c32 1194.0 -> 1253.2  +5.0%   (c2-c16 take the NVFP4 operand and never reach it)
#   SwiGLU into the down quantize   c32 1254.1 -> 1259.9  +0.5%   c16  999.1 -> 1003.5  +0.4%
#   QK-norm + RoPE + append         c32 1259.0 -> 1267.9  +0.7%   c16 1004.3 -> 1013.1  +0.9%

# nsys, steady-state c32 step: main issues q, gate and o as 157 tensor-core launches at 28.6 us plus 156
# epilogues, and k|v as 52 row-grid launches at 31.0 us (22.32 ms, 1205 kernels). With the one-launch
# projections: on the 26 layers whose attn_v is Q4_K, q|gate|k|v are one 52.5 us launch; on the 26 whose
# attn_v is Q6_K, q|gate are one 49.0 us launch and k|v stay on the 31.0 us grid (21.25 ms, 1075 kernels).
# The folds then replace 104 RMSNorm + 52 append launches with 52, and drop 52 SwiGLU launches.

# Unchanged -- main build vs this PR build, interleaved:
#   Qwen3.8 unsloth cb (bot env), six runs each:
#     c32  1295.1 / 1286.5 / 1296.3 / 1244.1 / 1293.6 / 1286.6 -> 1294.5 / 1246.6 / 1295.4 / 1293.3 / 1282.1 / 1293.5
#          (mean 1283.7 -> 1284.2; the ~1245 runs show up on both builds)
#     c16   842.8 /  843.1 /  843.4 /  843.9 ->  841.2 /  844.0 /  843.5 /  843.6
#   Qwen3.8 ModelOpt cb           c16 1124.2 / 1134.6 -> 1125.0 / 1134.0    c32 1720.9 / 1716.2 -> 1722.7 / 1715.9
#   Muse single stream, qwen3_gguf_bench sweep, reps=5, order PR/main/main/PR/PR/main, means of three:
#     128  decode 106.18 -> 106.29   prefill  7494 ->  7609   (one main run at 7313)
#     4k   decode 105.12 -> 105.17   prefill 15016 -> 15080
#     32k  decode 102.87 -> 102.94   prefill 12645 -> 12673
```

### Output equality

The two folds write the same bytes as the launches they replace, by construction (same expressions, same accumulation and reduction order, same bf16 rounding points). q and gate go through the same tensor-core kernel as before; k and v move from the dp4a row grid onto that kernel, a different reduction order, so they are not claimed bit-identical by construction (the same way #1085's k/v were not). Measured with a teacher-forced harness that drives `decode_packed` at a fixed row count and feeds each step a fixed token (so neither the scheduler nor sampling feedback can move the result), on main with #1089 applied:

```text
# 32 rows, 128-token prompts, 64 forced steps per row = 2048 argmaxes per run
#   every switch off (main's path)               hash 1d00c08067be8383
#   q|gate|k|v launch only                       hash 1d00c08067be8383
#   + SwiGLU fold                                hash 1d00c08067be8383
#   + QK-norm/RoPE/append fold (this PR)         hash 1d00c08067be8383
#   top-1 agreement, every pairing: 2048/2048
```

One c16 run with the SwiGLU fold on exited with SIGSEGV early in this work; it did not reproduce in 9 runs under gdb or in 10 plain runs per build afterwards (main 0/10, this PR 0/10). Single-stream decode and prefill, DSpark's verify (at most eight rows) and every Qwen3.8 path never reach the changed launches.

### Separate: a wrong v on main's wide packed path (not touched here)

Measuring the above turned up a correctness bug that is already on `main` and that this PR neither causes nor fixes -- flagging it rather than bundling a fix into a perf PR.

`launch_mmvq_q4k_rows_multi` fills its weight slots in order, but the kernel runs the Q6_K dot only for blocks that land in slot 3. Muse's wide packed in-projections call it as two matrices, `[k, v]`, so a Q6_K `attn_v` sits in slot 1 and its bytes are read by the Q4_K dot. That is every wide packed step on the 26 layers whose `attn_v` is Q6_K, i.e. whenever the NVFP4 q|gate|k|v copy is not resident -- c32 in the bot's own run. The four-matrix narrow call (<= 8 rows) already puts it in slot 3 and is correct, as is every path that projects v on its own.

```text
# teacher-forced harness, 32 rows, 128-token prompts, 64 forced steps, top-1 agreement
#   32-row default            vs 8-row chunked decode   32/2080 =  1.5%
#   32-row SPARKINFER_MUSE_V6_WIDE=0 (v on its own launch) vs the same  1793/2080 = 86.2%
#   8-row chunked             vs 4-row chunked          1793/2080 = 86.2%   <- correct-vs-correct noise
# Rows are independent sequences, so a row's argmax must not depend on the chunk that carries it;
# with random forced tokens many argmaxes are near-ties, which is what puts correct paths at ~86%.
```

Mapping the Q6_K matrix to slot 3 for any matrix count restores it (verified: 85.9% against the 8-row reference, 86.6% against the 4-row one). It is not in this PR because it is a separate change and it is not free: the correct Q6_K dot costs `main` about 2.4% at c32 (1192.7 -> 1164.6 tok/s here), and this PR's numbers are a like-for-like comparison in which both arms run main's current path. On a fixed baseline the same optimisations measure 1164.6 -> 1224.1, +5.1%. Happy to open that fix as its own PR.


skyrocket2026:

| metric | value |
|---|---|
| tier from | `muse-cb-decode@c32` (+6.2%) |
| commit | `dc41ea28c` |

|---|---|---|---|---|
| 128 | decode | 103.90 | 103.88 | -0.0% |
| 128 | prefill | 7873.42 | 7884.26 | +0.1% |
| 512 | decode | 103.40 | 103.24 | -0.2% |
| 512 | prefill | 10422.25 | 10420.10 | -0.0% |
| 4k | decode | 102.39 | 102.34 | -0.0% |
| 4k | prefill | 15729.85 | 15644.49 | -0.5% |
| 16k | decode | 101.50 | 101.54 | +0.0% |
| 16k | prefill | 14614.16 | 14573.19 | -0.3% |
| 32k | decode | 100.07 | 100.10 | +0.0% |
| 32k | prefill | 13204.83 | 13172.62 | -0.2% |
| 64k | decode | 97.84 | 97.87 | +0.0% |
| 64k | prefill | 11201.75 | 11176.54 | -0.2% |

|---|---|---|---|
| c2 | 159.80 | 161.10 | +0.8% |
| c4 | 299.40 | 302.60 | +1.1% |
| c8 | 617.40 | 625.60 | +1.3% |
| c16 | 1040.90 | 1058.70 | +1.7% |
| c32 | 1253.00 | 1330.60 | +6.2% |

ok

skyrocket2026:

PR #1100, opened 2026-09-17
perf(muse): run the narrow packed decode's attention and LM-head Q4_K rows kernels at the batch width (1.03x concurrent decode @c2)

kaivaryn:
## Summary

At 2 to 5 rows, Muse Glimmer's packed decode ran its Q4_K attention grid and its LM head through the 6-row kernel instantiations. Both now use the instantiation that matches the batch width. Each row does the same arithmetic in the same order, so outputs are unchanged. `SPARKINFER_MUSE_ROWS_EXACT=0` restores main.

## Proof of speedup

**Target model(s)**

- [x] **Muse Glimmer**
- [ ] **Shared / both**

| | decode tok/s |
|---|--:|
| before (main) | 176.5 |
| after (this PR) | 181.8 |

| | prefill pp tok/s |
|---|--:|
| before prefill (main) |  |

```text
# RTX 5090, Muse-Glimmer-30B-KQuant-17GB-Q4_K_M.gguf, main 5aa27de. bench.sh only runs prebuilt releases, so these are
# the bot's own command: qwen3_gguf_cb_bench <gguf> C 256 256 512. Decode table = cb-decode@c2.
  c2   176.5 -> 181.8   +3.0%
  c4   308.9 -> 314.4   +1.8%
# One binary, SPARKINFER_MUSE_ROWS_EXACT=0 -> 1:   c2 176.1 -> 181.6   c4 308.8 -> 314.1
# Accuracy: qwen3_gguf_score dumps main vs PR byte-identical (PPL 2.65977).
# c2 per-stream token hashes identical (at c4 two runs of main already differ from each other).
# Not reached: c8+ (gated to <= 5 rows), single-stream sweep (no Q4_K rows launch in its trace),
# Qwen3.6 / Qwen3.8 guards (gated to hidden size 6656). No extra VRAM. ctest 25/25.
```


skyrocket2026:

| metric | value |
|---|---|
| tier from | `muse-cb-decode@c2` (+3.8%) |
| commit | `edb116294` |

|---|---|---|---|---|
| 128 | decode | 104.03 | 104.04 | +0.0% |
| 128 | prefill | 7894.38 | 7894.97 | +0.0% |
| 512 | decode | 103.56 | 103.54 | -0.0% |
| 512 | prefill | 10473.02 | 10487.69 | +0.1% |
| 4k | decode | 102.62 | 102.58 | -0.0% |
| 4k | prefill | 15805.50 | 15820.52 | +0.1% |
| 16k | decode | 101.58 | 101.52 | -0.1% |
| 16k | prefill | 14742.34 | 14707.19 | -0.2% |
| 32k | decode | 100.17 | 100.13 | -0.0% |
| 32k | prefill | 13352.17 | 13302.01 | -0.4% |
| 64k | decode | 98.10 | 98.06 | -0.0% |
| 64k | prefill | 11372.72 | 11306.81 | -0.6% |

|---|---|---|---|
| c2 | 172.90 | 179.50 | +3.8% |
| c4 | 303.00 | 307.90 | +1.6% |
| c8 | 627.30 | 627.60 | +0.0% |
| c16 | 1061.20 | 1061.30 | +0.0% |
| c32 | 1336.20 | 1335.00 | -0.1% |

ok

skyrocket2026:

PR #1106, opened 2026-09-17
perf(muse): hold the sliding-window layers' KV in a per-sequence ring and spend the freed VRAM on the qkv-gate operand (1.11x cb decode @c32)

kaivaryn:
## Summary

Muse Glimmer slides 39 of its 52 layers over 2048 tokens, but every KV slice was sized for the whole context — a 33-session pool spent 7.31 GB holding windows worth 5.31 GB, and the NVFP4 qkv-gate operand a packed step wants (it replaces the Q4_K attention in-projections, 4.4 ms of a 19.4 ms c32 step) was refused for want of 1.7 GB. A windowed slot now gets a private ring of the window plus one prefill pass, whose ring-repeated block table means no append or attention kernel learns about rings, and the freed VRAM is spent by a budget that leaves the runtime its own allocation (the o-proj leg drops first, as the preflight already intended). `SPARKINFER_KV_SWA_CAP=0` restores full-context slices, `SPARKINFER_MUSE_NVFP4_RUNTIME_MB=0` the old budget.

## Proof of speedup

**Target model(s)**

- [x] **Muse Glimmer**
- [ ] **Shared / both**

| | decode tok/s |
|---|--:|
| before (main) | 1390.8 |
| after (this PR) | 1546.3 |

| | prefill pp tok/s |
|---|--:|
| before prefill (main) |  |

```text
# Muse-Glimmer-30B-KQuant-17GB-Q4_K_M.gguf, RTX 5090, bot: qwen3_gguf_cb_bench <gguf> C 256 256 512, agg_tok_s (bench.sh measures prebuilt releases). main = 4124b14 (#1105 merged).
  main       190.7   344.8   643.8  1093.1  1390.8
  this PR    191.0   346.8   651.9  1124.0  1546.3   +11.2% @c32, +2.8% @c16
  PR, cap=0  190.8   344.7   643.6  1093.0  1387.8   same binary
# Reversed order: c32 ring 1544.2 vs main 1388.6 (+11.2%), c16 1123.5 vs 1092.5 (+2.8%).
# Pool at 33 sessions 7.31 -> 5.31 GB (39 of 52 slices ringed at 2640 tokens vs 4096), free VRAM at load 2.23 -> 4.22 GB, so qkv-gate goes 0/52 -> 52/52 resident (down 2/52); peak VRAM 31.85 GB vs main's 31.92.
# Accuracy: bit-identical with the operand set pinned (SPARKINFER_MUSE_PREFILL_NVFP4=0, so both arms compute in Q4_K) -- cb c4 token dumps, cap on vs off, agree 256/256 on every stream. At c2 the arms differ from the second token, which is main's own c2 nondeterminism on this path (two outcomes run to run, reproduced on main alone). A pass past the ring is chunked by the windowed prefill driver, not wrapped: forcing the bound to 128 tokens with 256-token prompts gives zero token-loop fallbacks.
# Refusals, each rather than serving wrong data: cross-sequence prefix blocks (allocate_with_prefix returns false and the engine stops looking one up) and an over-long pass (prefill_batched_run declines as a backstop). New kv_swa_ring_gpu_test covers layout, ring repetition, per-sequence privacy, both refusals and both kill switches. Full build + ctest 26/26.
# Guards: Muse single-stream sweep 128..64k unaffected -- the ring is a cb-pool opt-in and that run activates it 0 times (PR 7983/10478/16698/16067/14492/12185 vs main 7893/10334/16417/15855/14326/12091, PR arm first in the chain). Qwen3.6 cb c4 968.1 -> 972.8; Qwen3.8 bot-env cb c4 269.9 -> 270.6. Neither model opts in, and both keep prefix sharing.
```


skyrocket2026:

| metric | value |
|---|---|
| tier from | `muse-cb-decode@c32` (+11.6%) |
| unsloth qwen3.8 guard | ❌ **FAILED** — unsloth qwen3.8 guard measurement unavailable; unsloth qwen3.8 decode@32k: PR measurement missing/zero (main 75.9) — treated as regression; unsloth qwen3.8 prefill@32k: PR measurement missing/zero (main 9082.8) — treated as regression — **verdict forced to REJECT regardless of speed/accuracy** |
| commit | `7ba4fa85c` |

|---|---|---|---|---|
| 128 | decode | 104.03 | 104.03 | -0.0% |
| 128 | prefill | 7906.16 | 7892.99 | -0.2% |
| 512 | decode | 103.55 | 103.51 | -0.0% |
| 512 | prefill | 10438.50 | 10481.81 | +0.4% |
| 4k | decode | 102.68 | 102.45 | -0.2% |
| 4k | prefill | 16781.74 | 16644.75 | -0.8% |
| 16k | decode | 101.52 | 101.50 | -0.0% |
| 16k | prefill | 16295.01 | 16195.76 | -0.6% |
| 32k | decode | 100.04 | 100.05 | +0.0% |
| 32k | prefill | 14611.03 | 14537.33 | -0.5% |
| 64k | decode | 98.13 | 98.01 | -0.1% |
| 64k | prefill | 12293.46 | 12220.64 | -0.6% |

|---|---|---|---|
| c2 | 186.00 | 185.50 | -0.3% |
| c4 | 336.40 | 339.40 | +0.9% |
| c8 | 627.70 | 635.50 | +1.2% |
| c16 | 1058.90 | 1091.00 | +3.0% |
| c32 | 1337.40 | 1492.80 | +11.6% |

unsloth qwen3.8 no-regression guard failed: unsloth qwen3.8 guard measurement unavailable; unsloth qwen3.8 decode@32k: PR measurement missing/zero (main 75.9) — treated as regression; unsloth qwen3.8 prefill@32k: PR measurement missing/zero (main 9082.8) — treated as regression | ok

skyrocket2026:

Measured **11.6%** vs the same-box `origin/main`, and regressed the unsloth Qwen3.8 no-regression guard (decode/prefill @ 32k) — closing automatically.

PR #1108, opened 2026-09-17
perf(muse): hold the sliding-window layers' KV in a per-sequence ring and spend the freed VRAM on the qkv-gate operand (1.11x cb decode @c32)

kaivaryn:
## Summary

Muse Glimmer slides 39 of its 52 layers over 2048 tokens, but every KV slice was sized for the whole context — a 33-session pool spent 7.31 GB holding windows worth 5.31 GB, and the NVFP4 qkv-gate operand a packed step wants (it replaces the Q4_K attention in-projections, 4.4 ms of a 19.4 ms c32 step) was refused for want of 1.7 GB. A windowed slot now gets a private ring of the window plus one prefill pass, whose ring-repeated block table means no append or attention kernel learns about rings, and the freed VRAM is spent by a budget that leaves the runtime its own allocation (the o-proj leg drops first, as the preflight already intended). `SPARKINFER_KV_SWA_CAP=0` restores full-context slices, `SPARKINFER_MUSE_NVFP4_RUNTIME_MB=0` the old budget.

## Proof of speedup

**Target model(s)**

- [x] **Muse Glimmer**
- [ ] **Shared / both**

| | decode tok/s |
|---|--:|
| before (main) | 1390.8 |
| after (this PR) | 1546.3 |

| | prefill pp tok/s |
|---|--:|
| before prefill (main) |  |

```text
# Muse-Glimmer-30B-KQuant-17GB-Q4_K_M.gguf, RTX 5090, bot: qwen3_gguf_cb_bench <gguf> C 256 256 512, agg_tok_s (bench.sh measures prebuilt releases). main = 4124b14 (#1105 merged).
  main       190.7   344.8   643.8  1093.1  1390.8
  this PR    191.0   346.8   651.9  1124.0  1546.3   +11.2% @c32, +2.8% @c16
  PR, cap=0  190.8   344.7   643.6  1093.0  1387.8   same binary
# Reversed order: c32 ring 1544.2 vs main 1388.6 (+11.2%), c16 1123.5 vs 1092.5 (+2.8%).
# Pool at 33 sessions 7.31 -> 5.31 GB (39 of 52 slices ringed at 2640 tokens vs 4096), free VRAM at load 2.23 -> 4.22 GB, so qkv-gate goes 0/52 -> 52/52 resident (down 2/52); peak VRAM 31.85 GB vs main's 31.92.
# Accuracy: bit-identical with the operand set pinned (SPARKINFER_MUSE_PREFILL_NVFP4=0, so both arms compute in Q4_K) -- cb c4 token dumps, cap on vs off, agree 256/256 on every stream. At c2 the arms differ from the second token, which is main's own c2 nondeterminism on this path (two outcomes run to run, reproduced on main alone). A pass past the ring is chunked by the windowed prefill driver, not wrapped: forcing the bound to 128 tokens with 256-token prompts gives zero token-loop fallbacks.
# Refusals, each rather than serving wrong data: cross-sequence prefix blocks (allocate_with_prefix returns false and the engine stops looking one up) and an over-long pass (prefill_batched_run declines as a backstop). New kv_swa_ring_gpu_test covers layout, ring repetition, per-sequence privacy, both refusals and both kill switches. Full build + ctest 26/26.
# The 2026-09-17 21:14 verdict on #1106 measured this change at +11.6% @c32 with accuracy, the qwen3.6 guard and the modelopt guard all green, and REJECTed it on the unsloth Qwen3.8 guard reporting "PR measurement missing/zero (main 75.9 / 9082.8)" -- a non-measurement on that box, not a slowdown. Re-run here three times, including the bot's own sequence (Muse cb c32, then modelopt 32k, then unsloth 32k at reps=5, wait-for-idle between): PR decode@32k 76.35 vs main 76.35, prefill@32k 9257.9 vs main 9225.9 (+0.3%), rc=0 with SWEEP_JSON present every time. The PR also SHORTENS that round -- the Muse cb c32 stage takes 15.1 s end to end against main's 30.9 s, since the ringed pool allocates 2 GB less. Asking for a re-run of that guard.
# Guards: Muse single-stream sweep 128..64k unaffected -- the ring is a cb-pool opt-in and that run activates it 0 times (PR 7983/10478/16698/16067/14492/12185 vs main 7893/10334/16417/15855/14326/12091, PR arm first in the chain). Qwen3.6 cb c4 968.1 -> 972.8; Qwen3.8 bot-env cb c4 269.9 -> 270.6. Neither model opts in, and both keep prefix sharing.
```


skyrocket2026:

| metric | value |
|---|---|
| tier from | `muse-cb-decode@c32` (+12.2%) |
| commit | `7ba4fa85c` |

|---|---|---|---|---|
| 128 | decode | 103.81 | 103.42 | -0.4% |
| 128 | prefill | 7879.94 | 7831.50 | -0.6% |
| 512 | decode | 103.29 | 102.79 | -0.5% |
| 512 | prefill | 10358.02 | 10336.32 | -0.2% |
| 4k | decode | 102.34 | 102.20 | -0.1% |
| 4k | prefill | 16566.59 | 16366.74 | -1.2% |
| 16k | decode | 101.30 | 101.25 | -0.0% |
| 16k | prefill | 16104.43 | 15939.63 | -1.0% |
| 32k | decode | 100.01 | 99.95 | -0.1% |
| 32k | prefill | 14495.84 | 14367.82 | -0.9% |
| 64k | decode | 98.08 | 97.98 | -0.1% |
| 64k | prefill | 12143.20 | 12121.55 | -0.2% |

|---|---|---|---|
| c2 | 185.30 | 185.00 | -0.2% |
| c4 | 336.80 | 339.10 | +0.7% |
| c8 | 627.90 | 634.30 | +1.0% |
| c16 | 1060.60 | 1092.20 | +3.0% |
| c32 | 1333.60 | 1496.30 | +12.2% |

ok

skyrocket2026:

PR #1110, opened 2026-09-18
perf(muse): serve two Q4_K tensor-core column blocks per CTA off one staged activation tile (1.03x cb decode @c32)

kaivaryn:
## Summary

At Muse Glimmer's 6656 x 19968 down projection every CTA stages the same activation slice out of L2 once per super-block, so it is re-read 208 times — a 32-row step moves 149.6 MB of activation to read 74.8 MB of weights, and the 202048 x 6656 LM head is worse at 1.51 GB against 756 MB. A CTA now serves two column blocks off one staged A tile, reloading Bs and its scales per block rather than widening the tile, so the shared footprint and the five CTAs per SM the 32-row tile allows are unchanged and the split-K grows to keep the grid a whole wave; the down projection, the o-projection and the packed LM head all take it, at the full 32-row packed width only. `SPARKINFER_MMA_COLGROUPS=1` restores main out of the same binary.

## Proof of speedup

**Target model(s)**

- [x] **Muse Glimmer**

| | decode tok/s |
|---|--:|
| before (main) | 1546.75 |
| after (this PR) | 1586.70 |

| | prefill pp tok/s |
|---|--:|
| before prefill (main) |  |

```text
Muse-Glimmer-30B-KQuant-17GB-Q4_K_M.gguf, RTX 5090, main 8d55f3f (measured on 4124b14 + #1108, its content here).
Bot invocation qwen3_gguf_cb_bench <gguf> <C> 256 256 512; bench.sh measures prebuilt releases, not this path.
  cb decode        main   this PR   knobs=0 (same binary)
  c32            1546.8    1586.7   1538.7   (+2.58%, single runs base/PR/PR/base, adjacent pair)
  c16/c8/c4/c2     flat      flat   -        (<=0.2%, the arm floors at 32 rows)
Sweep's 12 axes and Qwen3.8 cb at every width flat: this arm gates on M in [2,32], so prefill (M=N)
and single-stream decode (M=1) never reach it. Guards @32k PR/main decode/prefill: Qwen3.6
1.000/1.003, ModelOpt 1.000/1.002, unsloth 1.000/1.001. Accuracy: LM head argmax identical 32/32,
bit-identical (no split-K atomics there); down and o-proj rms identical. Peak VRAM identical (Muse
c32 31853, Qwen3.8 c32 31541 MiB). Full build + ctest 26/26.
```


skyrocket2026:

| metric | value |
|---|---|
| tier from | `muse-cb-decode@c32` (+2.9%) |
| commit | `775aec3a0` |

|---|---|---|---|---|
| 128 | decode | 104.23 | 104.10 | -0.1% |
| 128 | prefill | 7924.16 | 7902.50 | -0.3% |
| 512 | decode | 103.73 | 103.68 | -0.0% |
| 512 | prefill | 10488.26 | 10496.47 | +0.1% |
| 4k | decode | 102.68 | 102.73 | +0.0% |
| 4k | prefill | 16825.45 | 16803.87 | -0.1% |
| 16k | decode | 101.70 | 101.60 | -0.1% |
| 16k | prefill | 16372.77 | 16307.55 | -0.4% |
| 32k | decode | 100.10 | 100.22 | +0.1% |
| 32k | prefill | 14634.18 | 14604.33 | -0.2% |
| 64k | decode | 98.16 | 98.25 | +0.1% |
| 64k | prefill | 12271.36 | 12228.68 | -0.3% |

|---|---|---|---|
| c2 | 185.70 | 185.60 | -0.1% |
| c4 | 339.20 | 339.30 | +0.0% |
| c8 | 636.20 | 636.00 | -0.0% |
| c16 | 1090.80 | 1092.50 | +0.2% |
| c32 | 1495.30 | 1539.30 | +2.9% |

ok

skyrocket2026:

PR #1111, opened 2026-09-18
perf(qwen3.8): stage the int8 flash-decode K tile through shared memory so each token's bytes arrive whole (1.10x target decode @256k)

kaivaryn:
## Summary

The int8 tensor-core flash-decode wants a 16x16 K tile whose columns are TOKENS, so at `ldm = num_kv_heads * HEAD_DIM` its fragment load asks for sixteen 16-byte runs a kilobyte apart — 16 useful bytes out of every 32-byte sector — and nothing catches the other half, because the kernel's 31.25 KB of dynamic shared memory against the 100 KB an `sm_120` SM actually has caps residency at three blocks and leaves ~11 KB of L1 per block against a 32 KB per-group K working set. K now arrives through shared memory instead: a warp reads each token's 64 contiguous bytes — two whole sectors, nothing discarded — and scatters them into four 16-byte-strided tiles the mma reads at `ldm` 16, which is the same operand element for element. It costs no occupancy because `s_s` pays for it: that buffer is declared `[16][HEAD_DIM]` but every store into it uses `ldm` 128, the group token width rather than the head dim, so at hd256 half of its 16 KB was never addressed, and trimming it to `[16][128]` releases exactly the 8 KB the staging buffer needs. V keeps the direct load — its eight warps read adjacent 16-byte dim slices of the same tokens, so two warps already share each sector and there is nothing to recover. hd256 only.
`SPARKINFER_FA_KV_STAGE=0` restores main out of the same binary.

## Proof of speedup

**Target model(s)**

- [ ] **Muse Glimmer**
- [x] **Shared / both** — the change is in code both models use and should help either

| | decode tok/s |
|---|--:|
| before (main) | 47.06 |
| after (this PR) | 51.90 |

| | prefill pp tok/s |
|---|--:|
| before prefill (main) | 4113.0 |
| after prefill (this PR) | 4110.0 |

```text
unsloth/Qwen3.8-27B-NVFP4, RTX 5090 @525 W, main 4962fc0. Measured on 8d55f3f: 7fabb76 is eval/ only,
and 4962fc0 (#1110) gates on M in [2,32], which bs=1 decode never reaches, so this axis is unmoved.
Bot invocation qwen3_gguf_bench <ckpt> 128 sweep; bench.sh measures prebuilt releases, not this path.
  ctx 262144      main   this PR   knob=0 (same binary)    single runs base/PR/PR/base, adjacent pair
  target decode  47.06     51.90   47.07    (1.103x)
  target prefill 4113.0   4110.0   4101.6   (1.000x)
Scored axes, reversed adjacent pair: decode@128 1.002, prefill@128 1.004, decode@16k 1.005,
prefill@16k 1.009, decode@32k 1.022, prefill@32k 1.009; cb c2/c4/c8/c16/c32 1.003/0.999/1.000/1.000/
1.003. Guards @32k PR/main decode/prefill: Qwen3.6 1.000/1.002, unsloth Qwen3.8 1.022/1.009;
ModelOpt and Muse not run (checkpoints unavailable here) — Muse is hd128 and takes the unchanged arm
by construction. Accuracy: score dumps BIT-IDENTICAL to main at 16383 positions, top1 1.000, KL 0.000
(and at the 101-position default). Peak VRAM Qwen3.8 cb c32 31360 PR vs 31564 MiB main. ptxas 48
registers both arms, 4 B spill on the staged arm; a spill-free variant measured 1.099x, so it stays.
Full build + ctest 26/26.
```


skyrocket2026:

| metric | value |
|---|---|
| tier from | `muse-cb-decode@c16` (+0.1%) |
| commit | `9274ffc72` |

|---|---|---|---|---|
| 128 | decode | 104.26 | 104.02 | -0.2% |
| 128 | prefill | 7927.99 | 7907.28 | -0.3% |
| 512 | decode | 103.70 | 103.55 | -0.1% |
| 512 | prefill | 10468.43 | 10472.92 | +0.0% |
| 4k | decode | 102.74 | 102.61 | -0.1% |
| 4k | prefill | 16821.08 | 16750.21 | -0.4% |
| 16k | decode | 101.72 | 101.56 | -0.2% |
| 16k | prefill | 16335.33 | 16275.78 | -0.4% |
| 32k | decode | 100.13 | 100.11 | -0.0% |
| 32k | prefill | 14641.88 | 14599.96 | -0.3% |
| 64k | decode | 98.16 | 98.17 | +0.0% |
| 64k | prefill | 12269.74 | 12223.18 | -0.4% |

|---|---|---|---|
| c2 | 186.00 | 182.90 | -1.7% |
| c4 | 339.50 | 339.30 | -0.1% |
| c8 | 636.30 | 636.10 | -0.0% |
| c16 | 1090.70 | 1091.90 | +0.1% |
| c32 | 1535.30 | 1535.40 | +0.0% |

within significance gate — not a verified muse-cb-decode@c16 improvement

skyrocket2026:
## Closed: no verified speedup — `eval-museglimmer:none`

Measured on the pinned RTX 5090 against the same-box `origin/main`: **0.1%** on the best of 17 scored axes (decode + prefill @ 128/512/4k/16k/32k/64k, concurrent decode @ c2/c4/c8/c16/c32).

**This is not a finding that anything is wrong with your PR.** Nothing regressed, and every correctness gate passed — the change just did not move a number this bot measures. The queue is closed rather than left open so it reflects work that can still be scored.

Three ways forward, depending on which applies:

- **Nothing here measures your optimization yet?** That is a gap in the harness, not a verdict. Open an issue describing the axis you need — model, metric, context/concurrency, and the command that measures it — with your before/after numbers, then reopen this PR and ask for the [`hold`](../../labels/hold) label so it stays open while the axis is added. That path is real: `cb-decode@c2..c32` exists because #1026 asked for it.
- **Correctness fix, refactor, test or docs?** Those are welcome and reviewed by hand; they score 0 by design. Reopen as a **draft**, or ask for `hold`, and say so — being unscored is not grounds for closing that work.
- **Expected a speedup?** Re-measure against current `main` (it moves fast), and reopen with the new numbers — it re-evaluates on the next poll.

PR #1115, opened 2026-09-18
perf(muse): hold a prefix of NVFP4 o-proj copies when the whole set cannot fit (1.03x cb decode @c32)

carlh7777:
Summary
After #1108 the NVFP4 o-projection copy is the first Muse preflight drop: a 0.8 GB set that does not leave the runtime its graph pools instantiates into "out of memory" (51/52 copies, ~165 tok/s on a 32-GB 5090 at 33 sessions), so every packed o-proj stays on the Q4_K MMA even though packed decode and batched prefill already pick the FP4 arm per layer. The leftover after qkv-gate is only a couple of hundred MB — enough for a prefix, not the whole set — and ffn_down's prefix fill currently spends it on two Q4_K down layers the MMA already covers. A prefix [0, n) of o-proj copies is taken first, with the same keep margin the other fills use; 30 of 52 layers fit at keep 736 MiB. SPARKINFER_MUSE_NVFP4_WO_KEEP_MB=0 restores the all-or-nothing skip out of the same binary.

Proof of speedup

**Target model(s)**

- [x] **Muse Glimmer**

| | decode tok/s |
|---|--:|
| before (main) | 1557.6 |
| after (this PR) | 1604.1 |

| | prefill pp tok/s |
|---|--:|
| before prefill (main) |  |

```text
  main       189.8   341.7   637.2  1095.6  1557.6
  this PR    189.9   340.3   636.5  1101.7  1604.1   +3.0% @c32, c2–c16 flat (wo already 52/52)
  PR, KEEP=0 189.8   341.7   637.2  1095.6  1566.6   same binary
# Adjacent pair c32: 1557.6 -> 1604.1 / 1597.3 vs 1566.6 (base/PR/PR/base). o-proj 0/52 -> 30/52, down 2/52 -> 0/52, qkv-gate 52/52. Keep 1024 is 12/52 and flat; 640 is 36/52 and starves instantiate (~841 tok/s). Muse-only (c.muse_glimmer); Qwen3.6 / Qwen3.8 / AR decode / prefill never take the prefix.


skyrocket2026:
## 🚩 Flagged: eval-gaming

This PR involves an account blocked for gaming the SN74 emission mechanism (sybil / coordinated duplicate farming): `carlh7777`.

PR #1116, opened 2026-09-18
perf(qwen3.8): stage the int8 flash-decode K tile through shared memory so each token's bytes arrive whole (1.10x target decode @256k)

kaivaryn:
> Supersedes #1112 (same commit content, new head). #1112's run **scored this change**: `modelopt-decode@256k` **50.45 -> 55.97 tok/s, +10.9%**, `scored_dimension: modelopt-decode@256k`, with accuracy exact (top1 1.000, KL 0.0000), the Qwen3.6 and ModelOpt guards green, and every floor held (decode@128 -0.0%, prefill@128 -0.3%, prefill@16k -0.1%, cb c1-c32 within +/-0.4%).
> It was closed only because the **Muse Glimmer concurrent guard produced no measurement** — `GUARDCBMG_FAILED` -> "muse glimmer concurrent guard measurement unavailable", which `_check_model_guard` fails closed and treats as a regression. That guard cannot be affected by this change: `STAGE_K` is instantiated for `HEAD_DIM == 256` only, and Muse Glimmer is hd128, where `SS_W` falls back to `HEAD_DIM`, `STAGE_B` is 0 and the kernel is byte-identical to main. Note `_check_model_guard` also fails when the **main** leg is the one missing, so the absent reading is not necessarily on this PR's side.

## Summary

The int8 tensor-core flash-decode wants a 16x16 K tile whose columns are TOKENS, so at `ldm = num_kv_heads * HEAD_DIM` its fragment load asks for sixteen 16-byte runs a kilobyte apart — 16 useful bytes out of every 32-byte sector — and nothing catches the other half, because the kernel's 31.25 KB of dynamic shared memory against the 100 KB an `sm_120` SM actually has caps residency at three blocks and leaves ~11 KB of L1 per block against a 32 KB per-group K working set. K now arrives through shared memory instead: a warp reads each token's 64 contiguous bytes — two whole sectors, nothing discarded — and scatters them into four 16-byte-strided tiles the mma reads at `ldm` 16, which is the same operand element for element. It costs no occupancy because `s_s` pays for it: that buffer is declared `[16][HEAD_DIM]` but every store into it uses `ldm` 128, the group token width rather than the head dim, so at hd256 half of its 16 KB was never addressed, and trimming it to `[16][128]` releases exactly the 8 KB the staging buffer needs. V keeps the direct load — its eight warps read adjacent 16-byte dim slices of the same tokens, so two warps already share each sector and there is nothing to recover. hd256 only.
`SPARKINFER_FA_KV_STAGE=0` restores main out of the same binary.

## Proof of speedup

**Target model(s)**

- [ ] **Muse Glimmer**

| | decode tok/s |
|---|--:|
| before (main) | 50.45 |
| after (this PR) | 55.97 |

| | prefill pp tok/s |
|---|--:|
| before prefill (main) | 4113.0 |
| after prefill (this PR) | 4110.0 |

```text
The decode table above is the eval bot's OWN #1112 measurement of the scored axis, on the ModelOpt
NVFP4 checkpoint: modelopt-decode@256k 50.45 -> 55.97 (+10.9%), floors decode@128 -0.0%, prefill@128
-0.3%, prefill@16k -0.1%, cb c1..c32 within +/-0.4%, accuracy top1 1.000 / KL 0.0000.
My own runs below are unsloth/Qwen3.8-27B-NVFP4 on an RTX 5090 @525 W (the ModelOpt checkpoint is not
available to me), measured on 8d55f3f and re-measured on 4962fc0; main is now 4b6f4cd, whose #1113 is
eval/ only while #1110 and #1107 gate on packed widths (M in [2,32] / M >= 8) a bs=1 decode never
reaches. qwen3_gguf_bench <ckpt> 128 sweep; bench.sh measures prebuilt releases, not this path.
  ctx 262144      main   this PR   knob=0 (same binary)    single runs base/PR/PR/base, adjacent pair
  target decode  47.06     51.90   47.07    (1.103x, vs the bot's 1.109x on ModelOpt)
  target prefill 4113.0   4110.0   4101.6   (1.000x)
Scored axes, reversed adjacent pair: decode@128 1.002, prefill@128 1.004, decode@16k 1.005,
prefill@16k 1.009, decode@32k 1.022, prefill@32k 1.009; cb c2/c4/c8/c16/c32 1.003/0.999/1.000/1.000/
1.003. Guards @32k PR/main decode/prefill: Qwen3.6 1.000/1.002, unsloth Qwen3.8 1.022/1.009;
ModelOpt and Muse not run (checkpoints unavailable here) — Muse is hd128 and takes the unchanged arm
by construction. Accuracy: score dumps BIT-IDENTICAL to main at 16383 positions, top1 1.000, KL 0.000
(and at the 101-position default). Peak VRAM Qwen3.8 cb c32 31360 PR vs 31564 MiB main. ptxas 48
registers both arms, 4 B spill on the staged arm; a spill-free variant measured 1.099x, so it stays.
Full build + ctest 26/26.
```


skyrocket2026:

| metric | value |
|---|---|
| tier came from | `cb-decode@c8` |
| PR decode tok/s | 84.61 |
| main decode tok/s | 84.72 |
| PR prefill@128 pp | 5347.75 |
| main prefill@128 pp | 5391.95 |
| prefill@128 vs main | -0.8% |
| PR prefill@16k pp | 10274.71 |
| main prefill@16k pp | 10416.17 |
| prefill@16k vs main | -1.4% |
| decode@256k (ModelOpt NVFP4) | PR 57.53 / main 55.67 tok/s (+3.3%) |
| commit | `4e9a675c0` |

|---|--:|--:|--:|--:|
| c1 (floor) | 83.0 | 82.9 | -0.1% | 0 / 0 |
| c2 (scored) | 156.5 | 156.4 | -0.1% | 0 / 0 |
| c4 (scored) | 264.8 | 264.4 | -0.2% | 0 / 0 |
| c8 (scored) | 496.4 | 219.1 | -55.9% | 0 / 0 |
| c16 (scored) | 886.8 | 420.6 | -52.6% | 0 / 0 |
| c32 (scored) | 1353.7 | 1352.3 | -0.1% | 0 / 0 |

cb-decode@c8 regression: 219.10 < 98% of main 496.40 | cb-decode@c16 regression: 420.60 < 98% of main 886.80

skyrocket2026:
## Closed: regression or failed gate — `eval-qwen38:REJECT`

Measured **-55.9%** vs the same-box `origin/main`, (cb-decode@c8, cb-decode@c16 regression) — closing automatically.

Every measured axis is also a no-regression floor, and accuracy and the Qwen3.6, ModelOpt and Muse Glimmer guards are hard gates, so one failure closes the PR whatever it was aiming at. The verdict comment above names which one. Reopen once it is addressed and it re-evaluates on the next poll.

PR #1117, opened 2026-09-18
perf(muse): hold a prefix of NVFP4 o-proj copies when the whole set cannot fit (1.03x cb decode @c32)

milosde111:
Summary
After #1108 the NVFP4 o-projection copy is the first Muse preflight drop: a 0.8 GB set that does not leave the runtime its graph pools instantiates into "out of memory" (51/52 copies, ~165 tok/s on a 32-GB 5090 at 33 sessions), so every packed o-proj stays on the Q4_K MMA even though packed decode and batched prefill already pick the FP4 arm per layer. The leftover after qkv-gate is only a couple of hundred MB — enough for a prefix, not the whole set — and ffn_down's prefix fill currently spends it on two Q4_K down layers the MMA already covers. A prefix [0, n) of o-proj copies is taken first, with the same keep margin the other fills use; 30 of 52 layers fit at keep 736 MiB. SPARKINFER_MUSE_NVFP4_WO_KEEP_MB=0 restores the all-or-nothing skip out of the same binary.

Proof of speedup

**Target model(s)**

- [x] **Muse Glimmer**

| | decode tok/s |
|---|--:|
| before (main) | 1557.6 |
| after (this PR) | 1604.1 |

| | prefill pp tok/s |
|---|--:|
| before prefill (main) |  |

```text
  main       189.8   341.7   637.2  1095.6  1557.6
  this PR    189.9   340.3   636.5  1101.7  1604.1   +3.0% @c32, c2–c16 flat (wo already 52/52)
  PR, KEEP=0 189.8   341.7   637.2  1095.6  1566.6   same binary
# Adjacent pair c32: 1557.6 -> 1604.1 / 1597.3 vs 1566.6 (base/PR/PR/base). o-proj 0/52 -> 30/52, down 2/52 -> 0/52, qkv-gate 52/52. Keep 1024 is 12/52 and flat; 640 is 36/52 and starves instantiate (~841 tok/s). Muse-only (c.muse_glimmer); Qwen3.6 / Qwen3.8 / AR decode / prefill never take the prefix.


skyrocket2026:

| metric | value |
|---|---|
| tier from | `muse-prefill@4k` (+2.7%) |
| commit | `451a0d356` |

|---|---|---|---|---|
| 128 | decode | 103.29 | 103.99 | +0.7% |
| 128 | prefill | 7831.21 | 7890.08 | +0.8% |
| 512 | decode | 102.59 | 103.52 | +0.9% |
| 512 | prefill | 10251.02 | 10395.19 | +1.4% |
| 4k | decode | 102.01 | 102.55 | +0.5% |
| 4k | prefill | 16273.99 | 16714.96 | +2.7% |
| 16k | decode | 101.05 | 101.48 | +0.4% |
| 16k | prefill | 15834.59 | 16222.50 | +2.4% |
| 32k | decode | 99.78 | 100.00 | +0.2% |
| 32k | prefill | 14302.95 | 14525.61 | +1.6% |
| 64k | decode | 97.95 | 98.02 | +0.1% |
| 64k | prefill | 12067.95 | 12151.25 | +0.7% |

|---|---|---|---|
| c2 | 184.80 | 185.00 | +0.1% |
| c4 | 339.10 | 339.50 | +0.1% |
| c8 | 635.40 | 633.00 | -0.4% |
| c16 | 1092.00 | 1091.90 | -0.0% |
| c32 | 1536.90 | 1558.00 | +1.4% |

ok

skyrocket2026:

PR #707, opened 2026-08-07
Perf/dflash q5down nsplit fix

FranDev132:
## Summary

- [x] Tested on **RTX 5090** (`sm_120`)

| | decode tok/s |
|---|--:|
| before (main) | 804 |
| after (this PR) | 850 |

| | prefill pp tok/s |
|---|--:|
| before prefill (main) |  |
| after prefill (this PR) |  |

PR #1118, opened 2026-09-18
perf(qwen3.8): give the 24Q/4KV group the split-count correction the 8:1 and 4:1 groups already have (1.06x decode @256k)

Paral1995:
## Summary

`adaptive_nsplits_for` has carried an hd256 occupancy correction since #707 — 160 for the 8:1 group,
128/192/160 by band for the 4:1 group. **The 6:1 group (Qwen3.8-27B's full-attention layers) never
got one** and fell through to the raw `MAX_NSPLITS`, which costs it a whole CTA wave.

`fa_split_gqa_mma_i8_kernel<256,6>` holds **three blocks per SM**, so a 170-SM part runs **510 CTAs**
at once. The grid is `n_kv_heads * n_splits`, so 256 asks for **1024** CTAs — two full waves plus a
third carrying **four**. Dropping into the two-wave band deletes that tail. The cost is a slightly
longer chunk and one ragged 128-token group at the end of it, which is worth far less than the wave.

One line, no kernel change.

## Proof of speedup

- [ ] **Muse Glimmer**
- [ ] **Shared / both**

| | decode tok/s |
|---|--:|
| before (main) | 55.70 |
| after (this PR) | 58.86 |

| | prefill pp tok/s |
|---|--:|
| before prefill (main) | 5046.75 |
| after prefill (this PR) | 5031.76 |

Prefill does not use this kernel; it is listed only to show it is unmoved (the two differ by less
than the 1% run-to-run spread of the 256k prefill row).

### Measured, RTX 5090, ctx=262144, `modelopt-decode@256k`

Bot invocation, on current `main` (`2c1ebab`). The split count is the only variable:

| n_splits | CTAs | rep 1 | rep 2 | rep 3 | mean |
|--:|--:|--:|--:|--:|--:|
| 256 (before) | 1024 | 55.71 | 55.68 | 55.70 | 55.697 |
| 255 | 1020 | 58.81 | 58.81 | 58.81 | 58.810 |
| **252 (this PR)** | **1008** | **58.87** | **58.85** | **58.87** | **58.863** |

**decode@256k 55.697 → 58.863 tok/s (1.057×).** 252 beats 255 in every rep, so the peak of the band
is what this sets rather than the largest two-wave value. The band is not flat — 250 measures 57.32
and 248 measures 58.64 — so the constant is the measured optimum, not a round number.

### Scope

`want == MAX_NSPLITS` is reached at `seqlen > 16384` (and in the 7168–12288 band). Every shorter
context resolves to a smaller `want` and is not reached at all: `decode@128`, `prefill@128`, and
every concurrent-decode width, whose sequences top out at 768.

`ctx=32768` is the other context in the band, and it does not regress:

| ctx=32768 | before | after |
|---|--:|--:|
| decode tok/s | 86.955 | 86.870 |
| prefill pp | 11743.8 | 11697.8 |

No kernel change, so the attention math is identical — only how many splits the partials are
computed in, and empty splits contribute zero.


skyrocket2026:
## sparkinfer qwen38 auto-eval — `eval-qwen38:S`

| metric | value |
|---|---|
| **label** | `eval-qwen38:S` |
| tier came from | `modelopt-decode@256k` |
| PR decode tok/s | 84.75 |
| main decode tok/s | 84.72 |
| PR prefill@128 pp | 5392.19 |
| main prefill@128 pp | 5391.95 |
| prefill@128 vs main | +0.0% |
| PR prefill@16k pp | 10373.76 |
| main prefill@16k pp | 10416.17 |
| prefill@16k vs main | -0.4% |
| decode@256k (ModelOpt NVFP4) | PR 58.84 / main 55.67 tok/s (+5.7%) |
| commit | `8b12b6809` |

|---|--:|--:|--:|--:|
| c1 (floor) | 83.0 | 83.0 | +0.0% | 0 / 0 |
| c2 (scored) | 156.5 | 156.4 | -0.1% | 0 / 0 |
| c4 (scored) | 264.8 | 264.4 | -0.2% | 0 / 0 |
| c8 (scored) | 496.4 | 496.8 | +0.1% | 0 / 0 |
| c16 (scored) | 886.8 | 886.4 | -0.0% | 0 / 0 |
| c32 (scored) | 1353.7 | 1349.2 | -0.3% | 0 / 0 |

ok

skyrocket2026:

PR #1119, opened 2026-09-18
perf(qwen3.8): read 128 bytes of each weight row per K-tile in the packed FP8 GDN GEMM (1.08x cb decode @c8)

kaivaryn:
## Summary

A packed decode step's GDN in- and out-projections run `pf_gemm_fp8_kernel`, 31% of a linear-attention layer's critical path at c32, and with a 64-byte K tile each CTA reads 64 bytes of each of its 128 weight rows before returning to the first — 128 DRAM streams at 64-byte granularity, which is why it streams the checkpoint's FP8 weights at 0.93 TB/s where the row-at-a-time FP8 GEMV reads the same weights at 1.40. The tile is 128 rows tall where such a step has 8 to 32, so it is now 32 rows tall at those widths and spends the shared memory that frees on a 128-byte K tile, halving how often a CTA returns to the top of its row set. `SPARKINFER_FP8_GEMM_NARROW_M=0` restores the 128-row 64-byte tile, which is main byte for byte.

## Proof of speedup

**Target model(s)**

- [ ] **Muse Glimmer**
- [ ] **Shared / both**

| | decode tok/s |
|---|--:|
| before (main) | 502.9 |
| after (this PR) | 545.1 |

| | prefill pp tok/s |
|---|--:|
| before prefill (main) |  |

```text
unsloth/Qwen3.8-27B-NVFP4 dir, RTX 5090 sm_120 (525 W), main 89504f6; bench.sh measures prebuilt releases, not this path.
Bot invocation: SPARKINFER_QWEN38_{PREFILL,DECODE}_NVFP4=1 SPARKINFER_KV_INT8=1 qwen3_gguf_cb_bench <dir> <c> 256 256 512, agg_tok_s.
            main           this PR        PR knobs=0 (same binary)   delta
 cb @c8   502.9/503.8    545.1/545.0            -                    +8.2%
 cb @c16  889.6/888.5    940.6/940.9          891.9 (+0.3%)          +5.8%
 cb @c32  1359.5/1360.3  1401.5/1401.2          -                    +3.1%
Flat: c2/c4 never reach this GEMM (8-row floor); the 32k guards and decode@256k are single-request, so M=1 takes the FP8 GEMV, not this tile.
Guards PR/main decode|prefill: Qwen3.6 32k .997|.996, ModelOpt 32k .998|.994, Muse 32k .998|.996, qwen3.8 prefill@16k .995, decode@256k 1.000.
Accuracy (score-dump pair): top1 101/101=1.000, KL 0.0000, PPL 3.0472 vs 3.0472; that dump is one 102-token prefill so it exercises the unchanged 128-row tile, and the 32-row tile is bit-identical at BK=64, moving only the split-K boundary at BK=128.
VRAM peak c32: qwen3.8 31564->31440, Muse 32078->32080 MiB. Full build + ctest 26/26.
```


skyrocket2026:

| metric | value |
|---|---|
| PR decode tok/s | 48.51 |
| main decode tok/s | 48.52 |
| PR prefill@128 pp | 3465.14 |
| main prefill@128 pp | 3515.77 |
| prefill@128 vs main | -1.4% |
| PR prefill@16k pp | 6965.96 |
| main prefill@16k pp | 7002.18 |
| prefill@16k vs main | -0.5% |
| qwen3.6 guard | ❌ **FAILED** — qwen3.6 prefill@32k: 5935.4 < 98% of main 6072.6 (-2.3%) — **verdict forced to REJECT regardless of speed/accuracy** |
| commit | `373df4de2` |

|---|--:|--:|--:|--:|
| c1 (floor) | 48.0 | 50.7 | +5.6% | 0 / 0 |
| c2 (scored) | 96.3 | 94.8 | -1.6% | 0 / 0 |
| c4 (scored) | 170.5 | 170.0 | -0.3% | 0 / 0 |
| c8 (scored) | 337.4 | 343.3 | +1.7% | 0 / 0 |
| c16 (scored) | 98.5 | 922.9 | +837.0% | 0 / 0 |
| c32 (scored) | 601.4 | 1386.8 | +130.6% | 0 / 0 |

qwen3.6 no-regression guard failed: qwen3.6 prefill@32k: 5935.4 < 98% of main 6072.6 (-2.3%) | ok

skyrocket2026:
## Closed: regression or failed gate — `eval-qwen38:REJECT`

Measured **837.0%** vs the same-box `origin/main`, and regressed the Qwen3.6 no-regression guard (decode/prefill on shared code) — closing automatically.

Every measured axis is also a no-regression floor, and accuracy and the Qwen3.6, ModelOpt and Muse Glimmer guards are hard gates, so one failure closes the PR whatever it was aiming at. The verdict comment above names which one. Reopen once it is addressed and it re-evaluates on the next poll.

PR #1120, opened 2026-09-18
perf(muse): leave 1024 MiB after the o-proj NVFP4 prefix so packed graphs instantiate (2.12x cb decode @c32)

petermilord:
## Summary

#1117's o-proj prefix fill stops when cudaMemGetInfo says fewer than 736 MiB remain, which on a 32-GB 5090 at 33 sessions still converts 28/52 layers. Those copies fragment the heap: the 333 MB prefill scratch then declines with 16 MB free, CUDA-graph instantiate misses, and a packed c32 step spends 4.7 s on the first token (743 tok/s). The other two prefix legs already keep 1024 MiB. Using that same floor here holds 10/52 o-proj layers and leaves a contiguous arena the packed decode can actually allocate, 1575 tok/s. Packed also folds sigmoid(gate) and the o-proj Q8_1 into the combine AR already uses on the 42 layers that stay Q4_K, and the GQA16 MMA kernel asks for the four blocks/SM shared memory actually allows instead of five. c2–c16 already had the VRAM; AR decode and prefill stay on the skip path (0/52 o-proj). SPARKINFER_MUSE_NVFP4_WO_KEEP_MB=736 restores main, SPARKINFER_MUSE_PACKED_WO_GATE_Q8=0 the split combine, SPARKINFER_FAGQA16_MMA=0 the tile kernel.

## Proof of speedup

**Target model(s)** — tick every model this change is meant to speed up. This decides which bots
spend GPU time on it: a bot whose model you did **not** tick may skip your PR instead of spending a
~20-minute round proving a change it cannot move. Tick nothing and every bot evaluates it, as before.

- [x] **Muse Glimmer**

| | decode tok/s |
|---|--:|
| before (main) | 742.7 |
| after (this PR) | 1574.7 |

| | prefill pp tok/s |
|---|--:|
| before prefill (main) | 17486.53 |
| after prefill (this PR) | 17497.25 |

  main       189.7   342.5   642.9  1109.4   742.7
  this PR    189.6   342.7   642.5  1108.9  1574.7   +112% @c32
  PR, KEEP=736 189.7  342.5  642.9  1109.4   742.7   same binary
# main = KEEP=736 (28/52 o-proj, scratch alloc failed free=16 MB, max_itl 4712 ms @c32). this PR = KEEP=1024 (10/52 o-proj, graphs instantiate, max_itl 879 ms).
# qwen3_gguf_bench n=128 ctx=4096: decode 102.83 -> 102.96, prefill pp 17486.53 -> 17497.25 (AR path, 0/52 o-proj, not the scored win).


skyrocket2026:
## 🚩 Flagged: eval-gaming

This PR involves an account blocked for gaming the SN74 emission mechanism (sybil / coordinated duplicate farming): `petermilord`.

PR #1123, opened 2026-09-20
fix(qwen35): a concurrent request could be served from its recurrent state read at the wrong width

skyrocket2026:
A concurrent request could be served from its recurrent state read at the wrong width. This is on
the **default** path — no flag, no unusual checkpoint — for every hybrid (Gated-DeltaNet) model,
and it produces fluent output that degenerates a few tokens in rather than an error.

Continuous-batch decode compacts a session's GDN state from fp32 to bf16, packed from the
allocation base into the first half of the same buffer, the first time it packs that session.
Three things disagreed with that representation.

**1. The slot offset changed units and one caller did not.** `launch_qwen36_gdn_ar` took a pointer
the caller had already advanced to the layer's slot; `launch_qwen36_gdn_ar_batched` took the base
pointer plus a separate `state_off` and applied it inside the bf16 indexing. Under the compacted
form the offset counts bf16 elements, so `float*` arithmetic landed every slot at twice its byte
offset — layer 0 correct, every other GDN layer reading a slot it does not own. Both launchers now
take `(state, state_off)`.

**2. A decode CUDA graph outlived the representation it captured.** The graph bakes which kernel
instantiation runs (`state_compact_b16` selects one), but its validity key was
`(attn_mode, sparse, n_splits)`. A request that decodes alone, joins a batch, then outlives it
replays its fp32 capture over a compacted state. The representation is now in the key, and is
parked and restored with the graph so a session compacted while parked recaptures on its way back.

**3. A packed batch could be mixed.** `packed_state_b16` was "every row compacted", but rows
compacted on an earlier step stay compacted. A row that could not be — the shared prefix session,
a missing one, a failed conversion, or any row once `SPARKINFER_CB_GDN_STATE_B16=0` is set on a
process that already compacted some — left one kernel instantiation reading both kinds. It is
declined now; the per-row fallback consults each session's own flag and is correct for both.

### Why it went unnoticed

The unbatched kernel serves a row whenever the packed batch declines **or decays below two live
rows**, which happens in every concurrent batch as its requests finish at different lengths. But
`gdn_batched_gpu_test` already compared batched against unbatched at a non-zero slot offset and
passed throughout, because it only ever ran fp32 — it now runs the compacted form at slot 0 and at
later slots. And no accuracy gate here drives concurrency at all, so nothing above the kernels was
watching (2), which is a graph-replay fault and invisible to a unit test.

### Reproduction

Ask one prompt at temperature 0 twice against one server: once alone, once inside a batch that
decays to a single row (three short `max_tokens` beside one long). They must come back identical.
A *flat* concurrent batch where nothing finishes early does not reproduce it.

Measured on an RTX 5090, Qwen3.8-27B derivative, c4 decaying to one row:

| | before | after |
| --- | --- | --- |
| decayed row vs. asked alone | `'We need answer user's request: "List the first !!!!!!!!…'` | identical |
| flat c4 | correct | correct, ~51 tok/s/row |
| `gdn_batched_gpu_test` | PASS (fp32 only) | PASS incl. compacted at slot 0 and later slots |

### Also here

A verify decline was `fprintf`'d unconditionally, so a model the packed path cannot drive wrote
two stderr lines per decode token forever. Each site now reports itself once, keyed on the format
string's address because several carry a position that moves every step. Measured 78 lines → 3 on
one short run.

### Provenance

Found while bringing up a ternary checkpoint on `feat/ternary-bonsai-2-27b`, where the packed path
declines on every step and so hit (1) on all rows at once instead of only in a batch's tail. These
three commits are cherry-picked out of that branch because they fix the default path and should
not wait on model-support work. Original SHAs are named in each commit message.


skyrocket2026:
Verification notes, since the branch is cherry-picked and deserves its own evidence rather than the source branch's.

**Builds clean on `main`** (RTX 5090 box, full build from the cherry-picks), and `gdn_batched_gpu_test` passes there including the new compacted-state cases at slot 0 and at later slots.

**On the NVFP4 Qwen3.8-27B checkpoint this comparison cannot discriminate, and I want that on the record rather than presented as a pass.** Asking one prompt at temperature 0 three times *strictly sequentially*, with no concurrency anywhere, returned 354, 354 and 362 characters. The batched answer came back at 362 — i.e. one of the two answers that checkpoint gives on its own. So the alone-vs-decayed comparison says nothing either way there, and I am not claiming it as evidence.

The evidence for the fix is the ternary checkpoint, which is stable across repeated runs (identical 366-character answers on both paths over five runs):

| | before | after |
| --- | --- | --- |
| decayed row, folded path | `'We need answer user's request: "List the first !!!!!!!!…'` | identical to the same row asked alone |
| decayed row, native path | `'We need need\nUser'` | identical to the same row asked alone |

Worth knowing for anyone diffing server output on that checkpoint: it is non-deterministic at temperature 0 by default.

Issue #1093, opened 2026-09-16
GGUF Vision / mmproj support


tenten8401:
Hi! Could we get mmproj support for GGUF models? Wanting to use a Q4_K_M GGUF which ships a mmproj and has vision support on llama-cpp, but sparkinfer doesn't support them.

https://huggingface.co/unsloth/Qwen3.8-27B-GGUF/blob/main/Qwen3.8-27B-UD-Q4_K_M.gguf
https://huggingface.co/unsloth/Qwen3.8-27B-GGUF/blob/main/mmproj-BF16.gguf

maybe something like --mmproj? 

Thanks!

PR #1124, opened 2026-09-20
feat(qwen35): support Ternary-Bonsai-2-27B (PTQ1_0 ternary + Hadamard rotation)

skyrocket2026:
Adds support for **prism-ml/Ternary-Bonsai-2-27B-gguf**, a 1.75-bit ternary quantization of
Qwen3.8-27B whose weights live in a rotated basis. It loads, serves and scores; teacher-forced
over 107 positions of prose it reaches PPL 7.89, against the unquantized checkpoint's 4.46
through the same runtime.

No existing model changes path. The config, the layer stack and every kernel are the ones already
here — what is new is a weight format, a rotation to undo at load, and two orderings the file uses
that the runtime does not.

### The format

`PTQ1_0` (ggml type 143) packs 128 trits into a 28-byte block with one FP16 scale. The trits are
read in ggml's TQ1_0 order: 24 five-trit carrier bytes walked in two runs of 16 and 8, each run
emitting its trits **position-major**. Reading them carrier-major decodes every value correctly
and puts every one in the wrong place — which no round-trip test can see, because it packs with
the order it unpacks. It was caught by decoding against the unquantized checkpoint and fitting the
residual, which came out as a Hadamard with a permutation of period 128.

The checkpoint stores `R·W[o]` with `R = H·diag(s)`. Un-rotating each row with `R⁻¹ = diag(s)·H`
at load is the same arithmetic moved to the other operand, and leaves an ordinary model: no graph
needs a rotation inserted and every existing kernel applies unchanged. `token_embd` un-rotates
like everything else — the metadata's `inverse_weight_names` names a runtime obligation, not a
different storage direction.

### Defects this turned up, in the order they mattered

- **GDN v-head regrouping.** Everything that *produces* a v head — `attn_qkv`'s v rows,
  `attn_gate`, `ssm_conv1d`'s v channels, `ssm_a`, `ssm_dt.bias`, `ssm_alpha`, `ssm_beta` — stores
  its 48 heads transposed, while `ssm_out`, which consumes them, does not. Worth PPL 254,820
  against 8.
- **`gdn_qh_block` was never set from GGUF.** The safetensors loader sets it; the GGUF loader did
  not, so every GGUF of this family drove each v head from the wrong q/k head. PPL 139 against 8,
  and the reference scored fine throughout because it loads through the other path. The two
  loaders derive config independently — "the reference works through the same kernels" does not
  mean the same config.
- **fp16 subnormals in the Q4_K writer.** `d` is 1/63 of the weights it describes, so any group
  scale below ~3.8e-3 drove it under fp16's smallest normal and it was flushed to zero.
- **BF16 (type 30) had no GGUF block-table entry**, so every BF16 tensor sized to nothing and
  surfaced much later as an unrelated allocation failure.
- **A `dense()` returning nullptr on a failed `cudaMalloc`** was reported as a different tensor
  being absent.

### Native ternary path, off by default

`SPARKINFER_BONSAI_NATIVE` (`head`, `embed`, `proj`, or `all`) reads the 28-byte blocks as stored
and rotates the *activation* instead of folding the rotation into the weights. It is slightly
**better** than folding — PPL 7.89 against 8.11 — because the trits are read as they are rather
than refitted to Q4_K, and it holds 13.6 GB against 17.9.

It stays off, and the default is the point of the measurement rather than an oversight: it decodes
40.4 tok/s against 87.9 single-stream, and ~42 tok/s aggregate at four concurrent requests against
~204, because `proj()` cannot drive type 143, so the packed continuous-batch path declines every
step and falls back to one forward per row. A memory-and-quality trade for a single stream, not a
serving default.

### Tests

Five suites, all green: the PTQ1_0 codec (including an order test that does **not** go through the
packer, a subnormal-cliff test and a divergent-scale test), the Prism Hadamard, the v-head
regrouping, the rotation kernel, and the ternary GEMV against a host decoder.

`eval/bonsai_regression.py` is the guard — decoded tensors against the unquantized checkpoint
(cos ≥ 0.80; correct decode lands at 0.87–0.89), perplexity on both paths, greedy ids identical
between them, and a row that outlives its batch matching the same row asked alone. 3m40s, green.
`eval/run_bonsai_cron.sh` runs it daily off the PR-bot slots.

### Not supported

The FFN (~5.9 GB of the checkpoint) has no native path: every weight matrix goes through prefill's
`dq()`/`proj()` on N tokens at once, so it needs a batched ternary GEMM, not another per-tensor
slice. A per-layer attention slice was written and reverted for exactly this reason. `mmproj`
/vision is untouched.

### Note

Three commits here — the GDN state-slot offset, the decode-graph validity key, and the
once-per-site decline logging — fix a defect on the **default** path for every hybrid model, not
just this one, and are also open against `main` as #1123 so they need not wait on model support.
Expect them to drop out of this branch on rebase once that lands.


skyrocket2026:
**The FFN now reads natively too**, which was the largest gap in this branch — ~5.9 GB of the checkpoint was still being expanded to Q4_K at load because the fused expert kernel cannot drive type 143.

Its decode arm is three GEMVs plus an elementwise SwiGLU, the same shape the NVFP4 arm beside it already uses. It needs **two** rotations where a projection needs one: gate and up read the post-attention norm at the residual width, down reads SwiGLU's output at the FFN width — a different sign vector (17408 against 5120), applied in place. All three legs convert together or none do, since a Q4_K down leg beside native gate/up would read a correctly-rotated activation against un-rotated weights.

Prefill needed no batched ternary GEMM, which is what the branch description predicted. `dq()` already decoded ternary weights to bf16 scratch; it just matched the residual width alone, so the down leg fell through. Matching on the width was the whole change.

Measured on an RTX 5090, every configuration answering correctly:

| `SPARKINFER_BONSAI_NATIVE` | VRAM | single-stream | 4 concurrent | PPL |
| --- | --- | --- | --- | --- |
| unset (folded) | 17.9 GB | 87.4 tok/s | 214.4 tok/s | 6.946 |
| `head,embed,proj` | 13.6 GB | 40.4 tok/s | 44.3 tok/s | 6.449 |
| `all` | **8.2 GB** | 16.0 tok/s | 17.9 tok/s | **6.342** |

Perplexities are one passage, all three from the same build so they compare; the regression guard scores a longer one and reads 9.709 folded against 9.187 for `all`, with greedy ids identical between the paths and the decayed-batch check passing on both.

**Still off by default.** It is less than half the memory and the best-scoring of the three, at 5.5x the single-stream cost and 12x at concurrency — the packed continuous-batch path cannot drive a ternary projection, so every step falls back to one forward per row, and the FFN trades one fused kernel for four launches on each of 64 layers. That is the right option for a card that cannot hold 18 GB at all, and the wrong default for one that can.

The remaining gap on this branch is mmproj/vision, which is untouched and tracked separately as #1093.

skyrocket2026:
**The ternary GEMV was 1.9x off its own potential, and it was a kernel fault rather than a cost of the format.**

Two things on the line every weight in the model passes through:

- `pow3[m]` was a function-local array indexed by a value that differs across the lanes of a warp, so it could not stay in registers — nvcc spills it, and every trit extraction paid a **local-memory load**. It is a select chain now.
- The 28-byte block was re-read from global on each of four unrolled passes, as 32 scattered single-byte loads with a data-dependent index. It is staged in shared memory once per block.

The arithmetic is deliberately untouched, so it stays bit-identical — `gemv_ptq1_gpu_test` checks the kernel against a host decoder *and* against N separate GEMVs, and the guard requires native and folded to emit identical greedy ids. A speedup that had moved perplexity would have meant broken arithmetic, not a faster kernel; PPL is 9.187 before and after.

Updated serving numbers, every configuration answering correctly:

| `SPARKINFER_BONSAI_NATIVE` | VRAM | single-stream | 4 concurrent | PPL |
| --- | --- | --- | --- | --- |
| unset (folded) | 17.9 GB | 87.4 tok/s | 214.4 tok/s | 6.946 |
| `head,embed,proj` | 13.6 GB | 60.0 tok/s | 64.0 tok/s | 6.449 |
| `all` | **8.2 GB** | 29.8 tok/s | 34.3 tok/s | **6.342** |

So the full native path went from 5.5x the folded single-stream cost to 2.9x, and from 12x to 6.3x at concurrency, at less than half the memory and the best perplexity of the three. Still off by default, for the same reason as before.

A follow-up that held the block in the warp's registers via `__shfl_sync` instead of shared memory measured a dead heat (29.7 against 29.8) and was dropped rather than kept — the `__syncwarp`s it removed were not costing anything.

skyrocket2026:
**Packed continuous-batch decode now drives the ternary path**: 4-way aggregate went 34.3 → **82.7 tok/s**, and the decline count is 0.

It used to decline on every step. That was *correct* — a declined batch falls back to one forward per row — but it meant concurrency bought almost nothing (29.8 alone against 34.3 at four requests). Three arms were needed, because the packed path prepares activations three different ways: the projections, the dense FFN, and the LM head.

Two things were load-bearing and neither was obvious from the first attempt:

- **`proj_on()`, the side-stream variant, needed it too** and can't simply have its own rotation. It runs on `stream_k` while `proj()` runs on the main stream, and both want the single rotation scratch — a side stream rotating for itself is a write racing the main stream's read. The layer body now rotates `xn` once ahead of the fork and both read it, which is the rule the dp4a and fp8 activation staging two lines above already follows. Until this was done, layer 0 declined *silently* (no "unsupported type" message) because a launcher had simply returned false.
- **The staging has to be invalidated where the buffer is reused.** `xn` is the same pointer on every layer, so a staging left marked valid after the FFN repurposes the scratch would let the next layer's projections read the FFN's rotated SwiGLU output and treat it as a rotated `xn` — right pointer, wrong contents, nothing in the type to catch it.

Everything routes through one batched GEMM that is bit-identical per row to the single-row GEMV AR decode drives, so a packed step still reproduces the N separate steps it stands in for.

Current state of the option, every configuration answering correctly:

| `SPARKINFER_BONSAI_NATIVE` | VRAM | single-stream | 4 concurrent | PPL |
| --- | --- | --- | --- | --- |
| unset (folded) | 17.9 GB | 87.4 tok/s | 214.4 tok/s | 6.946 |
| `head,embed,proj` | 13.6 GB | 60.0 tok/s | 64.0 tok/s | 6.449 |
| `all` | **8.2 GB** | 29.8 tok/s | **82.7 tok/s** | **6.342** |

All six suites green at branch HEAD (`a156f1c`) plus the four-check guard: cosines 0.8912/0.8712/0.8706, native PPL 9.187 against folded 9.555, greedy ids identical between the paths, and the decayed-batch check passing on both.

Still off by default — 2.9x the single-stream cost — but the concurrency gap is down from 6.3x to 2.6x.

